PR #12, opened 2019-10-31
GKE single cluster configuration archetype

kumare3:
Demo purposes. To make it work, users will need to change some config

### DIFF
--- a/deployment/gke-single-cluster/flyte_generated.yaml
+++ b/deployment/gke-single-cluster/flyte_generated.yaml
@@ -592,38 +592,12 @@ data:
         name: domain
     registration:
       maxWorkflowNodes: 100
-    scheduler:
-      eventScheduler:
-        scheme: local
-        region: "my-region"
-        scheduleRole: "arn:aws:iam::abc123:role/my-iam-role"
-        targetName: "arn:aws:sqs:my-region:abc123:my-queue"
-      workflowExecutor:
-        scheme: local
-        region: "my-region"
-        scheduleQueueName: "won't-work-locally"
-        accountId: "abc123"
-    notifications:
-      type: local
-      region: "us-east-1"
-      publisher:
-        topicName: "foo"
-      processor:
-        queueName: "queue"
-        accountId: "bar"
-      emailer:
-        subject: "Notice: Execution \"{{ name }}\" has {{ phase }} in \"{{ domain }}\"."
-        sender:  "flyte-notifications@lyft.com"
-        body: >
-          Execution \"{{ name }}\" has {{ phase }} in \"{{ domain }}\". View details at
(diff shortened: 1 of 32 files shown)


deployment/gke-single-cluster/flyte_generated.yaml
```diff
@@ -0,0 +1,1550 @@
...
+      metrics-scope: "datacatalog"
+      profiler-port: 10254
+    application:
+      grpcPort: 8089
+    storage:
+      connection:
+        access-key: minio
+        auth-type: accesskey
+        disable-ssl: true
+        endpoint: http://minio.flyte.svc.cluster.local:9000
```
wild-endeavor: we're going to stick with using Minio for this?  should we move to gcs?

deployment/gke-single-cluster/flyte_generated.yaml
```diff
@@ -0,0 +1,1550 @@
...
+      - id: staging
+        name: staging
+      - id: production
+        name: production
+      - id: domain
+        name: domain
+    registration:
+      maxWorkflowNodes: 100
+    scheduler:
+      eventScheduler:
```
wild-endeavor: should we remove this section completely or is there a gcp equivalent we can move to?

deployment/gke-single-cluster/flyte_generated.yaml
```diff
@@ -0,0 +1,1550 @@
...
+        app: flyteadmin
+    spec:
+      containers:
+      - command:
+        - flyteadmin
+        - --logtostderr
+        - --config
+        - /etc/flyte/config/flyteadmin_config.yaml
+        - serve
+        image: docker.io/lyft/flyteadmin:v0.1.4
```
wild-endeavor: this says 0.1.4, but the overlay file has 0.1.1.  I'll remove that at some point.  no reason to specify if it's not being used.

deployment/gke-single-cluster/flyte_generated.yaml
ChethanUK: > we move to gcs?

Moving to GCS will be good..

kustomize/overlays/gke-single-cluster/propeller/propeller.yaml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  name: flytepropeller
+  namespace: flyte
+spec:
+  template:
+    spec:
+      containers:
+      - name: flytepropeller
+        env:
+        - name: QUBOLE_API_KEY
```
chethanu: What happen if we don't use QUBOLE?

mweagle:
Is this still valid @kumare3 ?

PR #14, opened 2019-11-11
pass argument to kustomize script

honnix:
By accepting an argument, users of this script would be able to specify
which deployment to build.

Internally we have customized overlay which doesn't make sense to
open source, but still we would like to utilize as much as possible from
this repo.

### DIFF
--- a/script/kustomize.sh
+++ b/script/kustomize.sh
@@ -1,10 +1,14 @@
+#!/usr/bin/env bash
+
 set -ex
+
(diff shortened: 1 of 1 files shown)


honnix:
@kumare3 PTAL whether this makes sense.

honnix:
Sure. Let's do that.

honnix:
Can someone from your side merge this? No permission here. Thanks.

PR #165, opened 2020-01-30
Add Statsd prometheus exporter to sandbox deployment

EngHabu:
This change adds a new daemonSet that uses https://github.com/prometheus/statsd_exporter to receive statsd metrics from user pods, aggregate and expose them as a scrape-able prometheus endpoint.

Flyte uses 2 different pipelines/SDKs to emit metrics; 

- System components (Propeller, Admin... etc.) expose prometheus metrics directly. These components are long lived and are suited for the prometheus pull model.
- User tasks (Python Tasks, Spark Tasks... etc.) use StatsD metrics to push metrics out. These tasks have un-defined runtime (from seconds to days) hence why they are more suited for StatsD push model.

Work in this PR:
- [X] Add DaemonSet deployment
- [X] Configure Plugin's default environment variables to point to the statsd endpoint.


mweagle:
Is this still valid @EngHabu ? /cc @wild-endeavor , @jonathanburns

kumare3:
@EngHabu should we keep this?

PR #172, opened 2020-02-12
Use Forbid restart policy for sync resources cron job

EngHabu:


mweagle:
Is this still valid @EngHabu ?

PR #184, opened 2020-02-20
upgrade to latest tag of datacatalog

honnix:
v0.1.1 is quite old and this PR upgrades to latest tag according to https://hub.docker.com/r/lyft/datacatalog/tags


honnix:
@kumare3 PTAL

EngHabu:
Can you run `make kustomize` to update the generated files.

honnix:
@EngHabu Right, I forgot about that. Good catch. Thanks.

PR #187, opened 2020-02-28
adding k8s-array plugin

jonathanburns:


kustomize/overlays/sandbox/propeller/config.yaml
```diff
@@ -64,3 +64,4 @@ tasks:
       - container
       - sidecar
       - spark
+      - k8s-array
```
EngHabu: I think we should also add it to the end2end tests.

And while we are at it, maybe also update all versions of components?

PR #199, opened 2020-03-09
Bump propeller version to latest

katrogan:


wild-endeavor:
oh you did... weird.

wild-endeavor:
you need to change the image too.

PR #200, opened 2020-03-09
check generated file as part of GitHub Workflow

honnix:
## Description

This PR configures a step of GitHub Workflow to ensure that in case of modification of kustomize code, `make kustomize` was executed and the newly generated `flyte_generated.yaml` was added as a part of the commit.

This PR is inspired by https://github.com/lyft/flyteidl/blob/master/generate_protos.sh#L58

## Motivation

It is very likely that a new contributor of this repo updated some kustomize file but didn't do `make kustomize`. Having an automated step telling what is missing would be useful.

## How did I test it

* https://github.com/lyft/flyte/pull/200/checks?check_run_id=496295498 as a successful workflow execution
* https://github.com/lyft/flyte/pull/200/checks?check_run_id=496307946 as a failed one

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,6 +15,8 @@ jobs:
         go-version: [1.10]
     steps:
     - uses: actions/checkout@v1
+    - name: Kustomize and diff
+      run: DELTA_CHECK=true make kustomize
     - name: Set up Go@${{ matrix.go-version }}
       uses: actions/setup-go@v1
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -15,6 +15,8 @@ jobs:
         go-version: [1.10]
     steps:
     - uses: actions/checkout@v1
+    - name: Kustomize and diff
```
wild-endeavor: cool.  it seems this runs as part of this PR too.

honnix: Yeah that was a successful execution.

EngHabu:
Thank you!

honnix:
_Sorry I didn't know force push would remove previous workflow executions._ :(

PR #247, opened 2020-04-07
Installing SageMaker operator

bnsblue:
This PR is to add the k8s operator of AWS SageMaker to Flyte deployment. SageMaker is AWS's hosted ML data transforming and training service (https://aws.amazon.com/sagemaker/) and has an [open-sourced k8s operator](https://github.com/aws/amazon-sagemaker-operator-for-k8s) for users to submit SageMaker jobs through from within a k8s cluster. Having this operator means to equip Flyte with the ability to submit well-formed SageMaker jobs from within Flyte


bnsblue:
🔨

bnsblue:
Just FYI @kumare3 @EngHabu @wild-endeavor , I have been able to run `make end2end` successfully on my laptop without a problem.

wild-endeavor:
Instead of copying, did you try something like this? https://github.com/kubernetes-sigs/kustomize/blob/master/examples/loadHttp/kustomization.yaml

wild-endeavor:
Also, can you describe what this PR does?  What is a sagemaker operator? Where is the code for it?  Can you like to a Flyte GH issue?

bnsblue:
> Instead of copying, did you try something like this? https://github.com/kubernetes-sigs/kustomize/blob/master/examples/loadHttp/kustomization.yaml

I did try it and it doesn't work now as there's currently no release that include this feature. I asked in their issue page to request a release that includes this feature but the request hasn't been addressed yet. 

And I also did leave a note in my comment to indicate that we should leverage that once it is available. See this https://github.com/lyft/flyte/pull/247/files#diff-67e4439f530f774154e4612cfa5abd2fR5

PR #249, opened 2020-04-08
K8s crons have issues on docker-for-mac deployments.

jonathanburns:
K8s crons have issues on docker-for-mac deployments. This moves cluster resource syncing from cron to loop. It also disables the logging for this function, since the logging is too verbose and quickly runs the node out of disk space.

This PR resolves https://github.com/lyft/flyte/issues/162

I've created an additional ticket to resolve verbose logging

### DIFF
--- a/deployment/sandbox/flyte_generated.yaml
+++ b/deployment/sandbox/flyte_generated.yaml
@@ -600,10 +600,10 @@ data:
         name: domain
     cluster_resources:
       templatePath: "/etc/flyte/clusterresource/templates"
-      refresh: 5m
+      refresh: 10s
 kind: ConfigMap
 metadata:
-  name: flyte-admin-config-4k466cfctf
+  name: flyte-admin-config-h889md4bc8
   namespace: flyte
 ---
 apiVersion: v1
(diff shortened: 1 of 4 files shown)


wild-endeavor:
I don't think mute is the right call.  We should be able to control the verbosity just through logging levels.  If Info logs alone are way too much and causing things to overload, then we should remove some of them.

Also, what does this refresh thing do?  It just runs every ten seconds perpetually?

jonathanburns:
I agree the logging should be less verbose. I created the following issue for the logging problem:

https://github.com/lyft/flyte/issues/250

The log_level is configured through the `flyteadmin_config.yaml`. We could probably change this one container to use a different log_level, but that could bite us later. It's probably better that all containers share the same log level that you can configure in one place. I think the best solution is to reduce the amount of logs that are spewing at this log level, as described by the issue above.

wild-endeavor:
I don't think having this run all the time in the background is the correct solution for this problem.

jonathanburns:
@wild-endeavor This is already running all the time in the background (using crons). It just doesn't work. This fixes that.

At one point, I had pitched kicking off a sync whenever projects are registered, but that's not how it works today. This change is definitely an improvement over the current state of the system (the current state is broken).

PR #252, opened 2020-04-08
Add admin service and entity docs

katrogan:


wild-endeavor:
I think you need to add the new file to one of the index.rst files also?  Maybe?

Also, what do you think the longer term split should be between what is written in the service/admin.proto file, and these docs?

PR #273, opened 2020-04-15
[debug don't merge] wait longer for k8s to start during e2e test

honnix:


honnix:
I managed to pass the waiting stage locally, but both flyteadmin init container and cron job tried to connect to wrong IP address and wrong port of postgres.

honnix:
I see, it was DNS IP and port. Took me a while to realise that. 😓

honnix:
OK it is how kubelet is started.

```
[root@4a0e8cf5750d kubernetes]# cat kubelet
###
# kubernetes kubelet (minion) config

# The address for the info server to serve on (set to 0.0.0.0 or "" for all interfaces)
KUBELET_ADDRESS="--address=127.0.0.1"

# The port for the info server to serve on
# KUBELET_PORT="--port=10250"

# You may leave this blank to use the actual hostname
KUBELET_HOSTNAME="--hostname-override=127.0.0.1"

# Add your own!
KUBELET_ARGS="--cgroup-driver=systemd --fail-swap-on=false"
KUBELET_ARGS="--cgroup-driver=systemd         --cgroups-per-qos=false         --enforce-node-allocatable=         --fail-swap-on=false         --kubeconfig=/etc/kubernetes/master-kubeconfig.yaml         --cluster-dns=10.254.0.10         --cluster-domain=cluster.local"```

kumare3:
Ohh man, you are awesome!

honnix:
The problem was `kube-proxy`. It wouldn't start, at least not on my laptop. Then `coredns.service` wouldn't start, so there is no DNS. :/

jonathanburns:
I have a feeling that Github might have made some changes to their CI VM's which either disallow or don't agree something we're doing here.

Possibly this line?

https://github.com/lyft/flyte/blob/master/end2end/launch_dockernetes.sh#L18

The reason this line is needed is because k8s adds the `bf_netfilter` kernel module, IIRC

honnix:
That might as well explain why it didn't work on my laptop (Mac). The tiny VM hosting docker doesn't have this kernel module.

wild-endeavor:
is this something we can install while the kernel is running?

honnix:
If there is no way to configure machine type, e2e might just won't work at all. I think.

PR #274, opened 2020-04-15
Docs process updated

kumare3:
- Better named makefile targets to generate docs
- Also using a pinned version of Flytekit and FlyteIDL for docs
- Added links to go repos documentation
- Fixed the index and introduction page (there was a duplicate)


script/generate_docs.sh
```diff
@@ -16,13 +16,13 @@ RSTS_DIR=`mktemp -d "${BASEDIR}/rsts_tmp/XXXXXXXXX"`
 
 # Copy all rst files to the same place
 cp -R rsts/* ${RSTS_DIR}
-cp -R _rsts/* ${RSTS_DIR}
+#cp -R _rsts/* ${RSTS_DIR}
```
wild-endeavor, line 19: I don't remember, what does this do?

wild-endeavor:
+1 with one comment.  approving, not sure if you want to change it

script/generate_docs.sh
kumare3: i think we need this, good catch, added back

kumare3:
> but the github workflow doesn't push to gh-pages automatically right? we still need to do that separately.

Right, that should be a follow up PR, I have something in progress, mind helping?

wild-endeavor:
yes ofc.

PR #284, opened 2020-04-20
Add support for sparkoperator webhook

akhurana001:
Ref https://github.com/lyft/flyte/issues/285

### DIFF
--- a/deployment/sandbox/flyte_generated.yaml
+++ b/deployment/sandbox/flyte_generated.yaml
@@ -881,6 +881,20 @@ spec:
     app: contour
   type: NodePort
 ---
+apiVersion: v1
+kind: Service
+metadata:
+  name: spark-webhook
+  namespace: sparkoperator
+spec:
+  ports:
+  - name: webhook
+    port: 443
+    targetPort: 8080
+  selector:
+    app.kubernetes.io/name: sparkoperator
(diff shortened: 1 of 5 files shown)


kumare3:
but this will enable the webhook automatically?

kumare3:
Also can you please create an issue and add it to the current milestone 0.3.0 and then close it, that way we can create better release notes

kumare3:
Please enable the certs

akhurana001:
> but this will enable the webhook automatically?

Added the service/init-job for certs etc. to support webhook. Also created/tagged an issue.

deployment/sandbox/flyte_generated.yaml
```diff
@@ -880,6 +881,20 @@ spec:
...
+  name: spark-webhook
+  namespace: sparkoperator
+spec:
+  ports:
+  - name: webhook
+    port: 443
+    targetPort: 8080
+  selector:
+    app.kubernetes.io/name: sparkoperator
+    app.kubernetes.io/version: v2.4.0-v1beta1
```
kumare3, line 896: does this need any plugin side changes?

akhurana001: no, The plugin is already on v1beta1 : https://github.com/lyft/flyteplugins/blob/master/go/tasks/plugins/k8s/spark/spark.go#L19

PR #292, opened 2020-04-23
SparkOperator: Run certs as an init-container

akhurana001:
Kustomize doesn't have a way to honor dependencies between the SparkOperator init-job which generates certs required by the SparkOperator deployment. 
This runs the init-job as an init-container on the deployment and mitigates this issue.


kumare3:
how does the OSS operator do this?

PR #354, opened 2020-06-16
Add single task execution docs

katrogan:


rsts/user/features/single_task_execution.rst
```diff
@@ -0,0 +1,90 @@
...
+Single task executions aren't limited to just tasks you've defined in your code. You can reference previously registered tasks and launch a single task execution like so:
+
+.. code-block:: python
+
+     from flytekit.common.tasks import task as _task
+
+     my_task = _task.SdkTask.fetch("my_flyte_project", "production", "workflows.my_task", "abc123")  # project, domain, name, version
+
+     my_task_exec = my_task.launch(project="my_other_project", domain="development", inputs={'plant': 'philodendron'})
+     my_task_exec.wait_for_completion()
```
wild-endeavor, line 49: Can Admin withstand potentially tens of thousands of these polling?

wild-endeavor: Do you think we should move to websockets?

katrogan: Following up offline

kumare3: Why websockets? These should be converted to long poll if needed. But we don’t need it at all I feel. This is a usecase only for testing

PR #424, opened 2020-07-21
kustomization for tensorflow plugin

pingsutw:
Add kustomization for tensorflow plugin


pingsutw:
> Do you want to deploy it to default sandbox deployment? It needs couple more things. 

Yes, What else needs to be added in this PR.

PR #527, opened 2020-09-28
Add documentation around interruptible

migueltol22:
Adding documentation around interruptible feature for users.


EngHabu (approved):
Thank you!

PR #556, opened 2020-10-15
Adding logo to sphinx docs

kumare3:


rsts/conf.py
```diff
@@ -93,6 +93,12 @@
 # a list of builtin themes.
 #
 html_theme = 'sphinx_rtd_theme'
+html_static_path = ['_static']
+html_logo = "flyte_lockup_on_dark.png"
```
katrogan, line 97: why is it called lockup?

katrogan (approved):

rsts/conf.py
kumare3: designer name?

PR #559, opened 2020-10-19
[Bug] fix; TFOperator manifest lead to a non-existent location

kumare3:


wild-endeavor (approved):
so kustomize knows to add the `/blob/` when actually downloading the files i assume?

kumare3:
> so kustomize knows to add the `/blob/` when actually downloading the files i assume?

you mean for github? Ya this is part of the `go-getter` module

katrogan (approved):

PR #571, opened 2020-10-27
Increase e2e test timeout

honnix:
Try 20 minutes timeout


honnix:
@wild-endeavor GitHub actions doesn't have strong enough machine to run e2e test fast enough. Increasing the timeout seems to work, but not so great.

katrogan (approved):

katrogan:
@honnix can we merge this in while we debug what caused the timeout?

honnix:
Sure I'm totally ok. I thought we would do some investigation first. Please go ahead.

PR #572, opened 2020-10-27
Spark CRD + Operator v1beta2

akhurana001:
Move spark plugin to v1beta2 multi-version of Operator.

Non-backward compatible: If the cluster already has Spark v1beta1 CRD & SparkOperator deployed, these will need to be manually deleted before the new CRD/operator can be deployed. This will cause any existing spark tasks/jobs running on the cluster to fail.  Ref: lyft/flyte#573 for more details.

Dependent on https://github.com/lyft/flyteplugins/pull/134

Updated FlytePropeller to version 0.5.0


kustomize/base/operators/spark/kustomization.yaml
```diff
@@ -1,6 +1,7 @@
 resources:
 - spark-operator.yaml
-- spark-dummy-crd.yaml
+- sparkapplications-crd.yaml
```
kumare3, line 3: we should probably add this to a CRD section

kumare3:
can you look at kustomize's CRD settings and add everything to that?

kustomize/base/operators/spark/kustomization.yaml
akhurana001: resolving as we decided to have this as a resource

kumare3 (approved):

kumare3:
Can we update the description to say that, this is a backwards incompatible change if you are using Spark!?

kumare3 (approved):

PR #628, opened 2020-11-30
Enable new event version in End2End

anandswaminathan:


EngHabu:
Can you run `make kustomize`?

PR #630, opened 2020-12-02
Update flyte to use container registry images instead

EngHabu:
https://github.com/lyft/flyte/issues/627


wild-endeavor (approved):
so we're bumping too?

kumare3:
should we bump, if we so we should create another release of flytekit?

EngHabu:
You mean bumping the components' versions? or the flyte version? the former yes... the latter we don't have to... whenever flytekit is ready, we can bump it separately and then create a flyte release... no?

PR #635, opened 2020-12-07
Mention Java as out of the box supported language

honnix:
@EngHabu PTAL thanks.


EngHabu (approved):

honnix:
Thanks for approving! Could you please help me merge it? It seems I don't have permission on this repo.

kumare3 (approved):

ArgentFalcon (approved):

kanterov:
What about Scala. I think it's supported as well. API is designed specifically to have good UX when using Scala, and there are extensions specific for Scala.

honnix:
This is complicated, unless we say `JVM based languages`, we will also miss something. But I'm OK adding Scala specifically.

EngHabu:
Maybe we worded something like `JVM based languages (e.g. Java, Scala... etc.)` ?

kumare3:
Not really all jvm languages right? As you can use the Java bridge but it may not be a native experience

honnix:
Yeah I think it is  reasonable to just list Java and Scala as first class citizens of JVM family.

PR #683, opened 2021-01-29
Updating kustomize and Contour

kumare3:
Contour version was very old. I replaced it with v0.11.1

Important note: we cannot directly use the remote yaml because it has
a namespace and kustomize's namespace transformer causes conflicts. So
i manually removed the namespace from the deployment

Important: Now console is simply available at
http://localhost/console if running locally
or the best way to get the console is

```bash
❯ kubectl get -n flyte service envoy -o wide
NAME    TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                      AGE     SELECTOR
envoy   LoadBalancer   10.96.225.227   localhost     80:32613/TCP,443:30037/TCP   2m18s   app=envoy
```

You can use the External IP
`http://<external-ip>/console`

OR

```bash
kubectl port-forward -n flyte svc/envoy 80:80
```
then use
http://localhost/console


EngHabu (approved):
Awesome! Thank you

PR #686, opened 2021-01-31
Access GRPC services using projectcontour

kumare3:
Between previous versions and later version the annotation to access a grpc service was change. Also during the previous change localhost:30084 was accidently removed. This change keeps the nodeport but also allows grpc to seamlessly work on port 80.

Thus to use the new sandbox config with flyte-cli use the following cli

```bash
$ flyte-cli -h localhost:80  -i list-projects
```
`Note the :80` this is required. Also assuming you are using docker for mac or running k8s locally`
here localhost can be replaced with envoy external IP

for flytectl use the following config
```yaml
admin:
  # For GRPC endpoints you might want to use dns:///flyte.myexample.com
  endpoint: localhost:80
  insecure: true
```

### DIFF
--- a/deployment/eks/flyte_generated.yaml
+++ b/deployment/eks/flyte_generated.yaml
@@ -8365,8 +8365,8 @@ apiVersion: v1
 kind: Service
 metadata:
   annotations:
-    contour.heptio.com/upstream-protocol.h2c: grpc
     external-dns.alpha.kubernetes.io/hostname: flyteadmin.subdomain.mydomain.com
+    projectcontour.io/upstream-protocol.h2c: grpc
     service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "600"
     service.beta.kubernetes.io/aws-load-balancer-extra-security-groups: sg-...,sg-...,sg-...
   name: flyteadmin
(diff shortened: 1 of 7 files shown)


wild-endeavor (approved):
i checked out this branch and ran make deploy sandbox and it still doesn't expose back 30081 for me.  I don't care that much, but we should just go through all the docs and make sure to remove that number if we're gonna stick with 80.

```(flyte) ytong@argus:~/go/src/github.com/lyft/flyte [docker-desktop] (fix-grpc-access) $ lsof -i :30081                                                                                           (flyte) ytong@argus:~/go/src/github.com/lyft/flyte [docker-desktop] (fix-grpc-access) $ lsof -i :80
COMMAND    PID  USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
com.docke 1240 ytong   58u  IPv6 0xf6c528374432f9ed      0t0  TCP *:http (LISTEN)
```
`flyte-cli -h localhost:30081 -i list-projects` also fails.

kumare3:
@wild-endeavor I think lets move away from using 30081 or other ports. With the ingress model the entire onus falls on envoy and the users wants to deploy envoy. This makes it extremely easy to get started with flyte.

PR #695, opened 2021-02-05
Update Link to Docs

apatel-fn:
Old link to the Docs was 404ing, updated to the `flyte.org` documentation. Hoping this is the correct reference for the README. :)


apatel-fn:
in favor of https://github.com/flyteorg/flyte/pull/692

PR #717, opened 2021-02-10
[Docs] - schedules how to

wild-endeavor:
Just copied what Katrina had from the old guide

### DIFF
--- a/rsts/howto/enable_and_use_schedules.rst
+++ b/rsts/howto/enable_and_use_schedules.rst
@@ -1,5 +1,108 @@
-.. _howto_enable_use_schedules:
+.. _howto_scheduling:
 
-###############################################
-How do I enable and use Schedules?
-###############################################
+#################################################
+How do I use Flyte scheduling?
+#################################################
+
+*******
+Usage
+*******
+
+Launch plans can be set to run automatically on a schedule if the Flyte platform is properly configured. There are two types of schedules, cron schedules, and fixed rate intervals.
+
+Cron Schedules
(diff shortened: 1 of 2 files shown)


rsts/howto/enable_and_use_schedules.rst
```diff
@@ -1,5 +1,108 @@
...
+
+    scheduler:
+      eventScheduler:
+        scheme: "aws"
+        region: "us-east-1"
+        scheduleRole: "arn:aws:iam::{{ YOUR ACCOUNT ID }}:role/{{ ROLE }}"
+        targetName: "arn:aws:sqs:us-east-1:{{ YOUR ACCOUNT ID }}:{{ YOUR QUEUE NAME }}"
+        scheduleNamePrefix: "flyte"
+
+* **scheme**: in this case because AWS is the only cloud back-end supported for scheduling workflows, only ``"aws"`` is a valid value. By default, the no-op scheduler is used.
```
katrogan, line 76: is this still true? not sure if spotify is running schedules

katrogan (approved):

rsts/howto/enable_and_use_schedules.rst
wild-endeavor: I'll merge and ask them later.

honnix: We do use a scheduler, but we deploy schedule config directly to it without going through flyteadmin, so we don't use this part of configuration.

PR #754, opened 2021-02-21
flyte-sandbox Dockerfile

jeevb:


.github/workflows/sandbox.yml
```diff
@@ -2,31 +2,46 @@ name: Build & Push Sandbox Docker Image
 
 on:
   pull_request:
-  create:
-    tags:
-      - v*
+  release:
```
jeevb, line 5: @EngHabu: How do you feel about using this trigger as opposed to `create`? This seems better if y'all are going to be cutting releases.

deployment/sandbox/docker/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+COPY flyte_generated.yaml /flyteorg/share/flyte_generated.yaml
+
+# Copy scripts
+COPY docker/kubectl docker/wait-for-flyte.sh /flyteorg/bin/
+
+
+FROM base_ AS default
+
+# Install dependencies
+RUN apk add --no-cache tini
```
jeevb, line 38: Should we be installing `git`, `make`, and `bash`?

.github/workflows/sandbox.yml
EngHabu: That's fine... is there a reason for the change? will one make it better for forking than the other?

deployment/sandbox/docker/Dockerfile
EngHabu: yeah.... that's alright... will optimize later... ;)

deployment/sandbox/docker/wait-for-flyte.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -e
+
+# Ensure cluster is up and running
+timeout 180 sh -c "until k3s kubectl explain deployment &> /dev/null; do sleep 1; done"
+
+# Wait for flyte deployment
+k3s kubectl wait --for=condition=available deployment/datacatalog deployment/flyteadmin deployment/flyteconsole deployment/flytepropeller -n flyte --timeout=180s
+
+timeout 180 sh -c "until [[ $(k3s kubectl get daemonset envoy -n flyte -o jsonpath='{.status.numberReady}') -eq 1 ]]; do sleep 1; done"
```
jeevb, line 11: There appears to be an issue where the entrypoint gets stuck here. Trying to figure it out.

.github/workflows/sandbox.yml
jeevb: I think they're both ok for forking. But this one better conveys the intent imo, since we explicitly cut releases for the repo. I'm fine with either.

EngHabu: awesome... release creation is fine!

deployment/sandbox/docker/wait-for-flyte.sh
EngHabu: Is this fixed now with the quotes?

jeevb: Yes it is!

.github/workflows/sandbox.yml
```diff
@@ -2,31 +2,46 @@ name: Build & Push Sandbox Docker Image
...
-      - name: Push Sandbox Plugin Docker Image to Github Registry
+      - name: Push Sandbox Docker Image to Github Registry
         uses: whoan/docker-build-with-cache-action@v5
         with:
           # https://docs.github.com/en/packages/learn-github-packages/publishing-a-package
           username: "${{ secrets.FLYTE_BOT_USERNAME }}"
           password: "${{ secrets.FLYTE_BOT_PAT }}"
           image_name: ${{ github.repository_owner }}/flyte-sandbox
           image_tag: latest,${{ github.sha }},${{ github.event.ref }}
-          push_git_tag: true
```
EngHabu, line 28: Are you removing this just for testing?

I think ideally we would replicate this workflow, one that get triggered with pull_request and does no push_git_tag and another with release creation that does have `push_git_tag: true`...

EngHabu: ah I just noticed the condition you added! awesome

jeevb: Well, no. I think this is unnecessary with `image_tag: ${{ github.sha }}`, right? Am I missing something here?

EngHabu (approved):

Issue #755, opened 2021-02-21
[BUG] ERROR:root:Error creating Flyte entity because of invalid arguments.


mneilly-et:
**Describe the bug**
I am trying to follow the getting started instructions and register the hello workflow per https://docs.flyte.org/en/latest/tutorials/run_on_flyte.html#running-your-flyte-workflows.

When running the following command:
```
FLYTE_PLATFORM_URL="http://flyte-host:30081" FLYTE_AWS_ENDPOINT=http://flyte-host:30083 FLYTE_AWS_ACCESS_KEY_ID=minio FLYTE_AWS_SECRET_ACCESS_KEY=miniostorage make fast_register FLYTE_HOST=flyte-host:30081
```

I get the following error:
> ERROR:root:Error creating Flyte entity because of invalid arguments. Create request:
> ...
> grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
        status = StatusCode.INVALID_ARGUMENT
        details = "failed to validate that project [flyteexamples] and domain [development] are registered, err: [entry not found]"

The full output follows:

> (venv) mneilly@MyHost$ FLYTE_PLATFORM_URL="http://flyte-host:30081" FLYTE_AWS_ENDPOINT=http://flyte-host:30083 FLYTE_AWS_ACCESS_KEY_ID=minio FLYTE_AWS_SECRET_ACCESS_KEY=miniostorage make fast_register FLYTE_HOST=flyte-host:30081
echo /home/mneilly/Projects/flyte/first_workflow/flyteexamples
/home/mneilly/Projects/flyte/first_workflow/flyteexamples
mkdir /home/mneilly/Projects/flyte/first_workflow/flyteexamples/_pb_output || true
mkdir: cannot create directory ‘/home/mneilly/Projects/flyte/first_workflow/flyteexamples/_pb_output’: File exists
rm /home/mneilly/Projects/flyte/first_workflow/flyteexamples/_pb_output/* || true
pyflyte --pkgs myapp.workflows serialize --in-container-config-path /root/flyte.config --image ghcr.io/flyteorg/flytekit-python-template:latest fast workflows -f _pb_output/
Configuration file 'flytekit.config' could not be loaded. Using values from environment.
Flyte Admin URL http://flyte-host:30081
Serializing Flyte elements with image ghcr.io/flyteorg/flytekit-python-template:latest
Writing output to _pb_output/
[...]

PR #756, opened 2021-02-21
FlyteExamples should be one of the seedprojects

kumare3:
- This is because the getting started guide uses this


EngHabu (approved):

kumare3:
Related to issue - https://github.com/flyteorg/flyte/issues/755

Issue #657, opened 2021-01-05
[Feature] Make OAuth Scopes configurable


EngHabu:
**Motivation: Why do you think this is important?**
Flyteadmin hardcodes the list of OAuth scopes it requests to profile, openid (common) and offline_access (only supported in some IdPs)

**Flyte component**
- [X] FlyteAdmin (Control Plane service)

**Success Criteria**
- [X] FlyteAdmin changes to make sure it works with Google OIdC, Okta and Github,
- [ ] Flyte docs updates to show examples of how to do that for these providers,
- [ ] FlyteKit changes if any to support the new flow.

PR #773, opened 2021-02-23
Add auth config examples

EngHabu:
Update documentation with steps on how to configure auth for a few popular services...

flyteorg/flyte#657
https://github.com/flyteorg/flyte/issues/925


kumare3:
Probably want to update this for changes needed to propeller?

wild-endeavor (dismissed):

kustomize/base/single_cluster/headless/config/admin/server.yaml
```diff
@@ -10,6 +10,23 @@ server:
...
+    thirdPartyConfig:
+      flyteClient:
+        clientId: flytectl
+        redirectUri: https://localhost:53593/callback
+        scopes:
+          - offline
+          - all
+  userAuth:
+    openId:
+      baseUrl: https://accounts.google.com
```
tnsetting, line 25: One issue with the base url is that https://accounts.google.com/.well-known/oauth-authorization-server does not exist.

EngHabu: If you put `https://accounts.google.com` under `appAuth -> baseUrl` then it won't work because Google doesn't offer An authorization server for personal use.. (outside of GCP).

if you put it (as in this file) under `userAuth -> openId -> baseUrl`, we won't try to hit that well known url, we will, instead, hit https://accounts.google.com/.well-known/openid-configuration which is available...

PR #783, opened 2021-03-01
Fix issues with kustomize templates

jeevb:
This PR reorganizes the kustomize templates, mainly the overlays. This should fix some of the weird namespaces issues we were having with the pytorch and spark operators.


deployment/test/flyte_generated.yaml
```diff
@@ -3,161 +3,1301 @@ kind: Namespace
 metadata:
   name: flyte
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
jeevb, line 6: This diff is large because I consolidated the contour-ingress-controller manifest between `test` and `sandbox`. @kumare3: Do you have context about why these were different to begin with? Should we keep them different?

jeevb:
Still have to work through why the e2e tests are failing...

kumare3:
@jeevb what's the hunch here - why do you think we're we getting the repeated namespace creation / conflicts? Or is there a way to avoid it?

jeevb:
> @jeevb what's the hunch here - why do you think we're we getting the repeated namespace creation / conflicts? Or is there a way to avoid it?

This is mostly because we need to package flyte resources separately (with its namespace override) from addons and operators. This PR bundles the flyte resources as a `base` named `flyte` in each of the overlays.

jeevb:
Test and working in sandbox!

deployment/test/flyte_generated.yaml
kumare3: yes, Sandbox is used by users to try out. test on the other hand has no flyteconsole and is used to run end to end tests.

kumare3 (dismissed):
Thank you for doing this. We might want to fix a few docs - especially the README

deployment/test/flyte_generated.yaml
jeevb: @kumare3: Hmm I see, we should probably tweak the test environment then. We can probably get rid of `contour_ingress_controller` there as well, then?

kumare3: ohh that would be great, because that will speed up the end to end. I think for the test environment eventually I prefer using the sandbox docker image itself as that would be much faster to startup and run tests

jeevb:
> Thank you for doing this. We might want to fix a few docs - especially the README

Should we do the README updates in this PR?

EngHabu (approved):
This looks great! I played with it a bit and while I've not yet attempted to apply EKS module, I think it looks right for it... I'll verify better later this week...

Issue #800, opened 2021-03-05
[Core Feature] Flyte Security Context


EngHabu:
**Motivation: Why do you think this is important?**
Flyte launches many entities through different execution engines, defining what identity each one should impersonate when executing has been one of the confusing aspects of using Flyte. Let's step back, define what the right model is based on what we know today and make sure it works across the stack.

**Goal: What should the final outcome look like, ideally?**

1. Works with Tasks that run on pods, remote containers (e.g. AWS Batch) where you have user's custom code that runs and need an identity to communicate with other resources as well as metadata tasks (e.g. SQL Queries, WebAPI calls... etc.) where the execution of user's code is more controlled.
2. Defaults can be set at Task level but should be overridable at LaunchPlan and Execution levels
3. Support passing secrets, IAM and k8s service accounts.

kumare3:
This is merged and will be available in the next platform release (it is available in the nightly build) - Example: https://github.com/flyteorg/flytesnacks/blob/master/cookbook/core/intermediate/use_secrets.py

PR #857, opened 2021-03-23
Pod Webhook deployment & Doc

EngHabu:
https://github.com/flyteorg/flyte/issues/800

### DIFF
--- a/deployment/eks/flyte_generated.yaml
+++ b/deployment/eks/flyte_generated.yaml
@@ -7654,7 +7654,7 @@ rules:
 - apiGroups:
   - '*'
   resources:
-  - mutatingwebhookconfiguration
+  - mutatingwebhookconfigurations
   - secrets
   - pods
   verbs:
--- a/deployment/eks/flyte_generated.yaml
+++ b/deployment/eks/flyte_generated.yaml
@@ -7654,14 +7654,13 @@ rules:
 - apiGroups:
   - '*'
   resources:
-  - '*'
+  - mutatingwebhookconfiguration
+  - secrets
+  - pods
   verbs:
   - get
-  - list
-  - watch
   - create
   - update
-  - delete
   - patch
 ---
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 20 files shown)


EngHabu:
@SandraGH5 Please review: https://github.com/flyteorg/flyte/pull/857/files?short_path=a45672c#diff-a45672c843e0c8ba29fb8cf8c2658e52f2f7ec2c6b37454badcb64b06eb25f8d

deployment/eks/flyte_generated.yaml
```diff
@@ -7610,6 +7610,12 @@ metadata:
 ---
 apiVersion: v1
 kind: ServiceAccount
+metadata:
```
kumare3, line 7613: cc @anandswaminathan this should not impact Lyft, but this is a new dependency, that if you want to use `secrets` in flytekit declaration then you have to enable

EngHabu:
CC @anandswaminathan Lyft shouldn't be affected but if your users want to start using the secrets stuff in FlyteKit, you will have to deploy the webhook described above

deployment/eks/flyte_generated.yaml
anandswaminathan: ohhh

rsts/howto/secrets.rst
```diff
@@ -0,0 +1,53 @@
...
+   CA Certificate as well as a derived TLS certificate and its private key. It stores them into a new secret ``flyte-pod-webhook-secret``.
+2) flyte-pod-webhook Deployment: This deployment creates the Webhook pod which creates a MutatingWebhookConfiguration
+   on startup. This serves as the registration contract with the ApiServer to know about the Webhook before it starts serving
+   traffic.
+
+*******************
+Scaling the Webhook
+*******************
+
+The Webhook does not make any external API Requests in response to Pod mutation requests. It should be able to handle traffic
```
kumare3, line 50: we will need to update this to say that you will most probably need to scale this for sufficiently large workloads.
Also that this is optional!!

SandraGH5: How about this:

Scaling the Webhook

Note: This is optional but highly recommended for sufficiently large workloads. (Do we have a measurement to include here to identify "large" ? )

The Webhook does not make any external API Requests in response to Pod mutation requests. It should be able to handle traffic...

EngHabu: Updated

kumare3 (approved):

PR #884, opened 2021-03-31
Update index.rst

SandraGH5:
Add Flyte users to Who's Using Flyte and Meet Flyte.

### DIFF
--- a/rsts/index.rst
+++ b/rsts/index.rst
@@ -43,7 +43,7 @@ Meet Flyte
 
 Flyte is an open-source, container-native, structured programming and distributed processing platform. It enables highly concurrent, scalable and maintainable workflows for machine learning and data processing.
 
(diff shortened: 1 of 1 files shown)


rsts/index.rst
```diff
@@ -94,6 +94,8 @@ Who's Using Flyte?
 
 At `Lyft <https://eng.lyft.com/introducing-flyte-cloud-native-machine-learning-and-data-processing-platform-fb2bb3046a59>`__, Flyte has served production model training and data processing for over four years, becoming the de-facto platform for the Pricing, Locations, ETA, Mapping teams, Growth, Autonomous and other teams.
 
+Flyte is also used in production by `Spotify <https://www.spotify.com/>`__, `Freenome <https://www.freenome.com/>`__, `USU Solutions <https://www.usu.com/>`__ and `Striveworks <https://striveworks.us/>`__.
```
kumare3, line 97: i dont know if we should add this, I think all this information should be in one place (maybe the README?) and we can just link to it.
Saying- 
Here is a list of users using Flyte in production

SandraGH5: Please check update below.

kumare3 (approved):

PR #905, opened 2021-04-09
Add init container that will pre-create buckets in minio

jeevb:


wild-endeavor (dismissed):

katrogan (approved):

katrogan (approved):

kustomize/base/addons/storage/kustomization.yaml
```diff
@@ -1,2 +1,6 @@
 resources:
 - storage.yaml
+
+images:
+  - name: minio/minio
+    newTag: RELEASE.2021-04-06T23-11-00Z
```
EngHabu, line 6: Can you move these lines to [this file](https://github.com/flyteorg/flyte/blob/master/kustomize/overlays/sandbox/kustomization.yaml#L24-L48).

EngHabu (commented):
Awesome! Thank you for doing that

kustomize/base/addons/storage/kustomization.yaml
jeevb: Done.

EngHabu (approved):
Great!

Issue #299, opened 2020-04-28
[Feature] Convert Flyte deployment from Kustomize to Helm!


wild-endeavor:
**Motivation: Why do you think this is important?**
https://github.com/lyft/flyte/pull/298 is incomplete.  Finish it.

**Goal: What should the final outcome look like, ideally?**
In no particular order:
- [ ] Documentation
- [ ] Create a separate RDS instance for data catalog, or figure out how to add a second database in the Admin RDS instance.
- [ ] Test that things actually work with gRPC clients through the ALB and fix if they don't.
- [ ] Refactor the terraform around the IAM webhook bits into a separate submodule
- [ ] Figure out a nicer way to get the output of terraform into the kustomize overlays.
- [ ] Figure out a nice or at least easier way for users to plug in their own information like aws account number into the kustomize overlays.
- [ ] Move the db password into a TF input variable.
- [ ] Create node groups and separate out the Flyte components to run on a separate set of servers.
- [ ] Create sample service accounts for users to allow the webhook to work (service account needs to be patched) and see if it's possible to get Admin to run that patch as part of the cluster resource manager.
- [ ] Figure out and document how to get the K8s dashboard running.

**Describe alternatives you've considered**
NA

**Flyte component**
- [x] Overall
- [x] Flyte Setup and Installation scripts
- [x] Flyte Documentation
- [ ] Flyte communication (slack/email etc)
- [ ] FlytePropeller
- [ ] FlyteIDL (Flyte specification language)
- [ ] Flytekit (Python SDK)
- [ ] FlyteAdmin (Control Plane service)
- [ ] FlytePlugins
- [ ] DataCatalog 
- [ ] FlyteStdlib (common libraries)
- [ ] FlyteConsole (UI)
- [ ] Other

**Additional context**
NA
**Is this a blocker for you to adopt Flyte**
NA

rstanevich:
discussed with @wild-endeavor in Slack

**_- Test that things actually work with gRPC clients through the ALB and fix if they don’t._**

From our experience with AWS - `gRPC` unfortunately doesn’t work with `ALB`. ALB will handle HTTP traffic, but it doesn't allow even register workflow in Flyte.
So, a few months we used `Contour Ingress` (like in flyte sandbox) attached to ELB only for Flyte.
And recently we connected Flyte ingress to cluster `Istio IngressGateway` attached to ELB.
As far as I know it should work correctly with `Nginx Ingress` too.

[...]

PR #916, opened 2021-04-14
Introduce Flyte Helm Chart

kumare3:
Issue: https://github.com/flyteorg/flyte/issues/299

Remaining Work:
- [ ] Validate and update values-gcp.yaml
- [ ] Publish Helm Chart in github actions
- [ ] Versioning
- [ ] Switch endtoend test to use helm sandbox
- [ ] Switch sandbox:dind image (used for flytesnacks getting started) to use helm sandbox..


helm/values.yaml
```diff
@@ -0,0 +1,715 @@
...
+          - spark.blacklist.timeout: "5m"
+          - spark.task.maxfailures: "8"
+
+  # Tasks logs plugin configuration
+  task_logs:
+    plugins:
+      logs:
+        kubernetes-enabled: false
+
+  # -- Template for namespaces resources
```
kumare3, line 656: should we separate all of these out into a separate section for
cluster-resource-controller config?

EngHabu (approved):

PR #921, opened 2021-04-17
[wip] Helm components

kumare3:
@sbrunk I have changed the config to match the dependencies based model. Let me know what you think.
I am still testing it.


sbrunk:
Didn't have the time to actually try it but it looks like a much better model in terms of coupling and coherence of components so :+1:

helm/templates/admin/deployment.yaml
```diff
@@ -115,9 +117,11 @@ spec:
       - configMap:
           name: flyte-admin-config
         name: config-volume
+      {{- if .Values.cluster_resource_manager.enabled }}
```
sbrunk, line 120: We should think about naming conventions at some point. Right now we have camel case and mixed.

sbrunk:
LGTM

PR #929, opened 2021-04-21
Add dynamic running state to node state machine

katrogan:


kumare3 (dismissed):

rsts/dive_deep/state_machine.rst
```diff
@@ -50,7 +50,7 @@ Nodes can be of different types, as follows, but all the nodes traverse through
 #. :std:ref:`gen/pb-protodoc/flyteidl/core/workflow.proto:flyteidl.core.tasknode`
 #. :std:ref:`gen/pb-protodoc/flyteidl/core/workflow.proto:flyteidl.core.branchnode`
 #. :std:ref:`gen/pb-protodoc/flyteidl/core/workflow.proto:flyteidl.core.workflownode`
-#. Dynamic node - which is just a task node that does not return outputs, but futures.
+#. Dynamic node - which is just a task node that does not return outputs, but futures. When a dynamic node is running the status trasitions to `DYNAMIC_RUNNING`.
```
kumare3, line 53: i think we should explain what a dynamic node running means here.
How about,
```suggestion
#. Dynamic node - which is just a task node that does not return outputs, but a dynamic workflow. When the task runs, it stays in a `RUNNING` state. Once the task completes and Flyte starts executing the dynamic workflow, the overarching node that contains both the original task and the dynamic workflow enters `DYNAMIC_RUNNING` state.
```

kumare3 (approved):

PR #937, opened 2021-04-24
Helm eks

kumare3:


helm/templates/common/ingress.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.common.ingress.enabled }}
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
```
sbrunk, line 2: `networking.k8s.io/v1` is only available in K8S 1.19+ and GKE is still on 1.17/1.18 (stable/regular channel). That's why I had reverted back to `v1beta1` in
https://github.com/flyteorg/flyte/pull/916/commits/f35a37486ec2026d550d0f253c2b330684a17a53

kumare3: aah ok

kumare3: it seems they were supposed to release 1.19 April / 18. seems slipping?

sbrunk: Oh you're right [1.19 is GA in GKE](https://cloud.google.com/kubernetes-engine/docs/release-notes#119_ga) since two weeks.

helm/values-eks.yaml
```diff
@@ -197,16 +210,16 @@ storage:
...
+    region: us-east-2
 
 db:
   database:
     port: 5432
-    username: postgres
-    host: <DB-NAME>
-    dbname: flyte-development
+    username: flyte
+    host: flyte-helm-test.cluster-cguibupmpitq.us-east-2.rds.amazonaws.com
```
sbrunk, line 221: Is this on purpose?

sbrunk:
Rebased against helm-working-branch locally and merged.

EngHabu:
Closing based on @sbrunk and Jd comments

PR #987, opened 2021-05-06
Include troubleshooting guide in the docs

samhita-alla:
Signed-off-by: Samhita Alla <aallasamhita@gmail.com>

* Added possible workarounds if an error pops up with the `make start` command
* Added `make register` besides `make fast_register` to specify its usage

### DIFF
--- a/rsts/getting_started.rst
+++ b/rsts/getting_started.rst
@@ -28,28 +28,6 @@ Steps
 
   make start
 
-.. note::
-  - ``make start`` usually gets completed within five minutes (could take longer if you aren't in the United States).
-  - If ``make start`` results in a timeout issue:
-     .. code-block:: bash
-
-       Starting Flyte sandbox
-       Waiting for Flyte to become ready...
-       Error from server (NotFound): deployments.apps "datacatalog" not found
-       Error from server (NotFound): deployments.apps "flyteadmin" not found
-       Error from server (NotFound): deployments.apps "flyteconsole" not found
-       Error from server (NotFound): deployments.apps "flytepropeller" not found
-       Timed out while waiting for the Flyte deployment to start
-     
-     You can run ``make teardown`` followed by the ``make start`` command.
-  - If the ``make start`` command isn't proceeding by any chance, check the pods' statuses -- run the command ``docker exec flyte-sandbox kubectl get po -A``.
-  - If you think a pod's crashing by any chance, describe the pod by running the command ``docker exec flyte-sandbox kubectl describe po <pod-name> -n flyte``. This gives a detailed overview of the pod's status.
-  - If Kubernetes reports a disk pressure issue:
-  
-    - Check the memory stats of the docker container using the command ``docker exec flyte-sandbox df -h``.
-    - Prune the images and volumes. 
-    - Given there's less than 10% free disk space, Kubernetes by default throws the disk pressure error.
-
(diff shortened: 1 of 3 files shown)


rsts/getting_started.rst
```diff
@@ -80,6 +101,9 @@ Steps
 
   REGISTRY=ghcr.io/flyteorg make fast_register
 
+.. note::
+   If the images are to be re-built, run ``make register`` command.
```
pmahindrakar-oss, line 105: register doesn't rebuild the images but just registers the serialized tasks and workflows.
May be you meant reregister here.
Or if you mean rebuilding the cookbook image then that should automatically happen if there are new commits and cookbook:<core | sage_maker|... > sha doesn't exist. Otherwise they can always delete the existing images to rebuild

rsts/getting_started.rst
```diff
@@ -28,6 +28,27 @@ Steps
...
+       Error from server (NotFound): deployments.apps "flyteconsole" not found
+       Error from server (NotFound): deployments.apps "flytepropeller" not found
+       Timed out while waiting for the Flyte deployment to start
+     
+     You can run ``make teardown`` followed by the ``make start`` command.
+  - If the ``make start`` command isn't proceeding by any chance, check the pods' statuses -- run the command ``docker exec flyte-sandbox kubectl get po -A``.
+  - If you think a pod's crashing by any chance, describe the pod by running the command ``docker exec flyte-sandbox kubectl describe po <pod-name> -n flyte``. This gives a detailed overview of the pod's status.
+  - If Kubernetes reports a disk pressure issue:
+  
+    - Check the memory stats of the docker container using the command ``docker exec flyte-sandbox df -h``.
```
pmahindrakar-oss, line 49: Probably also mention if < 10% disk free with default setting could cause this issue

rsts/getting_started.rst
samhita-alla: No, `register` does rebuild the cookbook image,`fast_register` doesn't. If you change anything that requires rebuilding the image (to pull the requirements.txt, say), we need to run `make register`.

rsts/getting_started.rst
samhita-alla: Added.

rsts/getting_started.rst
pmahindrakar-oss: Ok . I think i misunderstood this. Thanks for clarifying.

pmahindrakar-oss (approved):

kumare3:
Thank you for doing this, but I think we should point to a separate page, keep this page clean. We may have more than one thing

samhita-alla:
> Thank you for doing this, but I think we should point to a separate page, keep this page clean. We may have more than one thing

That's right. I thought about the same. Do you want me to include this on a page that we can call "Troubleshooting Guide"? I can show this in the main sidebar section if that's okay.

samhita-alla:
> > Thank you for doing this, but I think we should point to a separate page, keep this page clean. We may have more than one thing
> 
> That's right. I thought about the same. Do you want me to include this on a page that we can call "Troubleshooting Guide"? I can show this in the main sidebar section if that's okay.

@kumare3 Moved the section to a separate troubleshooting page.

pmahindrakar-oss (approved):

PR #988, opened 2021-05-06
Use beta version of ingress

juandiegopalomino:
"Hey so, I noticed that you are using the Ingress class out of beta, which is kind of an issue-- the Ingress class was moved out of beta in the 1.19 release, which happened in late august. If flyte is run in clusters older than this (e.g. 1.18), then this will fail. EKS only started offering 1.19 a couple of months ago, and the GKE regular channel defaults to 1.18, so this really won't work for most of your users. Do you guys mind if I file a pr to move back to the beta ingress version? I don't see you guys using anyhting required from the GA version, and beta has been serving everyone well for the past 6 years"


EngHabu:
Thank you! Mind fixing DCO check?

EngHabu (approved):

PR #1056, opened 2021-05-26
Helm template upgrade to auto-generate co-pilot configuration

kumare3:
Auto-generates a new cluster configuration template for copilot that has the storage configuration.

Signed-off-by: Ketan Umare <ketan.umare@gmail.com>

### DIFF
--- a/helm/templates/_helpers.tpl
+++ b/helm/templates/_helpers.tpl
@@ -164,7 +164,7 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end }}
 {{- end }}
 
-{{- define "storage" -}}
+{{- define "storage.base" -}}
 storage:
 {{- if eq .Values.storage.type "s3" }}
   type: s3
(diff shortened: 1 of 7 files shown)


helm/templates/_helpers.tpl
```diff
@@ -198,4 +198,46 @@ storage:
...
+{{- define "copilot.config" -}}
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: flyte-data-config
+  namespace: {{`{{ namespace }}`}}
+data:
+  config.yaml: |
+    storage:
+{{- if eq .Values.storage.type "s3" }}
```
EngHabu, line 212: Can't we do `{{ include storage }}` here?

sbrunk: Agreed if that's possible we should avoid the duplication here.

kumare3: Hmm let me try

kumare3: Done!

EngHabu (approved):

PR #1088, opened 2021-06-03
Helm automation Added

evalsocket:
1. Helm auto release added in release automation. 
2. Started bumping new version  of flyte component in helm `values.yaml`
3. Generated deployment file for both helm and kustomize

Note it is the part of same workflow that generate kustomize output and create pr


helm/.helm-template.yaml
```diff
@@ -0,0 +1,841 @@
+# ---------------------------------------------------------------------
```
EngHabu, line 1: Can you name it `.values-template.yaml` or something to indicate it's used to generate the values.yaml?

evalsocket: Done

EngHabu (approved):

PR #1112, opened 2021-06-09
Use cr.flyte.org to fetch flyte containers

aviaviavi:
This change updates ghcr.io container references to instead use flyte's new cr.flyte.org domain. All generated yaml files have been updated as well. Please let me know if I missed anything, changed something that shouldn't be, etc. Thanks!

### DIFF
--- a/deployment/test/flyte_generated.yaml
+++ b/deployment/test/flyte_generated.yaml
@@ -393,7 +393,7 @@ data:
       k8s:
         co-pilot:
           name: "flyte-copilot-"
-          image: "ghcr.io/flyteorg/flytecopilot:v0.5.28"
+          image: "cr.flyte.org/flyteorg/flytecopilot:v0.5.28"
           start-timeout: "30s"
   core.yaml: |
     propeller:
@@ -474,7 +474,7 @@ data:
 
 kind: ConfigMap
 metadata:
-  name: flyte-propeller-config-dtdt5gbgg2
+  name: flyte-propeller-config-d9cm925f8m
   namespace: flyte
 ---
 apiVersion: v1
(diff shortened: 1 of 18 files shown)


script/generate_helm.sh
```diff
@@ -21,7 +21,7 @@ then
     GO111MODULE=on go get github.com/norwoodj/helm-docs/cmd/helm-docs
 fi
 
-~/go/bin/helm-docs -t ${DIR}/../helm/README.md.gotmpl ${DIR}/../helm/
+${GOPATH}/bin/helm-docs -t ${DIR}/../helm/README.md.gotmpl ${DIR}/../helm/
```
aviaviavi, line 24: I ran into issues while running `make helm` as my GOPATH is somewhere else, so I bundled this change which seems like an improvement but let me know if you'd prefer this be reverted or put in its own PR.

deployment/test/flyte_generated.yaml
```diff
@@ -474,7 +474,7 @@ data:
 
 kind: ConfigMap
 metadata:
-  name: flyte-propeller-config-dtdt5gbgg2
+  name: flyte-propeller-config-d9cm925f8m
```
aviaviavi, line 477: I think these sums are supposed to get updated when generating these files, but let me know if this looks incorrect

EngHabu:
Great! Thank you for taking care of this. 
Mind clicking on "Details" for DCO action and applying the actions it recommends to sign your commits?

aviaviavi:
On it!

EngHabu (approved):
Thank you for doing this

EngHabu (approved):

PR #1145, opened 2021-06-15
Use HTTP instead of HTTPS for flytectl callback URL 

sbrunk:
As expected by `flytectl`. HTTPS doesn't make sense for localhost anyway.


sbrunk:
Helm `3.5.4` which I had installed locally seems to format some lines differently than `3.6.0`. That caused the delta check to fail. Something to be aware of.

EngHabu (approved):

Issue #1111, opened 2021-06-09
[Housekeeping] Publish Helm Chart


EngHabu:
**Describe the Issue**
We can use an action like this: https://github.com/marketplace/actions/helm-publish
to publish helm charts on release in flyte repo

evalsocket:
Should we use the same branch https://github.com/flyteorg/flyte/tree/gh-pages that we are using for documentation ?

EngHabu:
Can you clarify what's the branch is needed for?

evalsocket:
in helm we can publish our chart artifacts in `gh-pages`, gh-pages URL is publically accessible and user can use it as a registry. I found some documentation code in `gh-pages`, It's not a problem we can use the same branch for docs/helm. 

```
$ helm repo add https://flyteorg.github.com/flyte
$ helm install flyte 
```

Or else move helm out of flyte repo. 

```
$ helm repo add https://flyteorg.github.com/helm
$ helm install flyte 
```

kumare3:
@evalsocket / @EngHabu can we make a new domain helm.Flyte.org

evalsocket:
@kumare3 yes we can but first we need our artifacts in `gh-pages`

sbrunk:
There's also another official action for publishing charts: https://github.com/marketplace/actions/helm-chart-releaser

Once the publishing infra is in place we should also add the chart to [Artifact Hub](https://artifacthub.io/) ([Docs](https://artifacthub.io/docs/topics/repositories/)).

evalsocket:
@sbrunk yes I used the same GitHub action in my PR. It has one open issue related to dir structure. It needs `charts/flyte`.

sbrunk:
Ah I didn't see #1147 😄

sbrunk:
> @sbrunk yes I used the same GitHub action in my PR. It has one open issue related to dir structure. It needs `charts/flyte`.

So @evalsocket in order to use the action we'd need to move from

```
├── ...
├── helm
├── ...
```
to
```
├── ...
├── charts
│   └── flyte
├── ...
```
or, I guess since we can set `charts_dir`
```
├── ...
├── helm
│   └── flyte
├── ...
```
would also work?

evalsocket:
@sbrunk i used `helm/dir`

PR #1147, opened 2021-06-15
Added release steps for publishing helm artifacts in gh-pages

evalsocket:
- Added GitHub workflow for publishing helm artifact in gh-pages. 

Fix: https://github.com/flyteorg/flyte/issues/1111

Tested result : https://github.com/evalsocket/flyte/releases/tag/v0.1.222
Index of registry : https://github.com/evalsocket/flyte/blob/gh-pages/index.yaml
Github Action Test: https://github.com/evalsocket/flyte/runs/3002001581?check_suite_focus=true

Test Helm 
```bash
$ helm repo add flyte https://evalsocket.github.io/flyte
$ helm repo update
$ helm  install sandbox flyte/flyte
```


evalsocket:
Some notes:
1. https://github.com/helm/charts-repo-actions-demo/blob/main/.github/workflows/release.yaml is the demo of the helm github action but it only works if you run It on every push on master because in that case it creates a new release and publishes helm package. But in our case, we will publish it in release and then it failed (https://github.com/helm/chart-releaser-action/issues?q=No+changes+detected)

.goreleaser.yml
```diff
@@ -11,4 +11,5 @@ release:
   name_template: "Flyte v{{.Version}} milestone release"
   disable: false
   extra_files:
-    - glob: ./release/*
\ No newline at end of file
+    - glob: ./release/*
+    - glob: .cr-release-packages/
```
EngHabu, line 15: I don't see us copying anything into this... or is that something `cr helm...` expects/populates?

script/prepare_artifacts.sh
```diff
@@ -1,11 +1,31 @@
...
+
+# Download helm chart releaser
+wget -q -O /tmp/chart-releaser.tar.gz https://github.com/helm/chart-releaser/releases/download/v1.2.1/chart-releaser_1.2.1_linux_amd64.tar.gz 
+mkdir -p bin
+tar -xf /tmp/chart-releaser.tar.gz -C bin
+chmod +x bin/cr
+rm /tmp/chart-releaser.tar.gz \
+
+# Clean Git History
+git stash
```
EngHabu, line 25: Why is this needed?

script/prepare_artifacts.sh
```diff
@@ -1,11 +1,31 @@
...
+
+# Package helm chart
+bin/cr package helm
+
+# Commit Chart registry to github gh-pages
+bin/cr index --owner flyteorg --git-repo flyte --charts-repo="https://flyteorg.github.io/flyte" --push --token=${FLYTE_BOT_PAT} --release-name-template="{{ .Version }}"
```
EngHabu, line 31: Where does it get the `Version` value from? I think they come from here: https://github.com/flyteorg/flyte/blob/master/helm/Chart.yaml#L5-L6 

Will these values be updated as part of this: https://github.com/flyteorg/flyte/actions/workflows/generate-flyte-manifest.yml ?

.github/workflows/create-release.yml
```diff
@@ -16,6 +16,17 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: "0"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.4.0
+            
+      - name: Setup Git
```
EngHabu, line 25: nit:
```suggestion
      - name: Configure Git
```

EngHabu:
Looks so close! thank you for working on this

.goreleaser.yml
evalsocket: yes, they populate the tar of helm package

script/prepare_artifacts.sh
evalsocket: It can be removed because i added everything in gitignore

script/prepare_artifacts.sh
evalsocket: yes, they come from the chart.yaml

evalsocket: Yes, they are coming from chart.yaml and we can't add them in the above workflow because we need to add a version that we will create, and that workflow doesn't know that.


We need to update the value manually

EngHabu: Maybe what we should do is have a placeholder `#VERSION` in the chart.yaml... and we populate that here, temporarily, using the version passed in to the current workflow. We package the chart and push it to the gh-pages branch then we `git stash` that change.... how about that?

evalsocket: yes that will work

evalsocket: done

evalsocket:
@EngHabu do we need to be generated helm?

EngHabu:
Do you mean `do we need to run make helm in this WF?` or `do we still need make helm at all?`
It runs as part of [this WF](https://github.com/flyteorg/flyte/blob/master/.github/workflows/generate-flyte-manifest.yml#L18) Which I believe is the right thing to do

EngHabu (approved):

PR #1174, opened 2021-06-24
Make initial projects customizable

Dread1982:
This change allows customizing the projects created on helm install. Defaults to the previously hardcoded list of ["flytesnacks","flytetester","flyteexamples"].


sbrunk (approved):
Looks good to me!

PR #1182, opened 2021-06-27
Added conditional annotation for grpc to use HTTP2 as target group

pmahindrakar-oss:

This arn points to self signed cert generated for flyte.example.org
arn:aws:acm:us-east-2:832573439839:certificate/577f9991-a6cf-4d1c-b56d-447b24077f46

User can directly hit the ingress https endpoint to browse UI console and also hit grpc to query through flytectl

User has to ignore the warning when browsing the UI console which is expected and new flag in flytectl has been added to do the same skip. This won't be the case in production flyte deployment where a CA verified cert will be used.

### DIFF
--- a/helm/manifest.yaml
+++ b/helm/manifest.yaml
@@ -1945,8 +1945,7 @@ metadata:
   name: flyte
   namespace: flyte
   annotations: 
-    alb.ingress.kubernetes.io/actions.ssl-redirect: '{"Type": "redirect", "RedirectConfig":
-      { "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'
+    alb.ingress.kubernetes.io/actions.ssl-redirect: '{"Type": "redirect", "RedirectConfig": { "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'
     alb.ingress.kubernetes.io/certificate-arn: arn:aws:acm:us-east-2:832573439839:certificate/a7db79d4-4f8c-4c7b-a4f4-109bf2fb5efa
     alb.ingress.kubernetes.io/group.name: flyte
     alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS":443}]'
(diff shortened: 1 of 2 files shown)


helm/values-eks.yaml
```diff
@@ -172,11 +172,13 @@ common:
       kubernetes.io/ingress.class: alb
       alb.ingress.kubernetes.io/tags: service_instance=production
       alb.ingress.kubernetes.io/scheme: internet-facing
-      alb.ingress.kubernetes.io/certificate-arn: arn:aws:acm:us-east-2:832573439839:certificate/a7db79d4-4f8c-4c7b-a4f4-109bf2fb5efa
+      alb.ingress.kubernetes.io/certificate-arn: arn:aws:acm:us-east-2:832573439839:certificate/577f9991-a6cf-4d1c-b56d-447b24077f46
```
kumare3, line 175: This is just the arn? Can we as some comments here? As this converts to docs

pmahindrakar-oss: Sure. Done

helm/manifest.yaml
```diff
@@ -2063,16 +2070,25 @@ metadata:
...
     alb.ingress.kubernetes.io/scheme: internet-facing
     alb.ingress.kubernetes.io/tags: service_instance=production
     kubernetes.io/ingress.class: alb
+    alb.ingress.kubernetes.io/backend-protocol-version: HTTP2
     nginx.ingress.kubernetes.io/backend-protocol: GRPC
 spec:
   rules:
     - http:
         paths:
+          #
```
wild-endeavor, line 2086: what was this for?

pmahindrakar-oss: This is coming from the revert, but we can regenerate this file if needed in your branch or delete it since its not used directly for the installation.

pmahindrakar-oss: Specifically this is coming from
```
   # {{- if .Values.common.ingress.albSSLRedirect }}
          # - backend:
          #     serviceName: ssl-redirect
          #     servicePort: use-annotation
          #   path: /*
          #   pathType: ImplementationSpecific
          # {{- end }}
```
From templates/common/ingress.yaml as we have now enabled albSSLRedirect  with self signed certs

PR #1183, opened 2021-06-28
Add example code for Java and Scala in the "Meet Flyte" page

mayitbeegh:
The Java/Scala examples contain significantly more code than Python. I'm debating on whether I should add all the comments because they're hard to understand without them. Here's a comparison:
Without comments:
image
With comments:
image
The comments do make the supposedly introductory examples appear more intimidating.


rsts/index.rst
```diff
@@ -54,27 +54,208 @@ Created at `Lyft <https://www.lyft.com/>`__ in collaboration with Spotify, Freen
...
+                // uses the workflow input as the task input of the GreetTask
+                val greeting = builder.apply("greet", GreetTask(name)).getOutput("greeting")
+
+                // uses the output of the GreetTask as the task input of the AddQuestionTask
+                val greetingWithQuestion = builder.apply("add-question", AddQuestionTask(greeting)).getOutput("greeting")
+
+                // uses the task output of the AddQuestionTask as the output of the workflow
+                builder.output("greeting", greetingWithQuestion, "Welcome message")
+              }
+            }
```
EngHabu, line 257: Is there something similar to calling it like in python? https://github.com/flyteorg/flyte/pull/1183/files#diff-9a1147ef4c0aeaf0680900dbd2902c51d2a5475c0c21a606fbe1c9420c4261edR80-R81

```
           welcome("Traveler")
           # Output: "Welcome, Traveler! How are you?"
```

rsts/index.rst
```diff
@@ -54,27 +54,208 @@ Created at `Lyft <https://www.lyft.com/>`__ in collaboration with Spotify, Freen
...
 
-   @task
-   def add_question(greeting: str) -> str:
-       return f"{greeting} How are you?"
-   
-   @workflow
-   def welcome(name: str) -> str:
-       greeting = greet(name=name)
-       return add_question(greeting=greeting)
+           @task
```
EngHabu, line 63: While we are at it, can we also add comments to python 😬 ?

rsts/index.rst
```diff
@@ -54,27 +54,208 @@ Created at `Lyft <https://www.lyft.com/>`__ in collaboration with Spotify, Freen
...
 
-.. code:: python
-   
-   @task
-   def greet(name: str) -> str:
-       return f"Welcome, {name}!"
+.. tabs::
 
-You can compose one or more tasks to create a ``workflow``:
+    .. tab:: python
```
EngHabu, line 59: Is there an obvious link in the page that takes you to the rest of the python examples? if not, let's add one, if yes, then let's add a link to the "java examples page" (if that's a thing) as well...

Or maybe they should be at the end of teach of the tabs?

rsts/index.rst
kumare3: @EngHabu I am actually thinking of removing comments from Java. To make it more concise, but just use good code. But the actual example should have comments - and we should link to the actual example

EngHabu:
Can we simplify the comments to explain "what and why things need to be written the way it's for Flyte" as opposed to adding "java docs" everywhere?

rsts/index.rst
EngHabu: Can we simplify the comments to explain "what and why things need to be written the way it's for Flyte" as opposed to adding "java docs" everywhere?

rsts/index.rst
mayitbeegh: Yes, like a main() method

rsts/index.rst
mayitbeegh: I don't think there is.

rsts/index.rst
mayitbeegh: @EngHabu Most javadocs are not related to Flyte concepts, we can remove all javadocs and keep the comments in the workflow file.

EngHabu: excellent!

rsts/index.rst
mayitbeegh: Turns out that although it's easy to execute a task by itself, it's not trivial to execute a workflow similar to Python.

EngHabu (approved):

PR #1224, opened 2021-07-08
Added kustomize support in docker sandbox

evalsocket:
Context : In flytectl sandbox, I am introducing a new flag i.e. `--kustomize`, But for that container should build manifest before deployment in the entry point.`. Currently, Pytest-flyte maintains the logic for kustomize build-in entry point (https://github.com/flyteorg/pytest-flyte/tree/main/src/pytest_flyte/docker). 

- I used the same implementation that pytest has but after these changes flytectl and pytest can share the same dependency.

### DIFF
--- a/docker/sandbox/Dockerfile
+++ b/docker/sandbox/Dockerfile
@@ -78,7 +78,7 @@ COPY --from=base_ /flyteorg/ /flyteorg/
 COPY docker/sandbox/flyte-entrypoint-dind.sh /flyteorg/bin/flyte-entrypoint.sh
 
 # Copy kustomization template 
-COPY docker/sandbox/kustomization.yaml.tmpl /opt/flyteorg/share/deployment/kustomization.yaml
+COPY docker/sandbox/kustomization.yaml  /flyteorg/share/kustomization.yaml
 
 # Update PATH variable
 ENV PATH "/flyteorg/bin:${PATH}"
--- a/docker/sandbox/Dockerfile
+++ b/docker/sandbox/Dockerfile
@@ -22,13 +22,6 @@ ARG K3S_VERSION="v1.21.1%2Bk3s1"
 RUN wget -q -O /flyteorg/bin/k3s https://github.com/k3s-io/k3s/releases/download/${K3S_VERSION}/k3s \
     && chmod +x /flyteorg/bin/k3s
 
-# Install kustomize
-ARG KUSTOMIZE_VERSION="v4.0.5"
-RUN wget -q -O /tmp/kustomize.tar.gz https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2F${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64.tar.gz \
-    && tar -xf /tmp/kustomize.tar.gz -C /flyteorg/bin \
-    && chmod +x /flyteorg/bin/kustomize \
-    && rm /tmp/kustomize.tar.gz
-
 # Install flytectl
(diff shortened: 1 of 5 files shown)


EngHabu:
I don't think pytest *requires* kustomize... it was just whatever was available at the time... Let's hold off this PR until we discuss.

In fact, I would like to understand if there are anything special needed for pytest or whether we can just use sandbox as is... @jeevb @cosmicBboy @wild-endeavor may be able to help clarify that...

jeevb:
The goal was to eventually support setting flyte component versions programmatically in pytest. Could still be useful for that.

@evalsocket Will kubectl apply -k not suffice for this?

evalsocket:
@jeev yes we can use that. I can remove the kustomize install but is it ok if I move the logic from pytest to sandbox ?

jeevb:
> @jeev yes we can use that. I can remove the kustomize install but is it ok if I move the logic from pytest to sandbox ?

Fine by me, but will defer to others.

docker/sandbox/kustomization.yaml.tmpl
```diff
@@ -0,0 +1,9 @@
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+bases:
+  - flyte_generated.yaml
+
+# Image overrides
```
jeevb, line 8: Probably unnecessary since this file is not going to be modified directly right? Is the idea to mount a modified file here to override the default sandbox config?

docker/sandbox/kustomization.yaml.tmpl
```diff
@@ -0,0 +1,9 @@
+
```
jeevb, line 1: Probably worth removing the .tmpl suffix in the filename since you are already doing so with the copy.

docker/sandbox/Dockerfile
```diff
@@ -70,6 +77,9 @@ COPY --from=base_ /flyteorg/ /flyteorg/
 # Copy entrypoints
 COPY docker/sandbox/flyte-entrypoint-dind.sh /flyteorg/bin/flyte-entrypoint.sh
 
+# Copy kustomization template 
+COPY docker/sandbox/kustomization.yaml.tmpl /opt/flyteorg/share/deployment/kustomization.yaml
```
jeevb, line 81: How about copying into /flyteorg/share directly and installing via kubectl apply -k instead of -f?

docker/sandbox/kustomization.yaml.tmpl
evalsocket: yes, I will mount the user-provided kustomize file and the entry point script will handle the logic. By default, it will not override anything.

docker/sandbox/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+
```
jeevb, line 1: Whitespace

jeevb (approved):
LGTM.

docker/sandbox/Dockerfile
```diff
@@ -70,6 +70,9 @@ COPY --from=base_ /flyteorg/ /flyteorg/
 # Copy entrypoints
 COPY docker/sandbox/flyte-entrypoint-dind.sh /flyteorg/bin/flyte-entrypoint.sh
 
+# Copy kustomization template
```
EngHabu, line 73: I would actually prefer to move this to install using helm as we discussed... and maybe enable users to provide a values file to be overridden... basically reducing the places we use kustomize and making sure helm consistently work... I think introducing a kustomization.yaml here makes that job harder....

EngHabu (approved):

PR #1255, opened 2021-07-16
Extracting user input values into separate block for EKS Helm

wild-endeavor:
Collating and putting all the user-supplied information at the top of the `values-eks` file.  Then running the outputs of the values through tpl functions where necessary.  Note that the generated files haven't really changed (except for a postgres host - the gcp command wasn't working initially anyways, and still isn't).


evalsocket (dismissed):

sbrunk (dismissed):
Awesome! I wasn't even aware of the `tpl` function until now.

katrogan (approved):

PR #1259, opened 2021-07-19
Fix kubectl command in sandbox.rst

kanterov:


brucearctor:
LGTM

@samhita-alla and @kumare3 -- I can confirm this solves the issue (so should be OK to merge).  Closing PR #1271 as I tried the same fix.

kumare3 (approved):

PR #1348, opened 2021-08-17
Support test workflow runs on other branches

pingsutw:
Support test workflow runs on other branches, so we could check whether tests are failed before creating a PR.


EngHabu:
I think it's ok to require a PR for tests to run... you can create a PR in draft mode until it's ready for others to review...

PR #1366, opened 2021-08-20
Added helm chart for Flyte core

evalsocket:
- Added helm chart flyte-core

Workflow :- https://github.com/evalsocket/flyte/runs/3381628027?check_suite_focus=true
Package :- https://github.com/evalsocket/flyte/tree/gh-pages / https://github.com/evalsocket/flyte/blob/gh-pages/index.yaml

Test Helm charts:
```
helm repo add flyte https://evalsocket.github.io/flyte
helm install -n flyte -f values-eks.yaml --create-namespace flyte flyte/flyte-core
```

Install contour:
```
helm repo add bitnami https://charts.bitnami.com/bitnami
helm install gateway bitnami/contour -n flyte  
```


charts/flyte-core/templates/redis/service.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if .Values.redis.enabled }}
```
EngHabu, line 1: I see we are still installing redis... you will pull this out, right?

evalsocket: yes I removed the Redis from flyte-core

EngHabu (approved):

PR #1395, opened 2021-08-31
Update getting started guide for flytectl

evalsocket:
- Added flytectl upgrade and config command


rsts/getting_started_scale.rst
```diff
@@ -55,15 +55,19 @@ Install flytectl
...
-        brew upgrade flytectl
+        flytectl upgrade
 
    .. tabbed:: Other Operating systems
 
      .. prompt:: bash $
 
-         curl -s https://raw.githubusercontent.com/lyft/flytectl/master/install.sh | bash
-         export PATH=$(pwd)/bin:$PATH
+         curl -s https://raw.githubusercontent.com/flyteorg/flytectl/master/install.sh | sudo bash -s -- -b /usr/local/bin
```
pmahindrakar-oss, line 64: Can we also add details on if users wants to change from /usr/local/bin then he can just change that to new path.

And also if it a path which is not available on PATH variable then its better to keep that  command

kumare3 (commented):
Nice.

rsts/getting_started_scale.rst
evalsocket: done

pmahindrakar-oss (dismissed):

pmahindrakar-oss (approved):

PR #1398, opened 2021-08-31
helm cleanup for flyte-core

evalsocket:
- Added Redis config in `userSetting` of eks
- Remove unwanted plugins in core


EngHabu (approved):

charts/flyte-core/values-eks.yaml
```diff
@@ -276,6 +279,18 @@ configmap:
           rate: 100
           capacity: 1000
 
+  # -- Resource manager configuration
+  resource_manager:
+    # -- resource manager configuration
+    propeller:
+      resourcemanager:
+        type:
```
EngHabu, line 287: Why is type empty?

EngHabu (changes requested):

charts/flyte-core/values-eks.yaml
evalsocket: it should be redis

pmahindrakar-oss (dismissed):

charts/flyte-core/values.yaml
```diff
@@ -487,8 +487,7 @@ configmap:
     # -- resource manager configuration
     propeller:
       resourcemanager:
-        type:
-
+        type: redis
```
EngHabu, line 490: I think the default one shouldn't rely on redis... it should just be `noop`
https://github.com/flyteorg/flytepropeller/blob/eaf084934de5d630cd4c11aae15ecae780cc787e/pkg/controller/nodes/task/resourcemanager/config/config.go#L14

PR #1449, opened 2021-09-08
Refactor helm for database config 

evalsocket:
- Added different values for datacatlog & fkyteadmin. Now they have different values in values.yaml 
- Disable postgres/minio/contour from GCP.


wild-endeavor (approved):
so opta will be forced to use the same database table for now.  I think we should fix this too before the 0.18 milestone release.

PR #1462, opened 2021-09-10
[RR] RFC - flytekit learn

cosmicBboy:
This RFC proposes a new flytekit plugin called `flytekit-learn` that allows users to write model training, hyperparameter optimization, and prediction logic in python code that's automatically compiled into flytekit tasks and workflows.


SandraGH5 (approved):
I've just capitalized Flyte, Flytekit, and the first word of bulleted points.

PR #1470, opened 2021-09-13
Add Admin volumes to helm deployment

wild-endeavor:

This is useful to inject additional secrets which may already exist on the cluster, but for which we don't have the plaintext at helm installation time so we can't use https://github.com/flyteorg/flyte/pull/1087/


evalsocket (dismissed):
Can you also replicate these changes for flyte-core ?

evalsocket (approved):

PR #1485, opened 2021-09-15
Flyte should use official helm chart

juandiegopalomino:


NitinAgg (commented):
🎉

opta/flyte.yaml
```diff
@@ -43,7 +43,8 @@ modules:
     links:
       - s3: [ "write" ]
   - type: helm-chart
-    chart: "../helm" # NOTE: relative path to chart
+    repository: https://flyteorg.github.io/flyte
+    chart: flyte-core
```
EngHabu, line 47: flyte
```suggestion
    chart: flyte
```

juandiegopalomino: Not according to your docs: https://artifacthub.io/packages/helm/flyte/flyte

EngHabu: Thank you... will fix the docs..

opta/flyte.yaml
```diff
@@ -43,7 +43,9 @@ modules:
     links:
       - s3: [ "write" ]
   - type: helm-chart
-    chart: "../helm" # NOTE: relative path to chart
+    repository: https://flyteorg.github.io/flyte
+    chart: flyte
+    chart_version: "0.17.3-b2"
```
EngHabu, line 48: I think we may keep this PR open until we release the chart... with all the changes we need... ?
or else, you can pin it to the one currently "released":
```suggestion
    chart_version: "v0.17.2"
```

And we can open a new PR when a new one is released...

In fact, @evalsocket, can we have this line get updated in a github WF?
When a release is created, a github WF can get triggered to update this file and open a PR for us to merge... ?

evalsocket: Yes but why can't we use latest tag for deployment. Helm will install latest if version is not specified

EngHabu:
@juandiegopalomino is it possible to reference a remote `values-eks.yaml` file? or does that has to be local?

evalsocket:
I am closing this because we removed the opta config from the flyte https://github.com/flyteorg/flyte/issues/2740

PR #1535, opened 2021-09-30
Add wolt in deployments

stephen37:
We have it deployed in our cluster now so I guess it's time to add our name to the list 😄


kumare3:
@stephen37 +1 but needs a DCO sign-off!

PR #1565, opened 2021-10-05
ECR

EngHabu:

### DIFF
--- a/charts/flyte-core/templates/admin/deployment.yaml
+++ b/charts/flyte-core/templates/admin/deployment.yaml
@@ -20,7 +20,7 @@ spec:
       initContainers:
         {{- if .Values.postgres.enabled }}
         - name: check-db-ready
-          image: postgres:10.16-alpine
+          image: public.ecr.aws/ubuntu/postgres:13-21.04_beta
           command:
           - sh
           - -c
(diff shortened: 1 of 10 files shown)


charts/flyte-core/values.yaml
```diff
@@ -60,6 +59,8 @@ flyteadmin:
   secrets: {}
   additionalVolumes: []
   additionalVolumeMounts: []
+  # -- Deploys a Redoc container in Flyteadmin's pod
+  deployRedoc: true
```
katrogan, line 63: should we maybe always have this set to false and call out in the docs how and why you would want to enable this?

charts/flyte-core/templates/admin/deployment.yaml
```diff
@@ -20,7 +20,7 @@ spec:
       initContainers:
         {{- if .Values.postgres.enabled }}
         - name: check-db-ready
-          image: postgres:10.16-alpine
+          image: cr.flyte.org/ubuntu/postgres:13-21.04_beta
```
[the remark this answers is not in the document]
EngHabu: These all reference already uploaded digger images on ecr. Our donation only redirects to that..

charts/flyte-core/values.yaml
EngHabu: Good point about the docs. I'll do that... I only kept it on for backward compatibility...

Issue #1538, opened 2021-09-30
[Housekeeping] Helm charts for control/data plane deployment


evalsocket:
**Describe the Issue**
Idea is to create a separate helm chart for control plane & data plane deployment .

**What if we do not do this?**
Short description of potential consequences

***Related component***
Either Specific / all

Issue #1537, opened 2021-09-30
[Housekeeping] Deprecate duplicate helm chart


evalsocket:
**Describe the Issue**
Currently, we are maintaining both chart `flyte` and `flyte-core`  but idea is to convert `flyte` helm chart into a sandbox chart, It will depend on `flyte-core` for flyte core component deployment. 

Flyte charts 
- Flyte (It will install the sandbox deployment)
- Flyte-core (It will only deploy the core component of flyte)

**What if we do not do this?**
For any new changes, We need to replicate that change in `flyte-core`. It's a overhead and there can be a difference between both charts along the time. 

***Related component***
Either Specific / all

Pretty-19:
Hi @evalsocket  can you assign this task to me ?

evalsocket:
Awesome! Please do. :) 

Let us know if you have any questions. You can join our [slack](https://docs.google.com/forms/d/e/1FAIpQLScWPJZncyL-bBCWbuQ9HWuMzA8-r2RJc__CSWJoAYUQADN-BQ/viewform) channel.

samhita-alla:
Hi, @Pretty-19! Would you mind creating a PR within the next 2-3 days? :)

Pretty-19:
@samhita-alla  was a bit behind on schedule, yes will do that

Issue #1228, opened 2021-07-09
[Housekeeping] Modularize Flyte Helm Chart 


EngHabu:
**Describe the Issue**
The [Flyte Helm Chart](https://artifacthub.io/packages/helm/flyte/flyte) references 14 different images, only 5 of which are maintained by the Flyte team. I think it might be better to break it down into:
1. A Core helm chart that deploys flyte (FlyteAdmin, FlytePropeller, DataCatalog & FlyteConsole) only...
1. A Deployment helm chart that depends on that ^ plus other dependencies' helm charts (e.g. [contour](https://artifacthub.io/packages/helm/bitnami/contour), [minio](https://artifacthub.io/packages/helm/bitnami/minio), [postgres](https://artifacthub.io/packages/helm/nicholaswilde/postgres), ...)
   and if a dependency doesn't have a helm chart, it's ok to include its manifests like we do today...

EngHabu:
I know you've a lot on your plate now. Is this something you can take on, @evalsocket?

evalsocket:
As per discussion with @EngHabu on call. 

Action Items -

- [ ] Change Helm chart dir from `helm` to `charts/flyte`
- [ ] Use current helm charts without any change 
- [ ] Create a new helm chart (flyte-core). It only has flyte component. 
- [ ] Create a documentation for installing operators like a spark, PyTorch, etc (Each section should explain the process of operator installation and also describe the change in `value.yaml` of the helm).  Users can use different processes for installing these component 
- [ ] Create a documentation  for installing ingress (start with aws/Nginx/contour)
- [ ] Replace Sandbox docker image with helm deployment

PR #1759, opened 2021-10-28
Deprecate duplicate helm template

evalsocket:
- Imported `flyte-core` chart as a dependency in `flyte`
- Added option to enable/disable the specific components. By doing this user can install a control plane & data plane via a single chart.  

Fix: 
- https://github.com/flyteorg/flyte/issues/1537
- https://github.com/flyteorg/flyte/issues/1538
- https://github.com/flyteorg/flyte/issues/1228

Test Sandbox: 

Screenshot 2021-10-28 at 3 51 33 PM

Test GCP

Screenshot 2021-11-15 at 10 40 55 PM
:


deployment/eks/flyte_helm_generated.yaml
```diff
@@ -1,35 +1,4 @@
 ---
-# Source: flyte/templates/pytorch-operator/namespace.yaml
```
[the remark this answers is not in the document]
evalsocket: There is no change in eks deployment. Just removed the plugins config from the core chart. Now we have docs for plugin setup

deployment/gcp/flyte_helm_generated.yaml
```diff
@@ -10,8 +10,6 @@ metadata:
     app.kubernetes.io/instance: flyte
```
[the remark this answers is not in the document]
evalsocket: done please check description.

pmahindrakar-oss:
Can you add testing details for GCP . 
And also update the docs here https://docs.flyte.org/en/latest/deployment/gcp/index.html#installing-flyte

charts/flyte-core/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-name: flyte-core
-description: A Helm chart for Flyte
+name: flyte
```
EngHabu, line 2: We are keeping the existing names, right?

sbrunk: @EngHabu I brought that up already somewhere but I can't find the other comment.

helm can have weird issues when a chart with a `-` is used as a subchart. So my suggestion is to use this overhaul as an opportunity to rename `flyte-core` to just `flyte` so it can be used as a subchart. The other one is basically the sandbox chart right? So it's not going to be used as a subchart and we could name it `flyte-sandbox`. We could bump the major version of the chart for the change.

sbrunk: I gave it another try with the current `flyte-core` chart aliased as `flyte` and that seems to work fine. So perhaps things have improved since I tried last time.

```yaml
apiVersion: v2
appVersion: "1.0"
description: A Helm chart for Kubernetes
name: some-chart
version: 1.0.0

dependencies:
  - name: flyte-core
    alias: flyte
    version: 0.18.1
    repository: https://flyteorg.github.io/flyte
```

evalsocket: @sbrunk it's nice. I will try it

EngHabu (commented):
The changes look good to me. Let's fix the names of the charts and do one more round of testing as discussed and then merge

EngHabu (approved):
Once we validate on EKS, let's merge! wohooo

evalsocket:
Tested on eks....Hope so everything will work smoothly

Issue #1606, opened 2021-10-11
[Housekeeping] Containers should not run as root


frsann:
### Describe the issue

It seems like most Dockerimages related to the project does not define a user/group as which the service is run. This means that the services are run as root. We should fix this, to prevent security issues.

### What if we do not do this?

We run a risk of privilege escalation.  

### Related component(s)

At least flytepropeller, flyteadmin, flyteconsole, and datacatalog. 

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

EngHabu:
Thank you for filing this issue! I 100% agree.. We should update Dockerfiles of all of flyte components to run as user 1001 or something instead.. and a corresponding change in all deployments to let K8s set the right permissions for mounted volumes.

Is this something you might be able to help us do? even if it's just for one component so that people can start copying what you did into others? @frsann

frsann:
@EngHabu I made some PRs that hopefully fixes the issues.

PR #1766, opened 2021-10-29
Configure user and group in security context

frsann:
We set the user and group id in the security context to match the user and group defined explicitly in the docker images.

Note, the `console` uses [user id 1000](https://github.com/flyteorg/flyteconsole/blob/a64e1385852125bec9efe7f37c1d3930c7e2a434/Dockerfile#L30), while the rest of the images use user id `1001` and group id `1001` (see PRs below)

Related PRs:
* https://github.com/flyteorg/flytepropeller/pull/356
* https://github.com/flyteorg/flyteadmin/pull/280
* https://github.com/flyteorg/datacatalog/pull/54

Related issue:
* https://github.com/flyteorg/flyte/issues/1606

### DIFF
--- a/charts/flyte-core/templates/admin/deployment.yaml
+++ b/charts/flyte-core/templates/admin/deployment.yaml
@@ -83,11 +83,13 @@ spec:
           command: ["/bin/sh", "-c"]
           args:
             [
-                "flyteadmin --config={{ .Values.flyteadmin.configPath }} secrets init --localPath /etc/secrets/auth && flyteadmin --config=/etc/flyte/config/*.yaml secrets create --name flyte-admin-secrets --fromPath /etc/secrets/auth",
+                "flyteadmin --config={{ .Values.flyteadmin.configPath }} secrets init --localPath /etc/scratch/secrets && flyteadmin --config=/etc/flyte/config/*.yaml secrets create --name flyte-admin-secrets --fromPath /etc/scratch/secrets",
             ]
           volumeMounts:
             - name: config-volume
               mountPath: /etc/flyte/config
+            - name: scratch
+              mountPath: /etc/scratch
           env:
             - name: POD_NAMESPACE
               valueFrom:
@@ -145,6 +147,8 @@ spec:
       volumes: {{- include "databaseSecret.volume" . | nindent 8 }}
         - emptyDir: {}
           name: shared-data
+        - emptyDir: {}
+          name: scratch
         - configMap:
             name: flyte-admin-config
           name: config-volume
--- a/charts/flyte-core/templates/admin/deployment.yaml
+++ b/charts/flyte-core/templates/admin/deployment.yaml
@@ -18,13 +18,15 @@ spec:
       labels: {{ include "flyteadmin.labels" . | nindent 8 }}
     spec:
       securityContext:
-        fsGroup: 1001
+        fsGroup: 65534
         runAsUser: 1001
-        fsGroupChangePolicy: "OnRootMismatch"
+        fsGroupChangePolicy: "Always"
       initContainers:
         {{- if .Values.db.checks }}
         - name: check-db-ready
           image: ecr.flyte.org/ubuntu/postgres:13-21.04_beta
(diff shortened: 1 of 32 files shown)


charts/flyte-core/templates/admin/deployment.yaml
```diff
@@ -17,6 +17,10 @@ spec:
         {{- end }}
       labels: {{ include "flyteadmin.labels" . | nindent 8 }}
     spec:
+      securityContext:
+        fsGroup: 1001
+        runAsUser: 1001
+        fsGroupChangePolicy: "OnRootMismatch"
       initContainers:
         {{- if .Values.db.checks }}
         - name: check-db-ready
```
frsann, line 26: I dont have a easy way of testing how this user definition works together with the initContainer defined below.

pmahindrakar-oss:
@frsann i tested this change and admin pod doesn't come up.
```
kubectl logs -f flyteadmin-5cbc955498-6svq5 -n flyte generate-secrets
Using config file:  [/etc/flyte/config/cluster_resources.yaml /etc/flyte/config/db.yaml /etc/flyte/config/domain.yaml /etc/flyte/config/logger.yaml /etc/flyte/config/server.yaml /etc/flyte/config/storage.yaml /etc/flyte/config/task_resource_defaults.yaml]
Error: failed to create path [/etc/secrets/auth]. Error: mkdir /etc/secrets: permission denied
failed to create path [/etc/secrets/auth]. Error: mkdir /etc/secrets: permission denied
Usage:

```

Since admin init container generate secrets requires write access to /etc folder it fails.
Instead of docker file switching to use `USER flyte` , would it better to just create the user and the group in Dockerfile and have the orchestration yaml file contain the security context to use for each init container and the main container.

I have taken that approach in the admin change here 
https://github.com/flyteorg/flyteadmin/pull/278/

cc : @EngHabu @kumare3

pmahindrakar-oss:
Instead of using /etc/secrets/auth for the flyteadmin generate-secrets container , I was able to use ~/secrets/auth which doesn't require root permissions and the init container was able to run successfully.
This worked for kustomize based generated sandbox but it failed to work on helm generated sandbox due to subsitution issues. Helm generated file when applied was interpreting ~/secrets/auth as //secrets/auth and fails . Couldnt debug much why this was happening though.


Another issue is the init-container for check-db-ready uses pg_isready which fails to resolve the host 
```
postgres:5432 - no attempt
waiting for database
postgres:5432 - no attempt
waiting for database
```
Whereas the same init container when run with root user works fine . I believe this could be due to /etc/hosts file permissions.

These were the issue i found for flyteadmin kustomize , helm . Wasn't able to run flyte-core since that work is in progress  AFAIK

charts/flyte-core/templates/admin/deployment.yaml
EngHabu: I think the best way to do this is to:

```bash
flytectl sandbox start
helm list -a -A
helm uninstall -n flyte <helm chart name>
helm install flyte --debug ./charts/flyte -f ./charts/flyte/values-sandbox.yaml --create-namespace --namespace=flyte
```

and see what happens :-)

EngHabu:
I manually verified the changes in a separate deployment. I believe we can move ahead with the change...

Can you rerun `make kustomize && make helm` and commit the changes? @frsann

EngHabu:
It seems it's still complaining about changes in kustomize and helm... you might want to make sure you have the same versions of these tools as the ones used in CI

charts/flyte-core/templates/datacatalog/deployment.yaml
```diff
@@ -25,6 +25,8 @@ spec:
       {{- if .Values.db.checks }}
       - name: check-db-ready
         image: postgres:10.16-alpine
+        securityContext:
```
EngHabu, line 28: Perfect... yes that's what needed.. do you mind replicating that to kustomize?

EngHabu (approved):

PR #1801, opened 2021-11-05
Update the Getting started docs to use `pyflyte init`

eapolinario:
This is the corresponding changes to the Getting Started docs in order to reflect the introduction of[ cookiecutter templates in flytekit](https://github.com/flyteorg/flytekit/pull/738). 

This should only be merged after we cut the new `flytekit` release.

The Getting started docs now look like: https://flyte--1801.org.readthedocs.build/en/1801/getting_started.html

Signed-off-by: Eduardo Apolinario <eapolinario@users.noreply.github.com>

### DIFF
--- a/rsts/getting_started.rst
+++ b/rsts/getting_started.rst
@@ -44,7 +44,7 @@ Make sure you have `Git <https://git-scm.com/>`__, and `Python <https://www.pyth
 
 Start a new project / repository
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Install Flyte's Python SDK — `Flytekit <https://pypi.org/project/flytekit/>`__ (recommended using a  `virtual environment <https://docs.python.org/3/library/venv.html>`__), run a command to initialize your project, that command will prompt you for the name of directory where your example workflows are going to live. The ``init`` command comes with a help menu that can be shown by running ``pyflyte init --help``. For this guide we're going to use ``myflyteapp`` in the initial invocation.
+Install Flyte's Python SDK — `Flytekit <https://pypi.org/project/flytekit/>`__ (recommended using a  `virtual environment <https://docs.python.org/3/library/venv.html>`__), run ``pyflyte init {project_name}``, where ``{project_name}`` is the directory that will be created containing the scaffolding for a flyte-ready project. Feel free to use any name as your project name, however for this guide we're going to use ``myflyteapp`` in the  invocation:
 
 .. prompt:: bash (venv)$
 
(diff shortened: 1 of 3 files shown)


rsts/getting_started.rst
```diff
@@ -44,15 +44,15 @@ Make sure you have `Git <https://git-scm.com/>`__, and `Python <https://www.pyth
 
 Start a new project / repository
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Install Flyte's Python SDK — `Flytekit <https://pypi.org/project/flytekit/>`__ (recommended using a  `virtual environment <https://docs.python.org/3/library/venv.html>`__) and clone the `flytekit-python-template <https://github.com/flyteorg/flytekit-python-template>`__ repo.
+Install Flyte's Python SDK — `Flytekit <https://pypi.org/project/flytekit/>`__ (recommended using a  `virtual environment <https://docs.python.org/3/library/venv.html>`__), run a command to initialize your project, that command will prompt you for the name of directory where your example workflows are going to live. The ``init`` command comes with a help menu that can be shown by running ``pyflyte init --help``. For this guide we're going to use ``myflyteapp`` in the initial invocation.
```
kumare3, line 47: I dont think we need the extra text. That is the beauty of CLI's?

rsts/getting_started.rst
```diff
@@ -44,15 +44,15 @@ Make sure you have `Git <https://git-scm.com/>`__, and `Python <https://www.pyth
...
 
 .. prompt:: bash (venv)$
 
     pip3 install flytekit --upgrade
-    git clone https://github.com/flyteorg/flytekit-python-template.git myflyteapp
+    pyflyte init myflyteapp
```
kumare3, line 52: here is where we can add a comment saying, `myflyteapp` or your own app name?

kumare3: should we also say, it uses cookie-cutter underneath?

rsts/getting_started.rst
eapolinario: I guess I was trying to hand-hold too much. On a second read I agree with you, let me remove that part.

rsts/getting_started.rst
eapolinario: the paragraph right above mentions the choice of `myflyteapp`. I'll make it more explicit that it's actually part of the invocation (and that the user is free to pick whatever name they want).

As for letting them know that it's using cookie-cutter, do you really think it's needed? From a beginner user perspective they shouldn't need to know that it uses this or that templating engine. WDYT?

eapolinario: We can trust that the users that want to know what's going on under the covers will find out about cookie-cutter from the cli itself:
```
❯ pyflyte init --help
...
Usage: pyflyte init [OPTIONS] PROJECT_NAME

  Create flyte-ready projects.

Options:
  --template TEXT  cookiecutter template folder name to be used in the repo -
                   https://github.com/flyteorg/flytekit-python-template.git

  --help           Show this message and exit.
```

kumare3 (approved):

PR #1805, opened 2021-11-08
Kustomize update

illy:
the current yaml files use some beta version API, which are not supported any more by K8S, in turn, this causes failures when we run Kustomize to install flyte on customized server.

To fix the above issue, we update the apiVersion in the  following three yaml files according to https://kubernetes.io/docs/reference/using-api/deprecation-guide/, and successfully deployed to our server.

### DIFF
--- a/kustomize/base/wf_crd/wf_crd.yaml
+++ b/kustomize/base/wf_crd/wf_crd.yaml
@@ -8,22 +8,13 @@ spec:
   group: flyte.lyft.com
   # version name to use for REST API: /apis/<group>/<version>
   versions:
-    - name: v1alpha1
-      served: true
-      storage: true
-      schema:
-        openAPIV3Schema:
-          type: object
-          properties:
-            spec:
-              type: object
-              properties:
-                cronSpec:
-                  type: string
-                image:
-                  type: string
-                replicas:
-                  type: integer
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: 'FlyteWorkflow: represents one Execution Workflow object'
+        type: object
+    served: true
+    storage: true
   # either Namespaced or Cluster
   names:
     # plural name to be used in the URL: /apis/<group>/<version>/<plural>
--- a/kustomize/base/wf_crd/wf_crd.yaml
+++ b/kustomize/base/wf_crd/wf_crd.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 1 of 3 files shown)


kustomize/base/wf_crd/wf_crd.yaml
```diff
@@ -7,7 +7,23 @@ spec:
...
+          properties:
+            spec:
+              type: object
+              properties:
+                cronSpec:
+                  type: string
+                image:
+                  type: string
+                replicas:
+                  type: integer
```
EngHabu, line 26: I don't think we have these properties on the object... this can be left empty. right?

EngHabu (commented):
Are these changes backward compatible? will applying the new manifests on an existing deployment break or duplicate the objects?

kustomize/base/wf_crd/wf_crd.yaml
illy: We are not quite familiar with this part. However, if we don't include anything in the schema, it still broke. That's why we included an empty schema. If you think this can be left empty, please go ahead.

kustomize/base/wf_crd/wf_crd.yaml
```diff
@@ -7,7 +7,23 @@ spec:
...
   # either Namespaced or Cluster
```
EngHabu, line 27: This worked for me

```suggestion
  versions:
  - name: v1alpha1
    schema:
      openAPIV3Schema:
        description: 'FlyteWorkflow: represents one Execution Workflow object'
        type: object
    served: true
    storage: true
  # either Namespaced or Cluster
```

kustomize/base/wf_crd/wf_crd.yaml
EngHabu: this worked for me:
```
  versions:
  - name: v1alpha1
    schema:
      openAPIV3Schema:
        description: 'FlyteWorkflow: represents one Execution Workflow object'
        type: object
    served: true
    storage: true
```

kustomize/base/wf_crd/wf_crd.yaml
illy: thanks @EngHabu , i tested in my laptop, and it worked.

EngHabu: Thank you for confirming. Can you please update the PR so I can approve and merge?

illy: updated the changes, thanks again

EngHabu (approved):

EngHabu:
Please click on Details for "DCO" and follow instructions to resolve it

EngHabu:
Closing for lack of activity. Replacing with https://github.com/flyteorg/flyte/pull/1928

Issue #1808, opened 2021-11-09
[BUG] helm chart hardcodes flytepropeller's secret


EngHabu:
### Describe the bug

Flyte Propeller auth secret [here](https://github.com/flyteorg/flyte/blob/eca085f7c13c63b6ca6aa3ad2acf1f97af6b21b7/charts/flyte/templates/propeller/secret-auth.yaml#L8), [there](https://github.com/flyteorg/flyte/blob/1ede3750bfdeb307d1a81efd08674a68882bc873/kustomize/base/propeller/auth_secret.yaml#L8), [one more](https://github.com/flyteorg/flyte/blob/506408bb4d23287b9f0e1ab76af432aaa44dbd91/charts/flyte-core/templates/propeller/secret-auth.yaml#L8) is hardcoded.

This doesn't allow helm users to override it through values.

### Expected behavior

expose this value through values...

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #1812, opened 2021-11-10
Added client secret in helm values.yaml

evalsocket:
Flyte Propeller auth secret is hardcoded. This doesn't allow helm users to override it through values. Just added client secret in values.yaml

Fix https://github.com/flyteorg/flyte/issues/1808


evalsocket:
@EngHabu do we want kustomize changes ?

charts/flyte-core/values-eks.yaml
```diff
@@ -111,6 +111,7 @@ flytepropeller:
       ephemeral-storage: 1Gi
       memory: 2Gi
   cacheSizeMbs: 1024
+  client_secret: foobar
```
EngHabu, line 114: We only need to set it in values file, right? why do we also need to do that in vlaues-eks?

EngHabu (commented):
LGTM... one question

charts/flyte-core/values-eks.yaml
evalsocket: yes

EngHabu (approved):

Issue #1780, opened 2021-11-02
[BUG] Missing authentication options on flytescheduler deployment


ariefrahmansyah:
### Describe the bug

After we update to flyte 0.18.0 via helm chart, we got following error on flytescheduler pod:

```
{
  "json": {
    "src": "executor_impl.go:110"
  },
  "level": "error",
  "msg": "failed to create execution create request %+v due to %vproject:\"arief\" domain:\"development\" name:\"fb8e54574d3f61d72000\" spec:<launch_plan:<resource_type:LAUNCH_PLAN project:\"arief\" domain:\"development\" name:\"my_cron_scheduled_lp\" version:\"v1\" > metadata:<mode:SCHEDULED scheduled_at:<seconds:1632741000 > > > inputs:<literals:<key:\"kickoff_time\" value:<scalar:<primitive:<datetime:<seconds:1632741000 > > > > > >  rpc error: code = Unauthenticated desc = token parse error [JWT_VERIFICATION_FAILED] Could not retrieve id token from metadata, caused by: rpc error: code = Unauthenticated desc = Request unauthenticated with IDToken",
  "ts": "2021-11-02T02:44:29Z"
}
```

### Expected behavior

flytescheduler pod should be able to communicate with flyteadmin.

For my investigation, flytescheduler already has admin.yaml [configmap](https://github.com/flyteorg/flyte/blob/master/charts/flyte/templates/flytescheduler/configmap.yaml#L11) that consists of clientSecretLocation, however, flytescheduler's deployment itself doesn't mount the related secret. It should be configured similar to flytepropeller: https://github.com/flyteorg/flyte/blob/master/charts/flyte/templates/propeller/deployment.yaml#L42-L43

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #1836, opened 2021-11-17
Fix flytescheduler auth #1780 in flyte-core as well

sbrunk:
This has already been done in the flyte chart in #1810 but missing in `flyte-core`.


sbrunk:
Ah i think I know now why the docs were different again, causing the diff check to fail. My `helm-docs` was to old. After running
```bash
GO111MODULE=on go get github.com/norwoodj/helm-docs/cmd/helm-docs
```
manually I get the same output as in CI now.

EngHabu (approved):

PR #1846, opened 2021-11-19
Added helm registry in sandbox

evalsocket:
Currently for local testing if user wants to upgrade the sandbox then they need clone the flyte repo because of chart but with new changes user can use registry name for installing/upgrading the sandbox. 

For new release user can use `flyteorg/flyte` for installing/upgrading helm in sandbox, For older version user need to clone flyte repo for installing/upgrading flyte helm


EngHabu (approved):
Awesome!

PR #1850, opened 2021-11-19
Generate config docs only when running release workflow

pingsutw:
Signed-off-by: Kevin Su <pingsutw@apache.org>

Nit: Generate config docs only when running release workflow


EngHabu (approved):
Thank you

pingsutw:
Thanks~

PR #1855, opened 2021-11-23
rfc: type metadata

kennyworkman:


kumare3:
cc @eapolinario / @EngHabu. This looks pretty cool to me, just that the metadata field may be overloaded, especially in the dataclass case?

rfc/system/1855-type-metadata.md
```diff
@@ -0,0 +1,128 @@
...
+suggesting that support for metadata annotations for flyte types was in the
+works.
+
+We wish to add python-native support for this field by allowing flytekit to
+recognize and parse the use of
+[`typing.Annotation`](https://docs.python.org/3/library/typing.html#typing.Annotated).
+
+This could be to change the presentation of a form input ingesting typed values,
+surface descriptions to users or store indexes to control the order of
+parameters with custom frontends.
```
EngHabu, line 32: To be crisp clear, I would also like to list the things users won't be able to do with this...

```suggestion
This could be to change the presentation of a form input ingesting typed values,
surface descriptions to users or store indexes to control the order of
parameters with custom frontends.

What it doesn't give users:
1. At runtime, when a task is invoked, the metadata won't be sent to the container. If the container has access to the original code, it will be able to access/load metadata from that. This is particularly important to note for TaskTemplate-based flytekit plugins.
2. ?
```

rfc/system/1855-type-metadata.md
```diff
@@ -0,0 +1,128 @@
...
+- @maximsmol
+- @kennyworkman
+
+## 1 Executive Summary
+
+A small patch to introduce flytekit support for python-native type metadata.
+
+## 2 Motivation
+
+Users of flyte wish to have access to arbitrary parameter metadata specified in
```
EngHabu, line 14: I'm unclear why does it say "docstring" here when the proposal only includes typing.Annotation as the supported way of passing these annotations.

rfc/system/1855-type-metadata.md
```diff
@@ -0,0 +1,128 @@
...
+
+## 5 Drawbacks
+
+Do not see any.
+
+## 6 Alternatives
+
+Do not see any good ones.
+
+## 7 Potential Impact and Dependencies
```
EngHabu, line 118: Can you survey flytekit code for any existing use of metadata field? the risk is that this may clobber (or get clobbered by) an existing use-case.

rfc/system/1855-type-metadata.md
```diff
@@ -0,0 +1,128 @@
...
+                            }
+                        }
+                    }
+                ),
+            ]
+        ]
+    ],
+```
+
+## 3 Proposed Implementation
```
EngHabu, line 58: Can we include the work here to strip them in [here](https://github.com/flyteorg/flytepropeller/blob/master/pkg/compiler/transformers/k8s/node.go#L213).

EngHabu (commented):
Awesome! I've a few questions if you can please take a look

rfc/system/1855-type-metadata.md
kennyworkman: this was lazy copy-and-pasting from the sister RFC (https://github.com/flyteorg/flyte/pull/1856), just pushed a fix

rfc/system/1855-type-metadata.md
kennyworkman: i added your proposed anti-feature, but could not think of anymore. please add as many as you want with your greater familiarity of flyte

kennyworkman: btw, this seems to be the behavior of the proposed implementation

rfc/system/1855-type-metadata.md
kennyworkman: i did a grep for metadata over the whole codebase and could not find typing metadata related conflicts. i could not find any use of the metadata field other than where it is defined on the type models

rfc/system/1855-type-metadata.md
kennyworkman: can you help me understand why?

rfc/system/1855-type-metadata.md
```diff
@@ -0,0 +1,130 @@
...
+
+```
+amplicon_min_alignment_score: Optional[
+        List[
+            Annotated[
+                int,
+                FlyteMetadata(
+                    {
+                        "appearance": {
+                            "multiselect": {
```
kumare3, line 48: i this example, should this not be an enum instead?

kumare3: or is this purely cosmetic?

rfc/system/1855-type-metadata.md
kumare3: this is so that the backend representation at runtime does not bloat and is important to remain efficient and performant

rfc/system/1855-type-metadata.md
```diff
@@ -0,0 +1,130 @@
...
+        if hasattr(python_type, "__metadata__"):
+            for x in python_type.__metadata__:
+                if not isinstance(x, FlyteMetadata):
+                    continue
+                if x.data.get("__consumed", False):
+                    continue
+                meta = x.data
+                x.data["__consumed"] = True
+
+        if meta is not None:
```
kumare3, line 83: One problem with this approach is that `metadata` is currently used to store the json schema for struct types. I think we should either change that behavior (which will be breaking) or introduce a new field for storing documentation more uniformly in the LiteralType?

cc @EngHabu

My preference is to introduce a new substructure called Documentation to store

rfc/system/1855-type-metadata.md
kumare3: Also if we are storing this information, its more than just metadata - atleast in this example, it seems to be validation fields?

rfc/system/1855-type-metadata.md
kumare3: metadata is used for dataclasses to store the json schema. I am ok with storing all types of schema values associated with every type in the open api - json schema form. This can also introduce validations

rfc/system/1855-type-metadata.md
kennyworkman: this is purely cosmetic. it will be used by a custom frontend to saturate component props in this example

rfc/system/1855-type-metadata.md
kennyworkman: If you want to reserve metadata for internal flyte usage, Annotation makes more semantic sense to me ( as a field that can be overloaded by clients)

rfc/system/1855-type-metadata.md
```diff
@@ -0,0 +1,130 @@
...
+## 1 Executive Summary
+
+A small patch to introduce flytekit support for python-native type metadata.
+
+## 2 Motivation
+
+Users of flyte wish to have access to arbitrary parameter metadata specified in
+typing annotations.
+
+The `LiteralType` definition in flyteidl already has a `metadata` field in its
```
EngHabu, line 17: Per the discussion in the comments here and on chat, I think it's better to introduce a new message and a corresponding field (and potentially deprecating metadata altogether eventually)...

And the rest of the spec look great. 

```protobuf
message TypeMetadata {
  google.protobuf.Struct annotations = 1;
  // string type_hint = 2;
  // google.protobuf.Struct json_schema = 2;
 ...
}

message LiteralType 
{
...
   TypeMetadata metadata = 6; // For now I guess we have to call it type_metadata
}
```

EngHabu (commented):
Awesome! one comment (to introduce a new field/message) and then let's merge it!

rfc/system/1855-type-metadata.md
kennyworkman: what purpose does replacing  
``` 
google.protobuf.Struct metadata = 6;
```
with a nested message serve. Its just one more level of unwrapping for the exact same information?

Also, doesn't adding an _additional_ field make more sense.?We don't break support for existing use of `metadata` for flyte specific implementation but allows for use of annotations as intended in a way that doesn't conflict.

EngHabu (approved):
Awesome! #excited

Issue #1870, opened 2021-11-29
[Housekeeping] User common secret in place of flyte-propeller-auth in flyte-core helm chart


evalsocket:
### Describe the issue

Currently flytescheduler is using secret `flyte-propeller-auth` that's exist in propeller template [here](https://github.com/flyteorg/flyte/blob/master/charts/flyte-core/templates/propeller/secret-auth.yaml)

It should exist in common template, Currently it is creating issue while installing flyte in control plane mode 

### What if we do not do this?

Flytescheduler deployment will fail when user will try to install flyte-core in control plane mode

### Related component(s)

flytescheduler

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #1871, opened 2021-11-29
Added secret to common template

evalsocket:
Moved `flyte-propeller-auth` secret template in common template so flytescheduler can use it independently. 

Fix: https://github.com/flyteorg/flyte/issues/1870


sbrunk (dismissed):
Looks good to me.

charts/flyte-core/templates/flytescheduler/configmap.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.flytescheduler.enabled}}
 {{- if .Values.workflow_scheduler.enabled}}
   {{- if eq .Values.workflow_scheduler.type "native" }}
```
pmahindrakar-oss, line 3: Not your current change but we have flytescheduler and workflow_scheduler both here. 
Can we just keep flytescheduler .
Similarly instead of workflow_scheduler.type we should have flytescheduler.type

charts/flyte/values.yaml
```diff
@@ -237,6 +235,14 @@ flyte:
     # -- affinity for Flyteconsole deployment
     affinity: {}
 
+  #
+  # Common secret auth
+  #
+
+  adminOauthClientCredentials:
+    clientId: flytepropeller
```
EngHabu, line 243: [This value](https://github.com/flyteorg/flyte/blob/484c5b79ba614a5401cc24ca302358c871002d44/charts/flyte/values.yaml#L452) should be driven from `.Values.adminOauthClientCredentials.clientId`

Scheduler should have a similar one, not sure where is it... or maybe it's referencing the same one?

charts/flyte-core/templates/flytescheduler/configmap.yaml
evalsocket: Created a different PR for that https://github.com/flyteorg/flyte/pull/1896

charts/flyte/values.yaml
evalsocket: It is referencing the same one

EngHabu (commented):
Seems to need a rebase

EngHabu (dismissed):

EngHabu (approved):

PR #1881, opened 2021-12-01
Update dbname for datacatalog

wild-endeavor:

### DIFF
--- a/charts/flyte-core/values-eks.yaml
+++ b/charts/flyte-core/values-eks.yaml
@@ -209,7 +209,7 @@ db:
       username: flyteadmin
       host: "{{ .Values.userSettings.rdsHost }}"
       # -- Create a DB called datacatalog (OR change the name here)
-      dbname: flyteadmin
+      dbname: datacatalog
       passwordPath: /etc/db/pass.txt
   admin:
     database:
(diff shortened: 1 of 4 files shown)


wild-endeavor:
Word on the street is that one of datacatalog or admin will not create the pg database if it doesn't already exist.  whichever one it is will need to be fixed first before this pr can be merged.

charts/flyte-core/values-gcp.yaml
```diff
@@ -199,7 +199,7 @@ db:
       username: flyteadmin
       host: "{{ .Values.userSettings.dbHost }}"
       # -- Create a DB called datacatalog (OR change the name here)
-      dbname: flyteadmin
+      dbname: datacatalog
```
pmahindrakar-oss, line 202: Can we also add a step here on this doc https://docs.flyte.org/en/latest/deployment/gcp/index.html#deployment-gcp to create the datacatalog DB

charts/flyte-core/values-eks.yaml
```diff
@@ -209,7 +209,7 @@ db:
       username: flyteadmin
       host: "{{ .Values.userSettings.rdsHost }}"
       # -- Create a DB called datacatalog (OR change the name here)
-      dbname: flyteadmin
+      dbname: datacatalog
```
pmahindrakar-oss, line 212: Same here . https://docs.flyte.org/en/latest/deployment/aws/manual.html#create-rds-database

Issue #1864, opened 2021-11-25
[BUG] Flyte Native Scheduler does not expose metrics server


pradithya:
### Describe the bug

Flyte scheduler deployment does not expose `/metrics` endpoint for prometheus scraping

### Expected behavior

Flyte scheduler deployment should expose `/metrics` endpoint and publish necessary metrics for monitoring its operation.
Currently the helm chart / deployment has following configuration

```
Name:         flytescheduler-6b986459fd-szfk7
Namespace:    flyte
Labels:       app.kubernetes.io/instance=flyte
              app.kubernetes.io/managed-by=Helm
              app.kubernetes.io/name=flytescheduler
              helm.sh/chart=flyte-v0.18.1
              pod-template-hash=6b986459fd
Annotations:  configChecksum: a0c9f11fb2642375a4481c603a7c604c28e32728b175b768bee5b51ecee61b3
              prometheus.io/path: /metrics
              prometheus.io/port: 10254
              prometheus.io/scrape: true
```

Which expect the `:10254/metrics` to be accessible by prometheus, however it seems the port is not opened
```
curl: (52) Empty reply from server
```

Also from our prometheus 

<img width="1460" alt="Screenshot 2021-11-25 at 2 55 27 PM" src="https://user-images.githubusercontent.com/4023015/143394036-36b2d112-ece4-47a6-897e-fc00268237a9.png">

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

kumare3:
What I thought it did? This is a trivial change and important one. Cc @pmahindrakar-oss

EngHabu:
We will need to either update the default value to 10254 or change the helm manifests for scheduler to the correct annotation.

PR #1884, opened 2021-12-02
Added customizable metric port in helm chart

evalsocket:
Currently metric port is configurable but in helm there is no way to configure in helm, Added metric port in helm chart, Now user can configure the port. 

Tested All component with different port on sandbox, And it worked fine

Fix: https://github.com/flyteorg/flyte/issues/1864


charts/flyte-core/values.yaml
```diff
@@ -200,6 +206,8 @@ flytepropeller:
   tolerations: []
   # -- affinity for Flytepropeller deployment
   affinity: {}
+  # Metric port used for collecting metric
+  metricPort: 10254
```
EngHabu, line 210: You will have to put this in the config maps to take effect... it's not enough to list them here and in the container Ports...
e.g.: https://github.com/flyteorg/flytepropeller/blob/6a14e7fbffe89786fb1d8cde22715f93c2f3aff5/pkg/controller/config/config.go#L101

evalsocket:
@pmahindrakar-oss you can review the PR

pmahindrakar-oss:
@EngHabu  @evalsocket Do we have a prometheus deployment which can be configured to view these metrics. Also what usual scraping period of the metric port .

evalsocket:
@pmahindrakar-oss We have datadog deployment, 

Scraping period will be configure on prometheus side.

pmahindrakar-oss:
Ok . I wanted to see the scheduler metrics in datadog with this change.And also what's the scraping interval in datadog

EngHabu (dismissed):

kumare3:
We also have Prometheus deployed in the demo environment

charts/flyte/values.yaml
```diff
@@ -357,6 +357,12 @@ flyte:
         - id: production
           name: production
 
+    # Refer to the full [structure](https://pkg.go.dev/github.com/lyft/flyteadmin@v0.3.37/pkg/runtime/interfaces#ApplicationConfig) for documentation.
+    schedulerConfig:
+      scheduler:
+        metricsScope: 'flyte:'
+        profilerPort: 10254
+
```
EngHabu, line 365: We do not need this here, do we?
basically we do not need anything under `flyte:` in the flyte helm chart...

EngHabu (approved):

PR #1895, opened 2021-12-03
Update auth_setup.rst

bimtauer:
Update draft based on discussion on Flyte Slack where changing to "http://localhost..." was suggested as a fix. It works for us. Unsure if the comment on using https in case of secure: true is correct. 
Added unfinished explanation for obtaining to ClientSecret, since this is what we're looking for. We tried both internal auth server and external (Google). In the latter case obtaining the ClientSecret was obvious however Google does not support the `grant_type=client_credentials`. It would be good to know if the ClientSecret flow is at all possible with the internal auth server and which external providers acutally support it.

### DIFF
--- a/rsts/deployment/cluster_config/auth_setup.rst
+++ b/rsts/deployment/cluster_config/auth_setup.rst
@@ -217,8 +217,8 @@ Apply Configuration
                     # 3. Replace with a new Native Client ID provisioned in the custom authorization server
                     clientId: flytectl
 
-                    # This should not change
-                    redirectUri: https://localhost:53593/callback
+                    # Do not change to other callback. Switch to https when using TLS (secure: true)
+                    redirectUri: http://localhost:53593/callback
 
                     # 4. "all" is a required scope and must be configured in the custom authorization server
                     scopes:
@@ -282,6 +282,7 @@ If your organization does any automated registration, then you'll need to authen
    Flytectl's `config.yaml <https://docs.flyte.org/projects/flytectl/en/stable/#configure>`_ can be
    configured to use either PKCE (`Proof key for code exchange <https://datatracker.ietf.org/doc/html/rfc7636>`_)
    or Client Credentials (`Client Credentials <https://datatracker.ietf.org/doc/html/rfc6749#section-4.4>`_) flows.
+   If you are using the internal auth server, you can obtain the client credential by ... (This is what we are trying to figure out)
 
    Update ``config.yaml`` as follows:
 


rsts/deployment/cluster_config/auth_setup.rst
```diff
@@ -282,6 +282,7 @@ If your organization does any automated registration, then you'll need to authen
    Flytectl's `config.yaml <https://docs.flyte.org/projects/flytectl/en/stable/#configure>`_ can be
```
kumare3, line 282: We can use interSphinx to connect. Let me paste an update here

wild-endeavor:
```
:std:ref:`config.yaml <flytectl:flytectl_config>`
```

I think that should do it. Just searched for something that had to do with configuration
```
$ sphobjinv suggest https://flytectl.readthedocs.io config -us

No inventory at provided URL.
Attempting "https://flytectl.readthedocs.io/objects.inv" ...
Remote inventory found.

  Name                                             Score
------------------------------------------------  -------
:std:label:`flytectl_config`                        90
:std:label:`flytectl_config_discover`               90
:std:label:`flytectl_config_init`                   90
:std:label:`flytectl_config_validate`               90
:std:doc:`gen/flytectl_config`                      90
:std:label:`gen/flytectl_config:options`            90
:std:label:`gen/flytectl_config:see also`           90
:std:label:`gen/flytectl_config:synopsis`           90
:std:doc:`gen/flytectl_config_discover`             90
:std:doc:`gen/flytectl_config_init`                 90
:std:label:`gen/flytectl_config_init:options`       90
:std:label:`gen/flytectl_config_init:see also`      90
:std:label:`gen/flytectl_config_init:synopsis`      90
:std:doc:`gen/flytectl_config_validate`             90
:std:label:`index:basic configuration`              90
:std:label:`index:configure`                        90
```

i think this pr will build docs for us, we can check

rsts/deployment/cluster_config/auth_setup.rst
```diff
@@ -282,6 +282,7 @@ If your organization does any automated registration, then you'll need to authen
...
    configured to use either PKCE (`Proof key for code exchange <https://datatracker.ietf.org/doc/html/rfc7636>`_)
    or Client Credentials (`Client Credentials <https://datatracker.ietf.org/doc/html/rfc6749#section-4.4>`_) flows.
+   If you are using the internal auth server, you can obtain the client credential by ... (This is what we are trying to figure out)
```
wild-endeavor, line 285: cc @EngHabu

by internal auth server you mean the one provided by Admin?

wild-endeavor: or like, internal to your company?

rsts/deployment/cluster_config/auth_setup.rst
```diff
@@ -217,8 +217,8 @@ Apply Configuration
                     # 3. Replace with a new Native Client ID provisioned in the custom authorization server
                     clientId: flytectl
 
-                    # This should not change
-                    redirectUri: https://localhost:53593/callback
+                    # Do not change to other callback. Switch to https when using TLS (secure: true)
```
EngHabu, line 220: Until we implement SSL in flytectl, this should always be http://. 
```suggestion
                    # This should not change
```

rsts/deployment/cluster_config/auth_setup.rst
bimtauer: The one provided by admin. I assume even a company "internal" one would be configured as "external" in the flyteadmin config.

EngHabu: True on both accounts... 
So admin is deployed with a default [config here](https://github.com/flyteorg/flyteadmin/blob/12d6aa0a419ccec81b4c8289fd172e70a2ded525/auth/config/config.go#L86-L110). This can be overridden through the [config here](https://github.com/flyteorg/flyte/blob/1ede3750bfdeb307d1a81efd08674a68882bc873/kustomize/base/single_cluster/headless/config/admin/server.yaml#L26) by adding something like this:
```yaml
appAuth:
  selfAuthServer:
    staticClients:
      - newClient:
          id: newClient
          redirectUris:
            - http://myOtherApp...
          public: true
```

Issue #1891, opened 2021-12-03
[BUG] .Values.flytescheduler.enabled is not implemented


EngHabu:
### Describe the bug

Setting `.Values.flytescheduler.enabled` does not take effect because it's not implemented anywhere

### Expected behavior

Setting it to false should NOT install the scheduler

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #1896, opened 2021-12-03
Fix scheduler in helm

evalsocket:
Enable/disable scheduler with `workflow_scheduler.enabled`, If use want to use aws scheduler then he needs to define `workflow_scheduler.type: aws` or else helm will install the default scheduler.  

Sandbox: Sandbox will use native as default scheduler
```
workflow_scheduler:
  enabled: true
  type: native
```

GCP: GCP will use native as default scheduler
```
workflow_scheduler:
  enabled: true
  type: native
```

Eks: Eks will deploy use native as default scheduler
```
workflow_scheduler:
  enabled: true
  type: native

#To enable aws specific scheduler then

workflow_scheduler:
 enabled: true
  type: aws
  config:
    scheduler:
      # -- This is configured to use Cloudwatch schedules as explained [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/Create-CloudWatch-Events-Scheduled-Rule.html)
      eventScheduler:
        scheme: aws
        region: "{{ .Values.userSettings.accountRegion }}"
        scheduleRole: "arn:aws:iam::{{ .Values.userSettings.accountNumber }}:role/flyte_cron_scheduler_role"
        targetName: "arn:aws:sqs:{{ .Values.userSettings.accountRegion }}:{{ .Values.userSettings.accountNumber }}:flyte-helm-test-cron-scheduler-queue"
        scheduleNamePrefix: flyte
      workflowExecutor:
        scheme: aws
        region: "{{ .Values.userSettings.accountRegion }}"
        scheduleQueueName: flyte-helm-test-cron-scheduler-queue
        accountId: "{{ .Values.userSettings.accountNumber }}"
        reconnectAttempts: 10
        reconnectDelaySeconds: 30

```

Test:
- Tested on sandbox & eks. 

Fix: https://github.com/flyteorg/flyte/issues/1891


pmahindrakar-oss:
Having both workflow_scheduler and flytescheduler in helm causes confusion.
Can we juts keep one and have native or aws type and keep default as native in all deployments.

The additional config required for AWS based scheduler should be put only if aws scheduler is enabled and can be enabled through conditional template. And for native scheduler it has its own config
@EngHabu @evalsocket  let me know what you all think

evalsocket:
@pmahindrakar-oss done

evalsocket:
@pmahindrakar-oss It's ready for review

pmahindrakar-oss:
We seem to have removed the AWS scheduler config. Can we templatize and import it when aws type is enabled.

evalsocket:
@pmahindrakar-oss  yeah that is done, it's not there because if enable flytescheduler as default

pmahindrakar-oss:
I understand that we have flytescheduler as default which is basically enabling native type , but what happens if we use aws type . How does the user get that config. Does he have to type in this config?
```
#  config:
#    scheduler:
#      # -- This is configured to use Cloudwatch schedules as explained [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/Create-CloudWatch-Events-Scheduled-Rule.html)
#      eventScheduler:
#        scheme: aws
#        region: "{{ .Values.userSettings.accountRegion }}"
#        scheduleRole: "arn:aws:iam::{{ .Values.userSettings.accountNumber }}:role/flyte_cron_scheduler_role"
#        targetName: "arn:aws:sqs:{{ .Values.userSettings.accountRegion }}:{{ .Values.userSettings.accountNumber }}:flyte-helm-test-cron-scheduler-queue"
#        scheduleNamePrefix: flyte
#      workflowExecutor:
#        scheme: aws
#        region: "{{ .Values.userSettings.accountRegion }}"
#        scheduleQueueName: flyte-helm-test-cron-scheduler-queue
#        accountId: "{{ .Values.userSettings.accountNumber }}"
#        reconnectAttempts: 10
#        reconnectDelaySeconds: 30
```

evalsocket:
@pmahindrakar-oss It's there https://github.com/flyteorg/flyte/pull/1896/files#diff-bd478ed1475344bf71c6df6c1219602a634dff0691900b55d5f1a770f2a9b4faR331-R346, it's not in generated menifest because of type. We have one condition https://github.com/flyteorg/flyte/pull/1896/files#diff-01968ddd24a15b6d0092a1eee5652aa896d4866278f8d50bafd939d4f8f349acR39

EngHabu (dismissed):
Looks good to me but please wait for Praful to sign off

pmahindrakar-oss (approved):

script/generate_helm.sh
```diff
@@ -17,6 +17,9 @@ for deployment in ${DEPLOYMENT_CORE}; do
     helm template flyte -n flyte ${DIR}/../charts/flyte-core/ -f ${DIR}/../charts/flyte-core/values.yaml -f ${DIR}/../charts/flyte-core/values-${deployment}.yaml > ${DIR}/../deployment/${deployment}/flyte_helm_generated.yaml
 done
 
+# Generate manifest AWS native scheduler
```
pmahindrakar-oss, line 20: nit : AWS scheduler

charts/flyte-core/values-eks-override.yaml
```diff
@@ -0,0 +1,22 @@
+# -- # Flyte uses a cloud hosted Cron scheduler to run workflows on a schedule. The following module is optional. Without,
+# this module, you will not have scheduled launchplans/workflows.
+# Override eks value for using aws native scheduler
```
pmahindrakar-oss, line 3: We should add that we support native scheduler and this is for backward compatability to older scheduler only

pmahindrakar-oss (dismissed):

pmahindrakar-oss (approved):

PR #1905, opened 2021-12-07
fix sandbox values in entrypoint script

evalsocket:
Rename values file in sandbox entry point (This bug introduce because of helm changes Deprecate duplicate helm template #1759)

### DIFF
--- a/docker/sandbox/flyte-entrypoint-default.sh
+++ b/docker/sandbox/flyte-entrypoint-default.sh
@@ -26,7 +26,7 @@ echo "Deploying Flyte..."
 helm repo add flyteorg https://flyteorg.github.io/flyte
 helm repo update
 helm fetch flyteorg/flyte --version=$FLYTE_VERSION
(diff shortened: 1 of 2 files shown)


docker/sandbox/flyte-entrypoint-default.sh
```diff
@@ -26,7 +26,7 @@ echo "Deploying Flyte..."
 helm repo add flyteorg https://flyteorg.github.io/flyte
 helm repo update
 helm fetch flyteorg/flyte --version=$FLYTE_VERSION
-helm upgrade -n flyte -f /flyteorg/share/flyte/values-sandbox.yaml --create-namespace flyte flyteorg/flyte --kubeconfig /etc/rancher/k3s/k3s.yaml --install --version $FLYTE_VERSION
+helm upgrade -n flyte -f /flyteorg/share/flyte/values.yaml --create-namespace flyte flyteorg/flyte --kubeconfig /etc/rancher/k3s/k3s.yaml --install --version $FLYTE_VERSION
```
EngHabu, line 29: I think we should make it like this?
```suggestion
helm upgrade -n flyte --create-namespace flyte flyteorg/flyte --kubeconfig /etc/rancher/k3s/k3s.yaml --install --version $FLYTE_VERSION
```

EngHabu (approved):

Issue #1273, opened 2021-07-22
[Housekeeping] Upgrade to apiextensions.k8s.io/v1 CustomResourceDefinition in deployment resources (prepare for v1.22+)


brucearctor:
Describe the bug
```Warning: apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition``` is pointed out when running Kustomize deployment

Expected behavior
Current version of K8s is 1.21 -- we should upgrade that to refer to rbac.authorization.k8s.io/v1 by v.1.22+ release

**Comment by pradithya** [comment]
Warning when deploying flyte gave us list removed apis in 1.22

```
W1102 03:48:56.655502     110 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W1102 03:48:56.660052     110 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W1102 03:48:56.688648     110 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W1102 03:48:56.713382     110 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W1102 03:48:56.718868     110 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W1102 03:48:56.748158     110 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
[...]

PR #1972, opened 2021-12-21
Update apis to allow for compatibility with k8s 1.22

fsz285:
The CRD for the flyteworkflows is incomplete, atm. A schema should be
added in the near future.

See #1273

### DIFF
--- a/charts/flyte-core/templates/admin/rbac.yaml
+++ b/charts/flyte-core/templates/admin/rbac.yaml
@@ -43,7 +43,7 @@ rules:
   - '*'
 
 ---
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: {{ template "flyteadmin.name" . }}-binding
--- a/charts/flyte-core/templates/common/ingress.yaml
+++ b/charts/flyte-core/templates/common/ingress.yaml
@@ -3,46 +3,62 @@
 - path: /flyteidl.service.AdminService
   pathType: ImplementationSpecific
   backend:
-    serviceName: flyteadmin
-    servicePort: 81
+    service:
+      name: flyteadmin
+      port:
+        number: 81
 - path: /flyteidl.service.AdminService/*
   pathType: ImplementationSpecific
   backend:
-    serviceName: flyteadmin
-    servicePort: 81
+    service:
+      name: flyteadmin
+      port:
+        number: 81
 - path: /flyteidl.service.AuthMetadataService
   pathType: ImplementationSpecific
   backend:
-    serviceName: flyteadmin
-    servicePort: 81
+    service:
+      name: flyteadmin
+      port:
+        number: 81
 - path: /flyteidl.service.AuthMetadataService/*
   pathType: ImplementationSpecific
   backend:
-    serviceName: flyteadmin
-    servicePort: 81
+    service:
+      name: flyteadmin
+      port:
+        number: 81
 - path: /flyteidl.service.IdentityService
   pathType: ImplementationSpecific
   backend:
-    serviceName: flyteadmin
-    servicePort: 81
+    service:
+      name: flyteadmin
+      port:
+        number: 81
 - path: /flyteidl.service.IdentityService/*
   pathType: ImplementationSpecific
   backend:
-    serviceName: flyteadmin
-    servicePort: 81
+    service:
+      name: flyteadmin
+      port:
+        number: 81
 - path: /grpc.health.v1.Health
   pathType: ImplementationSpecific
   backend:
-    serviceName: flyteadmin
-    servicePort: 81
+    service:
+      name: flyteadmin
+      port:
+        number: 81
 - path: /grpc.health.v1.Health/*
   pathType: ImplementationSpecific
   backend:
-    serviceName: flyteadmin
-    servicePort: 81
+    service:
+      name: flyteadmin
+      port:
+        number: 81
   {{- end }}
   {{- if .Values.common.ingress.enabled }}
-apiVersion: networking.k8s.io/v1beta1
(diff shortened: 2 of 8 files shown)


sbrunk:
@EngHabu @kumare3 
Docker desktop is now at k8s 1.22 so this would help people installing sandbox directly inside the k8s coming with docker desktop.

Looking at the failing test it seems we also need to update Flytepropeller to be able to deal with `apiextensions.k8s.io/v1` for CRDs:

```
goroutine 681 [running]:
github.com/flyteorg/flytepropeller/pkg/apis/flyteworkflow/v1alpha1.(*WorkflowSpec).GetID(...)
	/go/src/github.com/flyteorg/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/workflow.go:278
github.com/flyteorg/flytepropeller/pkg/controller.(*Propeller).TryMutateWorkflow(0xc000b7b350, 0x2898430, 0xc000785ce0, 0xc00045ef00, 0x0, 0x0, 0x0)
```

EngHabu:
@sbrunk @fsz285 Do you have an idea of what are the changes needed in propeller?

sbrunk:
@EngHabu I haven't looked at the code yet (my go knowledge is still somewhat limited) but AFAIK propeller dynamically creates `flyteworkflow` custom resources when a workflow is executed.

Since 1.22 removes `apiextensions.k8s.io/v1beta1`, @fsz285 has changed the [CRD here in this PR](https://github.com/flyteorg/flyte/pull/1972/files#diff-2af5ce96b0560e6b4a9c1fa86113a8007729fa0c9d2b13990f114dc254a0cbf7) to conform to `apiextensions.k8s.io/v1` according to https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122

I initially thought we'd have to change propeller to create `flyteworkflow` custom resources conforming to that updated CRD version, but the structure of a `flyteworkflow` custom resource instance itself actually shouldn't change at all. 🤔

The only difference in our case is that a schema is [required](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition) in `v1` while it was optional in `v1beta1` so we allow any JSON object but perhaps we are doing it wrong.

sbrunk:
The error in the test is caused by [this line](https://github.com/flyteorg/flytepropeller/blob/c016dabbfef6037bead59590b42326dabe89f957/pkg/controller/handler.go#L273-L278) when we're trying to update the custom resource status:

```go
// TODO we will need to call updatestatus when it is supported. But to preserve metadata like (label/finalizer) we will need to use update

// update the GetExecutionStatus block of the FlyteWorkflow resource. UpdateStatus will not
// allow changes to the Spec of the resource, which is ideal for ensuring
// nothing other than resource status has been updated.
p.wfStore.Update(ctx, mutatedWf, workflowstore.PriorityClassCritical)
```

EngHabu:
@hamersaw thank you! You should be able to repro this locally by starting `flytectl sandbox start`, updating the CRD in the cluster (like, [this file](charts/flyte-core/templates/propeller/crds/flyteworkflow.yaml)). And trying to run any example Workflow...

I'm not sure where things go wrong... Updating the Api version may require regeneration of clientset in propeller using -maybe- some updated versions of k8s code-gen repo? but yeah I would start with a simple repo and looking at the FlyteWorkflow CRD Instance Admin creates after launching an execution...

hamersaw:
@fsz285  @sbrunk I think you missed the `x-kubernetes-preserve-unknown-fields` in the FlyteWorkflow CRD v1 schema. As I understand it, unless this is explicitly set all fields not conforming to the schema will be automatically pruned. In my testing, all CRDs made in k8s v1.22 with the previous setup (this PR) had completely empty specs which then makes sense that retrieving the workflow ID gives us an "invalid memory address or nil pointer dereference" error as you saw. Would you be willing to run a quick test including the `x-kubernetes-preserve-unknown-fields` as shown below? it worked for me.

```
{{- if .Values.flytepropeller.enabled }}
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: flyteworkflows.flyte.lyft.com
spec:
  group: flyte.lyft.com
  names:
    kind: FlyteWorkflow
    plural: flyteworkflows
    shortNames:
    - fly
    singular: flyteworkflow
  scope: Namespaced
  versions:
  - name: v1alpha1
    served: true
    storage: true
    schema:
      openAPIV3Schema:
        type: object
        x-kubernetes-preserve-unknown-fields: true
        properties:
{{- end }}
```

cc @EngHabu this, of course, should be a temporary workaround. I think we really should flesh out a fully defined schema for the FlyteWorkflow CRD moving forward. Additionally, although it seems to work with the 0.20 k8s.io/api dependencies in a majority of our repos (I suspect we are using fairly stable APIs), we should probably update these as well. Thoughts?

sbrunk:
@hamersaw yeah that totally makes sense. Thanks for looking into it. I just added `x-kubernetes-preserve-unknown-fields: true` to see if the tests pass now.

EngHabu (approved):
Thank you for all the work!

bigbitbus:
@EngHabu This change does break the helm chart for everyone who is on Kubernetes 1.18 or earlier (ingress v1beta1)... I know those are old versions but still a good number of folks use those versions, and EKS supports that until March 2022.

sbrunk:
If we really need this (March 2022 is only one month away though), it is possible to use helm [capabilities](https://helm.sh/docs/chart_template_guide/builtin_objects/) to define an ingress differently depending on the available API/k8s version:

https://stackoverflow.com/questions/66147091/is-it-possible-to-configure-a-helm-chart-to-dynamically-use-a-value-based-on-ano

EngHabu:
I've an attempt at that here: https://github.com/flyteorg/flyte/pull/2114
but trying to figure out why `make helm` is producing a diff

Issue #1601, opened 2021-10-11
[Docs] Fix Docker image unambiguity in the getting started guide


samhita-alla:
### Description

Registry URL isn't required for sandbox; mention the same in the getting started guide.

Reference: https://flyte-org.slack.com/archives/CNMKCU6FR/p1633816410395600

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #1975, opened 2021-12-22
Fix Docker image unambiguity in the getting started guide

pingsutw:
https://github.com/flyteorg/flyte/issues/1601

doc link: https://flyte--1975.org.readthedocs.build/en/1975/getting_started_scale.html#build-deploy-your-application-to-the-cluster

image


kumare3:
Can we use the same tabbing structure for the first part? 
I like what you have done, it automatically switches all tabs, as they are named the same. Can we do something similar for the first tab too?

Screen Shot 2021-12-22 at 9 07 31 AM

pingsutw:
+1, good idea. Updated it, thanks

kumare3 (approved):

Issue #1959, opened 2021-12-16
[Docs] FlytePropeller Sharding Deployment


hamersaw:
### Description

Document deployment of FlytePropeller Manager [here](https://docs.flyte.org/en/latest/deployment/cluster_config/performance.html#scaling-out-flytepropeller) to provide a horizontally scaling FlytePropeller environment.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

evalsocket:
@hamersaw do we need changes in helm to enable Sharding ?

hamersaw:
> @hamersaw do we need changes in helm to enable Sharding ?

@evalsocket yes we will. I was planning on taking a look at it and opening up a small PR. It seems like a pretty simple update and a nice introduction to helm. Expect a review shortly and probably a few questions!

PR #1982, opened 2021-12-23
Adding helm updates and docs for flytepropeller sharding

hamersaw:
TL;DR adding a new manager template for the helm chart and documenting deployment of sharded FlytePropeller.

fixes https://github.com/flyteorg/flyte/issues/1959


charts/flyte-core/templates/propeller/manager.yaml
```diff
@@ -0,0 +1,135 @@
+{{- if .Values.flytepropeller.enabled }}
+{{- if .Values.flytepropeller.manager }}
+apiVersion: apps/v1
```
evalsocket, line 3: flytepropeller-manager is very similar to flytepropeller deployment, Can we use one resource for both ?

hamersaw: Merged the deployment templates. Let me know if you like this more, I was trying to avoid the constant "if flytepropeller-manager" statements, seems messy to me.

charts/flyte-core/templates/propeller/deployment.yaml
```diff
@@ -2,29 +2,48 @@
...
       securityContext:
         fsGroup: 65534
         runAsUser: 1001
         fsGroupChangePolicy: "Always"
       containers:
       - command:
+        {{- if .Values.flytepropeller.manager }}
+        - flytepropeller-manager
+        - --manager.pod-template-namespace
+        - {{ template "flyte.namespace" . }}
```
EngHabu, line 43: Not a helm expert. Should this say `include` or `template` ?

charts/flyte-core/values-eks.yaml
```diff
@@ -249,6 +250,15 @@ configmap:
...
   core:
+    # -- follows the structure specified [here](https://pkg.go.dev/github.com/flyteorg/flytepropeller/manager/config#Config).
+    manager:
+      pod-application: "flytepropeller"
+      pod-template-container-name: "flytepropeller"
+      pod-template-name: "flytepropeller-template"
+      scan-interval: 10s
+      shard:
+        type: Hash
+        pod-count: 3
```
EngHabu, line 261: Looks like these are the defaults here: https://github.com/flyteorg/flytepropeller/blob/v0.16.7/manager/config/config.go#L13-L25.. unless you need to reference them in the deployment, you don't need to explicitly set them here... makes it a bit easier to change the defaults if needed...

charts/flyte-core/values-gcp.yaml
```diff
@@ -254,6 +255,15 @@ configmap:
...
   core:
+    # -- follows the structure specified [here](https://pkg.go.dev/github.com/flyteorg/flytepropeller/manager/config#Config).
+    manager:
+      pod-application: "flytepropeller"
+      pod-template-container-name: "flytepropeller"
+      pod-template-name: "flytepropeller-template"
+      scan-interval: 10s
+      shard:
+        type: Hash
+        pod-count: 3
```
EngHabu, line 266: Also no need to specify them at all for values-*.yaml... it's enough to specify the defaults in values.yaml

charts/flyte/values.yaml
```diff
@@ -460,6 +461,15 @@ flyte:
...
     core:
+      # -- follows the structure specified [here](https://pkg.go.dev/github.com/flyteorg/flytepropeller/manager/config#Config).
+      manager:
+        pod-application: "flytepropeller"
+        pod-template-container-name: "flytepropeller"
+        pod-template-name: "flytepropeller-template"
+        scan-interval: 10s
+        shard:
+          type: Hash
+          pod-count: 3
```
EngHabu, line 472: No need to specify them here... in fact, this entire file will need a lot of cleanup since it's duplicating a bunch of things from flyte-core

EngHabu (commented):
Awesome! got a few helm questions 😬

charts/flyte-core/templates/propeller/deployment.yaml
hamersaw: @evalsocket can probably provide more information on this. I just used the definition of the namespace I see elsewhere. I believe the 'template' is because of how templates are compiled [here](https://github.com/flyteorg/flyte/blob/0c4b9bd39a107a005227a763519ce5e49a97bfde/script/generate_helm.sh#L14) with the '-n flyte' flag.

charts/flyte-core/values-eks.yaml
hamersaw: I removed the "scan-interval" and "shard" configuration options, but kept the others. I think it makes sense to them in places within the PodTemplate (e.x., [here](https://github.com/flyteorg/flyte/blob/0ce73769842d8046174507daf726943c886c0c15/charts/flyte-core/templates/propeller/manager.yaml#L37) so that the linking between the manager and the PodTemplate is automatic and users don't have to worry about mis-configuration. If you think this might be confusing we can certainly change it?

charts/flyte-core/values-gcp.yaml
hamersaw: Sure, removed.

charts/flyte/values.yaml
hamersaw: I kept the same as ^^^ where the pod-application, pod-template-container-name, and pod-template-name are still set here. I might be misunderstanding but the [deployments template generation](https://github.com/flyteorg/flyte/blob/0c4b9bd39a107a005227a763519ce5e49a97bfde/script/generate_helm.sh#L17) uses the flyte-core/values.yaml file but when generating for the [sandbox](https://github.com/flyteorg/flyte/blob/0c4b9bd39a107a005227a763519ce5e49a97bfde/script/generate_helm.sh#L14) we only use the flyte/values.yaml file. Maybe this should be cleaned up?

charts/flyte-core/values-eks.yaml
EngHabu: Makes sense... anything you are referencing in the chart you can keep... that linkage makes sense! thank you

charts/flyte/values.yaml
EngHabu: My understanding is that helm will automatically pickup the `values.yaml` of the dependent chart + whatever overrides you have in your own `values.yaml`.. you can try removing them here and `make helm` to see if that generates the same...

hamersaw: Absolutely, you are right. Seems to me like pretty helm master knowledge to me, my 30 minute tutorial didn't cover this. I removed it.

charts/flyte-core/templates/propeller/deployment.yaml
hamersaw: @EngHabu It looks like the "include" keyword is meant to be used in [Go pipelines](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-include-function). I think we are using "template" correctly here?

EngHabu (approved):

PR #1987, opened 2021-12-23
Adding Office Hours

SandraGH5:
Adding Office Hours and making some editorial checks.


kumare3 (approved):
we need to add the newer talks to the conferences section

SandraGH5:
> we need to add the newer talks to the conferences section

PyData and ODSC 2021 talks and podcasts are already there. 
I was planning to add future talks once they've happened and the recordings are available, but will go ahead and add them here then update the videos later.

kumare3 (approved):

PR #1991, opened 2021-12-24
Added FlytePropeller architecture document

hamersaw:
Added a toctree and set of buttons to the "Component Architecture" section to dive into each individual component architecture beginning with FlytePropeller's architecture.


kumare3:
@hamersaw +1 and I am merging, but we should get the sharded propeller docs in and explain that this doc is only for one workflow execution at a time.

kumare3 (approved):

Issue #1827, opened 2021-11-12
[Bug] Mention editing charts/flyte/values-eks.yaml in setup with Opta


RustamovAzer:
### Description

When I tried deploying without editing values-eks.yaml I got this error:
```
Failed to convert return value for var get_taxi_output for function myapp.workflows.queries_on_flyte.get_taxi_dataset_task with error <class 'flytekit.common.exceptions.user.FlyteAssertion'>: Failed to put data from /tmp/flytezd5ikpdu/local_flytekit/95adb1bdc4a4d9defb7ce5c82e67f871 to s3://<BUCKET_NAME>/dn/zsr5ihyt7h-n0-3/a6ed27d28f0cbb809555fdae65ee1041 (recursive=True).
```
When I changed `bucketName` under `userSettings` in `values-eks.yaml` from `<BUCKET_NAME>` to the name of my bucket (I also filled `accountNumber`, `accountRegion` and `logGroup`) I managed to run my workflow successfully.

I think it should be mentioned [here](https://docs.flyte.org/en/latest/deployment/aws/opta.html).

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

kumare3:
Thank you, what ll be part of our doc sprint

evalsocket:
@kumare3 It's not the documentation issue, It's opta deployment bug, In set all these values from opta file and we are not override the propeller config for correct bucket name.

PR #2012, opened 2022-01-04
Added user setting in opta, 

evalsocket:
Currently in aws opta setup, We are not overriding the propeller config for bucket name, That's why execution is failing in cluster, 

To avoid the conflict, I added the user setting in opta so if we missed something then `values-eks.yaml` will override it 

Fix: https://github.com/flyteorg/flyte/issues/1827

### DIFF
--- a/opta/aws/flyte.yaml
+++ b/opta/aws/flyte.yaml
@@ -50,14 +50,6 @@ modules:
     values_file: "../../charts/flyte-core/values-eks.yaml" # NOTE: relative path to values yaml
     # Additional overrides to the values provided by the chart. Opta enables piping through produced outputs from prior modules/steps.
     values:
-      userSettings:
-        rdsHost: "${{module.postgres.db_host}}"
-        accountRegion: "{vars.region}"
-        bucketName: "{parent_name}-{layer_name}"
-        logGroup: /eks/opta-staging/cluster
-        dbUser: "${{module.postgres.db_user}}"
-        dbname: "${{module.postgres.db_name}}"
-        dbPassword: "${{module.postgres.db_password}}"
       db:
         datacatalog:
           database:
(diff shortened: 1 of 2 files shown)


pmahindrakar-oss:
Can you add details on how the executions are failing.
Are the all the opta deployments broken for aws and gcp.

AFAIK when we tested gcp using opta, it was working fine and executions were being launched correctly.

Please elaborate more on the issue.
Also the linked issue seems to require only doc changes. These are mentioned here for manual deployment
https://docs.flyte.org/en/latest/deployment/aws/manual.html#installing-flyte
So probably we need to just link to this ?

evalsocket:
@pmahindrakar-oss  GCP is working we didn't have any issue for that, Just added these value for the safe side if we miss anything

evalsocket:
Regarding issue, In opta user didn't need to add user setting manually. Opta will automatically override the values. In issue user deployed flyte with opta and he is facing issue in execution because opta didn't override the propeller config with correct bucket name https://github.com/flyteorg/flyte/blob/master/charts/flyte-core/values-eks.yaml#L254

pmahindrakar-oss:
The changes user tried on are whats available in the branch , Hence my question was are we facing the same issues in other opta deployments, specifically the ones we deploy to verify the workflows are working on aws using opta. Trying to understand if other users will also be facing the same issue or we do something different internally to verify opta deployments.

evalsocket:
In our internal deployment we are doing override of user settings and we are not facing any issue

evalsocket:
@pmahindrakar-oss reverted the extra changes

pmahindrakar-oss:
We can keep the internal charts and open source charts in sync so that we are able to check the issues that community faces on our internal deployment too.

EngHabu (approved):

Issue #1857, opened 2021-11-23
[Housekeeping] Expose flyteconsole's ENABLE_GA env var to helm chart


EngHabu:
### Describe the issue

As part of addressing https://github.com/flyteorg/flyte/issues/1442, Console's PR https://github.com/flyteorg/flyteconsole/pull/244 is adding an environment variable to control whether Google Analytics should be enabled or not. We need to make changes to the deployment charts to:

- [ ] Expose a flag in the values.yaml to control this env var.. Defaults to false
- [ ] Enable the flag by default in values-sandbox
- [ ] Update getting started docs to explain this and tell people how to turn it off.

### What if we do not do this?

N/A

### Related component(s)

flyte

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

EngHabu:
Thank you @evalsocket for picking this up!

PR #2019, opened 2022-01-05
Add Env variable in flyteconsole helm deployment for GA

evalsocket:
- Add Env variable in flyteconsole helm deployment for GA

Fix: https://github.com/flyteorg/flyte/issues/1857


charts/flyte-core/templates/console/deployment.yaml
```diff
@@ -30,6 +30,13 @@ spec:
             name: flyte-console-config
         ports:
         - containerPort: 8080
+        {{- if .Values.flyteconsole.ga.enabled }}
+        env:
+          - name: ENABLE_GA
+            value: {{ .Values.flyteconsole.ga.enabled }}
```
pmahindrakar-oss, line 36: Since this is already in if condition with the same value,  do we need to maintain this variable .

pmahindrakar-oss: Why can't the app just use the tracking id if it's not empty that means ga tracking is enabled.

I find another env variable redundant here

evalsocket: Check this https://github.com/flyteorg/flyteconsole/pull/244

pmahindrakar-oss: great thanks for additional background.The app could have used one but sure since it already implemented using two variables then we can go ahead with this.

pmahindrakar-oss (approved):

PR #2021, opened 2022-01-05
add platform events to Deployment > Cluster Config

cosmicBboy:


eapolinario (commented):
Platform events already have a card: https://github.com/flyteorg/flyte/blame/659c7f473a4b11e5acd08d9e3fb64754009580a7/rsts/deployment/cluster_config/index.rst#L74-L81. They are named "external events", which should be renamed.

PR #2042, opened 2022-01-10
Creating mermaid charts

SmritiSatyanV:


kumare3:
Lgtm, need to ensure accurate representation

katrogan (approved):
hey @SmritiSatyanV thanks for this change! it looks like the boxes for Running and NodeRunning lost their box-level label, see https://flyte--2042.org.readthedocs.build/en/2042/concepts/state_machine.html#divedeep-state-machine vs https://docs.flyte.org/en/latest/concepts/state_machine.html#divedeep-state-machine although I'll defer to Ketan for how important that is to maintain :)

Issue #2080, opened 2022-01-19
[Housekeeping] Release rollback


evalsocket:
### Describe the issue

Improvements around Flyte release rollback 

- [x] Force flytectl to use `latest version sha`  for sandbox docker image in place of `dind` (https://github.com/flyteorg/flytectl/pull/266)
- [x] Write a documentation for rollback helm manually
- [ ] Stop publishing `latest` and `dind` tag for sandbox (https://github.com/flyteorg/flyte/pull/2085)
- [ ] Verify helm upgrade process for beta release
- [x] Create Github workflow for all component to create patch release (https://github.com/flyteorg/flyteadmin/pull/327, https://github.com/flyteorg/flytepropeller/pull/387)
- [x] Add Upgrade Test in CI with sandbox (Already exist but currrently it's in broken stage)
- [ ] https://github.com/flyteorg/flyte/issues/2077 (https://github.com/flyteorg/flyteadmin/pull/329)

### What if we do not do this?

NA

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

evalsocket:
cc: @EngHabu

evalsocket:
@EngHabu 

# Rollback Helm Release

- Clone `flyte` Repo
- checkout `gh-pages`
- Remove helm package for flyte-core `rm flyte-core-v0.19.0.tgz`
- Remove helm package for flyte `rm flyte-v0.19.0.tgz`
- Remove version entry in `index.yml` for both chart. Entry look like this 
```yaml 
  - apiVersion: v2
    created: "2022-01-19T03:23:18.295085027Z"
    dependencies:
    - alias: flyte
      name: flyte-core
      repository: file://../flyte-core
      version: v0.19.1
    - condition: contour.enabled
      name: contour
      repository: https://charts.bitnami.com/bitnami
      version: 4.1.2
    - alias: sparkoperator
      condition: sparkoperator.enabled
      name: spark-operator
      repository: https://googlecloudplatform.github.io/spark-on-k8s-operator
      version: 1.0.6
    - condition: kubernetes-dashboard.enabled
      name: kubernetes-dashboard
      repository: https://kubernetes.github.io/dashboard/
      version: 4.0.2
    description: A Helm chart for Flyte Sandbox
    digest: 6c207f4545311aa6d4d249ec536e6a6f37d20bbbfc327fa761baf1ad68a5541b
    name: flyte
    type: application
[...]

PR #2085, opened 2022-01-20
Stop publishing dind and latest tag for sandbox

evalsocket:
- Removed non-dind sandbox image from sandbox 
- Stop publishing `non-dind` images and `dind` tag 
- Fix the logic for publishing sandbox image to registery, Now it will also publish images on pre release

Fix: https://github.com/flyteorg/flyte/issues/2080


docker/sandbox/Dockerfile
```diff
@@ -56,37 +56,6 @@ COPY charts/flyte-core/ /flyteorg/share/flyte-core
 # Copy scripts
 COPY docker/sandbox/kubectl docker/sandbox/cgroup-v2-hack.sh docker/sandbox/wait-for-flyte.sh /flyteorg/bin/
 
-
-FROM base_ AS default
```
wild-endeavor, line 60: was this just not used?  I don't see a `--from=default` anywhere.  Or do you not need to specify that because it's default?

docker/sandbox/flyte-entrypoint-default.sh
```diff
@@ -7,10 +7,27 @@ cgroup-v2-hack.sh
```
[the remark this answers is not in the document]
evalsocket: yes

.github/workflows/sandbox.yml
```diff
@@ -97,7 +33,6 @@ jobs:
           images: |
             ghcr.io/${{ github.repository_owner }}/flyte-sandbox
           tags: |
-            dind
             type=sha,format=long, prefix=dind-
```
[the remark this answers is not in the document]
evalsocket: It's open question, New flytectl doesn't need dind tag, it can be used by old flytectl client for running latest sandbox

docker/sandbox/Dockerfile
evalsocket: yeah we can simplify it

.github/workflows/sandbox.yml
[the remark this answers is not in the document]
evalsocket: we will push dind image until people will stop pulling it

.github/workflows/sandbox.yml
```diff
@@ -112,7 +49,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-single-buildx
       - name: Login to GitHub Container Registry
-        if: ${{ github.event_name == 'release' }}
+        if: ${{ github.event_name != 'release' }}
```
EngHabu, line 52: This should change, right?

.github/workflows/sandbox.yml
```diff
@@ -98,6 +34,7 @@ jobs:
             ghcr.io/${{ github.repository_owner }}/flyte-sandbox
           tags: |
             dind
+            ${{ steps.set_version.outputs.flyte_version }}
```
EngHabu, line 37: Should we also add this now that we will only publish the dind images?

```suggestion
            ${{ steps.set_version.outputs.flyte_version }}
            type=sha,format=long
```

.github/workflows/sandbox.yml
evalsocket: this is correct bu github is not showing it

evalsocket: https://github.com/flyteorg/flyte/blob/06532ae86cab41082950e7c1aec57107de23ff53/.github/workflows/sandbox.yml#L63

.github/workflows/sandbox.yml
evalsocket: we need prefix in sha, it's old structure

evalsocket: @EngHabu There is one issue https://github.com/flyteorg/flytectl/blob/master/pkg/util/githubutil/githubutil.go#L224

EngHabu (approved):

PR #2105, opened 2022-01-26
Fixes for flyte-core helm chart to allow deploying cluster resource sync as a standalone process

katrogan:
- Update flyteadmin image version to latest with standalone cluster resource controller code changes
- Use boolean switch in the cluster resource config to set standalone mode


charts/flyte-core/values.yaml
```diff
@@ -638,15 +638,17 @@ external_events:
 cluster_resource_manager:
   # -- Enables the Cluster resource manager component
   enabled: true
-  # -- Starts the cluster resource manager in standalone mode with requisite auth credentials to call flyteadmin service endpoints
-  standalone_deploy: false
```
EngHabu, line 642: Wouldn't moving this break people's existing values files, if any? is there a reason to move it?

katrogan: it's already part of the config, see: https://github.com/flyteorg/flyteadmin/blob/30862d414125eda4ef74455e4614100b522d2af8/pkg/runtime/interfaces/cluster_resource_configuration.go#L46

katrogan: and this has only existed in a beta release so far

EngHabu (dismissed):

.github/workflows/tests.yml
```diff
@@ -4,6 +4,8 @@ on:
     branches:
```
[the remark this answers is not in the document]
katrogan: I am not sure... @evalsocket is this necessary?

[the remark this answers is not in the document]
katrogan: removed.

katrogan:
PTAL @EngHabu

EngHabu (approved):

PR #2111, opened 2022-01-28
Mount admin auth creds for cluster resource sync run in standalone mode.

katrogan:


EngHabu (approved):
`make helm` ?

PR #2114, opened 2022-01-31
Add backward compatibility checks for api changes

EngHabu:
As described [here](https://github.com/flyteorg/flyte/pull/1972). The work done to move to the latest K8s APIs would break backward compatibility with older clusters. This PR adds compatibility check to use the corresponding K8s APIs


charts/flyte-core/templates/propeller/crds/flyteworkflow.yaml
```diff
@@ -1,5 +1,9 @@
 {{- if .Values.flytepropeller.enabled }}
+{{- if $.Capabilities.APIVersions.Has "apiextensions.k8s.io/v1/CustomResourceDefinition" }}
```
sbrunk, line 2: Do people still use Flyte with k8s versions `<1.16`? If not, this might not be necessary.

.github/workflows/tests.yml
```diff
@@ -24,7 +24,10 @@ jobs:
   end-to-end-kustomize:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
```
wild-endeavor, line 28: why did we not need go before?

charts/flyte-core/templates/propeller/crds/flyteworkflow.yaml
```diff
@@ -21,4 +26,7 @@ spec:
           type: object
           x-kubernetes-preserve-unknown-fields: true
           properties:
+{{- else }}
+  version: v1alpha1
```
wild-endeavor, line 30: should this be plural?  or the one above singular?

script/generate_helm.sh
```diff
@@ -38,7 +39,7 @@ if [ -n "$DELTA_CHECK" ]; then
     echo "diff detected: $DIFF"
     DIFF=$(git diff --name-only)
     echo "files different: $DIFF"
-    exit 1
+    exit 0
```
wild-endeavor, line 42: so always succeed?

wild-endeavor (approved):
have some questions though, i've been worried about this change for a while.  how will this affect existing CRDs on a cluster?

.github/workflows/tests.yml
EngHabu: This is in my attempt to keep the environment consistent for `make helm` in CI and locally until we introduce a docker image to build helm template :-))))

charts/flyte-core/templates/propeller/crds/flyteworkflow.yaml
EngHabu: This is the deprecated form.. you could only have one... the new api allows multiple versions declared in the CRD...

script/generate_helm.sh
EngHabu: Ops... nooo... that was me playing with the script... thanks for catching this!

wild-endeavor (approved):

PR #2165, opened 2022-02-15
explicitly grant replicasets finalizer permissions to flyte-pod-webhook

akumor:
without granting this permission, installation of the mutatingwebhookconfiguration fails with the following error:
```
Error: mutatingwebhookconfigurations.admissionregistration.k8s.io "flyte-pod-webhook" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>
```


EngHabu (dismissed):
Thank you for fixing this

EngHabu:
@akumor mind looking into why it's failing in CI?

EngHabu (approved):

PR #2196, opened 2022-02-21
Propagate namespace mapping to flyteadmin and clusterresourcesync

### LINKED ISSUE #2195: Custom Namespace Mappings are Ignored
### Describe the bug

It's possible to use customize the mapping from projects and domains to namespaces via config. For instance, in the GCP config the default `{{ project }}-{{ domain }}` is changed to just domain `{{ domain }}` which makes it easier to configure permissions via workload-identity.

https://github.com/flyteorg/flyte/blob/230ad922cfccaab7bf5a3461d8f337d3eab3c510/charts/flyte-core/values-gcp.yaml#L252-L255

This configuration does not propagate to flyteadmin and syncresource though, so they'll always use the [default mapping](https://github.com/flyteorg/flyteadmin/blob/4af6bd5f1b5e3ecc483f527547f140073fbdaf24/flyteadmin_config.yaml#L207-L208).

### Expected behavior

Changing the namespace mapping should propagate correctly to flyteadmin, syncresources and any other component that needs it.

_No response_

### Screenshots

_No response_

- [X] Yes

- [X] Yes

sbrunk:
Fixes #2195

### DIFF
--- a/charts/flyte-core/templates/admin/configmap.yaml
+++ b/charts/flyte-core/templates/admin/configmap.yaml
@@ -21,6 +21,9 @@ data:
 {{- with .Values.configmap.remoteData }}
   remoteData.yaml: | {{ tpl (toYaml .) $ | nindent 4 }}
 {{- end }}
+{{- with .Values.configmap.namespace_config }}
+  namespace_config.yaml: | {{ toYaml . | nindent 4 }}
+{{- end }}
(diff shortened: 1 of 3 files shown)


sbrunk:
Do we need to propagate this to another configmap besides flyteadmin and clusterresourcesync?

EngHabu (approved):

PR #2221, opened 2022-03-02
Add env to flyteadmin chart values.

ariefrahmansyah:
For our case, we need to specify SENDGRID_API_KEY via environment variable.
To support this, flyteadmin needs to be able to receive env supplied by user.

Signed-off-by: ariefrahmansyah <ariefrahmansyah@hotmail.com>


pmahindrakar-oss:
You will also need to do make helm to create generated files for aws gcp

ariefrahmansyah:
> You will also need to do make helm to create generated files for aws gcp

I've run `make helm` and it's only update README for both flyte-core and flyte charts. @pmahindrakar-oss

pmahindrakar-oss (approved):

evalsocket (approved):

PR #2229, opened 2022-03-03
Add app: flytepropeller to pod labels

mayitbeegh:
Together with this PR https://github.com/flyteorg/flytepropeller/pull/410, they try to use the same set of labels on the flytepropeller pods deployed by Helm or flytepropeller manager.


EngHabu:
We do not need this PR any more, right, @mayitbeegh ?

PR #2307, opened 2022-03-31
Increase flyteconsole memory limit to 275Mi for sandbox

EngHabu:


anrusina (approved):
Thanks!

PR #2311, opened 2022-04-01
remove priority class from gcp helm values

evalsocket:
- GCP helm deployment is failing because of priority class, Removed for GCP
- Replaced `go get` with `go install` in ci


evalsocket:
@EngHabu Any idea why is it failing ? https://github.com/flyteorg/flyte/runs/5789618228?check_suite_focus=true#step:5:177

EngHabu (approved):
Awesome!

Issue #2324, opened 2022-04-06
[BUG] Debug flyteadmin memory spikes


katrogan:
### Describe the bug

In certain scenarios, highly parallelized dynamic tasks reporting node execution events lead to huge memory spikes in flyteadmin. We should debug the source of this and fix it.

### Expected behavior

Flyteadmin memory consumption should not spike like crazy during the process of workflow execution and associated event updates.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #2335, opened 2022-04-08
Downgrade logs for cloud deployments

katrogan:

https://github.com/flyteorg/flyte/issues/2324

### DIFF
--- a/charts/flyte-core/values-eks.yaml
+++ b/charts/flyte-core/values-eks.yaml
@@ -302,7 +302,7 @@ configmap:
 
   logger:
     logger:
-      level: 5
+      level: 3
 
   # -- Section that configures how the Task logs are displayed on the UI. This has to be changed based on your actual logging provider.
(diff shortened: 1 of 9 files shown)


charts/flyte-core/values-eks.yaml
```diff
@@ -302,7 +302,7 @@ configmap:
 
   logger:
     logger:
-      level: 5
+      level: 3
```
EngHabu, line 305: Can we just drop this? the default is good enough, right?

charts/flyte-core/values-gcp.yaml
```diff
@@ -287,7 +287,7 @@ configmap:
 
   logger:
     logger:
-      level: 5
+      level: 3
```
EngHabu, line 290: ditto

EngHabu (commented):
And can we set it to 5 or something in the flyte helm chart (for sandbox) ?

charts/flyte-core/values-eks.yaml
katrogan: oh good point! done

katrogan:
PTAL @EngHabu

EngHabu (approved):

PR #2336, opened 2022-04-09
Added sandbox deployment docs

evalsocket:


evalsocket:
Question: In new sandbox deployment, We have sure that user is using eks cluster. 

- How do we want to expose minio console & API 
- How do we want expose k8s dashboard ?
cc: @EngHabu @kumare3

rsts/deployment/sandbox/index.rst
```diff
@@ -0,0 +1,41 @@
...
+Flyte Deployment - Sandbox
+**************************
+This guide helps you set up Flyte from scratch, on any kubernetes cluster, It will helpful in trying out flyte. It details step-by-step instructions to go from a bare Kubernetes cluster to a fully functioning Flyte deployment that members of your company can use.
+
+.. note::
+
+  Please make sure you have a working kubernetes cluster with write permission
+
+
+1. Add Helm repo for Flyte,Postgres & Minio
```
EngHabu, line 17: Can we add links to official minio and postgres docs as well here... ?

rsts/deployment/sandbox/index.rst
```diff
@@ -0,0 +1,41 @@
...
+.. code-block::
+
+ helm repo add flyteorg https://helm.flyte.org
+ helm repo add bitnami https://charts.bitnami.com/bitnami
+
+2. Install Minio deployment in kubernetes cluster
+
+.. code-block::
+
+ helm install minio bitnami/minio -n default --set auth.rootPassword="flyte"
```
EngHabu, line 28: Is this what we set the password to in our sandbox deployment??

EngHabu: I think you set the user to "admin" in the sandbox-values.yaml above...

rsts/deployment/sandbox/index.rst
```diff
@@ -0,0 +1,41 @@
...
+
+3. Install Postgres deployment in kubernetes cluster
+
+.. code-block::
+
+ helm install postgresql bitnami/postgresql -n default --set auth.postgresPassword="flyte"
+
+4. Create a values-override.yaml file
```
EngHabu, line 36: So we tell people to create a `values-override.yaml` but we don't tell them what to put in there or how to use it

EngHabu (commented):
RE: Minio
We need to add instructions for setting [this value](https://github.com/flyteorg/flyte/blob/master/charts/flyte-core/templates/_helpers.tpl#L161): 
Maybe we just say this needs to be a URL accessible from outside the cluster to enable fast registration (and add a link to a docs page about fast registration)

RE: Dashboard
[Instructions here](https://kubernetes.io/docs/tasks/access-application-cluster/web-ui-dashboard/) have multiple options about accessing the dashboard... we should just say follow these instructions, 
then set [this value](https://github.com/flyteorg/flyte/blob/eb86da98cd209ac57a51ddd73b3a46fbe71280c8/kustomize/overlays/sandbox/flyte/config/propeller/plugins/task_logs.yaml#L4) (the equivalent value for helm) with that value...

EngHabu:
Offline discussion notes:

Let's also add instructions to verify things are working..
* Maybe get the LB for admin and verify console works
* Get LB for admin and make sure flytectl works

Maybe add a section at the end to explain a bit about ingress... why it's needed and how to install Contour and maybe links for Inginx, ALB, GCP's equivalent... 
Inginx: https://docs.flyte.org/en/latest/deployment/gcp/manual.html#ingress

charts/flyte-deps/README.md
```diff
@@ -14,11 +14,19 @@ A Helm chart for Flyte dependency
 
 ### SANDBOX INSTALLATION:
 - [Install helm 3](https://helm.sh/docs/intro/install/)
+<<<<<<< HEAD
```
kumare3, line 17: merge conflicts

Issue #2116, opened 2022-01-31
[BUG] Clean-up opta flyte.yaml


kumare3:
### Describe the bug

The Opta deployment flyte.yaml, contains some unnecessary sections like
setting up SQS / SNS, granting Cloudwatch eventss role. 
These can be found here - https://github.com/flyteorg/flyte/blob/master/opta/aws/flyte.yaml#L14-L39

### Expected behavior

Ideally the only things user should be expected to configure is
1. Setup EKS (or reuse an existing one)
2. Setup S3 (or reuse an existing one)
3. Setup Postgres
4. Setup role to access S3

This is all, but the current deployment seems to be more complicated than it needs to be.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

EngHabu:
Thanks to Austin Bennett for catching this. We seem to have lingering references to these module outputs in the opta file 
e.g. https://github.com/flyteorg/flyte/blob/master/opta/aws/flyte.yaml#L115

@evalsocket this file is now broken...

PR #2337, opened 2022-04-09
Removed sqs helm values from aws opta

evalsocket:
fix: https://github.com/flyteorg/flyte/issues/2116

### DIFF
--- a/opta/aws/flyte.yaml
+++ b/opta/aws/flyte.yaml
@@ -88,7 +88,7 @@ modules:
           annotations:
             "eks.amazonaws.com/role-arn": "${{module.adminflyterole.role_arn}}"
       workflow_scheduler:
-        enabled: false
+        enabled: true
       workflow_notifications:
         enabled: false
       configmap:
--- a/opta/aws/flyte.yaml
+++ b/opta/aws/flyte.yaml
@@ -88,41 +88,9 @@ modules:
           annotations:
             "eks.amazonaws.com/role-arn": "${{module.adminflyterole.role_arn}}"
       workflow_scheduler:
-        enabled: true
-        config:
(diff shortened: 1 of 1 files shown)


opta/aws/flyte.yaml
```diff
@@ -88,41 +88,9 @@ modules:
...
-              targetName: "${{module.schedulesQueue.queue_arn}}"
-              scheduleNamePrefix: flyte
-            workflowExecutor:
-              scheme: aws
-              region: "{vars.region}"
-              scheduleQueueName: "${{module.schedulesQueue.queue_name}}"
-              accountId: "{vars.account_id}"
-              reconnectAttempts: 10
-              reconnectDelaySeconds: 30
+        enabled: false
```
EngHabu, line 91: We can use native scheduler, right?

evalsocket: yes

opta/aws/flyte.yaml
```diff
@@ -89,40 +89,8 @@ modules:
             "eks.amazonaws.com/role-arn": "${{module.adminflyterole.role_arn}}"
       workflow_scheduler:
         enabled: true
```
EngHabu, line 91: Is the native scheduler the default value?

EngHabu (approved):

opta/aws/flyte.yaml
evalsocket: yes

evalsocket: It is coming from eks values https://github.com/flyteorg/flyte/blob/master/charts/flyte-core/values-eks.yaml#L329

EngHabu (approved):

PR #2344, opened 2022-04-12
Added flyte-deps Helm chart

evalsocket:


charts/flyte-deps/templates/webhook/service.yaml
```diff
@@ -0,0 +1,13 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "flyte-pod-webhook.name" . }}
+  namespace: {{ template "flyte.namespace" . }}
+  {{- with .Values.webhook.service.annotations }}
+  annotations: {{ tpl (toYaml .) $ | nindent 4 }}
+  {{- end }}
+spec:
+  type: ExternalName
```
pingsutw, line 11: We should use clusterIP for sandbox, and ExternalName for single binary (flyte-sandbox-lite or demo). Probably make it configurable.

pingsutw: okay, sounds good.

PR #2351, opened 2022-04-13
Speed up building process of sandbox-lite image

pingsutw:
- Remove `kubectl`, we can use `k3s kubectl` instead 

The main reason why the sandbox lite dnd image takes an hour to build is that it's our first time building the image, and we don't even cache any image layers.

I built the image in my fork repo a second time, it only took 3 mins
https://github.com/pingsutw/flyte/actions/runs/2163267735

### DIFF
--- a/docker/sandbox-lite/Dockerfile
+++ b/docker/sandbox-lite/Dockerfile
@@ -4,15 +4,7 @@ FROM ghcr.io/flyteorg/flyteconsole:${FLYTECONSOLE_VERSION} AS flyteconsole
 FROM golang:1.18.0-alpine3.15 AS go_builder
 
 # Install dependencies
-RUN apk add --no-cache build-base git make
-
-# Create directory to store built artifacts
-ARG INSTALL_DIR="/install"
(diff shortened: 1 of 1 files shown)


docker/sandbox-lite/Dockerfile
```diff
@@ -98,7 +90,6 @@ VOLUME /var/lib/kubelet
 VOLUME /var/lib/rancher/k3s
 VOLUME /var/lib/cni
 VOLUME /var/log
-COPY --from=flyteconsole /app/dist /app/flyte/cmd/single/dist
```
pingsutw, line 101: This is duplicate

eapolinario:
This is pretty awesome. It cut 20 minutes from the build (looking at https://github.com/flyteorg/flyte/actions/workflows/sandbox.yml). 

@pingsutw , are you saying that going forward we'll have image layers cached and that we should see these low build times of single-digit minutes in the future?

pingsutw:
yeah, I think so. 
It took 3 mins to build the image in my GA workflow.
image

eapolinario (approved):

PR #2353, opened 2022-04-14
fix: added prefix in sandbox-lite image tag

evalsocket:
Currently sandbox-lite doesn't add `dind-` prefix in image tag, That's why `flytectl demo start` is failing. 

Added prefix in sandbox lite image tag


.github/workflows/sandbox.yml
```diff
@@ -98,7 +98,7 @@ jobs:
             ghcr.io/${{ github.repository_owner }}/flyte-sandbox-lite
           tags: |
             ${{ steps.set_version.outputs.flyte_version }}
-            type=sha,format=long
+            type=sha,format=long, prefix=dind-
```
EngHabu, line 101: Should we fix flytectl instead?

PR #2378, opened 2022-04-19
Single binary: keep assets prefix when fetching console files

katrogan:


katrogan:
no longer needed due to console refactoring

Issue #2284, opened 2022-03-24
[Core feature] Allow to choose Spot Instances at workflow start time


Antaxify:
### Motivation: Why do you think this is important?

At this point, we can set `interruptible=true` either at task or workflow level. I would be great if we could set this flag when we **start a workflow**, similiar to `max parallism`.  This would allow to use the very same workflow as follows:

- during development / testing, run with `interruptible=false` to get feedback asap
- for big production runs where cost efficiency is more important, run with `interruptible=true`
- for time critical executions (e.g. an approaching deadline) , run with `interruptible=false` 

This would be a very helpful addition to Flyte. :)

### Goal: What should the final outcome look like, ideally?

- `interruptible` can be set at workflow start, similar to `max parallism`
- can be set when a workflow ist started using flytectl, console, flyte_remove, or api

### Describe alternatives you've considered 

Registering the same workflow twice, once with `interruptible` set to true, once set to false

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

kumare3:
Cc @pmahindrakar-oss / @EngHabu

katrogan:
this could be a great candidate for the newly [expanded execution config](https://github.com/flyteorg/flyteidl/pull/253/files#diff-6c101b93284e15aa7ab1060e7cab2d403065a1b27529d01e0540092bcaa23139R104) as a matchable resource too

MorpheusXAUT:
Hi!
As briefly mentioned on Slack, I'd be willing to look into this, feel free to assign the issue to me :smile:

kumare3:
Assigned

PR #2390, opened 2022-04-21
Add script for setting up k3d cluster for local development

MorpheusXAUT:
This PR adds a bash script and Make target for setting up a basic k3d cluster with Flyte's dependencies for local development using the newly introduced single binary.

Whilst working on my [latest changes](https://github.com/flyteorg/flyte/issues/2284), I've had to work on flyteadmin and flytepropeller simultaneously, which turned out to be a bit tedious using the default sandbox setup, especially building/pushing a new flyteadmin image to the sandbox cluster on every change to test.
Using the new single binary, I was able to develop much faster locally, however still had to run the dependencies for Flyte separately (mainly minio/postgres).

I've opted to setting up a basic k3d cluster with only Flyte's basic dependencies running (flyte single binary is built/run locally to allow for easier debugging). To make setup easier for my colleagues and me, I've put together a script for setting up all requirements for running and starting the cluster as well as the required containers (using the `flyte-deps` helm chart of this repository).
A separate config file for accessing local k3d resources has been added as well. In theory, this could also be gitignored and `sed`-replaced from the `flyte_local.yaml` file.

I figured I'd offer this script/setup for critique here to see if there's better ways of achieving the same goal (running all Flyte components locally via the single binary). I'm (somewhat :smile: ) aware of the `sandbox-lite` Dockerfile, however it seems to me that would still require image rebuilds/pushes to test changes.

### DIFF
--- a/script/setup_local_dev.sh
+++ b/script/setup_local_dev.sh
@@ -1,0 +1,171 @@
+#!/usr/bin/env bash
+
+: "${ENSURE_K3D:="false"}" # will ensure k3d is installed if set to "true"
+: "${ENSURE_KUBECTL:="false"}" # will ensure kubectl is installed if set to "true"
+: "${ENSURE_HELM:="false"}" # will ensure helm is installed if set to "true"
+
+: "${K3D_VERSION:="v4.4.8"}" # version of k3d to install, empty value uses latest one available. currently pinned to latest known working (due to unresolved port mapping issue: https://github.com/k3d-io/k3d/issues/380)
+: "${K3D_INSTALL_URL:="https://raw.githubusercontent.com/rancher/k3d/main/install.sh"}" # URL to k3d installer script
+: "${K3S_VERSION:="v1.22.8-k3s1"}" # version of k3s to run in k3d cluster, empty value uses default specified by k3d install
+: "${K3D_CLUSTER_NAME:="flyte"}" # name of k3d cluster to be used
+: "${K3D_KUBECONFIG_FILE_PATH:="${HOME}/.k3d/kubeconfig-${K3D_CLUSTER_NAME}.yaml"}" # file path to store kubeconfig file for k3d cluster at
+
+: "${KUBECTL_VERSION:=""}" # version of kubectl to install, empty value uses latest available
+: "${KUBECTL_INSTALL_URL:="https://dl.k8s.io/release/VERSION/bin/linux/amd64/kubectl"}" # URL to kubectl binary, include VERSION to be replaced with KUBECTL_VERSION
+: "${KUBECTL_INSTALL_DIR:="/usr/local/bin"}" # local path to install kubectl to
+: "${KUBECTL_VERIFY_SIGNATURE:="true"}" # verifies signature of kubectl binary downloaded if set to "true"
+
+: "${HELM_VERSION:=""}" # version of helm to install, empty value uses latest v3 available
+: "${HELM_INSTALL_URL:="https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3"}" # URL to helm installer script
+: "${HELM_K8S_NAMESPACE:="flyte"}" # k8s namespace to use with helm
+
+: "${FLYTE_TIMEOUT:=600}"
+
+set -e
+
+HAS_CURL="$(type "curl" &> /dev/null && echo true || echo false)"
+HAS_WGET="$(type "curl" &> /dev/null && echo true || echo false)"
+
+if [ "${HAS_CURL}" != "true" ] && [ "${HAS_WGET}" != "true" ]; then
+  echo "Either curl or wget is required"
+  exit 1
+fi
+
+echo "Setting up local development environment"
+
+if [ "${ENSURE_K3D}" == "true" ]; then
+  echo -e "\nEnsuring k3d is installed"
+
+  if [ "${HAS_CURL}" == "true" ]; then
+    curl -sSL $K3D_INSTALL_URL | TAG=$K3D_VERSION bash
+  else
+    wget -q -O - $K3D_INSTALL_URL | TAG=$K3D_VERSION bash
+  fi
+else
+  if type "k3d" &> /dev/null; then
+    echo -e "\nUsing existing k3d install"
+  else
+    echo -e "\nk3d was not found in PATH, please install manually or run script with ENSURE_K3D=true"
+    exit 1
+  fi
+fi
+
+k3d version
+
+if [ "${ENSURE_KUBECTL}" == "true" ]; then
+  echo -e "\nEnsuring kubectl is installed"
+
+  if [ "${HAS_CURL}" == "true" ]; then
+    if [ -z "${KUBECTL_VERSION}" ]; then
+      FULL_KUBECTL_INSTALL_URL="${KUBECTL_INSTALL_URL/VERSION/$(curl -sL https://dl.k8s.io/release/stable.txt)}"
+    else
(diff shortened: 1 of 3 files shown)


hamersaw:
I really like this idea, will test out over the next week or so! The initial development experience is quite rough and certainly hinders onboarding new developers. I think this is a great first step, just want to make sure it's in the right direction. It's easier to get things right now than relearn in the future and cleanup the bloat that comes with committing a changing dev env continually.

@pingsutw wondering if you had any input here? When implementing single binary what did your workflow look like?

script/setup_local_dev.sh
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env bash
+
+: "${ENSURE_K3D:="false"}" # will ensure k3d is installed if set to "true"
+: "${ENSURE_KUBECTL:="false"}" # will ensure kubectl is installed if set to "true"
+: "${ENSURE_HELM:="false"}" # will ensure helm is installed if set to "true"
+
+: "${K3D_VERSION:="v4.4.8"}" # version of k3d to install, empty value uses latest one available. currently pinned to latest known working (due to unresolved port mapping issue: https://github.com/k3d-io/k3d/issues/380)
+: "${K3D_INSTALL_URL:="https://raw.githubusercontent.com/rancher/k3d/main/install.sh"}" # URL to k3d installer script
+: "${K3S_VERSION:="v1.22.8-k3s1"}" # version of k3s to run in k3d cluster, empty value uses default specified by k3d install
```
pingsutw, line 9: Could we use `v.1.21.1`? just make it consistent with demo and sandbox. https://github.com/flyteorg/flyte/blob/b7928c075225843d097a1c598ccd7f51b2e87069/docker/sandbox-lite/Dockerfile#L27

pingsutw (commented):
I love the idea too, we need a basic container that only has Postgres and minio. it can make local development much easier. 

In local development, I used to start a new sandbox cluster and use a single binary to start another flyte cluster locally with the config flyte_local.yaml, but it always required a large amount of resources.

btw, I just tested this script, and it works for me, thanks!!

script/setup_local_dev.sh
MorpheusXAUT: Sure thing, updated :+1: Sorry, I must've missed the `sandbox-lite` image used that version when skimming over the Dockerfile

script/setup_local_dev.sh
```diff
@@ -0,0 +1,171 @@
...
+: "${HELM_VERSION:=""}" # version of helm to install, empty value uses latest v3 available
+: "${HELM_INSTALL_URL:="https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3"}" # URL to helm installer script
+: "${HELM_K8S_NAMESPACE:="flyte"}" # k8s namespace to use with helm
+
+: "${FLYTE_TIMEOUT:=600}"
+
+set -e
+
+HAS_CURL="$(type "curl" &> /dev/null && echo true || echo false)"
+HAS_WGET="$(type "curl" &> /dev/null && echo true || echo false)"
```
hamersaw, line 27: looks like this checks for `curl` rather than `wget`

script/setup_local_dev.sh
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env bash
+
+: "${ENSURE_K3D:="false"}" # will ensure k3d is installed if set to "true"
+: "${ENSURE_KUBECTL:="false"}" # will ensure kubectl is installed if set to "true"
+: "${ENSURE_HELM:="false"}" # will ensure helm is installed if set to "true"
+
+: "${K3D_VERSION:="v4.4.8"}" # version of k3d to install, empty value uses latest one available. currently pinned to latest known working (due to unresolved port mapping issue: https://github.com/k3d-io/k3d/issues/380)
```
hamersaw, line 7: So I updated to using 5.4.1 with the following command:
`k3d cluster create -p "30082:30082@server:0" -p "30084:30084@server:0" -p "30089:30089@server:0" --image rancher/k3s:v1.21.8-k3s2 -v /dev/mapper/nvme0n1p3_crypt:/dev/mapper/nvme0n1p3_crypt flyte`

I think if we just use a single server with no agents and map our forwarded ports to the server instance there is no port mapping issue? Or is this a load-balancer issue? Not a high priority change, but it's nice to use more recent software.

hamersaw (commented):
A few comments to keep this from going stale.

script/setup_local_dev.sh
MorpheusXAUT: Oh, whoops :upside_down_face: copy paste strikes again

script/setup_local_dev.sh
MorpheusXAUT: I have admittedly not tried using newer versions as the old one _just worked_. Will try out the newer version this week to see if I run into any issues - otherwise I'll bump the version as suggested :+1:

MorpheusXAUT: Seems to be working fine for all tests I've done locally, the only downside is running traefik/a loadbalancer although not strictly necessary. I'm fine with that though, since the impact should be non-existent anyways.
I agree that being able to use a newer version of k3d would be preferable :+1:

hamersaw:
> We could also move the k3d config from the script/command line params to a config file, which would allow for easier customization, but would drop just another config file into the repo somewhere...
> 
> One thing I just realized is that this script does not include a `flytectl` setup/config for the k3d cluster. Is there an easy way to have that generated without adding _yet another_ template config to the repo? sweat_smile From what I can tell, we can `flytectl config init` and set the host, but cannot configure e.g. storage to point to minio. Should we generate a flytectl config for this dev cluster as well in this script? If so, I guess the only option would be to add a default config file we can copy to the appropriate dir/set `FLYTECTL_CONFIG`...
As far as I know, there is not an easier way. I prefer to keep the number of dev config files checked in to a minimum, but that's something that we can always easily add later. 

Moving forward it doesn't sound like there is not any strong opposition here. With the new single binary using k3d with the flyte-deps helm chart seems to provide a good initial development environment in terms of ease of setup, iteration speed, etc. There are a number of things that I ran into when setting this up, like manually copying the `app/dist` directory from flyteconsole to compile the single binary and manually creating the flytesnacks-* namespaces (I don't believe single binary handles this). However, these are things that can be added / modified later.

How do you feel about checking this in and then we (I'm happy to spearhead it if necessary) can get an initial document to outline using this to set up a dev environment? It's a great start, but we can't expect to get everything 100% right away or this PR will be open for months. I think then we can easily tweak little things (ex. flytectl config, etc) as we iterate.

MorpheusXAUT:
> As far as I know, there is not an easier way. I prefer to keep the number of dev config files checked in to a minimum, but that's something that we can always easily add later.

Agreed.

> There are a number of things that I ran into when setting this up, like manually copying the `app/dist` directory from flyteconsole to compile the single binary and manually creating the flytesnacks-* namespaces (I don't believe single binary handles this). However, these are things that can be added / modified later.

True, that's also something I ran into, but figured that's something we can add at a later time since we'll have to figure a way to ensure the frontend repo is either available or grab the build from somewhere else.
The namespaces not being registered is something I noticed as well, can be added to the script as well, although I somewhat expected that the single binary would handle this instead.

> How do you feel about merging this and then we (I'm happy to spearhead it if necessary) can get an initial document to outline using this to set up a dev environment? It's a great start, but we can't expect to get everything 100% right away or this PR will be open for months. I think then we can easily tweak little things (ex. flytectl config, etc) as we iterate.

Sounds like a good plan to me, I'm happy to continue improving the dev environment tooling as we continue along and help out with putting together some documentation for first steps, if you'd like.

hamersaw (approved):

PR #2399, opened 2022-04-22
Update pyflyte run commands

wild-endeavor:


cosmicBboy (approved):

cosmicBboy:
we should also update this https://docs.flyte.org/projects/cookbook/en/latest/userguide_setup.html

Issue #1698, opened 2021-10-18
[Housekeeping] Configure grpc_health_prob in admin


evalsocket:
### Describe the issue

Currently, flyteadmin doesn't have a health check prob in deployment. We need a health check prob for flyteadmin. We can use https://github.com/grpc-ecosystem/grpc-health-probe for running a health check

### What if we do not do this?

It's best practice

### Related component(s)

flyteadmin

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

EngHabu:
1. We have installed grpc healthcheck [in boilerplate Dockerfile template](https://github.com/flyteorg/boilerplate/blob/master/boilerplate/flyte/golang_dockerfile/Dockerfile.GoTemplate#L23-L27) but it hasn't been applied to [flyteadmin's Dockerfile](https://github.com/flyteorg/flyteadmin/blob/master/Dockerfile).

2. Docs for gRPC liveness probes: [this](https://cloud.google.com/blog/topics/developers-practitioners/health-checking-your-grpc-servers-gke) and [this](https://kubernetes.io/blog/2018/10/01/health-checking-grpc-servers-on-kubernetes/).

3. We should also add healthchecks for the REST endpoint

PR #2431, opened 2022-04-29
Added livesness and readiness probes for flyteadmin

pmahindrakar-oss:

# TL;DR

Checks for both GRPC service and HTTP service used for REST 

Tested this on sandbox .

Related admin PR https://github.com/flyteorg/flyteadmin/pull/417

Admin PR has to be checkedin first.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
 _How did you fix the bug, make the feature etc. Link to any design docs etc_

## Tracking Issue

https://github.com/flyteorg/flyte/issues/1698

## Follow-up issue
_NA_


charts/flyte-core/templates/admin/deployment.yaml
```diff
@@ -111,6 +111,15 @@ spec:
         - containerPort: 8088
         - containerPort: 8089
         - containerPort: {{ .Values.configmap.adminServer.flyteadmin.profilerPort }}
+        readinessProbe:
+          exec:
+            command: [ "sh", "-c", "reply=$(curl -s -o /dev/null -w %{http_code} http://127.0.0.1:8088/healthcheck); if [ \"$reply\" -lt 200 -o \"$reply\" -ge 400 ]; then exit 1; fi;","grpc_health_probe", "-addr=:8089"]
```
EngHabu, line 116: I'm not quite sure what does this go.. is this a command because that's checking for grpc health? If it's just checking for the http endpoint health, can this be an http probe instead? I think the gke ingress can automatically pickup http probes if I recall correctly.
Or maybe it's it possible to have a probe for each exposed port? Then we can do http for 8088 and a command for 8089?

EngHabu (commented):
Thank you for picking this up ♥️

charts/flyte-core/templates/admin/deployment.yaml
pmahindrakar-oss: The script has two parts
* Checks the health REST endpoint and checks the returns code if between 200 and 400 and passes the check
* Uses the grpc health probe directly

The result combines the two and fails if any of the checks fail 
We can't have probe for each exposed port unfortunately https://github.com/kubernetes/kubernetes/issues/37218
And the suggested way was to use a script

EngHabu (dismissed):

EngHabu (approved):

PR #2484, opened 2022-05-12
Added instruction for cross origin requests

pmahindrakar-oss:

### DIFF
--- a/rsts/deployment/security/security.rst
+++ b/rsts/deployment/security/security.rst
@@ -76,7 +76,7 @@ This can be done by changing the flyteadmin config in the following manner
       kubectl edit deployment flyteconsole -n <flyteconsole-ns>
 
    .. code-block::
-      .......
+
       - env:
         - name: ENABLE_GA
           value: "true"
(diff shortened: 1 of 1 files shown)


rsts/deployment/security/security.rst
```diff
@@ -58,3 +58,56 @@ There are two options to fix this:
...
+
+      kubectl edit deployment flyteconsole -n <flyteconsole-ns>
+
+   .. code-block::
+
+      - env:
+        - name: ENABLE_GA
+          value: "true"
+        - name: GA_TRACKING_ID
+          value: G-0123456789
```
EngHabu, line 84: These don't need to be here, right?

rsts/deployment/security/security.rst
```diff
@@ -58,3 +58,56 @@ There are two options to fix this:
...
+      kubectl edit configmap flyte-admin-config -n <flyteadmin-ns>
+
+   .. code-block::
+
+             security:
+               allowCors: true
+               ......
+               allowedOrigins:
+               - 'https://<flyte-console-domain>'
+               ......
```
EngHabu, line 107: Should we also add instructions for how to modify the helm chart to do this? or is that too much clutter?

rsts/deployment/security/security.rst
pmahindrakar-oss: I have put this in for the context so that the user understand where to put the new variable in. Let me know if there is some other way we can do this

rsts/deployment/security/security.rst
pmahindrakar-oss: yeah was trying to avoid helm instructions here and just refer them to the config directly first . Similar to how we do for auth and then for productionizing they can update the helm charts

EngHabu (approved):

Issue #2486, opened 2022-05-12
[BUG] flytekit fails on python 3.10 on a Mac M1


eapolinario:
### Describe the bug

`grpcio` , one of flytekit's dependencies, does not have a [valid arm64 wheel for python 3.10](https://github.com/grpc/grpc/issues/28387).

This impacts all uses of the library, with the exception of local executions.

The temporary workaround is to force the `grpcio` wheel to be built locally by running:
```
pip install --no-binary :all: grpcio --ignore-installed
```

### Expected behavior

Using a pre-built wheel for a package like `grpcio` should be the default case, in other words, we shouldn't require the user to build any software locally in order to use the flytekit library.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #2487, opened 2022-05-12
Add note about python 3.10 on Mac M1's

eapolinario:
Signed-off-by: Eduardo Apolinario <eapolinario@users.noreply.github.com>

`grpcio` python 3.10 wheel [is mislabeled as universal](https://github.com/flyteorg/flyte/issues/2486), which makes it so that all uses cases in flytekit that require a connection to admin are going to fail.

Here's how it renders:

image

The next step is to figure out a way of doing this automatically in flytekit itself, but just to unblock M1 users, let's add a note in the docs.


kumare3:
Can we make this collapsible

kumare3 (approved):

PR #2530, opened 2022-05-24
Helm chart fixes for multi-cluster deployments

MorpheusXAUT:
As already briefly discussed on Slack, this PR includes some fixes to the `flyte-core` Helm chart to enable multi-cluster deployments.

The main issue presented was the incorrect `cluster` config being applied, however some of the init containers (mainly `generate-secrets`) refuse to work with multiple clusters configured.
I have thus split the ConfigMap into two parts (`flyte-admin-base-config` contains everything except clusters, `flyte-admin-clusters-config` just contains the multi-cluster configuration) which get mounted as required. Although some other init containers might work with cluster configuration (by simply ignoring it), I opted to only mount the cluster config as required (since that also includes mounting the secrets used for authenticating) - happy to change that to always include the cluster config unless it _cannot_ be included if you'd prefer that.

I'm still a bit unsure about `generate-secrets` specifically - in our Slack conversation, there was a bit of a mix-up whether we need to run it once (on the control plane) or multiple times (for each data plan cluster)... For now, it's being run once, which seems to have worked fine for the two-cluster setup I was running.
Since `generate-secret` refuses to start if more than one cluster is specified, we'd have to come up with a way to run it multiple times with a separate cluster configured every time as multiple init containers - or adapt `flyteadmin generate-secrets` accordingly...

Additionally, some smaller additions/refactoring that happened during our deployment testing was added to this PR as well, rest of the image tag updated etc. were generated by `make kustomize && make helm`.


katrogan (dismissed):
LGTM, @pmahindrakar-oss do you mind taking a look?

pmahindrakar-oss (dismissed):

MorpheusXAUT:
> LGTM to me too . @MorpheusXAUT can you check on the merge conflicts

Sure thing, will push an update shortly!

MorpheusXAUT:
@pmahindrakar-oss Rebased onto `master` and ran `make kustomize & make helm` again, should be free of conflicts now 👍

pmahindrakar-oss (approved):

evalsocket (approved):

PR #2566, opened 2022-06-01
remove lb

brucearctor:
I actually didn't do 100% checking ... but ... are these Load Balancers needed? 

[ they cause problems with opta, and at a first pass I don't see why they are required ]

### DIFF
--- a/charts/flyte-core/values-eks.yaml
+++ b/charts/flyte-core/values-eks.yaml
@@ -30,17 +30,6 @@ flyteadmin:
       cpu: 50m
       ephemeral-storage: 200Mi
       memory: 200Mi
-
-  service:
-    annotations:
-      service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "600"
-      # TODO alter domain
-      external-dns.alpha.kubernetes.io/hostname: "flyte.example.com"
-    type: LoadBalancer
-    loadBalancerSourceRanges:
-      # TODO change source ip range if desired
-      - 0.0.0.0/0
-
   affinity:
     podAntiAffinity:
       requiredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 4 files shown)


EngHabu:
These are not needed..

@evalsocket, I recall you went ahead and cleaned these up? was that for a specific provider?

brucearctor:
@evalsocket -- ran `make helm` which looks like created more failures :-)

evalsocket (approved):
Thank you

brucearctor:
Needs someone to merge ( since I don't have permissions, even though @evalsocket already approved ).  

That invites the question -- what are merge policies for Flyte?  I don't see documentation specifying those criteria.

EngHabu:
Had to rerun the WFs,... seems there was an issue earlier today with github actions (based on this https://www.githubstatus.com/incidents/8xxms3q3qzgp) not sure if's related

jw0515:
This breaks the https://docs.flyte.org/en/latest/deployment/aws/manual.html#connecting-to-flyte

Now we will have to add: #      alb.ingress.kubernetes.io/target-type: 'ip' in the ingress annotation

@brucearctor 
@EngHabu

brucearctor:
@jw0515 -- based on my experience, that change did not break the UI, flytectl, nor flyteconfig/pyflyte.  But, that doesn't mean that it didn't create a problem outside of my workflows.

jw0515:
> @jw0515 -- based on my experience, that change did not break the UI, flytectl, nor flyteconfig/pyflyte. But, that doesn't mean that it didn't create a problem outside of my workflows [ hence the reason for more tests, and specifically for the steps in the deployment guide ].

Thanks for your reply! Yes the thing is ingress creation will have some issue:
when I do `kubectl -n flyte get ingress`, I won't get any address

brucearctor:
I wonder whether another way to 'solve' your issue might be updating docs?  

I think you can use the loadbalancer's DNS, and `kubectl -n flyte get ingress` ingress not required.

jw0515:
In this case the AWS ALB won't be created at all...so no DNS

brucearctor:
hmmm; I was working with https://docs.flyte.org/en/latest/deployment/aws/opta.html ( even though terraform > opta, opta was the pre-existing committment ).  And working with that, LB gets created.  So, again, probably a docs thing.  

I imagine it'll be better if there are just ready-to-go terraform modules.  But, seems worth also determining whether manual ( ex: by `aws ...` commands also needs to be updated, and automated tests written and maintained

PR #2572, opened 2022-06-02
added single binary component update in release ci

evalsocket:
fix: https://github.com/flyteorg/flyte/compare/


katrogan (dismissed):
awesome, thank you!

PR #2581, opened 2022-06-04
Updated link of swagger UI

pingsutw:
https://github.com/flyteorg/flyteadmin/blob/77b83f64ad06a1ddbf68d2e0bb2b8e3badf89d2d/pkg/server/service.go#L158

### DIFF
--- a/rsts/reference/swagger.rst
+++ b/rsts/reference/swagger.rst
@@ -29,4 +29,4 @@ Once sandbox setup is complete, a ready-to-explore message is shown:
    👨‍💻 Flyte is ready! Flyte UI is available at http://localhost:30081/console 🚀 🚀 🎉 
 
 
-Visit http://localhost:30081/openapi to view swagger UI for the service and documentation of the payload fields. To test, send sample queries!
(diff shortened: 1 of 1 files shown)


rsts/reference/swagger.rst
```diff
@@ -29,4 +29,4 @@ Once sandbox setup is complete, a ready-to-explore message is shown:
    👨‍💻 Flyte is ready! Flyte UI is available at http://localhost:30081/console 🚀 🚀 🎉 
 
 
-Visit http://localhost:30081/openapi to view swagger UI for the service and documentation of the payload fields. To test, send sample queries!
+Visit http://localhost:30081/api/v1/openapi to view swagger UI for the service and documentation of the payload fields. To test, send sample queries!
```
sbrunk, line 32: @pingsutw http://localhost:30081/api/v1/openapi serves the actual openapi spec as json, not the redoc UI. http://localhost:30081/openapi causes a redirect to http://localhost:8087/openapi/, which is not available without port-forwarding. But if we add a slash at the end (http://localhost:30081/openapi/), it seems to be working.

```suggestion
Visit http://localhost:30081/openapi/ to view swagger UI for the service and documentation of the payload fields. To test, send sample queries!
```

PR #2633, opened 2022-06-27
RFC for Eviction of cached task outputs

MorpheusXAUT:
https://hackmd.io/qOztkaj4Rb6ypodvGEowAg?view

Comments already present on the HackMD doc are from our internal team and have been left in for clarification/further discussion.

[Initial discussion on Slack](https://app.slack.com/client/TN89P6GGK/CNMKCU6FR/thread/CNMKCU6FR-1654169506.626249)


paulbes:
> to perform housekeeping and clean up their cache, cleanly removing the cached data of previously executions that are no longer relevant executions

We also see a use case for this cache eviction functionality from a privacy perspective; the right to be forgotten. 

GDPR (30 days to complete a deletion request):
> The data subject shall have the right to obtain from the controller the erasure of personal data concerning him or her without undue delay and the controller shall have the obligation to erase personal data without undue delay.

CCPA (45 days to complete a deletion request):
> A consumer shall have the right to request that a business delete any personal information about the consumer which the business has collected from the consumer.

When processing personally identifiable information (PII) to adhere to, e.g., GDPR, we generally only retain data for 30 days in our workflows. We do this to ensure that any deletion requests from users whose data we are processing will be fulfilled "automagically" within the expected timeframe for the "right to be forgotten".

With the cache eviction API we would be able to build a system that could evict the cache for certain types of workflows or tasks within a given timeframe.

With that said, and in fear of adding scope creep to this RFC, it might be even better to have the ability to set a TTL on the cache as an attribute, e.g., 

```python
 # Might make sense to only allow for a quite coarse-granularity
cache_ttl_hours = 30 * 24
cache_ttl_days = 30
```

MorpheusXAUT:
@paulbes Interestingly enough, we talked about automatic expiration of cached values after a certain timespan internally just yesterday :smile: Although our concern was mainly due to a housekeeping/cache size perspective rather than GDPR related, I agree it could be useful for that as well.

> With that said, and in fear of adding scope creep to this RFC, it might be even better to have the ability to set a TTL on the cache as an attribute, e.g.,
> 
> ```python
>  # Might make sense to only allow for a quite coarse-granularity
> cache_ttl_hours = 30 * 24
> cache_ttl_days = 30
> ```

I agree that'd be great to have, but I'm not sure either if we want to include it in this RFC or keep it separate/add afterwards so we don't extend the scope too much...
If people feel we should include this addition as well, I can take another look at what we've discussed internally and adapt the RFC accordingly.

hamersaw:
> @paulbes Interestingly enough, we talked about automatic expiration of cached values after a certain timespan internally just yesterday smile Our discussion was mainly focused on a housekeeping/cache size perspective rather than GDPR related, but I agree it could be useful for that as well.
> 
> > With that said, and in fear of adding scope creep to this RFC, it might be even better to have the ability to set a TTL on the cache as an attribute, e.g.,
> > ```python
> >  # Might make sense to only allow for a quite coarse-granularity
> > cache_ttl_hours = 30 * 24
> > cache_ttl_days = 30
> > ```
> 
> I agree that'd be great to have, but I'm not sure either if we want to include it in this RFC or keep it separate/add afterwards so we don't extend the scope too much... If people feel we should include this addition as well, I can take another look at what we've discussed internally and adapt the RFC accordingly.

A few thoughts on this. I know we have discussed including a `cache_expiration` parameter onto cached tasks similar to the `cache_version` and `cache_serialized` that current exist. That would simply define a duration after which the cache is expired (e.x. `cache_expiration=45d`) and overwrite the cache if we find it beyond expiration. It doesn't sound like the goals of either use case discussed (i.e. (1) general cache cleanup to reduce cache size or (2) GDPR forget) are covered with that solution because nothing is deleted until a new task is executed to overwrite the expired value.

I think implementation of a cache eviction API does open possibilities for automated eviction. To support automated eviction it sounds like a separate service as @paulbes suggested or an additional component into one of the Flyte core services that periodically scrapes and GCs the cache is required if I am understanding this correctly? IMO it is a deep enough topic to require a separate discussion.

I think it could be a very useful feature, maybe open a new issue to track it?

rfc/system/2633-eviction-of-cached-task-outputs.md
```diff
@@ -0,0 +1,125 @@
...
+
+The potential for malicious exploitation is deemed non-existant as no access to any data is provided and most additional communication will be performed intra-cluster. An attacker could potentially cause additional computational costs by removing cached outputs.
+
+## 8 Unresolved questions
+
+1. When should cache eviction happen during a repeated execution?
+   Should we evict the cache entry immediately before starting task execution, potentially leaving no cached output available if the execution fails, or postpone the removal until the task has finished (successfully) and new data can be stored right away?
+2. Should the `cache_override` flag also be added to `flytekit`'s task decorator?
+   This would allow users to define tasks/workflows which will automatically evict their cached results, however does not strictly fit with the actual task "definition".
+3. Which Flyte tools (`flyteconsole`/`flytectl`) should support the proposed `AdminService` API extension for `flyteadmin`, if any?
```
pmahindrakar-oss, line 121: Basic support to do this from flytectl can exist using `flytectl update execution <task/workflow execution identifier> --evictCache` which can call the new API's. 

And also if we do support this through execution existing archival then also this flag can be passed as part ExecutionUpdateRequest

MorpheusXAUT:
> A few thoughts on this. I know we have discussed including a `cache_expiration` parameter onto cached tasks similar to the `cache_version` and `cache_serialized` that current exist. That would simply define a duration after which the cache is expired (e.x. `cache_expiration=45d`) and overwrite the cache if we find it beyond expiration. It doesn't sound like the goals of either use case discussed (i.e. (1) general cache cleanup to reduce cache size or (2) GDPR forget) are covered with that solution because nothing is deleted until a new task is executed to overwrite the expired value.
> 
> I think implementation of a cache eviction API does open possibilities for automated eviction. To support automated eviction it sounds like a separate service as @paulbes suggested or an additional component into one of the Flyte core services that periodically scrapes and GCs the cache is required if I am understanding this correctly? IMO it is a deep enough topic to require a separate discussion.

Yes, you're correct, the addition mentioned by @paulbes (and the one we talked about internally) would require an additional service/component within Flyte to periodically check all cached entries and evict them should they have cross a certain expiry threshold.
Just adding a `cache_expiration` parameter to the task could also be useful, but doesn't solve the issue we're trying to solve per se.

> I think it could be a very useful feature, maybe open a new issue to track it?

I agree, I'll open another issue to track this idea once the RFC gets accepted! Would definitely be a useful feature to have in Flyte itself, I'd say.

rfc/system/2633-eviction-of-cached-task-outputs.md
```diff
@@ -0,0 +1,125 @@
...
+
+At the moment, the only way to force Flyte to ignore the (faulty) cached value of a previous execution during a re-run of the workflow would be to change the `cache_version` of the affected task, causing all other calculations to be performed again as well even though their results might've been correct in the first place.
+
+Depending on the workload executed, this might result in unnecessary and potentially high costs and time requirements to re-process. Evicting some of the cached values would allow re-runs of previously executed workflows to leverage caching where desirable while still re-calculating some of the data that has been deemed "invalid".
+
+In addition, exposing cache eviction functionality via a `flyteadmin` API call would allow Flyte users/administrators to perform housekeeping and clean up their cache, cleanly removing the cached data of previously executions that are no longer relevant executions and would otherwise incur continuous storage costs. By exposing the same functionality used during the cache eviction override, we can ensure the cache maintained by `datacatalog` is consistent with the underlying blob storage.
```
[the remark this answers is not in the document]
MorpheusXAUT: Whilst we didn't necessarily think about archive executions in the first place, you've definitely got a valid point there. I like the suggestion of optionally evicting the cache of tasks when archiving them, could also make it a default and add an opt-out flag (less clicking-effort when archiving multiple ones 😁 ).

We very rarely use the archiving function of Flyte internally. When talking about "no longer relevant" executions, I originally meant past executions that (while not being archived) are not relevant to any future calculations anymore because e.g. their input dataset turned out to be unusable.
At the moment, we're still keeping those around unarchived because their workflow/task is still relevant, however the output of certain executions is no longer of interest to us and basically just eats up blob storage space.

I can add cache eviction on archival to the draft as I believe it could be a useful addition that should be very little effort to include from the beginning, thanks for the suggestion, didn't think about that 🙂

rfc/system/2633-eviction-of-cached-task-outputs.md
MorpheusXAUT: > Basic support to do this from flytectl can exist using `flytectl update execution <task/workflow execution identifier> --evictCache` which can call the new API's.

Sounds like a good plan, that'll expose the new functionality in an easily accessible way that would allow basic scripting as well 👍 

> And also if we do support this through execution existing archival then also this flag can be passed as part ExecutionUpdateRequest

Sure, can do. Would you rather have that included in the existing `ExecutionUpdateRequest` only, not adding a new API at all? Or should we include it, but still add a separate endpoint dedicated to evicting the cache of a past execution of a workflow/task?

pmahindrakar-oss: > Would you rather have that included in the existing ExecutionUpdateRequest only, not adding a new API at all? Or should we include it, but still add a separate endpoint dedicated to evicting the cache of a past execution of a workflow/task?

I think we dont have ExecutionUpdateRequest for TaskExecutionIdentifier and only exists for WorkflowExecutionIdentifier , so we might need a separate API. Adding @katrogan for suggestion on this.

rfc/system/2633-eviction-of-cached-task-outputs.md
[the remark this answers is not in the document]
pmahindrakar-oss: Adding @katrogan

MorpheusXAUT:
> Do we see [Intratask Checkpoints](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/control_flow/checkpoint.html) as a cache? If so it might make sense to include them as part of an eviction.

@sbrunk Good point :thinking: Yes, we should probably also clear out all these values when we're evicting the cache for a task...

Since at least some of the handling is done outside flyteadmin/propeller's context (inside the actual code executed), we might have to remove these values _before_ the execution (instead of afterwards, as suggested for the cached output).
Alternatively, we could try to pass along the `cache_override` flag to flytekit/the `Checkpointer` somehow to have it skip any available entries, but I'm not sure how feasible that is...

rfc/system/2633-eviction-of-cached-task-outputs.md
katrogan: we might want an equivalent api method for task executions

MorpheusXAUT: > we might want an equivalent api method for task executions

You're talking about the equivalent of an `ExecutionUpdateRequest` API for task executions, correct?
If both are available and we add a flag to these calls, I agree that'd be preferred to introducing completely new, standalone APIs.

rfc/system/2633-eviction-of-cached-task-outputs.md
```diff
@@ -0,0 +1,125 @@
...
+
+Similar to the `interruptible` override flag provided for a single execution of a workflow or task, we propose adding a flag (e.g. `cache_override`) to evict the cached output of an execution and force its calculation to be performed (and cached) again.  
+This cleanup would be performed automatically by Flyte during the execution of a new workflow or task, returning the updated results:
+
+![cache eviction flowchart](https://i.imgur.com/rsJIlWt.png)
+
+* During the execution preparation, `flytepropeller` would query `datacatalog` as before, trying to find the cached output for a task and its inputs.
+* If no cache entry was found, the current behaviour remains the same - the task is executed and its result is stored (if caching has been enabled for the task).
+* Should a cache hit occur, `flytepropeller` would check the `cache_override` flag of the execution config:
+    * If the override has been set, `flytepropeller` would skip retrieving the entry from `datacatalog` and continue execution, removing the cached data before storing the updated results after the task finished successfully.
```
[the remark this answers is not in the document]
MorpheusXAUT: > question: why issue the call to remove existing data altogether? can't we just acquire the lease and write to the same key again and avoid a call?

Yes, that's a good point, we could actually just overwrite the existing data. That'd actually save us a blob storage call and make it less susceptible to interruptions 👍 

> what if a system failure occurs after we evict and before we can write the new contents of the cache?

In that case, we'd not have a cached result for the next execution and would have to run executions again. I'm not sure if that'd be preferred over still having the old result available, as we'd _make sure_ to run the task again (the next execution might not set the `cache_override` flag again), but it's also slightly bordering into undefined behaviour 😅 

I'd agree only evicting the cached entry if we have successful execution + reducing it to a single update call would be preferable and less inconsistent 👍

katrogan:
cc @kumare3 for the intra-task checkpointing bits

rfc/system/2633-eviction-of-cached-task-outputs.md
katrogan: @MorpheusXAUT yes, we could make the [id field](https://github.com/flyteorg/flyteidl/blob/e749b82a70027d3364f4e5b98ce2cfbffb4eef54/protos/flyteidl/admin/execution.proto#L353) a oneof and have it optionally be a task exec identifier although this is a stylistic departure from our other endpoints where they principally perform CRUD operation on the entity specified in the message name

MorpheusXAUT: Hmmm, I see. Yeah, that could work, but I'm not sure it's the best way to go forward if it's the only endpoint that'll have this "special treatment" and deviates from the rest... I'd prefer not breaking conventions just for the sake of using an endpoint for something it wasn't originally built to do :sweat_smile:

hamersaw:
> @kumare3 Any comment on the [Intratask Checkpoints](https://github.com/flyteorg/flyte/pull/2633#issuecomment-1176632765) topic? I believe we should remove/skip those values as well, but not sure what the best way to do so would be?

Intra-task checkpoints should only be applied during consecutive retries of a task within the same workflow execution. So if there was a new workflow execution with the `cache_override` parameter set, existing intra-task checkpoints from previous workflow executions would not affect correctness. If the concern is space and the goal is to delete the checkpoints this is another issue, and I think supporting this functionality is far beyond the scope of this proposal.

MorpheusXAUT:
> Intra-task checkpoints should only be applied during consecutive retries of a task within the same workflow execution. So if there was a new workflow execution with the `cache_override` parameter set, existing intra-task checkpoints from previous workflow executions would not affect correctness. If the concern is space and the goal is to delete the checkpoints this is another issue, and I think supporting this functionality is far beyond the scope of this proposal.

Ah, great, didn't know about that, thanks for the info @hamersaw :+1: 

I agree we should keep it out of the RFC, if possible, then. Might be worth to add a follow up issue for the sake of completeness when cleaning up cached data, but otherwise we might be increasing an already quite extended scope even more.

kumare3:
Checkpoints are localized to a single execution today not shared through cache - maybe we should do that in the future

pmahindrakar-oss:
> @katrogan @pmahindrakar-oss I've added some of the comments from this thread to the RFC doc. Please take another look at the [changes](https://github.com/flyteorg/flyte/pull/2633/commits/dd8abf67474e6e9c7b8479fef5bf9e17d1b4784b) to see if they satisfy your comments/questions or if there's something we should clarify in more detail.

Look good to me @MorpheusXAUT for the suggested changes

rfc/system/2633-eviction-of-cached-task-outputs.md
```diff
@@ -0,0 +1,125 @@
...
+Similar to the `interruptible` override currently available, we propose to add an optional execution launch setting to `flyteconsole` to easily trigger a single task to be executed again using identical inputs.
+
+### Evicting cache of a past execution
+
+In addition to the cache eviction override while launching a workflow, we propose to expose the same functionality via `flyteadmin` as an extension to the existing `AdminService`:
+
+* Similar to `GetExecutionData`, provide an endpoint such as `EvictExecutionCache`, removing all cached outputs of all tasks within the specified workflow execution.
+    * The workflow execution would be identified using a `WorkflowExecutionIdentifier`.
+    * Execution of the workflow must have already concluded.
+    * `flyteadmin` would automatically iterate all tasks included in the workflow and recursively clean up cached data as necessary.
```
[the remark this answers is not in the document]
katrogan: > about dynamic tasks, are we able to list them all within admin/datacatalog so we can retrieve their results?

Yes! this is how we render the nodes the UI. as you iterate through the nodes in a workflow, if you come across one that [is_dynamic](https://github.com/flyteorg/flyteidl/blob/aded40037d1a4d2ea32c6dd18644ccf16072124f/protos/flyteidl/admin/node_execution.proto#L116) or [is a workflow node](https://github.com/flyteorg/flyteidl/blob/aded40037d1a4d2ea32c6dd18644ccf16072124f/protos/flyteidl/admin/node_execution.proto#L108) you can call https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/service/admin.proto#L343 to list the child node executions and recursively evict the corresponding cache contents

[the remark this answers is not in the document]
katrogan: I think partial failures are acceptable as long as we propagate the overall failure back to the uesr and they can retry with idempotency

rfc/system/2633-eviction-of-cached-task-outputs.md
katrogan: 👍 for adding an equivalent method for task execution entities, would also prefer not to break convention.

katrogan (dismissed):
LGTM. Can we make sure to update to include details on recursively evicting dynamic and subworkflow nodes?

MorpheusXAUT:
> LGTM. Can we make sure to update to include details on recursively evicting dynamic and subworkflow nodes?

Sure thing, will update tomorrow morning 👍

MorpheusXAUT:
@katrogan Added some details about dynamic/workflow nodes and partial failures and slightly re-arranged the doc to emphasize we prefer extending the existing endpoints/adding a new `UpdateTaskExecution` instead of implementing completely new/independent endpoints.

hamersaw (approved):

katrogan (approved):

katrogan:
thank you @MorpheusXAUT looks great!

PR #2746, opened 2022-08-09
values file addition for on-prem k8s instance with ingress and tls

samuel-sujith:
Added values file for on-prem k8s installation of flyte with IDP user and app auth

### DIFF
--- a/charts/flyte-core/values-onprem.yaml
+++ b/charts/flyte-core/values-onprem.yaml
@@ -1,0 +1,841 @@
+# ---------------------------------------------------------------------
+# Core System settings
+# This section consists of Core components of Flyte and their deployment
+# settings. This includes FlyteAdmin service, Datacatalog, FlytePropeller and
+# Flyteconsole
+# ---------------------------------------------------------------------
+
+#
+# FLYTEADMIN SETTINGS
+#
+
+flyteadmin:
+  enabled: true
+  # -- Replicas count for Flyteadmin deployment
+  replicaCount: 1
+  image:
+    # -- Docker image for Flyteadmin deployment
+    repository: cr.flyte.org/flyteorg/flyteadmin # FLYTEADMIN_IMAGE
+    tag: v1.1.29-hotfix # FLYTEADMIN_TAG
+    pullPolicy: IfNotPresent
+  # -- Additional flyteadmin container environment variables
+  #
+  # e.g. SendGrid's API key
+  #  - name: SENDGRID_API_KEY
+  #    value: "<your sendgrid api key>"
+  #
+  # e.g. secret environment variable (you can combine it with .additionalVolumes):
+  # - name: SENDGRID_API_KEY
+  #   valueFrom:
+  #     secretKeyRef:
+  #       name: sendgrid-secret
+  #       key: api_key
+  env: []
+  # -- Default resources requests and limits for Flyteadmin deployment
+  resources:
+    limits:
+      cpu: 250m
+      ephemeral-storage: 100Mi
+      memory: 500Mi
+    requests:
+      cpu: 10m
+      ephemeral-storage: 50Mi
+      memory: 50Mi
+  # -- Default regex string for searching configuration files
+  configPath: /etc/flyte/config/*.yaml
+  # -- Initial projects to create
+  initialProjects:
+    - flytesnacks
+    - flytetester
+    - flyteexamples
+  # -- Service settings for Flyteadmin
+  service:
+    annotations:
+      projectcontour.io/upstream-protocol.h2c: grpc
+    type: ClusterIP
+    loadBalancerSourceRanges: []
+  # -- Configuration for service accounts for FlyteAdmin
+  serviceAccount:
+    # -- Should a service account be created for flyteadmin
+    create: true
+    # -- Should a service account always be created for flyteadmin even without an actual flyteadmin deployment running (e.g. for multi-cluster setups)
+    alwaysCreate: false
+    # -- Annotations for ServiceAccount attached to Flyteadmin pods
+    annotations: {}
+    # -- ImagePullSecrets to automatically assign to the service account
+    imagePullSecrets: []
+    # -- Should a ClusterRole be created for Flyteadmin
+    createClusterRole: true
+  # -- Annotations for Flyteadmin pods
+  podAnnotations: {}
+  # -- nodeSelector for Flyteadmin deployment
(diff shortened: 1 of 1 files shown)


charts/flyte-core/values-onprem.yaml
```diff
@@ -0,0 +1,841 @@
...
+# Common secret auth for propeller & scheduler
+#
+
+secrets:
+  adminOauthClientCredentials:
+    # -- If enabled is true, helm will create and manage `flyte-secret-auth` and populate it with `clientSecret`.
+    # If enabled is false, it's up to the user to create `flyte-secret-auth` as described in
+    # https://docs.flyte.org/en/latest/deployment/cluster_config/auth_setup.html#oauth2-authorization-server
+    enabled: true
+    clientSecret: "KjxCqOLkbw3rwrMjDatZmBI9B1zkLuoF" # put the secret for the confidential client flytepropeller defined in the IDP
```
pmahindrakar-oss, line 297: We can keep a dummy value here

charts/flyte-core/values-onprem.yaml
```diff
@@ -0,0 +1,841 @@
...
+  # -- affinity for Flyteadmin deployment
+  affinity: {}
+  secrets: {}
+  additionalVolumes: 
+    - configMap:
+        name: sslcerts
+      name: sslcerts
+  additionalVolumeMounts: 
+    - mountPath: /etc/ssl/certs/
+      name: sslcerts
```
pmahindrakar-oss, line 84: Can you add a few comments regarding this config and how is this different than the L78-L81

charts/flyte-core/values-onprem.yaml
```diff
@@ -0,0 +1,841 @@
...
+      username: postgres
+      password: password
+      host: postgres-flyte
+      dbname: "datacatalog"
+  admin:
+    database:
+      port: 5432
+      username: postgres
+      password: password
+      host: postgresflyte
```
pmahindrakar-oss, line 406: Are you using different host for admin and datacatalog (L399)

charts/flyte-core/values-onprem.yaml
```diff
@@ -0,0 +1,841 @@
...
+#        certPath: "/path/to/testcluster3/cert"
+
+  # -- Configuration for Flyte console UI
+  console:
+    BASE_URL: /console
+    CONFIG_DIR: /etc/flyte/config
+  
+  logger:
+    show-source: true
+    level: 6
```
pmahindrakar-oss, line 457: This config seems to be duplicated here https://github.com/flyteorg/flyte/pull/2746/files#diff-20cdbbc75e0c9937a9f82aa3318444cc58ede71498cb7c3ad67f44a42b38aec1R681

charts/flyte-core/values-onprem.yaml
```diff
@@ -0,0 +1,841 @@
...
+        - http://flyteadmin.flyte.svc.cluster.local:80 #assuming that flyte is installed in flyte namespace
+
+      # Controls app authentication config
+      appAuth:
+
+        authServerType: External
+        externalAuthServer:
+          allowedAudience: flytepropeller
+          baseUrl: "" # populate this value only if different from userauth baseUrl
+          metadataUrl: ".well-known/openid-configuration"
```
pmahindrakar-oss, line 522: this is specific to provider so would be good if we can mention that here. Currently we only saw this required for keycloak.

charts/flyte-core/values-onprem.yaml
```diff
@@ -0,0 +1,841 @@
...
+
+
+        thirdPartyConfig:
+          flyteClient:
+            audience: flytepropeller
+            clientId: flytepropeller
```
pmahindrakar-oss, line 528: This is only required in case flyteClients are being run from a machine which doesn't have browser access. 
We can mention this info here and also have an alternate config in comments which shows with browser access.

pmahindrakar-oss (commented):
I think this is more catered to keycloak  and flyteclient without browser access than being for generic on-prem . Can we instead rename the file to values-keycloak-idp-flyteclients-without-browser.yaml .

Please let me know if you think otherwise

charts/flyte-core/values-onprem.yaml
```diff
@@ -0,0 +1,841 @@
...
+          domain: "<flytedomain>" # add here the sitename for flyte
+          sameSitePolicy: DefaultMode
+        openId:
+          baseUrl: "<IDP URL for the realm>"
+          scopes:
+            - profile
+            - openid
+            - offline_access
+          clientId: <clientid declared in IDP for the UI access> 
+          clientSecretName: <client secret of the above client from the IDP>
```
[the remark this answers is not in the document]
samuel-sujith: Yes I htink we can rename to reflect eh keycloak IDP and yes the userauth does not need the client secret, it is picked up from the oidc_client_secret.

charts/flyte-core/values-onprem.yaml
```diff
@@ -0,0 +1,841 @@
...
+      metricsScope: "flyte:"
+      metadataStoragePrefix:
+        - "metadata"
+        - "admin"
+      eventVersion: 2
+      testing:
+        host: http://flyteadmin
+    # -- Authentication configuration
+    auth:
+      grpcAuthorizationHeader: "flyte-authorization"
```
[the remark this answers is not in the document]
samuel-sujith: Yes

charts/flyte-core/values-onprem.yaml
samuel-sujith: Yes

charts/flyte-core/values-onprem.yaml
samuel-sujith: Ok  I will do

charts/flyte-core/values-onprem.yaml
samuel-sujith: No its the same. Let me reconfirm

charts/flyte-core/values-onprem.yaml
samuel-sujith: Yes, I will mention

charts/flyte-core/values-onprem.yaml
samuel-sujith: yes

pmahindrakar-oss:
Can we keep the changes in same  PR @samuel-sujith   since it easier to track the feedback and its related changes

Issue #1763, opened 2021-10-29
[Docs] Document Flyte Resource Manager and its usage


kumare3:
### Description

Flyte has a powerful feature called FlyteResourceManager, which can be used to globally throttle requests to any downstream service. This needs to be documented and configuration explained.

This is why the REDIS instance is used, and wihtout this documentation users have no idea how to best use the Redis.

Moreover, this documentation is extremely pertinent to Map/Array jobs.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

SmritiSatyanV:
[Discussion](https://unionai.slack.com/archives/C01H0FN1NJX/p1658766729399059)

PR #2770, opened 2022-08-16
add ResourceManager doc

SmritiSatyanV:
[Issue](https://github.com/flyteorg/flyte/issues/1763)
Add ResourceManager docs, explain how resources are allocated and deallocated.
Signed-off-by: SmritiSatyanV smriti@union.ai

### DIFF
--- a/rsts/deployment/cluster_config/customizable_resources.rst
+++ b/rsts/deployment/cluster_config/customizable_resources.rst
@@ -87,3 +87,106 @@ Flytekit
 For convenience, add a FlyteCTL wrapper to update the new attributes. Refer to `this PR <https://github.com/flyteorg/flytectl/pull/65>`__ for the entire set of changes required.
 
 That's it! You now have a new matchable attribute to configure as the needs of your users evolve.
+
+Flyte ResourceManager
+---------------------
+
+**Flyte ResourceManager** is a configurable component that allows plugins to manage resource allocations independently. It helps track resource utilization of tasks that run on Flyte. The default deployments are configured as ``noop``, which means they are disabled and you would fallback to the platform specific resource management, like K8s, which tells you that you can't schedule more tasks.
+
+The below attribute is configurable within FlytePropeller, which can be disabled with:
+
+.. code-block:: yaml
+
+    resourcemanager:
+      type: noop
+
+The ResourceManager provides a task-type-specific pooling system for Flyte tasks. Optionally, plugin writers can request resource allocation in their tasks.
+
+A plugin defines a collection of resource pools using its configuration. Flyte uses tokens as a placeholder to represent a unit of resource.
+
+How does a Flyte plugin request for resources?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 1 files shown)


rsts/deployment/cluster_config/customizable_resources.rst
```diff
@@ -86,4 +86,107 @@ Flytekit
...
+During runtime, the ResourceManager:
+
+#. Allocates tokens to the plugin;
+#. Releases tokens once the task is completed.
+
+How are resources allocated?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+You can request resources using **ResourceName**, which is a unique token and a fully qualified **resource request** (which is typically an integer). If the resource pool has sufficient capacity to fulfil your request, then the resources requested are allocated, and the plugin proceeds further.
+
+When a Flyte task execution needs to send a request to an external service, the plugin claims a unit of the corresponding resource. This is done using a ResourceName, and a fully qualified resource request. The execution generates this unique token and registers this token with the ResourceManager by calling the ResourceManager’s **AllocateResource** function. The ResourceManager inspects the current utilization of resources and the allocation policy to determine if the request should be granted or not.
```
hamersaw, line 139: These two paragraphs seem to be covering the same information. Does it make sense to merge them? Maybe something like:

When a Flyte task execution needs to send a request to an external service, the plugin claims a unit of the corresponding resource. This is done using a **ResourceName**, which is a unique token and a fully qualified **resource request** (which is typically an integer). The execution generates this unique token and registers this token with the ResourceManager by calling the ResourceManager’s **AllocateResource** function. If the resource pool has sufficient capacity to fulfil your request, then the resources requested are allocated, and the plugin proceeds further.

rsts/deployment/cluster_config/customizable_resources.rst
```diff
@@ -86,4 +86,107 @@ Flytekit
...
 For convenience, add a FlyteCTL wrapper to update the new attributes. Refer to `this PR <https://github.com/flyteorg/flytectl/pull/65>`__ for the entire set of changes required.
 
-That's it! You now have a new matchable attribute to configure as the needs of your users evolve.
\ No newline at end of file
+That's it! You now have a new matchable attribute to configure as the needs of your users evolve.
+
+Flyte ResourceManager
+---------------------
+
+**Flyte ResourceManager** is a configurable component that allows plugins to manage resource allocations independently. It helps track resource utilization of tasks that run on Flyte. The default deployments are configured as ``noop``, which means they are disabled and you would fallback to the platform specific resource management, like K8s, which tells you that you can't schedule more tasks.
```
hamersaw, line 94: Can we elaborate on where the resource manager is beneficial? Maybe changing this portion:

"which means they are disabled and you would fallback to the platform specific resource management, like K8s, which tells you that you can't schedule more tasks."

To something like:

"which indicates the ResourceManager provided by Flyte is disabled and plugins rely on each independent platform to manage resource utilization. In situations like the k8s plugin, where the the platform has a robust mechanism to manage resource scheduling this may work well. However, in a scenario like a simple web API plugin, the rate at which Flyte sends requests may overwhelm a service and benefit from additional resource management."

hamersaw (commented):
I think this looks great! A few minor clarifications just to make sure we cover every question I can think of.

hamersaw (approved):

PR #2780, opened 2022-08-18
Adjust getting started docs page

iaroslav-ciupin:
# Synopsis

Looks like python3.8 has problems with flytekit. Better to use python3.9 and pip3.9.

### DIFF
--- a/rsts/getting_started/index.rst
+++ b/rsts/getting_started/index.rst
@@ -11,6 +11,7 @@ Requirements
 
 - Install `Docker <https://docs.docker.com/get-docker/>`__
 - Ensure Docker Daemon is running
+- Python 3.9
 
 Installation
 ^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


pingsutw:
What problems do we encounter when using Python3? We should fix that.

iaroslav-ciupin:
@pingsutw when installing with Python3.8 `pip3 install flytekit` I got following warning
```
WARNING: The scripts flyte-cli, pyflyte, pyflyte-execute, pyflyte-fast-execute and pyflyte-map-execute are installed in '/Users/yaroslav/Library/Python/3.8/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
```
when installing with Python3.9 `pip3.9 install flytekit` I got no warnings.

wild-endeavor:
installing with a virtualenv or a conda environment should resolve this, thanks.

PR #2799, opened 2022-08-22
Fallback to logs in propeller termination message by default

andrewwdye:
Add `terminationMessagePolicy` to the flytepropeller templates and set to `FallbackToLogsOnError` by default. This makes failure info available through the k8s pod status interface.


katrogan (approved):

hamersaw:
Does it make sense to add these to the [propeller manager template](https://github.com/flyteorg/flyte/blob/e3e4978838f3caee0d156348ca966b7f940e3d45/charts/flyte-core/templates/propeller/manager.yaml#L48) as well?

hamersaw:
> Does it make sense to add these to the [propeller manager template](https://github.com/flyteorg/flyte/blob/e3e4978838f3caee0d156348ca966b7f940e3d45/charts/flyte-core/templates/propeller/manager.yaml#L48) as well?

cc. @katrogan @andrewwdye does it make sense to propagate this change to [propeller manager](https://docs.flyte.org/en/latest/deployment/cluster_config/performance.html#automatic-scale-out) as well? I opened a [PR](https://github.com/flyteorg/flyte/pull/2833). If it's unnecessary, feel free to close.

andrewwdye:
Yes, good catch. Thanks!

PR #2916, opened 2022-09-23
Fix getting started example: pass float instead of int parameter

uschi2000:
Before:
```
flytekit.core.type_engine.TypeTransformerFailedError: Type of Val '1' is not an instance of <class 'float'>
```

After:
Can run main successfully

Signed-off-by: Robert Fink <rf@robertfink.de>


pingsutw (approved):

uschi2000:
I don't have merge permissions here @pingsutw , but obviously feel free to merge any time.

pingsutw:
merged it. Thank you for the fix.

PR #2961, opened 2022-10-06
Add explicit check for k3s server startup completion

andrewwdye:
# Overview
When deploying the flyte-sandbox-lite image I ran into races between `k3s server` and `helm upgrade`. Previously the entrypoint scripts waited for `k3s kubectl explain deployment` to complete successfully. With this change we now explicitly wait on the node to be up and ready.

# Testing
```
$ docker build -t flyte-sandbox-lite --target=dind --compress=true -f ./docker/sandbox-lite/Dockerfile .
$ docker run --rm -it --privileged flyte-sandbox-lite

$ docker build -t flyte-sandbox --target=dind --compress=true -f ./docker/sandbox/Dockerfile .
$ docker run --rm -it --privileged flyte-sandbox
```

### DIFF
--- a/docker/sandbox-lite/flyte-entrypoint-dind.sh
+++ b/docker/sandbox-lite/flyte-entrypoint-dind.sh
@@ -24,7 +24,7 @@ file="/var/run/docker.pid"
 if [ -f "$file" ] ; then
     rm "$file"
 fi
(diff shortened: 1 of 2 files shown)


hamersaw (approved):

eapolinario (approved):
Nice.

PR #2982, opened 2022-10-14
chart: also disable redoc service if not enabled

flixr:
If redoc is disabled also don't add the service...

And update some comments.


flixr:
Could this be reopened and merged?

PR #3007, opened 2022-10-20
Update README with list of features that provide reproducibility

ronaldosaheki:
Currently this phrase is interrupted as there is no list after the `via:` 
```
 - Guaranteed **[reproducibility](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/task_cache.html)** of pipelines via:
```

Two solutions could be:
- to remove totaly the `via:`
- to provide the list that was removed in commit 80baa791c02e82cc258c1dfd7184e1b9440a386d, which gives more hints about how reproducibility can be achieved

```
 - Guaranteed **[reproducibility](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/task_cache.html)** of pipelines via:
-  - Versioned data, code, and models
-  - Automatically tracked executions
-  - Declarative pipelines
```


pingsutw (approved):
@samhita-alla will merge it if you agree.

PR #3019, opened 2022-10-24
Updated default FlytePropeller subqueue base delay to 0

hamersaw:
## Describe your changes
This PR updates the subqueue default configuration for FlytePropeller. This construct is used to queue a workflow when FlytePropeller detects a changein a node status. This can happen using a k8s watch over Pods to detect state changes. Currently, the configuration defaults to a 5s base-delay, this means that everytime an item is added to the queue, FlytePropeller waits 5s to process it, which results in unnecessary delays in workflow execution.

The trade-off is more aggressive workflow evaluations. We have tested this change at some scale and do not believe this is too frequent, meaning FlytePropeller is overwhelmed by the number of workflows in the queue. However, if it does prove too small, this can be increased in the configuration.


charts/flyte-core/values.yaml
```diff
@@ -607,13 +607,13 @@ configmap:
         retry-period: 2s
       queue:
         type: batch
-        batching-interval: 2s
+        batching-interval: 1s
         batch-size: -1
         queue:
           type: maxof
           rate: 100
```
kumare3, line 614: How about increase the rate too

charts/flyte-core/values.yaml
```diff
@@ -607,13 +607,13 @@ configmap:
...
           capacity: 1000
```
kumare3, line 615: And capacity

PR #3061, opened 2022-11-08
Increase default project quota in the sandbox and demo

pingsutw:
Signed-off-by: Kevin Su <pingsutw@apache.org>

## Tracking issue
https://flyte-org.slack.com/archives/C049Q7GDWN9/p1667885776403149

## Describe your changes
People often fail to run the workflow in the sandbox or demo cluster because resource exceeds the task default resources or the project quota. Therefore, we increase the quota and default resource to address this issue.


charts/flyte-core/values-sandbox.yaml
```diff
@@ -102,3 +117,64 @@ redis:
   # --- enable or disable Redis Statefulset installation
   enabled: false
 
+# -- Configuration for the Cluster resource manager component. This is an optional component, that enables automatic
+# cluster configuration. This is useful to set default quotas, manage namespaces etc that map to a project/domain
+cluster_resource_manager:
```
eapolinario, line 122: Why do we need this block?

eapolinario (commented):
How did you test this, Kevin?

charts/flyte-core/values-sandbox.yaml
pingsutw: IIUC, If we don't add it, the sandbox will use the default config in [value.yaml](https://github.com/flyteorg/flyte/blob/173c409e2633a7fa5a33ad07491665ef35acfe6f/charts/flyte-core/values.yaml#L732)

> How did you test this, Kevin?

I rebuilt the sandbox image and ran the command `flytectl sandbox start --image pingsutw/sandbox-test` to create cluster. Then check if the resource is updated in the configmap.

charts/flyte-core/values-sandbox.yaml
```diff
@@ -49,6 +49,21 @@ configmap:
...
+    # -- Task default resources parameters
+    task_resources:
+      defaults:
+        cpu: 500m
+        memory: 1Gi
+        storage: 500Mi
+      limits:
+        cpu: 2
+        memory: 4Gi
+        storage: 20Mi
```
eapolinario, line 64: why is this so low? Do we have to set it?

pingsutw: That is the default value in [value.yaml](https://github.com/flyteorg/flyte/blob/f8178c1c093cf641c36c0f00614bd0fbe1581c31/charts/flyte-core/values.yaml#L549). Yup, it's optional. I just removed it to make the config simple.

eapolinario (approved):

gpgn:
Hi, I had some issues with the default demo environment you get when running `flytectl demo start`, see discussion here: https://flyte-org.slack.com/archives/CP2HDHKE1/p1688567606055059

I was able to prevent things from getting OOMKilled by updating the sandbox configmap according to above values. Would it make sense for you to use these as defaults for the demo environment as well? Let me know if I should open a separate issue for this.

Issue #3080, opened 2022-11-16
[Housekeeping] Issues with bundled demo image


wild-endeavor:
# New Bundled Sandbox Image
This ticket is to collect issues with the new bundled sandbox environment image.  This image is the result of the following PRs:
https://github.com/flyteorg/flyte/pull/3066
https://github.com/flyteorg/flytectl/pull/370
https://github.com/flyteorg/flyte/pull/2994
https://github.com/flyteorg/flyte/pull/3137

## Cluster Resource Cron
* Need more investigation but either it's not running, or not picking up additional configuration.
  * Manually added more templates to the configmap but none of them were created.
```
$ kf get cm cluster-resource-templates -o yaml
apiVersion: v1
data:
  ac_spark_role.yaml: |
    apiVersion: rbac.authorization.k8s.io/v1beta1
    kind: Role
    metadata:
      name: spark-role
      namespace: {{ namespace }}
    rules:
    - apiGroups: ["*"]
      resources: ["pods"]
      verbs: ["*"]
    - apiGroups: ["*"]
      resources: ["services"]
      verbs: ["*"]
    - apiGroups: ["*"]
      resources: ["configmaps", "persistentvolumeclaims"]
      verbs: ["*"]
  ad_spark_service_account.yaml: |
    apiVersion: v1
    kind: ServiceAccount
    metadata:
      name: spark
      namespace: {{ namespace }}
  ae_spark_role_binding.yaml: |
    apiVersion: rbac.authorization.k8s.io/v1beta1
    kind: RoleBinding
    metadata:
      name: spark-role-binding
      namespace: {{ namespace }}
    roleRef:
      apiGroup: rbac.authorization.k8s.io
      kind: Role
      name: spark-role
    subjects:
    - kind: ServiceAccount
      name: spark
      namespace: {{ namespace }}
  namespace.yaml: |
    apiVersion: v1
    kind: Namespace
    metadata:
      name: {{ namespace }}
kind: ConfigMap
```

## Nits
* Logging level should be increased.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

jeevb:
Error message:
```
[...]

PR #3112, opened 2022-11-29
Increase log level in flyte-sandbox-bundled from 1 -> 6

jeevb:
Partially addresses: https://github.com/flyteorg/flyte/issues/3080


wild-endeavor (approved):
but let's keep that issue open?  just want a place to keep track of everything

Issue #2993, opened 2022-10-18
[Housekeeping] Flyte Deployment journey and documentation


wild-endeavor:
# Revamp Flyte Deployment
This ticket encompasses all the work for revamping the Flyte deployment journey.  Specifically this covers

* Updates to the flyte sandbox-lite image
* Changes to flytectl
* Documentation updates
* Helm chart updates

More information to follow.

## Flyte Local Demo Experience
* Persist state to user's local workstation
* Run a registry so we can deprecate the --source flag
* Preload images

### New local demo sandbox
After this [PR](https://github.com/flyteorg/flyte/pull/3066/files) is merged, we will have the following structure.

#### Helm
* A Helm chart just for the Flyte single binary.
* A Helm chart for the sandbox that uses the Flyte single binary Helm chart and pulls in the open source Helm charts for dependencies (like minio/postgres)

#### Images
* New [bundled image](https://github.com/flyteorg/flyte/tree/master/docker/sandbox-bundled) will be created.  This is the core of the new experience.  It will contain
  * predownloaded images for all the dependencies including the flyte binary itself.
  * a registry so that the `--source` flag in flytectl is no longer necessary.

## Flytectl
* Allow users to start with an external flyte config
* Allow users to modify a config and reload
* Print current state of cluster

## Docs
Implement the new user journey, moving the more difficult error prone things like auth and ingress to later stages.

## Helm
Refactor and introduce new Helm charts that use the single binary.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

cosmicBboy:
fixed by #3116

PR #3116, opened 2022-11-29
Update Administrator docs

wild-endeavor:
# Deployment Documentation
Restructure and update some deployment docs.

Main changes:
* Remove all the cloud specific stuff.
  * Point users to a newly created GH Discussion topic for submitting articles on deployment experiences.
* Use the new `flyte-binary` Helm chart.
* Remove a lot of the sandbox stuff since the new `flytectl demo` environment is architected differently.
* Change instructions to use `flytectl demo` instead of `flytectl sandbox`
* Remove `ideal_flow.rst` - this is incomplete. We should eventually offer some basic gh workflow examples rather than just talking about it.

## Structure Changes
Bundling all Flyte configuration under the deployment moniker felt a bit of a stretch.  I think renaming it to a broader administrator's guide makes more sense.

* Currently the generated component configs are missing, but this will end up changing with the partial mono-repo work anyways.

## Deployment Updates
One of the things we want to do is change the deployment journey.  We want to make sure users are led comfortably through  the various stages that one might expect of something as complex as Flyte.

Basically the steps of the journey should be
1. flytectl demo sandbox
1. a simple cloud (eks/gke) based deployment with nothing tricky - just helm install.  users will have to port-forward to see anything.
1. a production ready deployment (ingress, auth, etc) (think stable enough for most companies).
1. a scalable multicluster setup (lots of deployments will never need this level).

The middle steps will replace the aws/gcp guides that we have today.  These have been completely deleted.

From there we should link to a revamped version of the `ideal_flow.rst` file I think.

https://github.com/flyteorg/flyte/issues/2993


rsts/index.rst
```diff
@@ -31,12 +31,12 @@
...
 .. toctree::
-   :caption: Deployment
+   :caption: Administrator's Guide
    :maxdepth: -1
-   :name: deploymenttoc
+   :name: administratortoc
    :hidden:
 
-   deployment/index
+   administrator/index
```
cosmicBboy, line 39: Can we consider keeping the `Deployment` name here?

I think we want to keep a product/functionality-based nomenclature in the docs as opposed to a role-based nomenclature, because the latter alienates users who don't play that role and discourages them from exploring that section of the documentation.

We want to help educate people who don't necessarily have a background in infra to be able to potentially set up their own Flyte cluster. Of course we can talk about _who_ the pages are primarily for, but I think we should let that be part of the introduction to this overall section of the docs.

Happy to discuss alternatives to "Deployment", e.g. "Cluster Deployment", etc.

wild-endeavor: Yeah I guess that's fine... as long as what we mean by that is all things that relate to a deployment of Flyte.  Because the initial deployment of Flyte is just one component of it.  There are things related to maintaining, scaling, plugins, etc.  All the observability and monitoring stuff should also go here.  Upgrade procedures should go here, etc.

wild-endeavor:
@ppiegaze

cosmicBboy (approved):

Issue #427, opened 2020-07-22
[Backend][Plugin]Support for Dask clustered tasks in Flyte


kumare3:
**Why would this plugin be helpful to the Flyte community**
Users could write very short running distributed array jobs using DASK. This makes it possible to have very small runtime jobs multi-plexed onto same set of nodes.

**Type of Plugin**
- [ ] Web Service (e.g. AWS Sagemaker, GCP DataFlow, Qubole etc...)
- [x] Kubernetes Operator (e.g. TfOperator, SparkOperator, FlinkK8sOperator, etc...)
- [ ] Customized Plugin using native kubernetes constructs
- [ ] Other

**Can you help us with the implementation?**
- [ ] Yes
- [ ] No

**Additional context**
This would really help express some ideas that are not Spark, or heavyweight like Flyte batch jobs.

kumare3:
Dask can be deployed to Kubernetes, the template is shown [here](https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/kubernetes.yaml). Allowing this would help the users a lot and enable writing really short tasks. This coupled with cluster re-use (coming later) or cluster gateways (daskhub) and support for a coiled task in the future would enable users to use dask more effectively and make Flyte + Dask work together.

```python
@task(config=Dask(
            workers=4,
            worker_resources=....,
            [worker-pod-template=...] # Also the command should probably be hard coded client side?
            ), resources=Resources(....) # Driver resource
)
def my_dask_program():
      pass
```

ossareh:
This was discussed a little on slack: https://app.slack.com/client/TN89P6GGK/CNMKCU6FR/thread/CNMKCU6FR-1648660418.322249

Currently we do not use dask, nor do we use [Coiled's hosted platform for Dask](https://coiled.io/). However, both are really interesting to us in terms of migrating away from our current, home rolled, workflow orchestration solution and having someone else run our work loads for us. 

The primary interest in Dask is its drop in nature w.r.t. dataframes and numpy arrays. We currently employ a streaming solution (using mmap) to allow us to do this work on one node; being able to scale up to multiple nodes without needing code changes in dependent areas of our project would be an instant win for us.

From an integration point of view I found the Dask+Prefect video informative:

- [Jump to 16minutes, watch until 23:30, everything else is unrelated](https://fast.wistia.com/embed/channel/pjv47wv0u1?wchannelid=pjv47wv0u1&wmediaid=zub5r4nxp4)
[...]

PR #3145, opened 2022-12-12
Add dask operator

bstadlbauer:
## Tracking issue

https://github.com/flyteorg/flyte/issues/427

## Describe your changes

Adds: 
- `dask` operator deployment to the `helm` chart
- Notes about `dask` to the top-level `README.md`
- Description on how to setup `dask` backend plugin

## Note to reviewers

Do I also need to add the `dask` operator deployment to kustomize?


bstadlbauer:
@hamersaw Is there also a need to a `kustomize` deployment for the `dask` operator?

hamersaw:
> @hamersaw Is there also a need to a `kustomize` deployment for the `dask` operator?

I believe we are deprecating `kustomize` support, so I do not think it is necessary.

cc @eapolinario can you confirm?

charts/flyte-core/values-eks.yaml
```diff
@@ -400,3 +400,10 @@ cluster_resource_manager:
 
 sparkoperator:
   enabled: false
+
+#
+# SPARKOPERATOR
```
wild-endeavor, line 405: i don't think we need these...

eapolinario:
> > @hamersaw Is there also a need to a `kustomize` deployment for the `dask` operator?
> 
> I believe we are deprecating `kustomize` support, so I do not think it is necessary.
> 
> cc @eapolinario can you confirm?

That's correct. We talked about it, so no need to update the kustomize deployments to support this.

charts/flyte-core/values-eks.yaml
eapolinario: This should say Dask operator, no?

charts/flyte-core/values-keycloak-idp-flyteclients-without-browser.yaml
```diff
@@ -832,3 +832,8 @@ sparkoperator:
           - spark.blacklist.enabled: "true"
           - spark.blacklist.timeout: "5m"
           - spark.task.maxfailures: "8"
+
+# -- Optional: Dask Plugin using the Dask Operator
+daskoperator:
+  # --- enable or disable the dask operator deployment installation
+  enabled: false
```
eapolinario, line 839: nit: can you add a newline?

charts/flyte-core/values.yaml
```diff
@@ -821,3 +821,8 @@ sparkoperator:
           - spark.blacklist.enabled: "true"
           - spark.blacklist.timeout: "5m"
           - spark.task.maxfailures: "8"
+
+# -- Optional: Dask Plugin using the Dask Operator
+daskoperator:
+  # --- enable or disable the dask operator deployment installation
+  enabled: false
```
eapolinario, line 828: ditto.

wild-endeavor (approved):

PR #3155, opened 2022-12-15
Add support for authentication in flyte-binary chart

jeevb:


charts/flyte-binary/templates/deployment.yaml
```diff
@@ -186,6 +231,15 @@ spec:
         {{- include "flyte-binary.renderTemplate" ( dict "value" .Values.deployment.sidecars "context" $ ) | nindent 8 }}
         {{- end }}
       volumes:
+        {{- if .Values.configuration.auth.enabled }}
+        - name: auth
+          projected:
+            sources:
```
wild-endeavor, line 237: oh so this allows us to merge two secrets into the same volume?  cool

charts/flyte-binary/templates/auth-client-secret.yaml
```diff
@@ -0,0 +1,19 @@
...
+    {{- include "flyte-binary.renderTemplate" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  annotations:
+    {{- if .Values.commonAnnotations }}
+    {{- include "flyte-binary.renderTemplate" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- end }}
+type: Opaque
+stringData:
+  client_secret: {{ required "Internal client secret required when authentication is enabled" .Values.configuration.auth.internal.clientSecret | quote }}
+  oidc_client_secret: {{ required "OIDC client secret required when authentication is enabled" .Values.configuration.auth.oidc.clientSecret | quote }}
```
wild-endeavor, line 18: would it make more sense to move this into the one above since this Secret is just for internal use right?  And the other one is all things admin auth?

charts/flyte-binary/templates/auth-client-secret.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.configuration.auth.enabled }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "flyte-binary.configuration.auth.clientSecretName" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "flyte-binary.labels" . | nindent 4 }}
+    {{- if .Values.commonLabels }}
+    {{- include "flyte-binary.renderTemplate" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
```
wild-endeavor, line 9: remind me to ask you how this works tomorrow.

wild-endeavor (approved):

charts/flyte-binary/templates/auth-client-secret.yaml
jeevb: I don't think so. `admin-auth-secret.yaml` is basically only for auto-generated secrets. Anything that we add in there in the helm chart will conflict with the output of the `gen-admin-auth-secret` init container. `auth-client-secret.yaml` is for user-provided secrets, but we're really only storing client secrets in here, so its named accordingly.

charts/flyte-binary/templates/auth-client-secret.yaml
jeevb: This is basically a helper for rendering template strings: https://github.com/flyteorg/flyte/blob/0ba5bc05c02f6d169543c2495f3c4ac36b5cf9f3/charts/flyte-binary/templates/_helpers.tpl#L65.

`( dict "value" .Values.commonLabels "context" $ )` generates a map with 2 keys: `value` and `context`. This is then passed into the template. The output is `.value` rendered with attributes from `.context`, with handling for simple strings and more complex structures. I saw this pattern in one of the bitnami charts to standardize how templates are rendered.

jeevb: In retrospect, this is needlessly verbose. I'll refactor this out a follow-up PR.

jeevb: https://github.com/flyteorg/flyte/pull/3156

PR #3191, opened 2022-12-28
Change base image for flyte-binary image

jeevb:


wild-endeavor (approved):
the same arg gets picked up twice right?

PR #3248, opened 2023-01-20
[charts/flyte-binary] Add provision for specifying configmaps for extra configs

jeevb:
- Adds a new value in `flyte-binary` chart for extra configuration to be passed in via an existing configmap (`inlineConfigMap`). If specified, this will be merged with the base configuration via a `projected` volume.
- Leverages the above option in sandbox-bundled, and creates an empty configmap that can be updated by users to inject additional configuration and cluster resource templates for the flyte-sandbox.
- Changed some values so that resources names are more concise in the sandbox. Updated the local single-binary flyte config example to reflect the new values.
- Add secrets/webhook test to the `make test` target for `sandbox-bundled`.


charts/flyte-binary/README.md
```diff
@@ -11,6 +11,7 @@ Chart for basic single Flyte executable deployment
 | clusterResourceTemplates.annotations | object | `{}` |  |
 | clusterResourceTemplates.externalConfigMap | string | `""` |  |
 | clusterResourceTemplates.inline | object | `{}` |  |
+| clusterResourceTemplates.inlineConfigMap | string | `""` |  |
```
jeevb, line 14: Does this name make sense? It mirrors the functionality of `inline` but allows extra configuration to be specified via a configmap.

jeevb:
Example to install `sparkoperator`:
- Add extra config files to the `flyte-sandbox-extra-config` configmap. Without changes to flytectl, this will require either a `kubectl edit` or a `kubectl apply`.
- Add extra cluster resource templates for spark roles to the `flyte-sandbox-extra-cluster-resource-templates` configmap.
- Restart flyte-sandbox deployment (`kubectl rollout restart deploy/flyte-sandbox`)

wild-endeavor (approved):

PR #3266, opened 2023-01-25
Update flyte-sandbox subcharts and add support for data persistence a…

jeevb:
…cross clusters

Signed-off-by: Jeev B <jeevb@users.noreply.github.com>

@wild-endeavor proposed that users might want their database and object storage data persisted across sandbox clusters. This is especially useful when upgrading to a newer sandbox version, which would require a teardown of the old version and creation of the new version.

This PR adds a new path `/var/lib/flyte` where data with persistence requirements might be stored. HostPath PV/PVCs are also created for PostgreSQL and Minio at `/var/lib/flyte/storage/db` and `/var/lib/flyte/storage/minio` respectively. Mounting a pre-created (via `docker volume create flyte-sandbox`) docker volume to `/var/lib/flyte` will ensure that data will persist across different cluster versions, for as long as the docker volume exists.

TODO: Logic will need to be added to `flytectl` to manage the lifecycle of the docker volume, and ensure that it is correctly mounted into the sandbox cluster.


eapolinario (commented):
Just for my own education, why are we sprinkling 63 all over the place?

jeevb:
> Just for my own education, why are we sprinkling 63 all over the place?

There is a requirement for certain resource names to not exceed 63 chars:
```
> kubectl run -it --rm --image=busybox $(for i in $(seq 64); do printf 'a'; done) bash
The Pod "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" is invalid:
* metadata.labels: Invalid value: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa": must be no more than 63 characters
* spec.containers[0].name: Invalid value: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa": must be no more than 63 characters
```

I believe this is only really valid for pods and services. I'll check and fix the ones that don't need that transformation.

eapolinario (approved):

Issue #3296, opened 2023-01-30
[BUG] Helm chart handles workflow_notifications region incorrectly


andrusha:
### Describe the bug

Applying helm chart with `workflow_notifications` enabled produces broken configmap yaml for the `region` parameter, thus admin and scheduler fails service deployment.

Eg:
```﻿
﻿notifications.yaml:
----
notifications:
  type: aws
  region:us-east-1
  publisher:
```

Notice `region:us-east-1` misses space after semicolon.

### Expected behavior

Valid yaml should be produced by the helm chart.

**Workaround** add space in your region template values like so: `region: " {{ .Values.userSettings.accountRegion }}"`

### Additional context to reproduce

1. Create `test-values.yaml`
```yaml
userSettings:
  accountRegion: us-east-1
  accountNumber: 123123123
  notifications:
    topicName: topic-name
    queueName: queue-name

workflow_notifications:
  enabled: true
  config:
    notifications:
      type: "aws"
      region: "{{ .Values.userSettings.accountRegion }}"
      publisher:
        topicName: "arn:aws:sns:{{ .Values.userSettings.accountRegion }}:{{ .Values.userSettings.accountNumber }}:{{ .Values.userSettings.notifications.topicName }}"
      processor:
        queueName: "{{ .Values.userSettings.notifications.queueName }}"
        accountId: "{{ .Values.userSettings.accountNumber }}"
      emailer:
        subject: "Flyte: {{ project }}/{{ domain }}/{{ launch_plan.name }} has '{{ phase }}'"
        sender: "{{ .Values.userSettings.notifications.sender }}"
        body: |
          "Execution {{ workflow.project }}/{{ workflow.domain }}/{{ workflow.name }}/{{ name }} has {{ phase }}.
           Details: [https://flyte.example.com/console/projects/{{](https://flyte.example.com/console/projects/%7B%7B) project }}/domains/{{ domain }}/executions/{{ name }}.
           {{ error }}"
```
2. Run `helm template admin flyteorg/flyte-core --version v1.2.1 --values test-values.yaml | grep -A 4 "notifications.yaml"`
3. See:
```yaml
  notifications.yaml: |
    notifications:
      type: aws
      region:us-east-1
      publisher:
```

Which misses space after semicolon.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #3297, opened 2023-01-30
Fixed notification region config

pmahindrakar-oss:
Signed-off-by: pmahindrakar-oss <prafulla.mahindrakar@gmail.com>

## Tracking issue
fixes https://github.com/flyteorg/flyte/issues/3296

Tested it and it generates the right region config with indentation

## Describe your changes

[x] - Correct; marked as done
[X] - Correct; marked as done
[ ] - Not correct; marked as **not** done

### DIFF
--- a/charts/flyte-core/templates/admin/configmap.yaml
+++ b/charts/flyte-core/templates/admin/configmap.yaml
@@ -45,9 +45,7 @@ data:
   notifications.yaml: |
     notifications:
       type: {{ .Values.workflow_notifications.config.notifications.type }}
-      {{- with .Values.workflow_notifications.config.notifications.region }}
-      region: {{- tpl (.) $ }}
-      {{- end }}
+      region: {{ tpl .Values.workflow_notifications.config.notifications.region $ }}
       {{- with .Values.workflow_notifications.config.notifications.publisher }}
       publisher: {{- tpl (toYaml .) $ | nindent 8 }}
       {{- end }}


pmahindrakar-oss:
@andrusha thanks for the review. Yes i agree removing that also works but found this to be more concise way to do it.

charts/flyte-core/templates/admin/configmap.yaml
```diff
@@ -45,9 +45,7 @@ data:
   notifications.yaml: |
     notifications:
       type: {{ .Values.workflow_notifications.config.notifications.type }}
-      {{- with .Values.workflow_notifications.config.notifications.region }}
-      region: {{- tpl (.) $ }}
-      {{- end }}
+      region: {{ tpl .Values.workflow_notifications.config.notifications.region $ }}
```
eapolinario, line 48: Just to confirm, region is a required field, right?

pmahindrakar-oss: Yes. These are currently being used and are required until we fully remove the usage of this deprecated field and use it from AWSConfig in  flyteadmin.

https://github.com/flyteorg/flyteadmin/blob/master/pkg/async/notifications/factory.go#L58
https://github.com/flyteorg/flyteadmin/blob/master/pkg/async/notifications/factory.go#L92
https://github.com/flyteorg/flyteadmin/blob/master/pkg/async/notifications/factory.go#L143

eapolinario (approved):

PR #3320, opened 2023-02-07
[Proposal]: Flyte System Tags and metadata

kumare3:
This RFC proposes a way to add execution Tags and description.
this would help the user in multple ways. Please read the rfc for more context.

Signed-off-by: Ketan Umare <ketan.umare@gmail.com>


goyalankit (commented):
very useful feature!

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,124 @@
...
+
+#### Approach 2: Certain label keys are treated special
+ - “group” will group everything
+ - “experiment” will also group everything with higher priority. 
+ - “name” will override the execution id with the name?
+ 
+![Grouping / Filtering UX](https://raw.githubusercontent.com/flyteorg/static-resources/main/flyte/rfc/tags/labels-filter.png)
+
+
+### Support for descriptions
```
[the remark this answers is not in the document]
kumare3: i think we should store the description in the blobstore? we could offload it, if its too large?

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,124 @@
...
+
+#### Approach 1: Treat all labels the same way and allow search/filter and click based grouping
+In this approach users will get the regular executions view with all the labels
+available on each execution. The users are allowed to filter an exection simply
+by clicking on a label and then all executions are filtered by that label. 
+
+#### Approach 2: Certain label keys are treated special
+ - “group” will group everything
+ - “experiment” will also group everything with higher priority. 
+ - “name” will override the execution id with the name?
```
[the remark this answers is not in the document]
kumare3: executionID cannot be changed - it is immutable and unique per project/domain.
name is just an alias. I will update the doc to reflect this

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,124 @@
...
+
+Once this is implemented the UI and CLI can be updated to support these
+queries.
+
+### CLI Interface 
+
+A workflow or task can be executed using
+
+```bash
+pyflyte run --remote --labels k:v --labels k1:v1 test.py wf --input1=10
```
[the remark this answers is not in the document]
kumare3: great point, but the rpc field is sadly already called label. and these will become k8s labels

rfc/system/0001-flyte-execution-tags.md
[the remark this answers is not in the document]
kumare3: but i do like the idea of immutable labels as well. once added you cannot change them

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,124 @@
...
+ -> 1.5 * 10^9 bytes (assuming one byte per character" -> 1.5GB
+
+This is not significant and will increase as executions increase
+
+## 6 Alternatives
+NA
+
+
+## 7 Potential Impact and Dependencies
+We this this is one of the most requested features in Flyte and will solve
```
fg91, line 116:
```suggestion
This is one of the most requested features in Flyte and will solve
```

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,124 @@
...
+### UI Interface
+
+Two approaches
+
+#### Approach 1: Treat all labels the same way and allow search/filter and click based grouping
+In this approach users will get the regular executions view with all the labels
+available on each execution. The users are allowed to filter an exection simply
+by clicking on a label and then all executions are filtered by that label. 
+
+#### Approach 2: Certain label keys are treated special
```
fg91, line 69: If this approach is chosen, I wonder whether it would be nicer for the user to do

```bash
pyflyte run --remote --group foo --experiment bar ...
```

instead of 

```bash
pyflyte run --remote --labels group:foo --labels experiment:bar ...
```

This doesn't mean that under the hood the labels mechanism couldn't be used.

fg91 (dismissed):
I like this proposal very much. Currently we use decks to link to Wandb runs corresponding to Flyte executions.
We then use tags in wandb to do the grouping by experiments, tags, ... that you describe in the RFC.
Would love to do this directly in Flyte.

davidmirror-ops:
03-30-2023 Meeting notes: 
KU: launchplan name could be the default grouping tag
Tim Sheiner: that's not very prominent right now in the UI
KU: we could add all to system tags but could be an overload
TS: make sure that this proposal is not redundant to the fact users are already naming workflows and tasks
TS: instead of treating this proposal as tags, treat it as a separate...
GG: their use case is running workflows from notebooks (using Quarto/Jupyter for reports)

tsheiner:
Slightly perpendicular to this proposal but related to the notion of making executions easier to identify:

I note that Prefect uses a system of nonsense ids for executions which is much friendlier looking and far easier to remember than Flyte alphanumeric ids.  For example ‘enigmatic-waxbill’ for ‘massive-antelope.’ Would something like this be possible for Flyte?

davidmirror-ops:
04-13-203 notes: no updates

rfc/system/0001-flyte-execution-tags.md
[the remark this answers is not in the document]
ggydush: So we'll support both mutable and immutable labels?

rfc/system/0001-flyte-execution-tags.md
bstadlbauer: Or do we wanna create `group` and `experiment` as CLI arguments and introduce them as a concept? 🤔

bstadlbauer (dismissed):
Looks good to me! 

There is an ongoing slack conversation related to MySQL tag storage [here](https://flyte-org.slack.com/archives/C04NJPLRWUX/p1683726030497939?thread_ts=1683703597.324409&cid=C04NJPLRWUX) - not 100% sure what that is about? cc @kumare3 @ByronHsu

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,124 @@
...
+isolation or running multiple co-horts of experiments. Also there are cases in
+which grouping executions by a group might make them easier to find and
+associate them better. This document provides a solution of how we could
+improve the experience of discovering executions within Flyte. It also provides
+motivation of how this feature could further improve other ecosystem projects.
+
+## 2 Motivation
+
+As a User I want to
+ - Group a certain number of executions into an experiment group - for ease of debugging and discovery.
```
fg91, line 21:
```suggestion
 - Group a certain number of executions into an experiment group - for ease of debugging and discovery - when launching the workflows or after they have already finished.
```

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,124 @@
...
+ - Users could name their executions without having to worry about the character limits, uniqueness constraints and limited characterset.
+ - Simplify filtering of certain executions
+
+## 3 Proposed Implementation
+
+### Support for tags
+
+We propose to solve the problem of discovery by supporting arbitrary metadata association with an entity. This is similar to conceppt of “tags” as in AWS. The tags are represented as `“key”: “value”` pairs. In kuberenetes this can be represented using “labels” and “annotations”. Labels and annotations are already supported per execution as documented in - [ExecutionCreateRequest](https://docs.flyte.org/projects/flyteidl/en/latest/protos/docs/admin/admin.html#executioncreaterequest) -> [ExecutionSpec](https://docs.flyte.org/projects/flyteidl/en/latest/protos/docs/admin/admin.html#executionspec). Moreover,  every project supports default labels [Project](https://docs.flyte.org/projects/flyteidl/en/latest/protos/docs/admin/admin.html#project). Thus the final execution will have a the union of the project default labels + the user specified labels as it exists today.
+
+Currently the resultant labels are not persisted and are only applied to the
```
fg91, line 34: Kubernetes labels also wouldn't be set when tagging an execution e.g. in the UI that already finished. Such tags that are set post execution should be saved the same way as tags that are set when launching a workflow.

rfc/system/0001-flyte-execution-tags.md
fg91: I personally prefer option 1: treat all labels the same way. Users might not want to follow the categories we deem sensible. Experiment tracking servers like Mlflow or Wandb, which also have such a tagging mechanism, simply allow users to assign arbitrary tags. I would argue that ML engineers are used to this and we should provide the same UX without imposing special naming conventions.

Only exception: execution name
I find it really helpful to have the pod names include customizable identifiers.
We have a registration script, similar to `pyflyte run` with has an `--execution_name` arg. The user provided value is appended with a random uuid, as is currently already chosen for the execution ids, and the result is checked against the execution name regex again and then passed to `FlyteRemote.execute(execution_name=...)` (already supported, see [here](https://github.com/flyteorg/flytekit/blob/26d1f290ab9e800acfd9af08595a82c3b9f962b6/flytekit/remote/remote.py#L1031)). So I wouldn't treat execution name with a pod label but the pods `metadata.name`.

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,124 @@
...
+execution in Kubernetes. As a first step, we recommend that these labels are
+persisted associated with an execution and then ListExecutions API is updated
+to return all 
+ - filtered executions by labels with supported `and` `or` queries
```
elibixby, line 38: Can we do this with tasks/workflows and their associated versions as well.

Particularly for versions it would be useful. For instance if you have multiple people working in different branches on the same workflow it's useful to have a floating "latest" tag for each branch, which may not be the "latest" execution.

rfc/system/0001-flyte-execution-tags.md
fg91: In the previous contributors synch we agreed that option 1 is preferable since we don't want to impose on users which label keys are considered special in their domain.

kumare3:
> Slightly perpendicular to this proposal but related to the notion of making executions easier to identify:
> 
> I note that Prefect uses a system of nonsense ids for executions which is much friendlier looking and far easier to remember than Flyte alphanumeric ids. For example ‘enigmatic-waxbill’ for ‘massive-antelope.’ Would something like this be possible for Flyte?

I think this is possible and I have thought about it. We should consider entropy considerations and cost of doing this and then all for it 👍🏽

rfc/system/0001-flyte-execution-tags.md
pingsutw: @bstadlbauer @fg91 @elibixby @flixr @goyalankit Some questions for you
1. Do you prefer key-value pair tags or tags that only have key? 
2. Should we add tags to Kubernetes label?

Currently, execution spec (with labels) is serialized to byte and is stored in the execution table. it's impossible to add / delete / update tags. if we use k8s client to filtered flyteworkflow (CRD) by labels. we cannot search a execution after CR is deleted.

I have a [PR](https://github.com/flyteorg/flyteadmin/pull/571) that adds tags table. it allows us easily add / update / delete tags, and even attach tags to task / workflow / project. however, it's not key-value pair tags for now. If we decide to use key-value pair tags, I just need to add a new column to the tags table and update the query. I'd like to know your thought first. 

btw, the current implementation works with both Mysql and Postgres.

tsheiner: My $0.02 is let's keep it simple and support what you call key-only tags.  

A person can 'hack' this to resemble key-value if needed (ie 'costcenter-12'), but we don't need to manage that complexity on the back end or in the UI when we get to figuring out how to let folks use tags to sort/group things.

fg91: * In my opinion key-only tags are perfectly fine and what ML engineers are used to from experiment tracking servers
* > Should we add tags to Kubernetes label

    I think being able to add/delete/update tags after the execution has already started or ended is an important feature. User story: an experiment is training/trained really well and I want to mark it for later. This is something that is not known when starting the execution. But updating/deleting/adding tags when the execution is already running would mean that the k8s labels are not in sync with what is stored in the tags table. I'd therefore say that I wouldn't apply the tags as labels to k8s.

bstadlbauer: Sorry for the late response here but agreed with what's been said above. Key only tags would also solve all our usecases 👍 

> I think being able to add/delete/update tags after the execution has already started or ended is an important feature 
+1 to this and the reasoning of not applying those to k8s

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,127 @@
+# Flyte Execution Tags and Metadata
+
+**Authors:**
+
+- @kumare3
+
+## 1 Executive Summary
+
+Flyte currently provides no visual ways of grouping executions and other
+entities with a “project” and “domain”. Usually a project is used by a team of
```
fg91, line 10:
```suggestion
entities apart from with a “project” and “domain”. Usually a project is used by a team of
```

fg91: Nit pick: Currently the sentence is wrong, right?

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,127 @@
...
+isolation or running multiple co-horts of experiments. Also, there are cases in
+which grouping executions by a group might make them easier to find and
+associate them better. This document provides a solution of how we could
+improve the experience of discovering executions within Flyte. It also provides
+motivation of how this feature could further improve other ecosystem projects.
+
+## 2 Motivation
+
+As a User I want to
+ - Group a certain number of executions into an experiment group - for ease of debugging and discovery - when launching the workflows or after they have already finished.
```
fg91, line 21:
```suggestion
 - Group a certain number of executions into an experiment group - for ease of debugging and discovery - when launching the workflows, while they are running or after they have already finished.
```

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,127 @@
...
+execution in Kubernetes. We'll create two new tables in the flyteadmin database.
+- ``execution_admin_tags`` is a join table, and it contains admin_tags ID and execution ID.
+- ``admin_tags`` saves all the tag names
+
+As a first step, we recommend that these tags are
+persisted associated with an execution and then ListExecutions API is updated
+to return all 
+ - filtered executions by tags with supported `and` `or` queries
+ - all associated executions with every execution
+ - limit total number of tags per execution to 10-15
```
fg91, line 47: Nit-pick (non blocking): this last bullet point appears to not fit "ListExecutions API is updated
to return all".

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,127 @@
...
+
+For the second step, we will support attach tags to the project, task, and workflow. In addition,
```
fg91, line 49:
```suggestion
For the second step, we will support attaching tags to the project, task, and workflow. In addition,
```

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,127 @@
...
+
+Once this is implemented the UI and CLI can be updated to support these
+queries.
+
+### CLI Interface 
+
+A workflow or task can be executed using
+
+```bash
+pyflyte run --remote --tags '["hello", "world"]' test.py wf --input1=10
```
fg91, line 61: Could we do `--tag hello --tag world` instead of providing a list of tags in string format?

rfc/system/0001-flyte-execution-tags.md
```diff
@@ -0,0 +1,127 @@
...
+
+## 5 Drawbacks
+It is important to understand that this may add a little more stress to the
+metadata database. Back of the envelope calculation
+
+ -> 1 million executions * 20 tags each. 
+ -> Each tag is a plain string, which has 64 characters.
+ -> 1.28 * 10^9 bytes (assuming one byte per character" -> 1.28GB
+
+This is not significant and will increase as executions increase
```
fg91, line 111:
```suggestion
This is not significant even though it will increase as executions increase
```

Was this the intended meaning? Otherwise the sentence appears to be contradicting itself.

fg91 (approved):
Thanks for incorporating what was discussed in the contributors' syncs. I like that now:

* tags will not be attached to k8s objects as labels but instead saved in the database so that they can be modified/deleted during/after the execution.
* all tags are equal and Flyte doesn't impose any special names.
* we have a clear distinction between tags and notes.

Please consider all comments below as nit-picks that you can just resolve in case you don't agree.

rfc/system/0001-flyte-execution-tags.md
pingsutw: maybe we can support both?

pingsutw:
```bash
pyflyte run --remote --tags '["hello", "world"]' 
# and 
pyflyte run --remote --tag hello --tag hello
# and 
pyflyte run --remote --tag hello --tags '["key1", "key2"]' 
```

fg91: My opinion about this is not so strong that I'd say we need to support two options in case others prefer `--tags '["hello", "world"]' `. I personally find lists or jsons in string representation as cli args a bit cumbersome.

fg91 (approved):
LG!

eapolinario (approved):

Issue #3385, opened 2023-03-01
[Docs] Add PodTemplate documentation: deployment configuration docs update and an example in flytesnacks


cosmicBboy:
### Description

This issue is for documenting example usage of the new `PodTemplate` configuration: https://github.com/flyteorg/flyte/issues/3123

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #3391, opened 2023-03-03
Proto-docs: Compile-time and Runtime PodTemplates

eapolinario:
## Tracking issue
fixes https://github.com/flyteorg/flyte/issues/3385

## Describe your changes
Documentation for Compile-time and Runtime PodTemplates feature. 

## Note to reviewers


hamersaw:
A few thoughts here:
(1) Should we be thinking of this as statically vs dynamically set rather than client-side vs server-side? In the realm of Flyte on k8s I think client / server boundaries are a little fuzzy, IMO a compile-time vs runtime differentiation is a little more understandable but maybe I'm too close to this.
(2) Is there anyway we can improve description on the order of application in this PR? With the original `PodTemplate` work that was a big concern that we had many issues on.

nit: stick to uniformity on `PodTemplate` in different sentences we're using `pod template` or `podtemplate`.

cosmicBboy:
Weird, sphinx thinks the subheader hierarchy is off https://github.com/flyteorg/flyte/actions/runs/4320063811/jobs/7539915673#step:5:121

Can you try using `^^^^^^^^^` instead of `------------`?

Also, we should also add a new example in flytesnacks under the Kubernetes Pod integration
https://docs.flyte.org/projects/cookbook/en/latest/auto/integrations/kubernetes/pod/index.html

hamersaw:
> under the Kubernetes Pod integration
should not be here as these will apply everywhere - ContainerTask, ShellTask, Dask / Ray plugins, etc

cosmicBboy:
> > under the Kubernetes Pod integration
> 
> that's not the correct place as these will apply everywhere - ContainerTask, ShellTask, Dask / Ray plugins, etc

In that case, thoughts on where this should go?

eapolinario:
> > > under the Kubernetes Pod integration
> > 
> > 
> > that's not the correct place as these will apply everywhere - ContainerTask, ShellTask, Dask / Ray plugins, etc
> 
> In that case, thoughts on where this should go?

How about we add those in each relevant section? For example:
- [ContainerTask](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/containerization/raw_container.html#sphx-glr-auto-core-containerization-raw-container-py) 
- [ShellTask](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/shell_task.html#sphx-glr-auto-core-flyte-basics-shell-task-py)
- [Dask](https://docs.flyte.org/projects/cookbook/en/latest/auto/integrations/kubernetes/k8s_dask/index.html#plugins-dask-k8s)
- [Ray](https://docs.flyte.org/projects/cookbook/en/latest/auto/integrations/kubernetes/ray_example/ray_example.html#sphx-glr-auto-integrations-kubernetes-ray-example-ray-example-py)

@cosmicBboy , what do you think?

eapolinario:
Here's how the little graph renders in the docs:
Screenshot 2023-06-01 at 10 17 48 AM

rsts/deployment/configuration/general.rst
```diff
@@ -374,14 +442,28 @@ which are applied when constructing a Pod. Typically, these options map one-to-o
...
+     D["@task pod_template_name"] --> B
+
 
-To better understand how Flyte constructs task execution Pods based on the default
-PodTemplate and K8s plugin configuration options, let's take an example. 
+To better understand how Flyte constructs task execution Pods based on Compile-time and Runtime PodTemplates,
+and K8s plugin configuration options, let's take a few examples.
 
-If you have the default PodTemplate defined in the ``flyte`` namespace
+Example 1: Runtime PodTemplate and K8s Plugin Configuration
```
hamersaw, line 463: I think some of these examples may be redundant. Example 3 covers Example 1 and Example 2 right? OK to leave all as well.

rsts/deployment/configuration/general.rst
```diff
@@ -304,10 +305,60 @@ of Flyte's task execution. It ensures complete control over Pod configuration,
...
+the name of the primary container, labels, and annotations.
+
+The term compile-time here refers to the fact that the pod template definition is part of the `TaskSpec <https://docs.flyte.org/projects/flyteidl/en/latest/protos/docs/admin/admin.html#ref-flyteidl-admin-taskclosure>`__.
+
+Runtime PodTemplate
+====================
 
 Set the ``default-pod-template-name`` in FlytePropeller
-========================================================
+--------------------------------------------------------
```
hamersaw, line 361: The `Set the default-pod-template-name in FlytePropeller` above this line is only required when using a default PodTemplate and not if it is set in the task decorator. Does it make sense to make this distinction?

rsts/deployment/configuration/general.rst
```diff
@@ -304,10 +305,60 @@ of Flyte's task execution. It ensures complete control over Pod configuration,
...
+    def t1() -> int:
+        ...
+
+Notice how in this example we are defining a new PodTemplate inline, which allows us to define a full
+`V1PodSpec <https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1PodSpec.md>`__ and also define
+the name of the primary container, labels, and annotations.
+
+The term compile-time here refers to the fact that the pod template definition is part of the `TaskSpec <https://docs.flyte.org/projects/flyteidl/en/latest/protos/docs/admin/admin.html#ref-flyteidl-admin-taskclosure>`__.
+
+Runtime PodTemplate
```
hamersaw, line 357: We should explicitly say that the `default-pod-template-name` is used unless it is overridden by the `pod-template-name` in the task decorator. Only one may be applied. Not sure where is the best fit.

rsts/deployment/configuration/general.rst
eapolinario: Yeah, I just want to make it clear that we can mix and match (up to a point).

rsts/deployment/configuration/general.rst
eapolinario: Good call.

rsts/deployment/configuration/general.rst
eapolinario: I'm going to add a little blurb explaining the difference between the default Runtime PodTemplate and a named one.

hamersaw (dismissed):
Had to read `in case (1) is active and FlytePropeller is building a Pod where (2) was used as part of the definition of the task, then only the values of the PodTemplate mentioned in (2) will be used to build the Pod.` multiple times to understand. Might consider rewording? If not, OK too.

hamersaw (approved):

PR #3394, opened 2023-03-03
Add support prepending or appending paths to flyte-binary ingresses

jeevb:


wild-endeavor (dismissed):
should we delete the file though? `charts/flyte-sandbox/charts/flyte-binary-v0.1.10.tgz`

jeevb:
I'll need to regenerate. I will.

wild-endeavor (approved):

PR #3411, opened 2023-03-07
Make clientIds and self auth server configurable in flyte-binary helm chart

jeevb:


wild-endeavor (approved):
but we can delete the tgz file right?

jeevb:
> but we can delete the tgz file right?

Not sure. Will try cleaning up all tgz files in another PR.

wild-endeavor (approved):

PR #3444, opened 2023-03-10
Fix local compile

jeevb:


wild-endeavor (approved):
note for posterity - this is only for post refactor flyteconsole versions

Issue #475, opened 2020-08-17
[Feature] Generic support for overrides during an execution


kumare3:
**Motivation: Why do you think this is important?**
As a user of Flyte, configuration like "resources (cpu/mem/gpu)", "catalog (enable/disable/version)", "retries",, "spark config", "hive cluster" are created only at registration time. The only way to update is to re-register. This is not desirable, as there are times, when the user may want to update these parameters for specific launchplans, or executions.

**Goal: What should the final outcome look like, ideally?**
A user should be able to optionally override any configuration information for an execution at the launchtime or launchplan creation time.

**Describe alternatives you've considered**
Re-register with the updated configuration

**Flyte component**
- [ ] Overall
- [ ] Flyte Setup and Installation scripts
- [ ] Flyte Documentation
- [ ] Flyte communication (slack/email etc)
- [ ] FlytePropeller
- [ ] FlyteIDL (Flyte specification language)
- [x] Flytekit (Python SDK)
- [x] FlyteAdmin (Control Plane service)
- [x] FlytePlugins
- [ ] DataCatalog 
- [ ] FlyteStdlib (common libraries)
- [ ] FlyteConsole (UI)
- [ ] Other

#### Goal
The purpose of this Allow users to override task metadatas at launch time and decide caching configuration and resource allocation without needing to push a new commit in their PR.

## Motivation
Users have requested cache toggling and cache version override at launch time. Other than that, users have also been wanting to be able to change the task resource quickly. So this is generally a highly requested feature as it allows users to change runtime configurations quickly without needing any code change, meaning that they don't need to go through the entire CI pipelines.

#### Brief description of proposal
We can achieve this by associating each execution with objects `StaticMetadata`,  `MetadataOverride`, and `MetadataFinal` using some simple, serializable format like JSON. 
UX/UI
At registration time we can extract all the metadata of every task and create an `StaticMetadata` object.
At launch time we let users specify `MetadataOverride` through the following ways:
If using UI, in the "Launch Workflow" panel, we have an "Input" tab and a "Execution Parameter Override" tab. The latter tab, by default, shows the `StaticMetadata` but the fields are editable, which allows users to specify overrides of the metadata
If using flyte-cli, we add an option, e.g., `--execution-parameter-override={"task1": {cache=false}}` when launching an execution
Every `MetadataOverride` object is also stored in Flyteadmin, and is versioned with the execution id. 
[...]

PR #3553, opened 2023-03-30
[RFC] Config Override

ByronHsu:
## Tracking issue

https://github.com/flyteorg/flyte/issues/475

## Describe your changes

Add RFC

## Note to reviewers

Need discussion on the UI part!

### DIFF
--- a/rfc/system/3553-config-override.md
+++ b/rfc/system/3553-config-override.md
@@ -1,0 +1,58 @@
+# Flyte Config Overrides RFC
+
+
+https://github.com/flyteorg/flyte/issues/475
+
+## Motivation
+As a user of Flyte, configuration like "resources (cpu/mem/gpu)", "catalog (enable/disable/version)", "retries", "spark config", "hive cluster" are created only at registration time. The only way to update is to re-register. This is not desirable, as there are times when the user may want to update these parameters for specific launchplans, or executions.
+
+Using with_overrides can provide node-level override for workflow, but this is still at compile time. Users need to use dynamic workflow if they want to pass the override values as inputs to the workflow. Moreover, with_overrides can not be applied to reference workflow because the workflow function body is not exposed.
+
+## Problem Statement
+Allow users to override task config at execution time without re-registering the workflows.
+- Support overriding config on both flyteconsole and CLI.  
+
+## API Design
+- flyteremote
+
+```python
+wf_overrides = WFOverride(
+    n0=TaskOverride(
+       container_image="repo/image:0.0.1"
+       limits=Resource(cpu="2")
+    ),
+    n1=WFOverride( # subwf
+        n0=TaskOverride(
+            container_image="repo/image:0.0.3"
+            limits=Resource(cpu="3")
+        )
+    )
+)
+
+flyteremote.execute(wf, wf_overrides)
+```
+
+- flytectl
+
+```python
+$ flytectl create execution --execFile execution_spec.yaml
+workflow: core.control_flow.merge_sort.merge
+wfOverride:
+  n0:
+    container_image: "repo/image:0.0.1"
+    limits:
+      cpu: 2
+  n1:
+    n0:
+      container_image: "repo/image:0.0.3"
+        limits:
+          cpu: 3
+```
+
+## UI design
+
+The hardest part. Need suggestion.
+
+## Implementation 
+
+WIP


bstadlbauer:
I've just read the linked issue (https://github.com/flyteorg/flyte/issues/475) and it seems like there is a lot of good conversation going on there.
From reading it, the trickiest part seems to be doing "per-node" configuration changes, mostly because node names are not trivially knowable before starting a workflow and especially for deeply nested graphs, those are hard to figure out.

Should we finish the discussion on how to best achieve that in https://github.com/flyteorg/flyte/issues/475 and once things are ready copy the result over to this RFC?

davidmirror-ops:
03-30-2023 Meeting notes. Byron Hsu: not sure how to implement this in the UI
FG: It would be a very complicated string. We can mark the tasks as overrideable
BH: what if your task resides at a deeper level
Eduardo: how does this work with Dynamic?
FG: in the dynamic subworkflow when I call a task could be .task.overrride. Whenever user register somethings dynamically with runtime.override: foo it could select runtime-level things like GPU
EA: basically naming overrides
FG: assign the name of the override at registration time
KU: create "override hooks" 
FG: for dynamic it would be impossible
Greg Gydush: regarding the UX, (see https://github.com/flyteorg/flyte/issues/475#issuecomment-1426963056) I'd love to pass promises to the override config. That is the UX that I'd want
KU: we should not confuse them with inputs, it's config. Executions should take a config
HSU: we can parameterize it and pass it to override config
KU: with async tasks you should be able to do something like this
Bernhard: for nested workflows it could be complicated. If you run the same task twice you couldn't run it with different resources
KU: we could call it dynamic_config. In the dynamic case
Dan Rammer: scalability concerns. Checking for naming conflicts would be necessary but complicated
KU: what about making this an experimental feature

yubofredwang:
Hi @fg91, you approach looks pretty neat. Can you add more details on how `model_1_resources` is defined? Is that a different entity type other than task/workflow?

bstadlbauer:
@fg91 I like this, the UX is great!

Trying to take a step back here, it seems like this is pretty much a special case of "tagging" + matching on a tag (option 2 from https://github.com/flyteorg/flyte/issues/475#issuecomment-1490006426), right? 

So in theory, if `.with_runtime_overrides("train_resources")` would add some sort of tag to the task, say `{"run_time_override": "train_resources"}` this could be one of many things to match on. 

I.e. on the backed, we could represent an override as something like (pseudocode; names all TBD): 

```python
class NodeMatchCriteria:
    task_name: str
    node_name: Optional[str] = None
    tags: Dict[str, str] = {}
    function_arguments: Dict[str, Any]
    ...

class ConfigOverride:
    match: NodeMatchCriteria
    resources: Resources
    ...
```

I think if we could keep the matching a bit more general we could set us up for future usecases without too much additional implementation effort. One thing this would also allow (on the backend) is to re-run a workflow with override based on a node-name (as the node name is known from the UI in the second run).

What do you think?

ByronHsu:
My concern with "named override" method is that users can only override the configs on UI/CLI, but cannot override them inside the code. There are two layers to override in the code.

1. constant value (already supported)

```
@task
def t1():
  ...
  
@workflow
def wf():
  t1.with_override(cpu=1)
```

2. promise value

```
@task
def t1():
  ...
  
@workflow
def wf(x):
  t1.with_override(cpu=x)
```

The second case would be extremely useful when reference workflows are shared across the team, but each team wants to tailor the config to their need.

```
@ref_launch_plan
def ref_wf(cpu:int):
  ...
```

```
# team1
@workflow
def wf():
   ref_wf(cpu=3)

# team2
@workflow
def wf():
   ref_wf(cpu=2)
```

ByronHsu:
To resolve to second case above, one possible way is to define a special input parameter (`task_config_override`) with the special dataclass (`TaskConfigOverride`)  that will be parsed in runtime and override `taskTemplate`. 

```
@task(requests=Resources(cpu="3"), limits=Resources(cpu="4"))
def python_task(task_config_override: TaskConfigOverride):
    print("hello python task")
 
cfg = PythonTaskConfigOverride(
        requests=ResourcesOverride(cpu=requests_cpu),
        limits=ResourcesOverride(cpu=limits_cpu)
 )
 
@workflow
def python_wf(requests_cpu: str, limits_cpu: str):
    python_task(task_config_override=cfg)

```

fg91:
I'm personally not a fan of passing task resources (or their overrides) as worklow/task arguments since it mixes logic with resource configs:

> ```python
> @task
> def t1():
>  ...
>  
>@workflow
> def wf(x):
>  t1.with_override(cpu=x)
> ```

When being able to modify the code and not having to rely on overrides at runtime, what is the reason for not using the existing `.with_overrides`?

ByronHsu:
https://docs.google.com/document/d/1gaWU3lsa66APG2aD95_BeL9hFEsEsFiNGGj7KcQmt8c/edit?usp=sharing

l listed down our usecase to better step back and think from the user side

ByronHsu:
To resolve the concern that **users cannot override them inside the code** regarding @fg91's idea. Me and @pingsutw came up with a solution.
If they don't want to specify on UI, they can pass the override config through workflow definition. 

```
@workflow(
  override={
    "model_1_resources": <override obj 1>
    "model_2_resources" <override obj 2>
  }
)
def my_wf():
  ...
```

note: `<override obj x>` will be the default value of override values of the UI.

fg91:
That looks like a very nice UX to me @ByronHsu 👏

ByronHsu:
yeah that would be nice @fg91

kumare3:
@ByronHsu can we update the rfc itself with the proposal?

ByronHsu:
@goyalankit What do you think ^?

rfc/system/3553-config-override.md
```diff
@@ -0,0 +1,102 @@
...
+    runtime_override={
+      "task-yee": TaskNodeConfigOverride(...),
+      "task-ketan": TaskNodeConfigOverride(...)
+    }
+)
+```
+
+## Concerns
+
+1. How to resolve when two hooks have the same names?
```
kumare3, line 101: both will allow the same override?

rfc/system/3553-config-override.md
```diff
@@ -0,0 +1,102 @@
...
+- The flytekit compiler should error out
```
kumare3, line 102: it should be the backend compiler.

rfc/system/3553-config-override.md
ByronHsu: we can do that

bstadlbauer (commented):
Thanks @ByronHsu for being persistent here even though this took longer than anticipated! Let's make sure to get this across the finish line and start implementing soon 👍 

I think the UX is great now! The part that is still unclear to me is how the `TaskNodeConfigOverride` is tied to a particular task on the backend.

Two options would come to mind:
1. `.with_runtime_override("task-yee")` would fill a new field (something like `string runtime_override_name`) in the [`TaskNode`](https://github.com/flyteorg/flyteidl/blob/9247ff0474e9bee94a1ef47a0ab228044816e536/protos/flyteidl/core/workflow.proto#L50-L58) proto. Then the [`WorkflowMetadata`](https://github.com/flyteorg/flyteidl/blob/9247ff0474e9bee94a1ef47a0ab228044816e536/protos/flyteidl/core/workflow.proto#L179), the [`LaunchPlanSpec`](https://github.com/flyteorg/flyteidl/blob/9247ff0474e9bee94a1ef47a0ab228044816e536/protos/flyteidl/admin/launch_plan.proto#L80) as well as the  [`ExecutionSpec`](https://github.com/flyteorg/flyteidl/blob/9247ff0474e9bee94a1ef47a0ab228044816e536/protos/flyteidl/admin/execution.proto#L253) would have a new field `map<string, TaskNodeConfigOverride> runtime_overrides`.

2. We add a new field `map<string, string> tags` to [`TaskNode`](https://github.com/flyteorg/flyteidl/blob/9247ff0474e9bee94a1ef47a0ab228044816e536/protos/flyteidl/core/workflow.proto#L50-L58). Then `.with_runtime_override("task-yee")` would set a `"runtime_override_name": "task-yee"` tag on the node. The `TaskNodeConfigOverride` would then need to know what it applies to (similar to `NodeMatchCriteria` in [this comment](https://github.com/flyteorg/flyte/pull/3553#issuecomment-1491684354)).
While the UX would be the same, this would also allow for other matches (e.g. based on node name or like [here](https://github.com/flyteorg/flyte/issues/475#issuecomment-1460965254)) in the future, without us needing to change the flyteidl API.

Also, should we add a sentence about precedence of these? E.g. something like UI (=Execution) > LaunchPlan > Workflow?

rfc/system/3553-config-override.md
```diff
@@ -0,0 +1,102 @@
+# Flyte Config Overrides RFC
+
+
+https://github.com/flyteorg/flyte/issues/475
+
+## Motivation
+As a user of Flyte, configuration like "resources (cpu/mem/gpu)", "catalog (enable/disable/version)", "retries", "spark config", "hive cluster" are created only at registration time. The only way to update is to re-register. This is not desirable, as there are times when the user may want to update these parameters for specific launchplans, or executions.
+
+Using with_overrides can provide node-level override for workflow, but this is still at compile time. Users need to use dynamic workflow if they want to pass the override values as inputs to the workflow. Moreover, with_overrides can not be applied to reference workflow because the workflow function body is not exposed.
```
fg91, line 9:
```suggestion
Using `with_overrides` can provide node-level override for workflow, but this is still at compile time. Users need to use dynamic workflow if they want to pass the override values as inputs to the workflow. Moreover, with_overrides can not be applied to reference workflow because the workflow function body is not exposed.
```

Minor nit

rfc/system/3553-config-override.md
```diff
@@ -0,0 +1,102 @@
...
+  t1() # this task node cannot be overridden
+  t1().with_runtime_override("task-yee") # can be overriden under the name "task-yee"
+  t2().with_runtime_override("task-ketan") # can be overriden under the name "task-ketan"
+  t3() # this task node cannot be overridden
+```
+
+We will have a proto called `taskNodeConfigOverride`, which contains all the overridable fields.
+
+```proto
+message TaskNodeConfigOverride {
```
fg91, line 35: Does this duplicate logic that already exists for `with_overrides()`? I think it would be nice if these were handled "as similarly as possible".

rfc/system/3553-config-override.md
```diff
@@ -0,0 +1,102 @@
...
+def wf():
+  t1() # this task node cannot be overridden
+  t1().with_runtime_override("task-yee") # can be overriden under the name "task-yee"
+  t2().with_runtime_override("task-ketan") # can be overriden under the name "task-ketan"
+  t3() # this task node cannot be overridden
+```
+
+### 3. launch plan
+
+Users can also provide predefined values by `LaunchPlan.get_or_create`. If users provide both workflow decorator and launch plan predefined values, launch plan one will override workflow decorator one.
```
fg91, line 86:
```suggestion
Users can also provide overrides to `LaunchPlan.get_or_create`. If users provide both workflow decorator and launch plan predefined values, launch plan one will override workflow decorator one.
```

rfc/system/3553-config-override.md
```diff
@@ -0,0 +1,102 @@
...
+def t2():
+  ...
+
+@workflow
+def wf():
+  t1() # this task node cannot be overridden
+  t1().with_runtime_override("task-yee") # can be overriden under the name "task-yee"
+  t2().with_runtime_override("task-ketan") # can be overriden under the name "task-ketan"
+  t3() # this task node cannot be overridden
+```
```
fg91, line 30:
```suggestion
```

One main motivation for introducing such a "hook mechanism" based on identifiers/names is that we couldn't come up with a good answer how a good UX can be achieved when having to specify overrides *in a nested workflow graph* in the UI, in code, or via the CLI. This way, the overrides can be specified in a simple map and the structure of the workflow graph does not matter/does not have to be shown.
```

**Review by fg91 — APPROVED** [review]
I very much like the proposed UX now 👏 

`with_runtime_overrides` was the first name that came to my mind when I first had this idea. @kumare3 called it `with_override_hook`(?) in the contributors meeting. Which name should we write into the RFC in the end?

**Comment by davidmirror-ops** [comment]
2023-05-11: Agree on naming before merge, also address implementation concerns. Byron might need help to distribute implementation tasks. The suggestion is to work on backend first (propeller and flytekit first). TSC happy to help kickoff implementation.

**Comment by pingsutw** [comment]
Here are all the different ways to override the config. 
- wf2.py
```python
from flytekit import task, workflow, Resources
@task
def t1():
    print("hello world")

@workflow
def sub_wf():
    t1()
```
- wf1.py
```python
from flytekit import task, workflow, Resources
from wf2 import sub_wf

@task
def t0():
    print("hello world")

@workflow()
def wf():
    t0()
    sub_wf()
```
1. To override the `t0` at runtime
```python
# wf1.py
@workflow()
def wf():
    t0().with_runtime_override("model_1_resources", runtime_override_config(cpu=1, mem="1Gi"))
    ...
```
`cpu=1, mem="1Gi"` is default execution config, people are still able to update the cpu and mem on flyteConsole.



2. To override the `t1` at runtime. update `wf2.py`

```python
# wf2.py
@workflow
def sub_wf():
    t1().with_runtime_override("model_2_resources", runtime_override_config(cpu=1, mem="1Gi"))
```
3. To override the default execution config of `t1` from `wf1.py`. Add override config to @workflow decorator. Then `t1` task will use 2 cpu by default.

```python
# wf1.py
@workflow(override={"model_2_resources": runtime_override_config(cpu=2, mem="2Gi")})
def wf():
    ...  
```

FlyteIDL: 
1. Add more config to [TaskOverrides](https://github.com/flyteorg/flyteidl/blob/b4947af200685e6eb1d43e82e50cd51006c02219/protos/flyteidl/core/workflow.proto#L249-L252)
2. Add string `runtime_override_name` to `TaskNode`
3. Add `runtimeOverrideConfigs` (map<string, TaskOverrides>) to [ExecutionSpec](https://github.com/flyteorg/flyteidl/blob/44ae4010d76267dfa2fbcd42148781be067ed3bb/protos/flyteidl/admin/execution.proto#L253)
4. Add `runtimeOverrideConfigs` (map<string, TaskOverrides>) to [LaunchPlanSpec](https://github.com/flyteorg/flyteidl/blob/44ae4010d76267dfa2fbcd42148781be067ed3bb/protos/flyteidl/admin/launch_plan.proto#L80)

FlyteAdmin:
1. Iterate `runtimeOverrideConfigs`. if the `runtime_override_name` in the `TaskNode` matches one of the elements in the runtimeOverrideConfigs, then update the task config (cpu, role, etc). https://github.com/flyteorg/flyteadmin/blob/610451d21ca2fa8f598ebc26b8ebb96e423e270a/pkg/manager/impl/execution_manager.go#L780-L783

Flytekit: 
1. Add `runtimeOverrideConfigs` to [LaunchPlanSpec](https://github.com/flyteorg/flytekit/blob/c363881c4aeb4fe9a4479e5ff2eef7c03c13e87e/flytekit/tools/translator.py#L343-L357)
2. Add `runtimeOverrideConfigs` to [ExecutionSpec](https://github.com/flyteorg/flytekit/blob/faf02b11fc4e32c2652e63296f5b735be36cdcb5/flytekit/remote/remote.py#L967)
3. Update Flytekit remote and CLI (`pyflyte run --remote --override "{'model_1_resources': {'cpu': 3}}" wf1.py wf`). 

@ByronHsu is going to update IDL. cc @fg91 @bstadlbauer would you like to help work on flytekit and admin?

ByronHsu:
@pingsutw do we also want to override on `launchplan.get_or_create`? https://github.com/flyteorg/flyte/blob/c77eeaa89a1ff7fffdc30537868b19cc98bb3ea6/rfc/system/3553-config-override.md#3-launch-plan

fg91:
@pingsutw @ByronHsu what do you think of this:

```python
# wf1.py
@workflow
def wf():
    t0()
    sub_wf().override={"model_2_resources": runtime_override_config(cpu=2, mem="2Gi")}
```

instead of 

```python
# wf1.py
@workflow(override={"model_2_resources": runtime_override_config(cpu=2, mem="2Gi")})
def wf():
```

?


> @pingsutw do we also want to override on launchplan.get_or_create?

Yes, I think this would be very useful :+1:

pingsutw:
@fg91 I like that, too. It will be much easier to implement.

> do we also want to override on launchplan.get_or_create?

yup, we should support it as well

ByronHsu:
what if users want to override a task under the subworkflow's subworkflow? and users want to do that inside `wf` function. I think its more generic if users can override everything down the workflow on the top level.

```
# wf1.py
@workflow
def sub_wf():
   sub_sub_wf()
   ...

@workflow
def wf():
    t0()
    sub_wf().override={"model_2_resources": runtime_override_config(cpu=2, mem="2Gi")}
```

pingsutw:
@ByronHsu You still do it at top-level workflow. 

old one:
```python
@workflow(override={"model_2_resources": runtime_override_config(cpu=2, mem="2Gi")})
def wf():
    t0()
    sub_wf()
```
Fabio suggested:
```python
@workflow
def wf():
    t0()
    sub_wf().override={"model_2_resources": runtime_override_config(cpu=2, mem="2Gi")}

```

ByronHsu:
got it. So you meant if `model_2_resources` is under `sub_wf`'s subflow's task, we can still override it. 
yeah i agree @fabio's idea is cleaner because we don't have to mess up `@workflow` argument.

pingsutw:
`with_runtime_override` can return promise. so it could be
```python
sub_wf().with_runtime_override(...).with_runtime_override(...).with_runtime_override(...) 

```

ByronHsu:
We can maybe provide 'with_runtime_override' to override a single task and 'with_runtime_overiddes' for multiple

bstadlbauer:
I'm in favor of chaining (e.g. `sub_wf().with_runtime_override(...).with_runtime_override(...).with_runtime_override(...)`) instead of two different methods. 

@pingsutw Happy to help out!

fg91:
Instead of chaining or two methods, one with `s`, we could also simply pass a dict?

```python
sub_wf().with_runtime_override({"model_2_resources": runtime_override_config(cpu=1, mem="1Gi"), "model_3_resources": ...})

```

Also happy to take a ticket @pingsutw :)

bstadlbauer:
Also works of course 👍

pingsutw (approved):

rfc/system/3553-config-override.md
```diff
@@ -0,0 +1,102 @@
...
+
+We provide multiple ways to override the task nodes at runtime, including UI, workflow decorator, launch plan, etc.
+
+### 1. UI
+
+The registered task nodes will prompt on UI for users to assign config override values.
+
+<img src="../images/config-override-ui.png" style="width: 50%; height: 50%">
+
+### 2. workflow decorator
```
fg91, line 66: @ByronHsu could you please update this with the latest change? 🙏

rfc/system/3553-config-override.md
```diff
@@ -0,0 +1,102 @@
...
+https://github.com/flyteorg/flyte/issues/475
+
+## Motivation
+As a user of Flyte, configuration like "resources (cpu/mem/gpu)", "catalog (enable/disable/version)", "retries", "spark config", "hive cluster" are created only at registration time. The only way to update is to re-register. This is not desirable, as there are times when the user may want to update these parameters for specific launchplans, or executions.
+
+Using with_overrides can provide node-level override for workflow, but this is still at compile time. Users need to use dynamic workflow if they want to pass the override values as inputs to the workflow. Moreover, with_overrides can not be applied to reference workflow because the workflow function body is not exposed.
+
+## Design
+
+By default, all task nodes are not overridable at runtime. However, users can call `.with_runtime_override(name: str)` on task nodes to register a "hook" on tasks, which allows the task to be indentified by `name` and overridden with new config at runtime.
```
eapolinario, line 13: typo: indentified -> identified.

rfc/system/3553-config-override.md
```diff
@@ -0,0 +1,102 @@
...
+  ...
+
+@task
+def t2():
+  ...
+
+@workflow
+def wf():
+  t1() # this task node cannot be overridden
+  t1().with_runtime_override("task-yee") # can be overriden under the name "task-yee"
```
eapolinario, line 27: Can we clarify what happens if `t1` defines a pod template? What takes precedence?

eapolinario (commented):
This is shaping up to be a great feature!

rfc/system/3553-config-override.md
ByronHsu: Will use the original one instead

bstadlbauer:
@ByronHsu Also ok with this after all changes discussed there are incorporated in the RFC

rfc/system/3553-config-override.md
fg91: @davidmirror-ops let's discuss this point in tomorrows contributors' sync?

bstadlbauer:
Also, one question that came up during a first stab at implementing this - what would the UX for overriding `task_config` look like?

PR #3609, opened 2023-04-20
Use additional ca-certificates in flyte sandbox configuration directory

jeevb:
Following k3s instructions here: https://ranchermanager.docs.rancher.com/v2.5/reference-guides/single-node-rancher-in-docker/advanced-options#custom-ca-certificate

Placing extra CA certs in PEM format within `~/.flyte/sandbox/ca-certificates` will suffice to add those to the sandbox trust store.


wild-endeavor (approved):

jeevb:
In retrospect, I think `~/.flyte/sandbox/ca-certificates` is a more appropriate location for storing extra CA certificates. That should be mounted into `/var/lib/flyte/ca-certificates`. Unfortunately, this will require a `flytectl` change as well.

PR #3631, opened 2023-05-01
feat: allow seedProjects to be set via inline configuration

### LINKED ISSUE #3545: [Core feature] Allow flyte-binary to specify SeedProjects

I'd like to be able to set [adminServer.SeedProjects](https://github.com/flyteorg/flyte/blob/507a466077883bd2c34677c928ac277be7be9c28/cmd/single/start.go#L68-L70) via the flyte-binary helm chart values.yaml so that I can make our flyte installation more declarative.

This allows a separation of concerns so that a system operator (me) can create k8s resources (namespaces, etc) and flyte resources (project/domain) in one configuration file which is under SCM.

A user of the flyte-binary helm chart should be able to set `configuration.admin.seedProjects` to a list of project names and those projects should be passed to [flyte admin on startup](https://github.com/flyteorg/flyte/blob/507a466077883bd2c34677c928ac277be7be9c28/cmd/single/start.go#L68).

(default value = ["flytesnacks"]) 

1. Creating a `flyte-operator` which knows how to read a CRD of desired flyte projects
2. Hooking flytectl into our [pulumi](https://www.pulumi.com/) powered k8s management

The first option is.. just a lot of work!

The second option is far less work, viable and a completely valid path forward. After discussing with jeev and Yee it seemed both thought the proposed solution is acceptable.

Discussed here: https://flyte-org.slack.com/archives/C01P3B761A6/p1679937303586009

I'll go ahead and create a PR for it in the coming days.

- [X] Yes

- [X] Yes

ossareh:
## Tracking issue

Closes #3545 

## Describe your changes

Add a `SeedProjects` key to the Admin configuration struct in `cmd/single`. This allows users to specify which projects should be created at start up. This allows management of the flyte installation to be more declarative via the helm chart installation path.

## Note to reviewers

This is not currently ready for review; marking as draft via github


ossareh:
suggestion: I should rename `seedProjects` to `initialProjects` - this keeps it inline with `flyte-core`'s options.

cmd/single/config.go
```diff
@@ -5,7 +5,13 @@ import "github.com/flyteorg/flytestdlib/config"
 //go:generate pflags Config --default-var=DefaultConfig
 
 var (
-	DefaultConfig = &Config{}
+	DefaultConfig = &Config{
+		Propeller: Propeller{},
```
wild-endeavor, line 9: i think the empty ones you don't have to specify...

deployment/sandbox-binary/flyte_sandbox_binary_helm_generated.yaml
```diff
@@ -115,6 +115,19 @@ data:
...
           auth_type: iam
       container: my-organization-flyte-container
+  005-flyte-binary-config.yaml: |
+    flyte:
+      admin:
+        disableClusterResourceManager: false
+        disableScheduler: false
+        disabled: false
+        seedProjects:
+        - flytesnacks
```
ossareh, line 125: is this indentation incorrect? I'd have expected it to be two spaces to the right.

Everything else seems indented correctly 🤔

charts/flyte-binary/Chart.yaml
```diff
@@ -7,7 +7,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: v0.1.10 # VERSION
+version: v0.1.11 # VERSION
```
jeevb, line 10: This is a magic number I believe. It is set on release. You can just leave this as is.

charts/flyte-binary/values.yaml
```diff
@@ -119,6 +119,28 @@ configuration:
       repository: cr.flyte.org/flyteorg/flytecopilot # FLYTECOPILOT_IMAGE
       # tag CoPilot sidecar image tag
       tag: v0.0.28 # FLYTECOPILOT_TAG
+  flyte-binary:
```
jeevb, line 122: Could this block be moved under the `configuration` block instead? `flyte-binary` might not be necessary here.

charts/flyte-binary/templates/configmap.yaml
```diff
@@ -190,6 +190,8 @@ data:
       security:
         useAuth: true
   {{- end }}
+  005-flyte-binary-config.yaml: |
```
jeevb, line 193: Would it make sense for this to be merged into `000-core.yaml` instead?

deployment/sandbox-binary/flyte_sandbox_binary_helm_generated.yaml
jeevb: LGTM!

charts/flyte-binary/Chart.yaml
ossareh: I've rolled this back.

charts/flyte-binary/templates/configmap.yaml
ossareh: moved up to `000-core.yaml`; I agree that is a better place for it to live.

wild-endeavor:
I'd rather move the configuration out of the `configuration` block to a top level key in the values file.  Call it something like `flyte-single-binary-component-configuration`.  I like to be long and clear in my naming.

wild-endeavor (approved):

PR #3727, opened 2023-05-25
Adapt execution tags RFC with results from contributor's sync discuss…

fg91:
## Describe your changes

@kumare3 this PR amends the system tags RFC according to what we discussed in the last contributors sync.

### DIFF
--- a/rfc/system/0001-flyte-execution-tags.md
+++ b/rfc/system/0001-flyte-execution-tags.md
@@ -18,12 +18,13 @@ motivation of how this feature could further improve other ecosystem projects.
 ## 2 Motivation
 
 As a User I want to
- - Group a certain number of executions into an experiment group - for ease of debugging and discovery.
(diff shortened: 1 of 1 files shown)


rfc/system/0001-flyte-execution-tags.md
```diff
@@ -58,19 +59,11 @@ flytectl get execution -p flytesnacks -d development --filter.labels="k:v"
 ```
 
 ### UI Interface
+All labels are treated the same way and allow search/filter and click based grouping.
+Users will get the regular executions view with all the labels available on each execution.
+The users are allowed to filter an exection simply by clicking on a label and then all
+executions are filtered by that label. 
 
-Two approaches
```
fg91, line 62: In the previous contributors synch we discussed that option 1 is preferable since we don't want to impose on users which label keys are considered special in their domain.

kumare3:
+1

kumare3 (approved):

fg91:
@kumare3 can this be merged? Don't have perms.

PR #3755, opened 2023-06-07
RFC - Flyte Console UI Upgrade

tsheiner:
## Describe your changes

The proposal is to do a modest branding, orientation & navigation enhancement to the Flyte Console:
1. move the application header to the left side of the viewport
1. introduce consistent bread crumb element on all pages
1. introduce consistent page header on all pages
1. enable the bread crumb & page header elements to also function as navigation controls

## Screenshots

Please review the RFC for details and illustrations.


fg91 (approved):
The proposed changes look good to me.

I would like to propose one additional change:

Screenshot_2023-06-08_at_19_20_18

I often work on a 13 inch macbook without an external monitor. On such a screen size the version string is only partly visible. When one hovers over it with the mouse, the full string is not shown. Initially I didn't know a better way other than zooming out in the browser to see the full version. Now I discovered that one can mark the visible part of the version but copy the entire version. I wished that there was a better way to see the full version when the screen size is limited.

tsheiner:
@fg91 Thank you for your response and feedback. I am in the early stages of a redesign of the layout of the execution page view and will include your concerns in my thinking.

bstadlbauer (approved):
Looks good to me, thank you!

Slightly perpendicular, but I can only second what Fabio has said about it being nice to also test against small (13 inch?) screens :-)

davidmirror-ops:
03-08-2023 Contributors sync notes: already implemented.

PR #3779, opened 2023-06-13
Update agent endpoint in the single binary

pingsutw:
## Describe your changes

update endpoint. `flyte-sandbox-http` -> `flyte-binary-http`


eapolinario:
We should keep an override in the case of sandbox. More specifically, we need to add something like https://github.com/flyteorg/flyte/blob/master/charts/flyte-binary/values.yaml#L151-L154 to the [sandbox values.yaml](https://github.com/flyteorg/flyte/blob/master/charts/flyte-sandbox/values.yaml) file.

eapolinario (approved):

PR #3788, opened 2023-06-16
Add service account for agent

pingsutw:
As the title, add a service account without a role for the agent. people can add annotations, and image secrets for SA by updating the helm chart value.yaml

### DIFF
--- a/charts/flyte-core/templates/agent/rbac.yaml
+++ b/charts/flyte-core/templates/agent/rbac.yaml
@@ -1,0 +1,17 @@
+{{- if .Values.flyteagent.enabled }}
+---
+{{- if .Values.flyteagent.serviceAccount.create }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "flyteagent.name" . }}
+  namespace: {{ template "flyte.namespace" . }}
+  labels: {{ include "flyteagent.labels" . | nindent 4 }}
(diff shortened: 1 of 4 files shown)


pmahindrakar-oss (approved):

charts/flyte-core/templates/agent/rbac.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: ServiceAccount
+metadata:
+  name: {{ template "flyteagent.name" . }}
+  namespace: {{ template "flyte.namespace" . }}
+  labels: {{ include "flyteagent.labels" . | nindent 4 }}
+  {{- with .Values.flyteagent.serviceAccount.annotations }}
+  annotations: {{ tpl (toYaml .) $ | nindent 4 }}
+  {{- end}}
+{{- with .Values.flyteagent.serviceAccount.imagePullSecrets }}
+imagePullSecrets: {{ tpl (toYaml .) $ | nindent 2 }}
```
pmahindrakar-oss, line 14: Shouldn't this be under spec

pingsutw: No, https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-image-pull-secret-to-service-account. We add it to the spec if it's a pod

Issue #3769, opened 2023-06-11
[Core feature] Configuration containing secrets should use Opaque Secrets over ConfigMaps


EternalDeiwos:
### Motivation: Why do you think this is important?

The way secrets are handled in deployment cause several headaches:

1. Secrets are supplied in plain text in Helm configuration; this is a problem when deployments and Helm `values.yaml` contents are stored in a git repository and deployed using an external tool.
1. There are no existing configuration to pull configuration and/or secrets from existing configmaps or secrets; this makes it difficult inject secrets and configuration from a central vault (e.g. Hashicorp Vault).
1. Potentially long-lived secrets are being kept in ConfigMaps instead of Opaque Secrets; this is generally not idiomatic even if the security of the two option were roughly equivalent.

This is creating a significant barrier to me deploying a _test_ instance in our environment (bare metal K8s).

### Goal: What should the final outcome look like, ideally?

I would like to supply names of existing secrets to the Helm chart and/or Kustomization without directly exposing secrets in configuration files or Helm values. This will allow incremental improvement of my deployment configuration without needing to expose secrets directly as part of the deployment.

### Describe alternatives you've considered 

I've been working on a Kustomization with significant patching to move existing configmaps into secrets... however I am considering a completely parallel deployment because writing the patches takes more time than editing the original files.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

jeevb:
@EternalDeiwos: Do you already have a list of long-lived secrets are incorrectly stored in a configmap right now?

EternalDeiwos:
@jeevb the ones that come to mind based on my currently limited experimentation are:

- `client_id` and `client_secret` for OIDC.
- S3 credentials
- Postgres credentials &mdash; this is already read from a mounted secret, which is ideal, however this could use some better documentation for providing that secret externally.

Other than that, the same would apply to literally anything in the configuration that constitutes a secret or externally issued credential.  One would need to scan through all existing deployment configuration to check that.

Also, another caveat is that I have been looking at the Kustomize scripts not necessarily the Helm chart output &mdash; which I understand the project is moving away from? Unsure of the precise state of the Helm charts but based on conversations on Slack it is clearly different to what I understood.

Will provide further feedback as soon as I can get back to this.

jeevb:
[...]

PR #3807, opened 2023-06-26
Add support for leveraging pre-created secrets in flyte-binary helm chart

jeevb:
Adds support for referencing pre-created Secret objects for sensitive configuration variables (e.g. DB password, auth client secrets, etc.). Also moves sensitive configuration into a Secret object as opposed to ConfigMap.

Addresses: #3769

### Secrets in Flyte configuration

1. Create an external secret containing info such as DB password, S3 access/secret key, client secret hash, etc as follows:
```
$ cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Secret
metadata:
  name: flyte-binary-inline-config-secret
  namespace: flyte
type: Opaque
stringData:
  202-database-secrets.yaml: |
    database:
      postgres:
        password: <DB_PASSWORD>
  203-storage-secrets.yaml: |
    storage:
      stow:
        config:
          access_key_id: <S3_ACCESS_KEY>
          secret_key: <S3_SECRET_KEY>
  204-auth-secrets.yaml: |
    auth:
      appAuth:
        selfAuthServer:
          staticClients:
            flytepropeller:
              client_secret: <CLIENT_SECRET_HASH>
EOF
```

2. Reference the newly created secret in `.Values.configuration.inlineSecretRef` in `values.yaml` as follows:
```
configuration:
  inlineSecretRef: flyte-binary-inline-config-secret
```

The above secrets will be merged into the base configuration created by the chart, and values will be available, as expected, to the Flyte platform.

### OIDC/Internal Client secrets

1. Create an external secret containing the client secret values:
```
$ cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Secret
metadata:
  name: flyte-binary-client-secrets-external-secret
  namespace: flyte
type: Opaque
stringData:
  client_secret: <INTERNAL_CLIENT_SECRET>
  oidc_client_secret: <OIDC_CLIENT_SECRET>
EOF
```

2. Reference the newly created secret in `.Values.configuration.auth.clientSecretsExternalSecretRef` in `values.yaml` as follows:
```
configuration:
  auth:
    clientSecretsExternalSecretRef: flyte-binary-client-secrets-external-secret
```


charts/flyte-binary/templates/config-secret.yaml
```diff
@@ -0,0 +1,54 @@
...
+  {{- if .Values.configuration.auth.internal.clientSecretHash }}
+  014-auth-secrets.yaml: |
+    auth:
+      appAuth:
+        selfAuthServer:
+          staticClients:
+            flytepropeller:
+              # Instructions to generate:
+              # https://unionai.slack.com/archives/C04HP9E6WNT/p1678308193819939
+              client_secret: {{ .Values.configuration.auth.internal.clientSecretHash | quote }}
```
EternalDeiwos, line 50: It would be preferable to have `client_id` here as well

jeevb: I don't believe those are intended to be secrets. Should they be?

EternalDeiwos: They are not, however it is inconvenient to have to configure a single credential in two separate locations, for example: you wouldn't configure `AWS_ACCESS_KEY_ID` separately to `AWS_SECRET_ACCESS_KEY`.

jeevb: This particular `Secret` is generated by the chart from config specified in `values.yaml` - for basic use cases, or when installing the chart from another framework like Terraform, this will probably suffice. For cases where you are committing a `values.yaml` to git, you are better off generating a `Secret` with sensitive configuration overrides externally, and referencing that in `.Values.configuration.inlineSecretRef`.

```
$ cat <<EOF | kubectl apply -f -
 apiVersion: v1
 kind: Secret
 metadata:
   name: flyte-binary-inline-config-secret
   namespace: flyte
 type: Opaque
 stringData:
   202-database-secrets.yaml: |
     database:
       postgres:
         password: <DB_PASSWORD>
   204-auth-secrets.yaml: |
     auth:
       appAuth:
         selfAuthServer:
           staticClients:
             flytepropeller:
               client_secret: <CLIENT_SECRET_HASH>
 EOF
 ```

Comment (EternalDeiwos) [reply charts/flyte-binary/templates/config-secret.yaml:50]: This is perfect. I know you provided on Slack already but thanks for recording here as well.

**Comment by wild-endeavor** [comment]
can we add some more context around secrets in general for the documentation?  maybe this is something that david can help with in the near future.  something like the following (but someone needs to go through and flesh out the rest of it).  Note that the below is true only after merging this PR.

The secrets that the flyte binary relies on all ultimately come through in through files.  There are two types of files that can contain secrets.
* Configuration files.  These are a series of yaml files that get loaded via command line args to the executable.  Configuration for the same key (for maps, not lists) can be split across different files.  At run time, flyte's configuration system will merge the files to form one cohesive view of the configuration.  By default these are mounted under `/etc/flyte/config.d/` in the container.  This is an example for the database password for instance.
  ```
    database:
      postgres:
        username: postgres
        password: mypassword
  ```
  This would get picked up by Admin's config logic that looks for the [db password](https://github.com/flyteorg/flyteadmin/blob/53282fe979b628447e72d0e3f8fd0e2b9235d929/pkg/runtime/interfaces/application_configuration.go#L45).
* Secret files.  These are files that contain just a straight value that gets read as a string value and should be the actual password you're looking for.  By default these are mounted under `/etc/secrets/` in the container.


If you're trying to get a secret into the flyte executable's set of configuration files, there's two ways of going about this.
* Set the value as plaintext in your Helm values file.  For example, the database password above is hooked up to the `chart/flyte-binary` Helm chart's value under
  ```
  configuration:
  database:
    password: "mypassword"
  ```
  If you do this, the Helm chart will create a Secret whose value mimics the configuration structure that the flyte binary expects:
  ```
  012-database-secrets.yaml: |
    database:
      postgres:
        password: "mypassword"
  ```
  At run time, the binary will read this in and merge it into the configuration that ultimately gets used.

* Set the value separately as a secret.  The advantage of doing this is that you don't have to have a secret sitting as plaintext in your Helm values file.
  << insert the block titled Secrets in Flyte configuration in the main PR description above >>

  This option basically mirrors the first option, except that you're creating the secret out of band (using TF or some other process).  This style of specifying secrets mimics the "inline" method for specifying additional configuration, hence the name.

* For the secrets that always expect a path *(we should probably list these out)* you can do << insert the second block titled OIDC/Internal Client secrets in the main PR description >>.  The `flyte-binary` Helm chart is set up such that these get mounted directly under the `/etc/secrets/` folder.

**Review by wild-endeavor — APPROVED** [review]

File: charts/flyte-binary/templates/config-secret.yaml [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+  {{- if .Values.configuration.auth.enableAuthServer }}
+  {{- if .Values.configuration.auth.internal.clientSecretHash }}
+  014-auth-secrets.yaml: |
+    auth:
+      appAuth:
+        selfAuthServer:
+          staticClients:
+            flytepropeller:
+              # Instructions to generate:
+              # https://unionai.slack.com/archives/C04HP9E6WNT/p1678308193819939
```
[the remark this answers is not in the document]
jeevb: Good point. Let's drop it.

eapolinario (approved):

charts/flyte-binary/templates/config-secret.yaml
gvashishtha: Does the key before the `.yaml` thing matter? I see you do `202-database-secrets.yaml` in your comment but in the code it is `012-database-secrets.yaml`

jeevb: It gives us control over merge order. The internal range is 1xx, so 2xx is used as the range for externally provided values. Values that are provided in subsequent files will override any existing ones.

PR #3828, opened 2023-07-04
Streamline sandbox and functional tests to avoid disk pressure on CI VM

jeevb:


charts/flyte-binary/values.yaml
```diff
@@ -180,15 +181,6 @@ clusterResourceTemplates:
   # annotations Add annotations to created ConfigMap
   annotations: {}
 
-flyteagent:
```
jeevb, line 183: Removing this will be breaking, if this config is already being used.

PR #3837, opened 2023-07-05
Update auth config instructions

### LINKED ISSUE #3029: [Docs] Add steps for configuring self auth server
### Description

Current docs for authentication https://docs.flyte.org/en/latest/deployment/cluster_config/auth_setup.html for flyte don't mention how to setup self Auth server and related config changes.

Add a section where we details steps for these for setting up auth for 
* user auth using flyteconsole
* app auth for flytescheduler, flytepropeller

https://flyte-org.slack.com/archives/CP2HDHKE1/p1666278679120739

- [X] Yes

- [X] Yes

davidmirror-ops:
## Tracking issue

Closes #3029 

## Describe your changes

A first attempt to provide updated and easier to follow instructions to setup auth for both `flyte-binary` and `flyte-core`

I haven't been able to reproduce the Azure AD config as external auth server, so it is left as a separate section to be addressed in an upcoming PR

### DIFF
--- a/rsts/deployment/configuration/auth_setup.rst
+++ b/rsts/deployment/configuration/auth_setup.rst
@@ -571,14 +571,14 @@ Follow the steps in this section to configure `flyteadmin` to use an external au
    
          secrets:
            adminOauthClientCredentials:
-             enabled: true
+             enabled: true # see the section "Disable Helm secret management" if you require to do so
              # Replace with the client_secret provided by your IdP for flytepropeller.
              clientSecret: <client_secret>
              # Replace with the client_id provided by provided by your IdP for flytepropeller.
              clientId: <client_id>
       
-      1. Save your changes
-      2. Upgrade your Helm release with the new configuration:
+      2. Save your changes
+      3. Upgrade your Helm release with the new configuration:
 
       .. prompt:: bash $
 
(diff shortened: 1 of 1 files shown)


davidmirror-ops:
I don't know how to fix those build errors. I struggle quite a bit with RST syntax

davidmirror-ops:
Thanks @wild-endeavor for your help!

Your reviews are appreciated

rsts/deployment/configuration/auth_setup.rst
```diff
@@ -51,178 +41,346 @@ Authentication Setup
...
+Custom Authorization Server
+***************************
 
-An OAuth2 Authorization Server allows external clients to request to authenticate and act on behalf of users (or as their own identities). Having
-an OAuth2 Authorization Server enables Flyte administrators control over which apps can be installed and what scopes they are allowed to request or be granted (i.e. what privileges can they assume).
 
-Flyte comes with a built-in authorization server that can be statically configured with a set of clients to request and act on behalf of the user.
-The default clients are defined `here <https://github.com/flyteorg/flyteadmin/pull/168/files#diff-1267ff8bd9146e1c0ff22a9e9d53cfc56d71c1d47fed9905f95ed4bddf930f8eR74-R100>`__
-and the corresponding section can be modified through configs.
+Flyte ships with an internal authorization server, hence setting up an external Authorization Server is optional but recommended. Having an external Authorization Server enables Flyte administrators control over which apps can be installed and what privileges can be granted to users.
```
wild-endeavor, line 381: this isn't true... having an external auth server does not enable Flyte administrators to control over which apps can be installed.  In any case, what do you mean by an app here?  Apps aren't installable in Flyte at all and this auth server is not the rbac that people often ask about ('i want to limit some users to only be able to access certain flyte projects').  Maybe just truncate this line to after "is optional".

davidmirror-ops: That line is on the current version of the docs ([here](https://docs.flyte.org/en/latest/deployment/configuration/auth_setup.html#oauth2-authorization-server)) but agree that it can be misleading. Nevertheless, the docs should provide some reasoning on what are the reasons/situations when using an external auth server is preferred.

davidmirror-ops:
> good for the most part. but there is a crucial bit that i think is misleading. the auth server that comes stock with Flyte is good enough to run in production environments. Is there a reason why we're saying it's not?

In part, it comes from this sentence on the current version of the docs:

>it is not necessary to set up an external Authorization Server. However, it is recommended to do so to maintain the security of the configured apps in a secure location

While the OAuth2.0 protocol doesn't grade the security of an auth server by where it runs, and while the internal auth sever is great, probably for environments with more strict security guidelines, maintaining the auth server separate from the resource server could be helpful, instead of doing the opposite. 

Probably where I'm failing here is by elevating this to the category of "recommendation". Both auth server options are ready for production environments. Choosing one over another just depends on the features and level of control or integration with current security mechanisms our users want/need.

samhita-alla (approved):

PR #3844, opened 2023-07-06
[RFC] Artifact service

wild-endeavor:
RFC for Artifact service.


rfc/system/3844-artifact-service.md
```diff
@@ -0,0 +1,148 @@
...
+Type checking will happen in Admin as soon as the CreateExecutionRequest comes in, it can retrieve the underlying workflow/task output types and perform the type check.
+
+### Introduction of Artifact
+If you squint at the above from a one-thousand meter view, the behaviour of the URL starts to look like a cataloging service. There's a unique identifier for everything produced by Flyte, it should be usable across the platform and it should be persisted. Since the catalog name was already taken, we are calling this Artifact. An Artifact is nothing more than a tracked piece of metadata associated with a Literal and LiteralType.
+
+* Artifacts have a unique identifier per Flyte installation 
+* All outputs in Flyte automatically come with an Artifact (every output of every task/workflow execution).
+* Artifacts can have additional metadata data
+  * Can keep type information (dataset, model, maybe a "value" which is used for Scalars that are important)
+  * Can have aliases, so you should be able to alias an artifact with `latest` and then reference it.
```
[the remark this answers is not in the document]
fg91: Update: this is partially answered below but leaving by thoughts here for visibility. Feel free to resolve.

rfc/system/3844-artifact-service.md
```diff
@@ -0,0 +1,148 @@
...
+* Run the Python object through the flytekit Type Engine.
+* Return the Artifact object containing a possibly different flyte URL to the user. Attributes of the created Artifact might contain
+  * an address that's universal per Artifact service (aka Flyte installaion).
+  * a name that was autogenerated using the item stored.
+  * aliases that make sense that are autogenerated
+  * it will be associated with a Literal and a Literal type and also store any schema or metadata information.
+  * bookkeeping information, created_at, user who created it, etc.
+
+#### Queryable and bindable queries
+Use case: Identify the artifact with such and such name and the "latest" alias. I have a workflow whose default input I want to be whatever is returned by that search when the workflow is kicked off.
```
fg91, line 102: :+1:

rfc/system/3844-artifact-service.md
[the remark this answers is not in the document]
fg91: Update: I like `overwrite_alias=True` which makes this customisable for each k-v pair.

rfc/system/3844-artifact-service.md
```diff
@@ -0,0 +1,148 @@
...
+* Given a set of aliases or tags, return Artifact objects.
+
+On the authoring side, we would like users to be able to do the following
+```python
+@task
+def t1() -> Annotated[nn.Module, Artifact(name="models.nn.lidar", alias="latest", overwrite_alias=True)]: ...
+
+# Possibly in a completely different repo
+@workflow
+def wf(model: nn.Module = Artifact.get_query(name="models.nn.lidar", alias="latest")): ...
```
fg91, line 116: Would you say I need to specify the project/domain here as part of the name or are these shared across projects/domains? In production I probably don't accidentally want to end up with the latest development version.

rfc/system/3844-artifact-service.md
```diff
@@ -0,0 +1,148 @@
...
+```
+The binding is made at compilation time but it is not resolved until run time. That is, when `wf` is run, Admin will have to call out to the Artifact service, get back precisely one Artifact, and use the Literal therein contained as the input. If the type is not a match, then it can result it in a user run time error, which is okay.
```
fg91, line 118: 👍

rfc/system/3844-artifact-service.md
```diff
@@ -0,0 +1,148 @@
...
+
+## 3 Proposed Implementation
```
fg91, line 120: Use case:
As a user I would like to visualize the lineage graph of all these artifacts, links being created by a task that has an artifact as input and producing another artifact.

I don't think Flyte would have to implement this visualization itself but there could be integrations with e.g. https://openlineage.io later on.

rfc/system/3844-artifact-service.md
```diff
@@ -0,0 +1,148 @@
...
+
+## 4 Metrics & Dashboards
+Will address last.
+
+
+## 5 Drawbacks
+None.
+
+## 6 Alternatives
+One alternative not to the idea presented here but to the implementation is to use the existing data catalog service instead, either in whole or in part. Everything from using the grpc service itself with well-crafted calls, to implementing the new service but with the data catalog code.
```
fg91, line 138: Cannot judge whether a good project exists that could be used.

rfc/system/3844-artifact-service.md
```diff
@@ -0,0 +1,148 @@
...
+
+## 7 Potential Impact and Dependencies
+Everything will be implemented before merging anything major or anything that can't be pulled back out.
+
+## 8 Unresolved questions
+If you look in the IDL change, you will see that the Artifact message comes with a `source` field. This is nice but doesn't really make sense if you think about it. There's a better name for the act of identifying where an Artifact came from, and also for identifying things in the search capability like what other executions it was used in. That term is lineage. We're not sure if lineage should be a different service or is an integral part of this Artifact service yet. I can see arguments both ways. It is clear that they are closely intertwined though.
```
fg91, line 144: My personal opinion is that this should be a lineage service and not a separate one.

fg91 (commented):
I really like this effort, thanks a lot.
Left questions and thoughts, happy to discuss.

rfc/system/3844-artifact-service.md
```diff
@@ -0,0 +1,148 @@
...
+Since the Artifact service can take a link and return a Literal, this part is pretty simple since remote.execute already can take Literals. Will just have to fetch and add it to the LiteralMap.  When remote runs, it’ll have to fetch the Literal.
+
+Note: with this construct, two different versions/aliases/etc of the same dataset/artifact can be used in one workflow - think about a workflow that compares the performance of two models.
+
+#### Specify metadata at Artifact creation time
+Use case: When this task or workflow runs, I want to tag it with certain tags and give it a version alias.
+
+```python
+@task
+def t1() -> Annotated[nn.Module, Artifact(name="my.artifact.name", alias={"version": "latest"}, type=Model)]
```
dylanwilder, line 76: how will `alias` be populated at runtime? will it always be `"latest"` or can it be overridden for a specific execution?  How will this work for cached workflows. 

In an ideal world, we'd want to fetch artifacts by some name + execution id.

PR #3849, opened 2023-07-07
Run specific set of tests in functional tests suite

eapolinario:
## Describe your changes
We pluck a specific set of tests and register those as opposed to registering all flytesnacks examples. 

This PR uses the default flytekit image as base image and fast-registers the flyte entities mentioned in the P0 functional tests.


.github/workflows/single-binary.yml
```diff
@@ -175,19 +175,38 @@ jobs:
...
+                  control_flow/control_flow/subworkflows.py \
+                  type_system/type_system/custom_objects.py \
+                  type_system/type_system/schema.py \
+                  type_system/type_system/typed_schema.py ;
+          do
+              pyflyte --config ./boilerplate/flyte/end2end/functional-test-config.yaml \
+                  register \
+                  --project flytesnacks \
+                  --domain development \
+                  --image cr.flyte.org/flyteorg/flytekit:py3.9-latest \
```
jeevb, line 206: Use latest version of `flytekit` image instead?

.github/workflows/single-binary.yml
```diff
@@ -175,19 +175,38 @@ jobs:
...
       - name: Install Python dependencies
         run: |
           python -m pip install --upgrade pip
           pip install flytekit
           pip freeze
+      - name: Checkout flytesnacks
+        uses: actions/checkout@v3
+        with:
+          repository: flyteorg/flytesnacks
+          path: flytesnacks
```
jeevb, line 187: Should we `FLYTESNACKS_VERSION` in this checkout step? We can update the variable with the latest commit sha if the original value was a branch like `master`.

eapolinario: yup, left a comment in the next update explaining this.

.github/workflows/single-binary.yml
eapolinario: which image?

eapolinario: sync'd offline. Meant latest python version image.

PR #3363, opened 2023-02-22
Updates to auth docs

davidmirror-ops:
## Tracking issue

This PR completes the work initiated on #3062 

## Check all the applicable boxes 

- [X] I updated the documentation accordingly.
- [X] All new and existing tests passed.
- [X] All commits are signed-off.

## Note to reviewers

Content as per the lastest review in #3062.

PR #3858, opened 2023-07-10
[Docs] flyte-binary readme and some discussion around secrets

wild-endeavor:
Docs only change.  Adding a readme to the `flyte-binary` Helm chart and pulling in some comments from https://github.com/flyteorg/flyte/pull/3363/files


jeevb:
@davidmirror-ops: cc

eapolinario (approved):

davidmirror-ops:
I see that #3837 contains most what was added here to auth_setup plus a bit more details. How can we proceed?

PR #3860, opened 2023-07-11
Bundle a buildkit instance within sandbox to facilitate local image spec builds

jeevb:
Related: https://github.com/flyteorg/flytectl/pull/413

Tested by running the following command:
```
envd context create --name flyte-sandbox --builder tcp --builder-address localhost:30003 --use
```

and executing (`pyflyte run --remote wf.py my_wf`) the following workflow:
```

image = ImageSpec(
    name="imagespec",
    registry="localhost:30000",
    packages=["smart-open[gcs]"],
    apt_packages=["git"],
)


@task(container_image=image)
def my_task() -> str:
    return "hello, world!"


@workflow
def my_wf() -> str:
    return my_task()
```


jeevb:
@eapolinario @cosmicBboy: For discussion.

docker/sandbox-bundled/Makefile
```diff
@@ -56,6 +56,7 @@ start:
 			--publish "30000:30000" \
```
wild-endeavor, line 56: can we add comments here saying what these ports are for?

wild-endeavor (commented):
can you explain what's happening?  what's the interaction between envd and buildkit?  I don't remember having to run a buildkit container when running envd outside of the sandbox.  why do i need it now when running something inside?

jeevb:
> can you explain what's happening? what's the interaction between envd and buildkit? I don't remember having to run a buildkit container when running envd outside of the sandbox. why do i need it now when running something inside?

`envd` uses `buildkit` under the hood to build images. If you run `envd context ls`, you can see that it defaults to using a local docker container for running builds. This container is created lazily when processing builds. The issue with using a separate container is that docker networking is not set up to allow this container to talk to the registry at `localhost:30000`. So it basically can't read the registry to check for the presence of an existing image. Running buildkit inside the sandbox allows us to fully manage the networking so that the experience with using `ImageSpec` locally is seamless.

wild-endeavor (approved):

pingsutw (approved):

PR #3881, opened 2023-07-24
[Monorepo] - Step 1 - Import golang repos

eapolinario:
## Describe your changes
This PR is the first step in the consolidation of the Flyte components into a monorepo. 

In this PR we only import the code for the golang components via [git subtrees](https://manpages.debian.org/testing/git-man/git-subtree.1.en.html). By using git subtrees, as opposed to simply cherry-picking commits, we allow for [code spelunking](https://queue.acm.org/detail.cfm?id=945136) while maintaining the tooling simple (i.e. no need for git submodules).


eapolinario:
Redone in https://github.com/flyteorg/flyte/pull/4012.

PR #3912, opened 2023-08-01
Add support ingress ingressClassName and tls configuration in flyte-b…

jeevb:
…inary chart


docker/sandbox-bundled/manifests/complete.yaml
```diff
@@ -456,7 +456,7 @@ data:
         stackdriver-enabled: false
       k8s:
         co-pilot:
-          image: "cr.flyte.org/flyteorg/flytecopilot:v0.0.32"
+          image: "cr.flyte.org/flyteorg/flytecopilot:v0.0.30"
```
jeevb, line 459: @eapolinario: Looks like this happened because the copilot image update was not persisted in `charts/flyte-binary/values.yaml` during the last release.

eapolinario (approved):

PR #3930, opened 2023-08-08
Add support for specifying a tls configuration block in both the http…

jeevb:
… and grpc ingress resources


eapolinario (approved):
Amazing. Thank you.

PR #3935, opened 2023-08-09
Flyte Agent Helm Chart

pingsutw:
## Describe your changes
Add a flyte agent helm chart, so that people can easily to install flyte agent independently.
```bash
git clone git@github.com:flyteorg/flyte
git checkout agent-chart
helm install flyte-agent
helm install flyte-agent-dev . --set nameOverride=flyte-agent-dev  # Install another agent
```

- Add a new helm chart for flyte agent
- `flyte-core` / `flyte` / `flyte-binary` import `flyte-agent`

## Screenshots
image


charts/flyte-agent/templates/agent/deployment.yaml
```diff
@@ -1,53 +1,55 @@
-{{- if .Values.flyteagent.enabled }}
+{{- if .Values.enabled }}
```
jeevb, line 1: Is this still necessary? If the chart is being installed, the intent is clear.

charts/flyte-agent/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| additionalContainers | list | `[]` | Appends additional containers to the deployment spec. May include template values. |
+| additionalVolumeMounts | list | `[]` | Appends additional volume mounts to the main container's spec. May include template values. |
+| additionalVolumes | list | `[]` | Appends additional volumes to the deployment spec. May include template values. |
+| affinity | object | `{}` | affinity for flyteagent deployment |
+| agentSecret.name | string | `""` | Specify name of K8s Secret. Leave it empty if you don't need this Secret |
+| agentSecret.secretData | object | `{"data":{"username":"User"}}` | Specify your Secret (with sensitive data) or pseudo-manifest (without sensitive data). See https://github.com/godaddy/kubernetes-external-secrets |
```
jeevb, line 16: This is the updated link: https://github.com/external-secrets/external-secrets

charts/flyte-core/Chart.yaml
```diff
@@ -3,3 +3,8 @@ name: flyte-core
 description: A Helm chart for Flyte core
 type: application
 version: v0.1.10 # VERSION
+dependencies:
```
jeevb, line 6: We can specify the conditional here as opposed to everywhere in the child chart:
https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags

jeevb: Also note how `flyte-binary` uses the `flyteagent` service: https://github.com/flyteorg/flyte/blob/809c6572cc2a62f5b16c9e52ec70b827120c66be/charts/flyte-binary/templates/configmap.yaml#L69

charts/flyte-agent/values.yaml
```diff
@@ -0,0 +1,77 @@
+# ---------------------------------------------------------------------
```
jeevb, line 1: Will there be value in consolidating the format of our `values.yaml` files so that users can start getting familiar. Right now, all charts are very different from each other.

charts/flyte-agent/templates/agent/deployment.yaml
pingsutw: Still necessary. For example, flyte agent is the `flyte-binary` dependency, but people can still disable it if they don't want to install the agent in their cluster. We disable the agent in the demo cluster in the CI.

jeevb: Would using a condition in Chart.yaml solve this problem?

pingsutw: Just realized we can use condition in helm, will update it.

charts/flyte-agent/README.md
pingsutw: sorry, forgot to remove it

charts/flyte-core/Chart.yaml
pingsutw: good catch, updated it, thanks

jeevb (approved):

PR #3964, opened 2023-08-15
Fix: Set flyteadmin gRPC port to 80 in ingress if using TLS between load balancer and backend

fg91:
## Describe your changes

Flyte can be configured to use TLS between the ingress (or rather load balancer) and the flyteadmin backend.

*This is required for instance in GKE clusters when using the GCE ingress controller (instead of nginx) since gRPC requires http2 and http2 between GCE load balancers and backends in GKE clusters requires TLS ([source](https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-http2#:~:text=Note%3A%20To%20ensure%20the%20load%20balancer%20can%20make%20a%20correct%20HTTP2%20request%20to%20your%20backend%2C%20your%20backend%20must%20be%20configured%20with%20SSL.)):*

> Note: To ensure the load balancer can make a correct HTTP2 request to your backend, your backend must be configured with SSL.

Whether flyteadmin uses TLS (and with which certificate) is controlled in the helm values file via:

```yaml
configmap:
  adminServer:
    security:
      secure: true
      ssl:
        certificateFile: "/etc/tls/cert.pem"
        keyFile: "/etc/tls/key.pem"
```

(The certificate and key needs to be mounted into flyteadmin e.g. via a secret. Can be self-signed.)

However, in case TLS is enabled, flyteadmin **doesn't** seve the http server on port 80 and the gRPC server on port 81 but actually a single http(s) server that wraps both of them on port 80!

---

Details:

In flyteadmin's main `serve` entrypoint there is a [decision gate](https://github.com/flyteorg/flyteadmin/blob/d492640dbd9e72b9a50259799aba76ce85b16313/pkg/server/service.go#L55) whether `serverConfig.Security.Secure`.

* If no, we go into [`serveGatewayInsecure`](https://github.com/flyteorg/flyteadmin/blob/d492640dbd9e72b9a50259799aba76ce85b16313/pkg/server/service.go#L276)
    * In this case, the gRPC server is started in a go-routine [here](https://github.com/flyteorg/flyteadmin/blob/d492640dbd9e72b9a50259799aba76ce85b16313/pkg/server/service.go#L330) while the separate http server is started afterwards [here](https://github.com/flyteorg/flyteadmin/blob/d492640dbd9e72b9a50259799aba76ce85b16313/pkg/server/service.go#L367). 
* If, however, `serverConfig.Security.Secure` is true, we instead go into [`serveGatewaySecure`](https://github.com/flyteorg/flyteadmin/blob/d492640dbd9e72b9a50259799aba76ce85b16313/pkg/server/service.go#L389)
    * Instead of separate http and gRPC servers, a [single one wrapping both](https://github.com/flyteorg/flyteadmin/blob/d492640dbd9e72b9a50259799aba76ce85b16313/pkg/server/service.go#L459) is started [here](https://github.com/flyteorg/flyteadmin/blob/d492640dbd9e72b9a50259799aba76ce85b16313/pkg/server/service.go#L468) with TLS. This server routes to the actual http or gRPC server [here](https://github.com/flyteorg/flyteadmin/blob/d492640dbd9e72b9a50259799aba76ce85b16313/pkg/server/service.go#L378). 

In the second case, all requests for flyteadmin, including gRPC, have to be sent to port 80.

---

This is not accounted for in the ingress template of the helm chart. Activating TLS for flyteadmin, thus, breaks the deployment.

This PR fixes this.

### DIFF
--- a/charts/flyte-core/templates/common/ingress.yaml
+++ b/charts/flyte-core/templates/common/ingress.yaml
@@ -1,4 +1,8 @@
 {{- define "grpcRoutes" -}}
+{{- $grpcPort := 81 -}}
+{{- if eq .Values.configmap.adminServer.server.security.secure true -}}
+  {{- $grpcPort = 80 -}}
+{{- end }}
 # NOTE: Port 81 in flyteadmin is the GRPC server port for FlyteAdmin.
 - path: /flyteidl.service.SignalService
   pathType: ImplementationSpecific
@@ -7,10 +11,10 @@
     service:
       name: flyteadmin
       port:
-        number: 81
+        number: {{ $grpcPort }}
 {{- else }}
     serviceName: flyteadmin
-    servicePort: 81
+    servicePort: {{ $grpcPort }}
 {{- end }}
 - path: /flyteidl.service.SignalService/*
   pathType: ImplementationSpecific
@@ -19,10 +23,10 @@
     service:
       name: flyteadmin
       port:
-        number: 81
+        number: {{ $grpcPort }}
 {{- else }}
     serviceName: flyteadmin
-    servicePort: 81
+    servicePort: {{ $grpcPort }}
 {{- end }}
 - path: /flyteidl.service.AdminService
   pathType: ImplementationSpecific
@@ -31,10 +35,10 @@
     service:
       name: flyteadmin
       port:
-        number: 81
+        number: {{ $grpcPort }}
 {{- else }}
     serviceName: flyteadmin
-    servicePort: 81
+    servicePort: {{ $grpcPort }}
 {{- end }}
 - path: /flyteidl.service.AdminService/*
   pathType: ImplementationSpecific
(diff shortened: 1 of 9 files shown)


fg91:
Not needed anymore for https://github.com/flyteorg/flyte/issues/3965 as the recommended deployment with IAP now uses [this](https://cloud.google.com/architecture/exposing-service-mesh-apps-through-gke-ingress#security) reference architecture using an istio ingress behind the GCP load balancer.
Still a bug though.

davidmirror-ops:
@fg91 I haven't reproduce the behavior you describe but the code is clear. I'd say that even if not needed for GKE Ingress controller, it's still a bug. 

Thanks

davidmirror-ops:
@fg91 could you `make helm` and push again please?

fg91:
> @fg91 I haven't reproduce the behavior you describe but the code is clear. I'd say that even if not needed for GKE Ingress controller, it's still a bug.

I wonder whether somebody actually lets flyteadmin handle TLS itself instead of doing this at the ingress. I only tried this because the GCE ingress requires TLS betwenn ingress and backend when http2 is used which in turn is required by grpc.

That being said, when I couldn't reach flyteadmin anymore after having turned on TLS in the flyteadmin config, I would have never guessed that the port changed (not documented) and only realised this happens when stepping through flyteadmin with a debugger. Which was definitely not the first thing I tried 😄 
So in case somebody else does in fact need this, let's finish this PR ..

> @fg91 could you `make helm` and push again please?

Done

davidmirror-ops (approved):
Thank you

PR #3969, opened 2023-08-18
Add LimitRange resource permission to flyteadmin serviceaccount

mucahitkantepe:
We deploy use LimitRanges for flyte created namespaces. It would be practical if syncresources can also create this, otherwise we would need to create the rbac resources ourselves


kumare3:
I think it is fair to create this resource- but should we make the RBAC role configurable?

wild-endeavor (approved):

Issue #2975, opened 2022-10-10
[Core feature] serve k8s controller-runtime manager metrics in flytepropeller


hamersaw:
### Motivation: Why do you think this is important?

Currently FlytePropeller uses a k8s controller-runtime manager to aid in k8s informers, caches, etc. This tool includes a metrics endpoint which serves prometheus metrics. In [this PR](https://github.com/flyteorg/flytepropeller/pull/492) that metrics endpoint was disabled to enable Flyte deployments in environments where host networking is required on webhooks.

### Goal: What should the final outcome look like, ideally?

FlytePropeller already includes a prometheus endpoint to serve Flyte specific prometheus metrics. We can include the registry from the k8s controller-runtime manager in this endpoint and serve them alongside the Flyte metrics.

### Describe alternatives you've considered 

Leaving this as is means that these verbose metrics are unavailable.

###  Propose: Link/Inline OR Additional context 

This is where the k8s controller-runtime manager uses the metrics listener to serve prometheus metrics - https://github.com/kubernetes-sigs/controller-runtime/blob/8ad090e0663e046c5a9dad68a572af0e30626c85/pkg/manager/internal.go#L325-L338

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #3981, opened 2023-08-21
Starting profiler for single binary with k8s metrics

hamersaw:
## Tracking issue
https://github.com/flyteorg/flyte/issues/2975

## Describe your changes
Starting the profiler http endpoint for the single binary and exposing k8s metrics using a separate handler.

## Screenshots
_NA_

## Note to reviewers
_NA_

### DIFF
--- a/cmd/single/start.go
+++ b/cmd/single/start.go
@@ -2,39 +2,39 @@ package single
 
 import (
 	"context"
-
-	"k8s.io/client-go/rest"
-	"sigs.k8s.io/controller-runtime/pkg/cache"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	"github.com/flyteorg/flytepropeller/pkg/controller/executors"
-	"sigs.k8s.io/controller-runtime/pkg/manager"
-
-	"github.com/flyteorg/flyteadmin/pkg/common"
-	"github.com/flyteorg/flyteadmin/plugins"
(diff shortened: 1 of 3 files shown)


eapolinario (dismissed):

eapolinario:
This looks reasonable, but CI check is failing. We're retrying in https://github.com/flyteorg/flyte/actions/runs/5944214374?pr=3981.

cmd/single/start.go
```diff
@@ -147,6 +147,20 @@ func startPropeller(ctx context.Context, cfg Propeller) error {
 	}
 
 	if !cfg.DisableWebhook || !cfg.Disabled {
+		handlers := map[string]http.Handler{
+			"/k8smetrics": promhttp.HandlerFor(metrics.Registry, promhttp.HandlerOpts{
```
kumare3, line 151: why a separate endpoint? we want to do it so that scraping can be specialized?

hamersaw: The `controller-runtime` package automatically populates a prometheus `Registry` as `metrics.Registry`. So this was just the easiest way to surface that registry (ie. separate HTTP endpoint). We could wrap both the `controller-runtime` prometheus registry and our registry (buried in flytestdlib) into a single entity, but it would require a good deal of refactoring. Also it does allowed for specialized scraping too.

eapolinario: Is there any advantage to unifying the registries?

eapolinario (approved):
Won't block the PR on the question I asked.

Issue #3970, opened 2023-08-18
[Housekeeping] Enable `flyte-core` to consume pre-created secrets


davidmirror-ops:
### Describe the issue

In a multi-cluster deployment using `flyte-core`, [one of the steps in the docs](https://docs.flyte.org/en/latest/deployment/deployment/multicluster.html#id1) indicate to manually create a secret containing the tokens and certs for each dataplane cluster and then submit it to the control plane cluster.

This is inconvenient for automated deployments.

Instead, similar to the `flyte-binary` chart, if `flyte-core` is capable of consuming pre-created secrets, automation could override the values and mount the necessary secrets for each dataplane cluster.

### What if we do not do this?

Otherwise, users will still have to either rely on an external tool or create secrets manually.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #3993, opened 2023-08-28
Fix flyte-core helm charts for multi cluster configuration

gdabisias:
While testing the flyte-core deployment, I found that there is an issue with secrets access describe in the [Flyte multicluster guide](https://docs.flyte.org/en/latest/deployment/deployment/multicluster.html#deployment-deployment-multicluster).
Basically we are creating manually a secret called `cluster-secrets`, but this does not get mounted properly by the services that are supposed to read from it, resulting in a PANIC error in `syncresources` service and in the `sync-cluster-resources` init container of the flyte admin service:
```
panic: Failed to get auth token: open /var/run/credentials/cluster_1_token: no such file or directory
failed to read k8s bearer token from configured path
```

I described the pods and notice that the secrets don't get mounted correctly and that we have a special case for when we actually add a multi cluster configuration
```
{{- if gt (len .Values.configmap.clusters.labelClusterMap) 0 }}
...
```
but that section is loading the `flyte-admin-secrets` secret instead of the new `cluster-secret`.

This is a fix to have things working for how they are set up right now, but longer term there should not be this custom secret and we should have a pre-created secret in the charts that gets filled in by users from the `values-eks.yaml` file or something similar (e.g. reuse `flyte-admin-secrets`)

## Tracking issue

This issue is linked to https://github.com/flyteorg/flyte/issues/3970 but is not fixing that

- [x] I updated the documentation accordingly. -> No documentation to update
- [x] All new and existing tests passed -> Tested this in a real world multicluster setup


gdabisias:
I see `FAILED: helm code updated without commiting generated code.` for the generate_helm code.
What is needed here?

charts/flyte-core/templates/clusterresourcesync/deployment.yaml
```diff
@@ -55,9 +55,9 @@ spec:
             name: flyte-clusterresourcesync-config
           name: config-volume
         {{- if gt (len .Values.configmap.clusters.labelClusterMap) 0 }}
-        - name: flyte-admin-secrets
+        - name: cluster-secrets
           secret:
-            secretName: flyte-admin-secrets
+            secretName: cluster-secrets
```
wild-endeavor, line 60: where is this secret created?  is the name of the secret "cluster-secrets"?  would you mind linking me to where that's set?

wild-endeavor (commented):
thank you for the help!

charts/flyte-core/templates/clusterresourcesync/deployment.yaml
gdabisias: Sure, it comes here from the guide
https://docs.flyte.org/en/latest/deployment/deployment/multicluster.html#deployment-deployment-multicluster

gdabisias: Ideally that should not be created manually but as part of the helm chart and only filled in with values by users, but for now this fixes the current setup

charts/flyte-core/templates/admin/deployment.yaml
```diff
@@ -72,6 +72,10 @@ spec:
             name: clusters-config-volume
           - mountPath: /etc/secrets/
             name: admin-secrets
+          {{- if gt (len .Values.configmap.clusters.labelClusterMap) 0 }}
+          - mountPath: /var/run/credentials
+            name: cluster-credentials
```
wild-endeavor, line 77: but this is a volumeMount right?  doesn't that mean there needs to be a volume declared with the same name in the Pod?

gdabisias: that will mount the secret to that path, no?

gdabisias: Also no, since this is an init container and the main container is already declaring the volume for the secret

wild-endeavor: oh i see it's in additionalVolumes

wild-endeavor: but wait shouldn't that come through in the additional mounts?
```
  additionalVolumeMounts:
  - name: cluster-credentials
    mountPath: /var/run/credentials
```
which gets injected here: https://github.com/flyteorg/flyte/blob/a071bade39bdad80fff042c235a1c3c046a82a09/charts/flyte-core/templates/admin/deployment.yaml#L137C37-L137C59

charts/flyte-core/templates/clusterresourcesync/deployment.yaml
wild-endeavor: i'm still confused by this.  i don't see the text "cluster-secrets" on that page.  I only see the `cluster-credentials` which should be covered by the additional values.

charts/flyte-core/templates/admin/deployment.yaml
davidmirror-ops: @gdabisias I just found that if we change the process to, instead of creating a new secret,  editing the existing `flyte-admin-secrets` and adding the data plane cluster token and cert there, the `syncresources` Pod works just fine. 
Right, no surprises there.

charts/flyte-core/templates/clusterresourcesync/deployment.yaml
davidmirror-ops: I think `cluster-credentials` is what you mean here @gdabisias right?

davidmirror-ops: Besides naming, I think this change makes sense. It enables `syncresources` to mount a pre-created Secret containing only the credentials for the multiple clusters the control plane will connect to, instead of having to manually edit the `flyte-admin-secrets` which mounts additional sensitive data that doesn't seem to be needed by `syncresources`

gdabisias: yeah this should be cluster-credentials; let me update that.

I think this should be fine as a first fix, but then we should have secrets that get created independently of the user code and they just get filled in by the user, like we do for the database secrets

gdabisias: Fixed

charts/flyte-core/templates/admin/deployment.yaml
gdabisias: @davidmirror-ops as you mentioned in the other comment, we should have a separate secret

gdabisias: @wild-endeavor no because here we are updating the syncresouces init container and not the admin one. For the admin one, we add the secret mounting point to the values-eks.yaml

wild-endeavor: should we reference `Values.flyteadmin.additionalVolumeMounts` also instead of directly adding the mountpath?

The issue is that this only works if the user specifies that string in the values file for the additional volumes. if they use a different string than "cluster-credentials" then this won't work right?

gdabisias: yes, it won't work. My idea was to fix things according to the guide, but we can also just add that to the values-eks.yaml.
Up to you, I don't mind. Either way, we should remove this secret creation and mounting completely and it should be part of the general chart, with the user only adding the secret value itself

gdabisias: @wild-endeavor I checked again and we can't sure that field, since here we are mounting things for the sync-cluster-resources init container, which is not the admin container.
the admin container is already loading the volume in the additional section that users have to add to the deployment to make things work, e.g.
```
flyteadmin: {
			additionalVolumes: [{
				name: "cluster-credentials"
				secret: secretName: "cluster-credentials"
			}]
			additionalVolumeMounts: [{
				name:      "cluster-credentials"
				mountPath: "/var/run/credentials"
			}]
```

but this can't include the values for the init container (if you mount something for the flyteadmin container, it won't be exported automatically to the sync-cluster-resources container

gdabisias: I made the change but notice that we are mounting anything specified in the additional volumes section, so there might be also some other stuff (Still better than what we had before and I don't see why something mounted by the admin container should not be mounted by it's init container)

davidmirror-ops (approved):
I reproduced this and it works

PR #4014, opened 2023-09-08
[Monorepo] - Proclaim monorepo

eapolinario:
## Describe your changes

This PR proclaims the 7 modules introduced by https://github.com/flyteorg/flyte/pull/4012 as the new place for contribution.

The list of changes:
1. Rename the modules
2. Build single-binary and sandbox (both demo and old sandbox) using the local modules
3. Component CI checks moved from each repo
4. Changes to releases

Let's talk about each change separately.

### Rename the modules
The rules that govern the publishing of multi-module repositories are described in [Go Modules Reference - The Go Programming Language](https://go.dev/ref/mod#vcs-version), more specifically:

> If a module is defined in a subdirectory within the repository, that is, the module subdirectory portion of the module path is not empty, then each tag name must be prefixed with the module subdirectory, followed by a slash. For example, the module http://golang.org/x/tools/gopls  is defined in the gopls subdirectory of the repository with root path http://golang.org/x/tools . The version v0.4.0 of that module must have the tag named gopls/v0.4.0 in that repository.

This essentially means that in a subsequent PR, the one where we deal with release, we'll need to generate a few more tags, one per-component.

This PR simply fixes the names of the modules and ensures that `go replace` directives point to each module. I manually tested the compilation of each component. A subsequent PR is going to fix the single-binary and the components Dockerfiles.

### Build single-binary using local modules
Done in this stacked [PR](https://github.com/flyteorg/flyte/pull/4015).

### Component CI checks
Done as part of this stacked [PR](https://github.com/flyteorg/flyte/pull/4027).

### Changes to releases
edit: this is going to come in a separate PR. For now merges to master will *not* trigger image builds nor releases.

We're going to tag merges to master using semver versionsing, so these new tags will update the patch version and use that to create a new release. Exactly how we do currently in each component. These automatic releases are *not* going to create new helm releases.


pingsutw (approved):

pingsutw (dismissed):

pingsutw (commented):
LGTM, but there is a failing test

pingsutw (approved):

PR #4054, opened 2023-09-20
add codespell: config, workflow and have typos fixed

yarikoptic:
## Describe your changes

I was listening a talk today featuring flyte, decided to check it out...

This PR introduces https://github.com/codespell-project/codespell to this repo -- codespell has a list of most common typos, and this way allows for quite robust check/fixing of the codebases so looks like "spellchecking".   Here I provide config to ignore some few ad-hoc variables etc, github action which would fail if new typo is introduced, and fixing a good number of typos. Note that a few typos were fixed in the code but I do not think those fixes are 'functional', but please check.

Also there was a commit (currently e6098758df3e3aab6920b2a146f8a6800f89061e) which fixed up .gitignore so that actually tracked by git folder is not ignored


flyteadmin/scheduler/snapshoter/versioned_snapshot.go
```diff
@@ -12,7 +12,7 @@ import (
 // Including a version provides compatibility check
 type VersionedSnapshot struct {
 	Version int
-	Ser     []byte
+	Set     []byte
```
yarikoptic, line 15: ooops -- needs to be ignore... will be done

pingsutw (approved):
Thank you so much for fixing it

Issue #3936, opened 2023-08-09
[Core feature] Flyte Agent Ecosystem


pingsutw:
### Motivation: Why do you think this is important?

Create this issue to track all the PRs and issues related to the Flyte agent.

### Goal: What should the final outcome look like, ideally?

More agents, more reliable, easy to use and deploy.

### Describe alternatives you've considered 

NA

###  Propose: Link/Inline OR Additional context 

Todo:
- [x] [Databricks Agent](https://github.com/flyteorg/flytekit/pull/1797)
- [x] [Snowflake Agent](https://github.com/flyteorg/flytekit/pull/1799)
- [x] [Memverge Agent](https://github.com/flyteorg/flytekit/pull/1821)
- [x] [File sensor](https://github.com/flyteorg/flytekit/pull/1790)
- [ ] Sagemaker Agent
- [x] ChatGPT Agent
- [ ] LainChain Agent (TBD)

PRs:
- https://github.com/flyteorg/flyte/pull/3935
- https://github.com/flyteorg/flytekit/pull/1790
- https://github.com/flyteorg/flyteplugins/pull/384
- https://github.com/flyteorg/flytekit/pull/1781
- https://github.com/flyteorg/flytekit/pull/1778
- https://github.com/flyteorg/flytekit/pull/1782
- https://github.com/flyteorg/flytekit/pull/1773
- https://github.com/flyteorg/flytekit/pull/1725
- https://github.com/flyteorg/flytekit/pull/1712
- https://github.com/flyteorg/flyte/pull/3558
- https://github.com/flyteorg/flyteplugins/pull/381
- https://github.com/flyteorg/flyteplugins/pull/368
- https://github.com/flyteorg/flyteplugins/pull/353
- https://github.com/flyteorg/flyteplugins/pull/330
- https://github.com/flyteorg/flytepropeller/pull/564
- https://github.com/flyteorg/flyte/pull/3788
- https://github.com/flyteorg/flyte/pull/3454
- https://github.com/flyteorg/flyte/pull/3779
- https://github.com/flyteorg/flytesnacks/pull/971
- https://github.com/flyteorg/flytekit/pull/1815
- https://github.com/flyteorg/flyteplugins/pull/392
- https://github.com/flyteorg/flyteplugins/pull/398
- https://github.com/flyteorg/flyteplugins/pull/391
- https://github.com/flyteorg/flyteplugins/pull/402
- https://github.com/flyteorg/flytestdlib/pull/164
[...]

PR #4107, opened 2023-09-30
Agent Sync Plugin

Future-Outlier:
## Tracking issue
https://github.com/flyteorg/flytekit/pull/1822

## Describe your changes
- flyteidl
1. add a `DoTask` function in `agent.proto`
2. change the purpose of `runtime metadata's flavor` in `tasks.proto`

https://github.com/flyteorg/flyte/pull/4107/files#diff-36edb1a8db489fabd420007c4542f657ed7e6d4f9d11052f81536c35f422ac14R91-R104
https://github.com/flyteorg/flyte/pull/4107/files#diff-d9c6122ef023db0768b904e6d6307684dfc35b4d03f0b2152ff7791344178a14R85

- Add `sync plugin` `routing mechanism` to support sync plugins
1. by adding the `type Plugin interface` here
https://github.com/flyteorg/flyte/pull/4107/files#diff-3119631e7cd5a6d4fb999fc50bbb8509bb685347c4a6e3cceda14acec6aa0eb3R100-R103

- Add a helper function to get `sync/async plugin` and handle the error
https://github.com/flyteorg/flyte/pull/4107/files#diff-da53a6f262afa680bad44d5bc3e5a6a95f1e8b8764002e0db9a7f015af7d2831R70-R86

- Implement `webapi/agent's sync plugin` method to support `api_task` like ChatGPT
Note: you can find similar logic in other functions in this file.
(Create, Get, Delete)

https://github.com/flyteorg/flyte/pull/4107/files#diff-5e3f52f33e7ff5f718bc03e33dd574152be1a784901745e7c652b37e0bfc46a7R63-R114

- Split `Handle` function into 2 functions, `syncHandle` and `asyncHandle`

- Add test for the above changes

- Finish above without introducing breaking changes

Note: `api_task` is a task type which has only `Do Task Function`, the sync plugin meets it's need.

###  Configuration Example
To set up agent sync plugins, users can now utilize the following configuration:

```yaml
tasks:
  task-plugins:
    enabled-plugins:
      - agent-service
    default-for-task-types:
      api_task: agent-service
plugins:
  agent-service:
    supportedTaskTypes:
      - api_task
```
### Run in Remote Environment
```bash
pyflyte run --remote --image futureoutlier/flytekit:chatgpt-v2 chatgpt_example.py wf
```

###  Routing Mechanism
To determine the route to the sync plugin, we're leveraging the `task's metadata flavor`.
The logic can be found in the PR's changes:

Here's the code
`if taskTemplate.GetMetadata().GetRuntime().GetFlavor() == syncPlugin`
https://github.com/flyteorg/flyte/pull/4107/files#diff-da53a6f262afa680bad44d5bc3e5a6a95f1e8b8764002e0db9a7f015af7d2831R77

## Note to reviewers
Please also take a look at this PR.
https://github.com/flyteorg/flytekit/pull/1822
Both PRs are interconnected and address related functionalities.

## Screen Shot
image
Note: You can find that the ChatGPT task is an `api_task`.

## How it works at high level
image

## Additional Information
The typecasting for the `webapi.Plugin` to `webapi.SyncPlugin` or  `webapi.AsyncPlugin` might be a little bit tricky, `webapi.Plugin`'s interface has a common function with `webapi.SyncPlugin`  and `webapi.AsyncPlugin`.

In the file `flyte/flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go`, since the function `p.GetConfig()` performances the same no matter the plugin is `webapi.AsyncPlugin` or `webapi.SyncPlugin`, and all the parts for `webapi.AsyncPlugin` and `webapi.SyncPlugin` are the same, it is OK to share the same snippets in the function, but don't need to create more `PluginLoaders` interface as the conversation below.

https://github.com/flyteorg/flyte/blob/c5bcd9939b99ba24f4e4d8134c8600a2a0518ab4/flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go#L231-L283


flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
```diff
@@ -68,7 +66,22 @@ func (c CorePlugin) GetProperties() core.PluginProperties {
...
+		return core.UnknownTransition, err
+	}
+
+	return core.DoTransition(core.PhaseInfoSuccess(&core.TaskInfo{})), nil
+}
+
 func (c CorePlugin) Handle(ctx context.Context, tCtx core.TaskExecutionContext) (core.Transition, error) {
+	taskTemplate, err := tCtx.TaskReader().Read(ctx)
+
+	if taskTemplate.Type == "dispatcher" {
```
honnix, line 81: Isn't `taskTemplate.Type` being used for agent routing in in `agent/plugin.go`?

Future-Outlier: Yes, taskTemplate.Type is indeed utilized there. 
However, we also need to ensure the plugin is directed to use the sync plugin interface. 
This is why we have to make the routing here.

I chose not to incorporate a cache mechanism for the sync plugin.
And all cases of do task can go through the `dispatcher` task type.
You can refer this pr: https://github.com/flyteorg/flytekit/pull/1822

honnix: Thanks for replying.

I'm not sure how routing is supposed to be done for a sync agent. If I understood correctly, it would be utilizing `custom` of task template: https://github.com/flyteorg/flytekit/pull/1822/files#diff-cac07a79ac0ced0778cd505ccfb5f8f1b4e33f78e1c40a91bba926171f9d2246R39-R41, is that correct?

I'm also confused by https://github.com/flyteorg/flytekit/pull/1822/files#diff-9f7af27264f8773b069e8200804c224fe19a6fcaaf9dc33edc644f5351cbb3beR137. Does it mean you are assuming there is at most one "dispatcher" (i.e task_template.type == "dispatcher") agent connected to flyte? If so, I think this is a very strong assumption which at least won't work in our setup, because we won't have a centralized agent but multiple ones developed by different teams.

Future-Outlier: 1. Yes you are right!
In the case of the chatgpt task, the task type is designated as `dispatcher`. We then utilize the ChatGPTDispatcher as specified in the configuration.

2. To illustrate a practical scenario:
Consider a situation where there are two dispatchers - the ChatGPTDispatcher and the LangChainDispatcher. 
When either is invoked to call the agent-service, the task type remains `dispatcher`. 
However, the dispatcher engine ensures each is used separately, depending on the specific requirements and configurations.

Future-Outlier: We can have multiple dispatchers at the same time!
Dispatcher Engine can handle multiple kinds of dispatchers.

Future-Outlier: I think your idea is great, too!
I am just a beginner in large-scale backend systems, so I don't have enough experience yet.
@pingsutw  Please take a look, thanks a lot!

Future-Outlier: Here is a screenshot for you to better understand how `Dispatcher Engine` works!
image

honnix: I think we are not referring to the same thing. In our case, we will have multiple gRPC services, each implementing the handler, and that would mean they cannot claim to support "dispatcher" type at the same time. We don't use the implementation in flytekit, and I think it should not be an assumption that everyone would be using flytekit implementation.

honnix: More clarification: I meant we have multiple gRPC endpoints, with everything being "dispatcher", agent plugin will not be able to know which endpoint to route to.

Future-Outlier: @pingsutw and I is currently trying to fix this!
Maybe we can have a discussion on slack.
And yes, you are right, we will improve the implementation of it.
We will definitely not use only one "dispatcher" typed agent endpoint in the system.

honnix: That is great! Sure please feel free to pull me to a Slack chat and I will follow up tomorrow. Sorry I am not able to discuss today.

Future-Outlier: No problem, I will book a time with you, thanks for your attention!

Future-Outlier:
@honnix if possible, please take a look at this PR and the relevant flytekit PR.
Thanks a lot!
https://github.com/flyteorg/flytekit/pull/1822
Feel free to mention us to improve the PR, thanks really much!

flyteplugins/go/tasks/pluginmachinery/webapi/plugin.go
```diff
@@ -20,7 +20,7 @@ import (
 
 // A Lazy loading function, that will load the plugin. Plugins should be initialized in this method. It is guaranteed
 // that the plugin loader will be called before any Handle/Abort/Finalize functions are invoked
-type PluginLoader func(ctx context.Context, iCtx PluginSetupContext) (AsyncPlugin, error)
+type PluginLoader func(ctx context.Context, iCtx PluginSetupContext) (Plugin, error)
```
honnix, line 23: Is this a breaking change?

Future-Outlier: Hi, thanks for the review.
with this change, users who previously used AsyncPlugin will now need to modify their code in the same way as the BigQuery plugin does.
BigQuery example: https://github.com/flyteorg/flyte/pull/4107/files#diff-6a9d13d1a2c7f030da6eaa91b837eee26b780d7c26e1084d1a76f2349977ee2dR562

I get it might feel a bit forceful making users adjust their private plugins. But I genuinely think this route is smoother. Going the other way, like suggesting a change where existing plugins (take BigQuery, for example) need deeper tweaks, doesn't seem as clean. Imagine this:
`type PluginLoader func(ctx context.Context, iCtx PluginSetupContext) (AsyncPlugin, SyncPlugin, error)`
https://github.com/flyteorg/flyte/pull/4107/commits/adc1c2c9d0d496a872b1164ac6b51e1f015bab57#diff-3119631e7cd5a6d4fb999fc50bbb8509bb685347c4a6e3cceda14acec6aa0eb3R25

In comparison, the current plugin interface is far more easier for users to tweak their code.

honnix: I think in general we should be very cautious with breaking changes given the number of flyte deployments and possible plugins in the wild. Unless it is absolutely necessary, we may not want to do that. This case to me doesn't seem to be so convincing.

honnix: Have you considered this approach?

```go
// Deprecated, please use AsyncPluginLoader instead for new plugin development
type PluginLoader func(ctx context.Context, iCtx PluginSetupContext) (AsyncPlugin, error)

type AsyncPluginLoader func(ctx context.Context, iCtx PluginSetupContext) (AsyncPlugin, error)
type SyncPluginLoader func(ctx context.Context, iCtx PluginSetupContext) (SyncPlugin, error)
```

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
```diff
@@ -66,24 +67,39 @@ func (c CorePlugin) GetProperties() core.PluginProperties {
 }
 
 func (c CorePlugin) Handle(ctx context.Context, tCtx core.TaskExecutionContext) (core.Transition, error) {
```
honnix, line 69: Maybe we can break this function into two, one handling sync and one async, for better readability?

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
```diff
@@ -66,24 +67,39 @@ func (c CorePlugin) GetProperties() core.PluginProperties {
...
+	taskTemplate, err := tCtx.TaskReader().Read(ctx)
+	if err != nil {
+		return core.UnknownTransition, err
+	}
+
+	// Use the sync plugin to execute the task if the task template has the sync plugin flavor.
+	if taskTemplate.GetMetadata().GetRuntime().GetFlavor() == syncPlugin {
+		phaseInfo, err := c.p.(webapi.SyncPlugin).Do(ctx, tCtx)
```
honnix, line 77: What would happen if the contract is broken? For example the tasks does not correctly set the flavour that is expected by the plugin?

Future-Outlier: The task will be executed by AsyncPlugin.
It is implemented by flytekit this PR.
We will set the variable `use_sync_plugin` to True.
https://github.com/flyteorg/flytekit/pull/1822/files#diff-63787311eeb5747eb3170136f7c05abe69fe01da23d94222a39baf798273e687R20-R52
And then eventually set the flavor variable to here.
https://github.com/flyteorg/flytekit/pull/1822/files#diff-d123cf5b0acf27c386c1ceb74fd4b0de0775378fe1864c76908b5a334f58fffaR173-R440

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
Future-Outlier: This is a good advice.
@pingsutw How do you think?
The implementation is comfortable for me.
Feel free to tell me the final decision, currently I think `break this function into two`  is more readable.

flyteplugins/go/tasks/pluginmachinery/webapi/plugin.go
Future-Outlier: good advice, looking

Future-Outlier: Your advice I think is doable too.
But users still have to modify their private plugins, because they need to modify the section like this.
```
func newSnowflakeJobTaskPlugin() webapi.PluginEntry {
  return webapi.PluginEntry{
    ID:                 "snowflake",
    SupportedTaskTypes: []core.TaskType{"snowflake"},
    AsyncPluginLoader: func(ctx context.Context, iCtx webapi.PluginSetupContext) (webapi.AsyncPlugin, error) {
      return Plugin{
        metricScope: iCtx.MetricsScope(),
        cfg:         GetConfig(),
        client:      &http.Client{},
      }, nil
    },
    SyncPluginLoader: func(ctx context.Context, iCtx webapi.PluginSetupContext) (webapi.SyncPlugin, error) {
      return Plugin{
        metricScope: iCtx.MetricsScope(),
        cfg:         GetConfig(),
        client:      &http.Client{},
      }, nil
    },
  }
}
```
source code: https://github.com/flyteorg/flyte/blob/master/flyteplugins/go/tasks/plugins/webapi/snowflake/plugin.go#L272-L285

There are still some implementation overhead

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
Future-Outlier: I think the agent plugin task will return an error.
As we know, flytepropeller will call the function `invokePlugin`, and if the plugin is agent service, it will send a grpc request to `agent server`, start by `pyflyte serve`.
I think we can write 2 things to help users understand how to use the sync and async plugins easily.

1. add `logger.errorf ` when the plugin doesn't execute successfully, tell users to check whether they use the correct ones or not.
2. write more annotation for the function and the `use_sync_plugin` variable in both flytekit and flyteplugins.

Do you think this is a good solution?
Or there's any other ideas?
I would like to improve it, thanks a lot for your time.

flyteplugins/go/tasks/pluginmachinery/webapi/plugin.go
Future-Outlier: Let me clarify, do you mean you want to change the interface like this?
`flyteplugins/go/tasks/pluginmachinery/core/plugin.go`
source code: https://github.com/flyteorg/flyte/blob/master/flyteplugins/go/tasks/pluginmachinery/core/plugin.go#L17-L31

Before
```go
type PluginLoader func(ctx context.Context, iCtx PluginSetupContext) (AsyncPlugin, error)

type PluginEntry struct {
	// System wide unique identifier for the plugin
	ID TaskType
	// A list of all the task types for which this plugin is applicable.
	RegisteredTaskTypes []TaskType
	// A Lazy loading function, that will load the plugin. Plugins should be initialized in this method. It is guaranteed
	// that the plugin loader will be called before any Handle/Abort/Finalize functions are invoked
	LoadPlugin PluginLoader
	// Boolean that indicates if this plugin can be used as the default for unknown task types. There can only be
	// one default in the system
	IsDefault bool
}
```

After
```go
type AsyncPluginLoader func(ctx context.Context, iCtx PluginSetupContext) (AsyncPlugin, error)
type SyncPluginLoader func(ctx context.Context, iCtx PluginSetupContext) (SyncPlugin, error)

type PluginEntry struct {
	// System wide unique identifier for the plugin
	ID TaskType
	// A list of all the task types for which this plugin is applicable.
	RegisteredTaskTypes []TaskType
	// A Lazy loading function, that will load the plugin. Plugins should be initialized in this method. It is guaranteed
	// that the plugin loader will be called before any Handle/Abort/Finalize functions are invoked
	LoadPlugin AsyncPluginLoader

	SyncLoadPlugin SyncPluginLoader
	// Boolean that indicates if this plugin can be used as the default for unknown task types. There can only be
	// one default in the system
	IsDefault bool
}
```

Based on that assumption, users who write plugins will need to modify their code too, because the interface in `core/plugin.go` has changed.
Example code in Flyteplugins.
https://github.com/flyteorg/flyte/blob/master/flyteplugins/go/tasks/plugins/hive/executor.go#L158-L166

Am I right?
Or there's a better alternative that I didn't notice?
Thanks a lot!

honnix: I have not got time to put much thought on it, but on top of my head, we could have a "one-of" semantics for  `<Async>PluginLoader` and `SyncPluginLoader`, so there is no compilation-time enforcement but runtime when loading the plugin. In this case, existing plugins will not need to be modified, hence we are introducing breaking changes.

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
honnix: Sure we should provide good ways for users to figured out the error.

Maybe a more concrete question, what happens if the cast `c.p.(webapi.SyncPlugin)` fails here?

flyteplugins/go/tasks/pluginmachinery/webapi/plugin.go
honnix: Please do not take my words as the proposal, I was just trying to explore how to ship this feature without introducing breaking changes, because I believe we should be able to achieve that.

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
Future-Outlier: I didn't try it before, but I totally agree that we should add a cast to check if it is correct!
Maybe something like this. (please ignore the implementation error here)
```
plugin, err := c.p.(webapi.SyncPlugin)
if err != nil {
                       logger.Errorf("please check if the sync plugin interface is implemented or not")
			return core.UnknownTransition, err
		}
```

flyteplugins/go/tasks/pluginmachinery/webapi/plugin.go
Future-Outlier: I think your thought is pretty awesome, I didn't think of this solution before.
I will discuss this with @pingsutw and tell you our thoughts.
If the implementation doesn't work, we will tell you why, and we will update the latest discussion with you tomorrow, thanks a lot for the review!

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
honnix: Yeah, this is the direction to go, and there might be other things to consider as well. Generally, if there exists a contract we need to be more protective on the backend side to avoid a total failure and have a better isolation.

flyteplugins/go/tasks/pluginmachinery/webapi/plugin.go
Future-Outlier: Currently, we face a problem that one of flytekit sync plugin, chatgpt, will need more than 10 seconds for the reply if we use openai API and gpt-4 as the model.
However, the grpc agent server shouldn't wait that long, so we need to either add a new server to handle a task like this, or we need to set the timeout limit in our request to the openai server.
If you have time, can you help us to think about the solution?
I think you are really worth learning for me, thanks a lot!
If you don't have time, I can understand your difficulty, since you might be busy.

The below are the 2 solutions' example:

(solution 1)
Create: 
- return a task ID to get the result from the API_TASK server, 
- API_TASK server create a thread to execute `OPENAI REST API REQUEST`

GET: 
- Use the task ID to check if the thread finish the execute, and store the output to a place.
- if `finished` 
return `result` 
else 
return `RUNNING`

DELETE: 
- Kill the Thread

image

(solution 2)
```python
timeout = aiohttp.ClientTimeout(total=10)

async with aiohttp.ClientSession(timeout=timeout) as session:
    async with session.post(
        url=openai_url, headers=get_header(openai_organization=self._openai_organization), data=data
    ) as resp:
        if resp.status != 200:
            raise Exception(f"Failed to execute chatgpt job with error: {resp.reason}")
        response = await resp.json()
```

Future-Outlier: We will solve this in the future, reviewers can ignore this now.

Future-Outlier:
@pingsutw, @honnix  This PR is ready for review, and I solved every problem you review in this PR.
Thanks for your time!

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
```diff
@@ -123,7 +168,17 @@ func (c CorePlugin) Finalize(ctx context.Context, tCtx core.TaskExecutionContext
...
-	return c.tokenAllocator.releaseToken(ctx, c.p, tCtx, c.metrics)
+
+	plugin, ok := c.p.(webapi.AsyncPlugin)
+	if !ok {
+		taskTemplate, err := tCtx.TaskReader().Read(ctx)
+		if err != nil {
+			return err
+		}
+		return fmt.Errorf("%s does not implement required plugin interface methods", taskTemplate.GetType())
+	}
```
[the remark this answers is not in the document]
Future-Outlier: No Problem

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core_test.go
```diff
@@ -2,16 +2,56 @@ package webapi
...
+	p.On("Do", ctx, tCtx).Return(core.PhaseInfo{}, nil)
+
+	c := CorePlugin{
+		id: "test",
+		p:  p,
+	}
+
+	transition, err := c.Handle(ctx, tCtx)
+	assert.NoError(t, err)
+	assert.NotNil(t, transition)
```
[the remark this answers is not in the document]
Future-Outlier: Nice advice, thanks a lot

flyteplugins/go/tasks/plugins/webapi/agent/plugin.go
```diff
@@ -60,6 +60,58 @@ func (p Plugin) ResourceRequirements(_ context.Context, _ webapi.TaskExecutionCo
...
+
+	client, err := p.getClient(ctx, agent, p.connectionCache)
+	if err != nil {
+		return core.PhaseInfoUndefined, fmt.Errorf("failed to connect to agent with error: %v", err)
+	}
+
+	finalCtx, cancel := getFinalContext(ctx, "DoTask", agent)
+	defer cancel()
+
+	res, err := client.DoTask(finalCtx, &admin.DoTaskRequest{Inputs: inputs, Template: taskTemplate})
```
honnix, line 101: I think we should send more information, similarly as:
```
res, err := client.CreateTask(finalCtx, &admin.CreateTaskRequest{Inputs: inputs, Template: taskTemplate, OutputPrefix: outputPrefix, TaskExecutionMetadata: &taskExecutionMetadata})
```

Future-Outlier: The reason why I didn't send more information like `CreateTask` is because currently, the ChatGPT plugin doesn't need it.
After a discussion with @pingsutw, we decided to remove the unused arguments.
If necessary, I can add it back.
I will reply in 3 hours!
Thanks a lot

Future-Outlier: @honnix We will support it now.
Will finish it in 2 hours

Future-Outlier: Add it, please review, thanks a lot!

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
```diff
@@ -65,25 +68,91 @@ func (c CorePlugin) GetProperties() core.PluginProperties {
 	return core.PluginProperties{}
 }
 
+func (c CorePlugin) getPlugin(pluginType string) (webapi.AsyncPlugin, webapi.SyncPlugin, error) {
```
honnix, line 71: I would prefer having two functions: toSyncPlugin and toAsyncPlugin. The call site knows which to call already so it seems weird passing a string to this function to control what to cast to. This function has to either return sync or async interface, which is also a bit strange.

Future-Outlier: No problem, I think this is better!
I will do it now

Future-Outlier: Thanks really much

Future-Outlier: Done, I also split the `Handle` function into `syncHandle` and `asyncHandle`, which is more clean and readable.
Thanks for you and Kevin's review!

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
```diff
@@ -65,25 +68,66 @@ func (c CorePlugin) GetProperties() core.PluginProperties {
...
+		return core.UnknownTransition, err
+	}
+
+	phaseInfo, err := plugin.Do(ctx, tCtx)
+	if err != nil {
+		taskTemplate, err := tCtx.TaskReader().Read(ctx)
+		if err != nil {
+			return core.UnknownTransition, err
+		}
+		logger.Errorf(ctx, "please check if [%v] task type has implemented sync plugin method or not", taskTemplate.GetType())
```
pingsutw, line 99: the error should be something like `failed to run do task with err....`

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
```diff
@@ -97,6 +141,36 @@ func (c CorePlugin) Handle(ctx context.Context, tCtx core.TaskExecutionContext)
...
+
+	return false
+}
+
+func (c CorePlugin) Handle(ctx context.Context, tCtx core.TaskExecutionContext) (core.Transition, error) {
+	taskTemplate, err := tCtx.TaskReader().Read(ctx)
+	if err != nil {
+		return core.UnknownTransition, err
+	}
+	// Assume the plugin is an async plugin if not explicitly specified as sync.
```
pingsutw, line 164: nit: could we move it to `useSyncPlugin`? line 155

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
```diff
@@ -28,11 +29,13 @@ const (
 	maxBurst           = 10000
 	minQPS             = 1
 	maxQPS             = 100000
+	asyncPlugin        = "async_plugin"
```
honnix, line 32: Are these still being used?

flyteplugins/go/tasks/pluginmachinery/internal/webapi/core.go
```diff
@@ -65,25 +68,66 @@ func (c CorePlugin) GetProperties() core.PluginProperties {
...
+	if !ok {
+		return nil, fmt.Errorf("Core plugin does not implement the async plugin interface")
+	}
+	return plugin, nil
+}
+
+func (c CorePlugin) syncHandle(ctx context.Context, tCtx core.TaskExecutionContext) (core.Transition, error) {
+	plugin, err := c.toSyncPlugin()
+	if err != nil {
+		return core.UnknownTransition, err
```
honnix, line 90: Does it make sense to fail the task/workflow execution immediately? I don't know how `core.UnknownTransition` drives the state machine.

Future-Outlier:
I am going to demo the sync plugin in flyte community sync today, so I will finish the update later, thanks a lot for you both.
:sparkling_heart:

flyteidl/protos/flyteidl/service/agent.proto
```diff
@@ -12,4 +12,6 @@ service AsyncAgentService {
   rpc GetTask (flyteidl.admin.GetTaskRequest) returns (flyteidl.admin.GetTaskResponse){};
   // Delete the task resource.
   rpc DeleteTask (flyteidl.admin.DeleteTaskRequest) returns (flyteidl.admin.DeleteTaskResponse){};
+  // Do a job and get response from the agent server.
+  rpc DoTask (flyteidl.admin.DoTaskRequest) returns (flyteidl.admin.DoTaskResponse){};
```
honnix, line 16: Sorry I did not pay much attention to this change. This service is named as `AsyncAgentService` but this RPC is for sync execution. Do you think it makes sense to have another service or renaming `AsyncAgentService` to `AgentService` (we will again need to think about breaking change if we rename it)?

Future-Outlier:
@honnix 
This is ready for be reviewed and approved.
We use a proto `plugin_metadata` instead, and will handle the breaking changes in flytekit!

Future-Outlier:
Please take a look, thank you really much!

Future-Outlier:
> Nice! Thank you for addressing all the comments and being patient about my nitpicking! Good job.

Thanks a lot!
I learned really much from you and Kevin in this PR.
It is a really important experience for me!

flyteidl/protos/flyteidl/core/tasks.proto
```diff
@@ -84,6 +84,14 @@ message RuntimeMetadata {
 
     //+optional It can be used to provide extra information about the runtime (e.g. python, golang... etc.).
     string flavor = 3;
+
+    //+optional It can be used to provide extra information for the plugin.
+    PluginMetadata plugin_metadata = 4;
+}
+
+message PluginMetadata {
+    //+optional It can be used to decide use sync plugin or async plugin during runtime.
```
pingsutw, line 93: we should move it to agent proto

pingsutw (changes requested):

flyteidl/protos/flyteidl/core/tasks.proto
honnix: I think it makes sense to be here? This is not strict only for agent, and it could be extended in the future to include more fields for other purposes. Also moving this to agent proto will result in circular imports that we will need to resolve somehow.

flyteidl/protos/flyteidl/admin/agent.proto
```diff
@@ -91,3 +91,21 @@ message DeleteTaskRequest {
 // Response to delete a task.
 message DeleteTaskResponse {
 }
+
+// A message used to do a task
+message DoTaskRequest {
```
kumare3, line 96: Lets merge the DoTaskRequest and DoTaskResponse into CreateTask
Make 
```
message CreateTaskResponse{
  oneof res {
   bytes resource_meta;
   Resource resource;
  }
}
```

After this we should modify the entire 
message Resource to have more fields that map TaskExecutionEvent

Add fileds like
https://github.com/flyteorg/flyte/blob/b14f1ac1c69b5925440ca847f6a85083b64bb854/flyteidl/protos/flyteidl/event/event.proto#L254
```
    // log information for the task execution
    repeated core.TaskLog logs = 6;

    // This timestamp represents when the original event occurred, it is generated
    // by the executor of the task.
    google.protobuf.Timestamp occurred_at = 7;

    // A predefined yet extensible Task type identifier. If the task definition is already registered in flyte admin
    // this type will be identical, but not all task executions necessarily use pre-registered definitions and this
    // type is useful to render the task in the UI, filter task executions, etc.
    string task_type = 14;

    // Phase associated with the event
    core.TaskExecution.Phase phase = 4;

    // id of the process that sent this event, mainly for trace debugging
    string producer_id = 5;

```

we could skip `task_type` and `occured_at` as this is know to propeller

kumare3: Most importantly i would remove the State and move it to Phase

Future-Outlier:
Update:
1.  Use `getClientFunc` back, since async task and sync task will use the same client. (Connect to the same servicer.)
2.  `CreateTaskResponse` proto change
(1) add `log_links`
(2) use `one_of` to support async task (return resouce_metadata) and sync task (return Resource)
3. add annotations in proto
4. remove DoTask proto
5. keep async/sync routing mechanism the same


Here are the new implementaion of sync agent task.
(1) agent-service
https://github.com/flyteorg/flytekit/pull/1822/files#diff-9f7af27264f8773b069e8200804c224fe19a6fcaaf9dc33edc644f5351cbb3beR93-R96
(2) return `Resouce` in CreateTaskRequest
- TaskExecutor (api_task agent)

https://github.com/flyteorg/flytekit/pull/1822/files#diff-61929e8f8c1a31b47f99c615311a9154f3fbdc50613f3bd92da01fcf46e6c897R30-R51

- Chatgpt Task Implementation
https://github.com/flyteorg/flytekit/pull/1822/files#diff-735f823d1e8d66245b78021b7e4ed7ca98493b15728bd0a9ae7a143d405bd55dR43-R66

Future-Outlier:
@honnix @kumare3 @pingsutw @EngHabu 
Please take a look, thank you all really much!

flyteidl/protos/flyteidl/core/tasks.proto
EngHabu: @honnix why should the is_sync be a feature/property of the task rather than the agent/plugin?

honnix: Sorry I have been able to follow up recent changes. This was needed for routing https://github.com/flyteorg/flyte/pull/4107/files#diff-da53a6f262afa680bad44d5bc3e5a6a95f1e8b8764002e0db9a7f015af7d2831R137-R154. Not sure whether this is still valid.

honnix: It's hard for me to see the full picture, especially combined with #4511 . I have a feeling that metadata service is used to have a reflection type of thing so the routing here won't be needed, but I'm not entirely sure.

It would be great if you can outline the overall architecture of these changes.

Future-Outlier:
```
pyflyte run --remote --image futureoutlier/flytekit:chatgp(dev) root@googler:/mnt/c/code/dev/example/plugins# pyflyte run --remote --image futureoutlier/flytekit:chatgpt-0306 chatgpt_example.py wf
```

image

flyteplugins/go/tasks/plugins/webapi/agent/plugin.go
```diff
@@ -112,12 +112,21 @@ func (p Plugin) Create(ctx context.Context, taskCtx webapi.TaskExecutionContextR
...
 
+	resource := ResourceWrapper{State: admin.State_RUNNING}
+
+	if res.GetResource() != nil {
+		resource.State = res.GetResource().State
+		resource.Outputs = res.GetResource().Outputs
+		resource.Message = res.GetResource().Message
+		resource.LogLinks = res.GetResource().LogLinks
+	}
+
```
kumare3, line 123: else?

Future-Outlier:
update:
sync task
image
async task
image

pingsutw (dismissed):

Future-Outlier:
I just test it again, it is correct.
image

pingsutw (approved):

PR #4117, opened 2023-10-02
Update the item only if it exists in the cache

pingsutw:
# TL;DR
I found that Propeller keeps sending the request to the agent even workflow is done.

The work queue only has unique items, so we add `itemID` to it. The workers won't process the item again after the task is done. 

Before:
||enqueueBatches| workqueue | sync (10 workers) |
|----|----|-----|----|
| t1 | obj(1) | obj(1) |  |
| t2 |  |  |  |
| t3 | obj(2)  |  obj(1), obj(2) |  |
| t4 |  |  |  |
| t5 | obj(1) | obj(1), obj(2), obj(1) |  |
| t6 |  |  | obj(1), obj(2), obj(1) |

After:
||enqueueBatches| workqueue | sync (10 workers) |
|----|----|-----|----|
| t1 | 1 | 1 |  |
| t2 |  |  |  |
| t3 | 2  |  1, 2 |  |
| t4 |  |  |  |
| t5 | 1 | 1, 2 |  |
| t6 |  |  | 1, 2 |

100 workflows, each has 100 tasks.
Before:
image

After:
image

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flytestdlib/cache/auto_refresh.go
+++ b/flytestdlib/cache/auto_refresh.go
@@ -332,7 +332,7 @@ func (w *autoRefresh) sync(ctx context.Context) (err error) {
 
 			for _, item := range updatedBatch {
 				if item.Action == Update {
-					// Add adds the item if it has been evicted or updates an existing one.
+					// Updates an existing item.
 					w.Update(item.ID, item.Item)
 				}
 			}
--- a/flytestdlib/cache/auto_refresh.go
+++ b/flytestdlib/cache/auto_refresh.go
@@ -304,24 +304,25 @@ func (w *autoRefresh) sync(ctx context.Context) (err error) {
 			w.workqueue.Forget(batch)
 			w.workqueue.Done(batch)
 
-			t := w.metrics.SyncLatency.Start()
 			newBatch := make(Batch, 0, len(*batch.(*Batch)))
 			for _, b := range *batch.(*Batch) {
 				itemID := b.GetID()
 				item, ok := w.lruMap.Get(itemID)
 				if !ok {
 					logger.Debugf(ctx, "item with id [%v] not found in cache", itemID)
-					t.Stop()
 					continue
 				}
 				if item.(Item).IsTerminal() {
 					logger.Debugf(ctx, "item with id [%v] is terminal", itemID)
-					t.Stop()
 					continue
 				}
 				newBatch = append(newBatch, b)
 			}
+			if len(newBatch) == 0 {
+				continue
+			}
 
+			t := w.metrics.SyncLatency.Start()
 			updatedBatch, err := w.syncCb(ctx, newBatch)
 
 			if err != nil {
--- a/flytestdlib/cache/auto_refresh.go
+++ b/flytestdlib/cache/auto_refresh.go
@@ -3,6 +3,7 @@ package cache
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/flyteorg/flyte/flytestdlib/contextutils"
@@ -122,6 +123,7 @@ type autoRefresh struct {
 	syncPeriod      time.Duration
 	workqueue       workqueue.RateLimitingInterface
 	parallelizm     int
+	lock            sync.RWMutex
 }
 
 func getEvictionFunction(counter prometheus.Counter) func(key interface{}, value interface{}) {
@@ -173,6 +175,24 @@ func (w *autoRefresh) Start(ctx context.Context) error {
 	return nil
 }
 
+// Update updates the item in the cache only if it exists, return true if we updated the item.
+func (w *autoRefresh) Update(id ItemID, item Item) (ok bool) {
+	w.lock.Lock()
+	ok = w.lruMap.Contains(id)
+	if ok {
+		w.lruMap.Add(id, item)
+	}
+	w.lock.Unlock()
+	return ok
+}
+
+func (w *autoRefresh) Delete(key interface{}) {
+	w.lock.Lock()
+	w.toDelete.Remove(key)
+	w.lruMap.Remove(key)
+	w.lock.Unlock()
+}
+
 func (w *autoRefresh) Get(id ItemID) (Item, error) {
 	if val, ok := w.lruMap.Get(id); ok {
 		w.metrics.CacheHit.Inc()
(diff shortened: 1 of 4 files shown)


pingsutw:
Updates:
- Still add the item to the workqueue
- Worker only updates the item if it exists in the cache
- To prevent items from being added back to the cache after deletion, update the item with a lock.

If the Item is added back to the cache after deletion, worker will run forever. 
1. Add item to the workqueue (there is no status in the item since we didn't call `plugin.status` in the [sync](https://github.com/flyteorg/flyte/blob/a218a0b683f9f77a77c7bf2e36d285831cdce41b/flyteplugins/go/tasks/pluginmachinery/internal/webapi/cache.go#L59) function)
2. Update LRU cache 
3. Repeat 1

Screenshot 2023-10-03 at 3 20 25 PM

image

flytestdlib/cache/auto_refresh.go
```diff
@@ -275,17 +295,34 @@ func (w *autoRefresh) sync(ctx context.Context) (err error) {
...
-			updatedBatch, err := w.syncCb(ctx, *item.(*Batch))
-
 			// Since we create batches every time we sync, we will just remove the item from the queue here
 			// regardless of whether it succeeded the sync or not.
 			w.workqueue.Forget(item)
 			w.workqueue.Done(item)
 
+			t := w.metrics.SyncLatency.Start()
+			batch := *item.(*Batch)
+			if len(batch) == 1 {
```
EngHabu, line 309: I think this block should instead do something like:
```
newBatch := make(Batch, 0, len(batch))
for _, itemID := range batch {
    item, ok := w.lruMap.Get(itemID)
    if !ok {
     ...
     continue
    }

    if is terminal {
    ..... continue
    }

    b = append(b, item)
}

syncB(...b)

File: flytestdlib/cache/auto_refresh.go [hunk 2]
```diff
@@ -173,6 +175,25 @@ func (w *autoRefresh) Start(ctx context.Context) error {
...
 }
 
+// Update updates the item only if it exists in the cache, return true if we updated the item.
+func (w *autoRefresh) Update(id ItemID, item Item) (ok bool) {
+	w.lock.Lock()
+	ok = w.lruMap.Contains(id)
+	if ok {
+		w.lruMap.Add(id, item)
+	}
+	w.lock.Unlock()
```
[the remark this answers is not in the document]
pingsutw: done

flytestdlib/cache/auto_refresh.go
```diff
@@ -173,6 +175,25 @@ func (w *autoRefresh) Start(ctx context.Context) error {
...
+	return ok
+}
+
+// Delete deletes the item from the cache if it exists.
+func (w *autoRefresh) Delete(key interface{}) {
+	w.lock.Lock()
+	w.toDelete.Remove(key)
+	w.lruMap.Remove(key)
+	w.lock.Unlock()
```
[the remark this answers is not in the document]
pingsutw: done

flytestdlib/cache/auto_refresh.go
pingsutw: done

flytestdlib/cache/auto_refresh.go
```diff
@@ -273,18 +293,36 @@ func (w *autoRefresh) sync(ctx context.Context) (err error) {
...
+			w.workqueue.Done(batch)
+
+			t := w.metrics.SyncLatency.Start()
+			newBatch := make(Batch, 0, len(*batch.(*Batch)))
+			for _, b := range *batch.(*Batch) {
+				itemID := b.GetID()
+				item, ok := w.lruMap.Get(itemID)
+				if !ok {
+					logger.Debugf(ctx, "item with id [%v] not found in cache", itemID)
+					t.Stop()
```
EngHabu, line 314: We shouldn't stop the timer here, should we? just log and continue? the stop should happen after sync is called

flytestdlib/cache/auto_refresh.go
```diff
@@ -273,18 +293,36 @@ func (w *autoRefresh) sync(ctx context.Context) (err error) {
...
+				}
+				if item.(Item).IsTerminal() {
+					logger.Debugf(ctx, "item with id [%v] is terminal", itemID)
+					t.Stop()
+					continue
+				}
+				newBatch = append(newBatch, b)
+			}
+
+			updatedBatch, err := w.syncCb(ctx, newBatch)
```
EngHabu, line 325: We should add a check for if the len(newBatch) == 0, we don't need to sync at that point...

flytestdlib/cache/auto_refresh.go
```diff
@@ -173,6 +175,25 @@ func (w *autoRefresh) Start(ctx context.Context) error {
...
+	if ok {
+		w.lruMap.Add(id, item)
+	}
+	return ok
+}
+
+// Delete deletes the item from the cache if it exists.
+func (w *autoRefresh) Delete(key interface{}) {
+	w.lock.Lock()
+	defer w.lock.Unlock()
```
EngHabu, line 192: I wanted to avoid locking the entire dataset for these operations... any reason to?

flytestdlib/cache/auto_refresh.go
```diff
@@ -173,6 +175,25 @@ func (w *autoRefresh) Start(ctx context.Context) error {
 	return nil
 }
 
+// Update updates the item only if it exists in the cache, return true if we updated the item.
+func (w *autoRefresh) Update(id ItemID, item Item) (ok bool) {
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	ok = w.lruMap.Contains(id)
+	if ok {
+		w.lruMap.Add(id, item)
```
EngHabu, line 184: Why only update if it's already there, what if it has been evicted since?

flytestdlib/cache/auto_refresh.go
pingsutw: updated it

flytestdlib/cache/auto_refresh.go
pingsutw: updated it

flytestdlib/cache/auto_refresh.go
pingsutw: If we don't check if it exist in the cache, we might have this scenario.

`worker1` removes item from the caches -> `worker2` adds the item to cache again. (workerqueue can have duplicate item)

If we don't use lock, we might have this scenario.

1. w.toDelete.Remove (worker 1)
2. w.lruMap.Contains(id) (woker 2)
3. w.lruMap.Remove(key) (woker 1)
4. w.lruMap.Add(id, item) (woker 2) # we add item back to the cache again.

EngHabu (approved):

PR #4135, opened 2023-10-03
Moving from flyteplugins - Feat: Expose nodeID to container

eapolinario:
# TL;DR
Exposes the NodeId to the kubernetes Pod as environment variable.

```
- name: FLYTE_INTERNAL_NODE_ID
  value: n0/dn0/dn0/dn0/dn0/dn0/dn0/dn0/dn0/dn0
```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Currently the environment variables do not provide enough information to know the exact task being executed. 

For example we intended to create a unique ID for all retries of a task. Without the nodeID this is not possible as tasks might be reused. 

One can retrieve this information also from the hostname but this fails for long names due to k8s pod name length limits.

### DIFF
--- a/flyteplugins/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
+++ b/flyteplugins/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
@@ -31,6 +31,15 @@ func GetContextEnvVars(ownerCtx context.Context) []v1.EnvVar {
 			},
 		)
 	}
+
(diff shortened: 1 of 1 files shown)


fg91:
Original PR was https://github.com/flyteorg/flyte/pull/4135. Closing for now as we discussed with @hamersaw that a different approach should be chosen.

fellhorn:
Will open a new PR once I were able to refactor it in such a way that the ID is the same as displayed in the UI (without truncation)

PR #4157, opened 2023-10-03
Add KeepAliveParameters to agent client

pingsutw:
## Tracking issue

## Describe your changes
Agents use headless services to balance client load. However, If [HPA](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/) is used, propeller needs to be restarted to get a new set of IP addresses.

We could add `KeepAliveParameters` to gRPC client, so that propeller can get a new set of IPs every 10s by default.

- https://discuss.kubernetes.io/t/how-to-do-load-balance-with-grpc-connection-when-hpa-autoscaling-is-enabled/16612

### DIFF
--- a/flyteplugins/go/tasks/plugins/webapi/agent/config.go
+++ b/flyteplugins/go/tasks/plugins/webapi/agent/config.go
@@ -43,6 +43,11 @@ var (
 			Endpoint:       "dns:///flyteagent.flyte.svc.cluster.local:80",
 			Insecure:       true,
 			DefaultTimeout: config.Duration{Duration: 10 * time.Second},
+			KeepAliveParameters: &KeepAliveParameters{
+				Time:                config.Duration{Duration: 10 * time.Second},
+				Timeout:             config.Duration{Duration: 5 * time.Second},
+				PermitWithoutStream: true,
+			},
 		},
 	}
 
@@ -67,6 +72,22 @@ type Config struct {
 	AgentForTaskTypes map[string]string `json:"agentForTaskTypes" pflag:"-,"`
 }
 
+// KeepAliveParameters defines keepalive parameters on the client-side. For more details, check https://pkg.go.dev/google.golang.org/grpc/keepalive#ClientParameters
+type KeepAliveParameters struct {
+	// After a duration of this time if the client doesn't see any activity it
+	// pings the server to see if the transport is still alive.
+	// If set below 10s, a minimum value of 10s will be used instead.
+	Time config.Duration `json:"time"`
+	// After having pinged for keepalive check, the client waits for a duration
+	// of Timeout and if no activity is seen even after that the connection is
+	// closed.
+	Timeout config.Duration `json:"timeout"`
+	// If true, client sends keepalive pings even with no active RPCs. If false,
+	// when there are no active RPCs, Time and Timeout will be ignored and no
+	// keepalive pings will be sent.
+	PermitWithoutStream bool `json:"permitWithoutStream"`
+}
+
 type Agent struct {
 	// Endpoint points to an agent gRPC endpoint
 	Endpoint string `json:"endpoint"`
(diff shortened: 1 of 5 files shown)


pingsutw:
cc @honnix mind taking a look

honnix:
Nice finding! I did not know there are no reasonable defaults for these. I guess we need the same thing for https://github.com/flyteorg/flytepropeller/blob/0fcc1da879333af1282ee1142651e191eb1d6bb4/pkg/controller/nodes/catalog/config.go#L44, and https://github.com/flyteorg/flyteidl/blob/85bfc8ff1c36b2d37a4fcc68a2be1911a9fc3940/clients/go/admin/client.go#L173 ?

flyteplugins/go/tasks/plugins/webapi/agent/config.go
```diff
@@ -43,6 +43,11 @@ var (
 			Endpoint:       "dns:///flyteagent.flyte.svc.cluster.local:80",
 			Insecure:       true,
 			DefaultTimeout: config.Duration{Duration: 10 * time.Second},
+			KeepAliveParameters: &KeepAliveParameters{
+				Time:                config.Duration{Duration: 10 * time.Second},
+				Timeout:             config.Duration{Duration: 5 * time.Second},
+				PermitWithoutStream: true,
```
honnix, line 49: What is the reason we set this to `true`? It defaults to `false` I think.

flyteplugins/go/tasks/plugins/webapi/agent/config.go
```diff
@@ -43,6 +43,11 @@ var (
 			Endpoint:       "dns:///flyteagent.flyte.svc.cluster.local:80",
 			Insecure:       true,
 			DefaultTimeout: config.Duration{Duration: 10 * time.Second},
+			KeepAliveParameters: &KeepAliveParameters{
```
honnix, line 46: What do you think we keep existing defaults defined by https://pkg.go.dev/google.golang.org/grpc/keepalive#ClientParameters unchanged, and only give a value for `Time`? Also 10s seems a bit aggressive. In a large setup, the overhead of DNS resolution is not negligible, plus 10s might be even smaller than DNS cache timeout so in many cases the refresh won't give any new IPs. In our backend production setup, we default this to 5 minutes, just for reference.

flyteplugins/go/tasks/plugins/webapi/agent/config.go
```diff
@@ -43,6 +43,11 @@ var (
...
+				Time:                config.Duration{Duration: 10 * time.Second},
+				Timeout:             config.Duration{Duration: 5 * time.Second},
```
honnix, line 48:
```suggestion
				Timeout:             config.Duration{Duration: 20 * time.Second},
```

flyteplugins/go/tasks/plugins/webapi/agent/config.go
honnix, line 49:
```suggestion
				PermitWithoutStream: false,
```

honnix:
After reading the [doc](https://grpc.io/docs/guides/keepalive/) more in-depth, I'm wondering whether this keepalive is intended for IP refreshing. It seems not to me. I think the problem might be more related to name resolver. The ping itself may not trigger name resolution refreshing.

There are more info:
* https://grpc.io/docs/guides/custom-name-resolution/
* https://learn.microsoft.com/en-us/aspnet/core/grpc/loadbalancing?view=aspnetcore-7.0#dns-address-caching (it's dotnet but should be general enough)

That being said, I think this configuration still makes sense. It's just I'm not sure it solves the problem. Internally we have a custom name resolver goes together with client-side load balancing so we have not seen this problem.

pingsutw:
It did solve the issue after adding these config. do you use HPA internally? The problem is that I need to restart the propeller if I scale up the agent. otherwise, propeller won't send the request to the new pod.

pingsutw:
> The DNS resolver can also optionally refresh itself on a periodic interval. This can be useful for quickly detecting new pod instances.

https://github.com/grpc/grpc/issues/18743 You're right. we can configure refresh frequency for dns resolver in the URL.  let me test it, thanks.

pingsutw:
https://github.com/grpc/grpc/issues/12295#issuecomment-893618782 you're right. updating resolver might a better solution. let me add one and test it.

honnix:
> It did solve the issue after adding these config. do you use HPA internally? The problem is that I need to restart the propeller if I scale up the agent. otherwise, propeller won't send the request to the new pod.

I actually can't explain how this regular ping would trigger name resolution and why it would be different than normal traffic when coming to name resolution. Without these configuration, if there is no normal traffic there is no resolution, and when traffic comes back it should work the same as sending a ping, triggering resolution if DNS cache timed out. Do you find any doc stating that `ping` explicitly forces resolution?

Yes we use HPA.

PR #4165, opened 2023-10-03
Add version control for ArrayNode eventing

hamersaw:
## Tracking issue
_NA_

## Describe your changes
Add a version control flag for ArrayNode eventing. `0` means use the existing approach (ie. populating `ExternalResourceInfo` field on the `TaskExecutionEvent`) as a drop-in replacement for maptask. `1` means to write separate `NodeExecutionEvents` for each subNode execution (linked by parent node ID) for full metadata. 

## Screenshots
_NA_

## Note to reviewers
_NA_


flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -541,24 +491,25 @@ func New(nodeExecutor interfaces.Node, eventConfig *config.EventConfig, scope pr
...
+		return nil, nil, nil, nil, nil, nil, err
 	}
 
 	inputReader := newStaticInputReader(nCtx.InputReader(), inputLiteralMap)
 
 	// wrap node lookup
 	subNodeSpec := *arrayNode.GetSubNodeSpec()
 
-	subNodeID := fmt.Sprintf("%s-n%d", nCtx.NodeID(), subNodeIndex)
+	//subNodeID := fmt.Sprintf("%s-n%d", nCtx.NodeID(), subNodeIndex)
```
pingsutw, line 511: remove?
```suggestion
```

hamersaw: thanks!

pingsutw (approved):

flytepropeller/pkg/controller/nodes/array/handler_test.go
```diff
@@ -379,7 +380,7 @@ func TestHandleArrayNodePhaseNone(t *testing.T) {
 			},
 			expectedArrayNodePhase:         v1alpha1.ArrayNodePhaseExecuting,
 			expectedTransitionPhase:        handler.EPhaseRunning,
-			expectedExternalResourcePhases: []idlcore.TaskExecution_Phase{idlcore.TaskExecution_QUEUED, idlcore.TaskExecution_QUEUED},
+			expectedExternalResourcePhases: []idlcore.TaskExecution_Phase{idlcore.TaskExecution_UNDEFINED, idlcore.TaskExecution_UNDEFINED},
```
eapolinario, line 383: just curious, why are we changing this?

eapolinario (approved):

flytepropeller/pkg/controller/nodes/array/handler_test.go
hamersaw: This test case is the scenario where the ArrayNode has not yet started. We write an single node and task event for each subnode to alert admin of it's existence - mainly so the UI can display `undefined x20` to indicate that there are 20 subtasks. We previously initially reported these with the `Queued` phase, because that skips input resolution downstream allowing us to just use a mocked in-memory input value from the array rather than copy each individual subnode input to a separate blobstore location. However, we can still use `Queued` as the initial phase during downstream evaluations and use `Undefined` here to more accurately report that these tasks have not yet started.

fg91:
@hamersaw when running map tasks with 10k pods I observe that sometimes the pod status is not updated in admin for hours.
When searching for knobs to tune I found this `event-version` flag you introduced. On latest master `0` still seems to be the default value.
Can you make a recommendation whether `event-version: 0` or `event-version: 1` should work better at large number of pods?

PR #4170, opened 2023-10-04
Override Ray config with fields for namespace, service account, pod resources

guozhen-la:
Recreates PR that was moved in the monorepo 
see #4136

As per our conversation with the Union AI, we are making the following changes to the Ray plugin to enable more parameters to be configurable on the user side using flytekit SDK.

see: https://github.com/flyteorg/flyteidl/pull/429


eapolinario:

Issue #4105, opened 2023-09-29
[Core feature] Add PodTemplate support for Spark driver and executor pods


andrewwdye:
### Motivation: Why do you think this is important?

Today Spark driver and executor pods are not configurable on a per task basis; they use the [shared propeller k8s plugin config](https://github.com/flyteorg/flyte/blob/b35cc95afa15034146c5ab403ea990d45bd1a66e/flyteplugins/go/tasks/plugins/k8s/spark/spark.go#L102-L133). This disallows, for instance, setting additional tolerations when running a Spark job.

This extends work PodTemplate investments made in 

### Goal: What should the final outcome look like, ideally?

flytekit should accept optional driver and executor [pod templates](https://docs.flyte.org/en/latest/deployment/configuration/general.html#using-k8s-podtemplates) in Spark task config
```
@task(
    task_config=Spark(
        ...
        driver_pod_template=PodTemplate(...),
        executor_pod_template=PodTemplate(...),
    ),
    ...
)
```

The Spark plugin should use these optional templates and existing [ToK8sPodSpec](https://github.com/flyteorg/flyte/blob/b35cc95afa15034146c5ab403ea990d45bd1a66e/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L256) logic to populate the [SparkApplicationSpec](https://github.com/flyteorg/flyte/blob/b35cc95afa15034146c5ab403ea990d45bd1a66e/flyteplugins/go/tasks/plugins/k8s/spark/spark.go#L102-L133).

### Describe alternatives you've considered 

The Spark on K8s operator supports a bunch of [k8s pod settings](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/5f2efd4ff97e7c0bfdb726a066118d3401576730/pkg/config/constants.go#L63-L152) via `spark_conf`, which is already available in flytekit. While it doesn't support tolerations, for example, we could extend this concept with the addition of keys like `spark.kubernetes.driver.tolerations` and either get support added to Spark on K8s operator or parse out separately in flyteplugins and apply to the SparkApplicationSpec when building it.

###  Propose: Link/Inline OR Additional context 

[...]

PR #4179, opened 2023-10-06
Add driver and executor pod template fields to SparkJob

andrewwdye:
## Tracking issue
https://github.com/flyteorg/flyte/issues/4105

## Describe your changes
This change defines a common pattern for per-role pod configuration and adds RoleSpec fields for driver and executor pods, accordingly. 
* Adds a `RoleSpec` to common.proto with initial fields for `pod` and `pod_template_name`. This mirrors the interface we have for container tasks spread across TaskTemplate and TaskMetadata. Fields are optional.
* Adds `driverSpec` and `executorSpec` to SparkJob

This PR was manually migrated from https://github.com/flyteorg/flyteidl/pull/450. See initial comment discussions there.

**How should a plugin handle TaskTemplate/TaskMetadata pod templates vs. these RoleSpecs?**
For Spark, the task-level pod template is not consumed today and can be treated as an error if set (not supported). The plugin can inspect optional per-role specs for configuration.

For existing plugins that consume the task-level pod template (Dask, MPI, PyTorch, Ray), they can continue doing so. If/when we are ready to introduce per-role pod configuration, we can add `RoleSpec` fields to the plugin's config, deprecate the task-level pod template, and rely on a runtime check to validate that only task-level pod template OR per-role pod templates are specified.


wild-endeavor (approved):

wild-endeavor:
@eapolinario i see that one +1 was enough to merge this pr.  should we continue to make idl a +2?  I think we should.

not sure if there was resolution to @hamersaw's comment.

eapolinario:
> @eapolinario i see that one +1 was enough to merge this pr. should we continue to make idl a +2? I think we should.
> 
> not sure if there was resolution to @hamersaw's comment.

As far as I can tell CODEOWNERS does not support that feature (as per https://github.com/orgs/community/discussions/22522). I opened https://github.com/flyteorg/flyte/issues/4182 to track this.

pingsutw (commented):
Should we follow the same pattern as [tensorflow.proto](https://github.com/flyteorg/flyte/blob/master/flyteidl/protos/flyteidl/plugins/kubeflow/tensorflow.proto#L11)

like
```proto
message SparkJob {
    ...
    sparkReplicaSpec driver = 11;
    sparkReplicaSpec worker = 12;
}

message sparkReplicaSpec {
    oneof PodValue {
        core.K8sPod podTemplate = 1;
    }
    string PodTemplateName = 2;
    string image = 3;
     ...
}
```

Maybe we should put repliaSpec to `common.proto`, all other plugins can import this proto in the future, like Ray.
```proto
# common.proto
message ReplicaSpec {
    oneof PodValue {
        core.K8sPod podTemplate = 1;
    }
    string PodTemplateName = 2;
    string image = 3;
     ...
}
```
cc @eapolinario @wild-endeavor @hamersaw wdyt

andrewwdye:
Wanted to also cross reference [this comment thread](https://github.com/flyteorg/flyteidl/pull/450#discussion_r1348845580) from the original PR, proposing using the `TaskTemplate.PodTemplate` for all roles, unless specifically overridden with a role-specific template/spec.

```
@task(
    task_config=Spark(
        spark_conf={
           "spark.driver.memory": "4G",
           "spark.executor.memory": "8G",
        },
        # driver only pod template overrides (takes precedence over top level pod_template)
        driver_pod_template=...
    ),
    # default pod template overrides for BOTH driver and executor
    pod_template=...
)
```

andrewwdye:
Pulling together references from existing leader/worker patterns for various plugins
* [Dask](https://github.com/flyteorg/flyte/blob/97a79c030555eaefa3e27383d9b933ba1fdc1140/flyteplugins/go/tasks/plugins/k8s/dask/dask.go#L103-L112) scheduler/worker uses a common spec w/ scheduler treated as non-interruptible
* [MPI](https://github.com/flyteorg/flyte/blob/97a79c030555eaefa3e27383d9b933ba1fdc1140/flyteplugins/go/tasks/plugins/k8s/kfoperators/mpi/mpi.go#L61-L143) uses a common pod spec with some per role args
* [PyTorch](https://github.com/flyteorg/flyte/blob/97a79c030555eaefa3e27383d9b933ba1fdc1140/flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go#L61-L133) uses a common pod spec with some per role args
* [Tensorflow](https://github.com/flyteorg/flyte/blob/c88f8f50c291a3f1c5c936781f606bf956aeaecd/flyteplugins/go/tasks/plugins/k8s/kfoperators/tensorflow/tensorflow.go#L106-L152) uses separate replica specs for each of chief/worker/ps
* [Ray](https://github.com/flyteorg/flyte/blob/97a79c030555eaefa3e27383d9b933ba1fdc1140/flyteplugins/go/tasks/plugins/k8s/ray/ray.go#L60-L117) uses a common pod spec for headgroup and worker

andrewwdye:
Based on comments from the [initial PR thread](https://github.com/flyteorg/flyteidl/pull/450#discussion_r1347463428) and above, I've moved pod and pod template name behind a `RoleSpec` definition in common.proto. See latest PR summary for the pattern this establishes and expected handling of `TaskTemplate.K8sPod`.

pingsutw (approved):

wild-endeavor (approved):
+1 but is this the right repo @eapolinario ?

flyteidl/protos/flyteidl/plugins/spark.proto
```diff
@@ -31,4 +33,8 @@ message SparkJob {
     // Domain name of your deployment. Use the form <account>.cloud.databricks.com.
     // This instance name can be set in either flytepropeller or flytekit.
     string databricksInstance = 9;
+    // The driver spec, used in place of the task's pod template
+    RoleSpec driverSpec = 10;
+    // The executor spec, used in place of the task's pod template
+    RoleSpec executorSpec = 11;
```
hamersaw, line 39: Is there every a scenario where both of these differ from the values set in the task decorator?

hamersaw:
> Maybe we should put repliaSpec to `common.proto`, all other plugins can import this proto in the future, like Ray.

In theory I think this is a great idea. But in practice I would bet there will be corner case plugins where not all options are supported.

flyteidl/protos/flyteidl/plugins/spark.proto
andrewwdye: Curious what you mean here. With this change we will fail
```
@task(
  pod_template=...
)
```
in favor of specifying one or both values in
```
@task(
  task_config=Spark(
    executor_pod_template=...
    driver_pod_template=...
  )
)
```

andrewwdye (commented):
> In theory I think this is a great idea. But in practice I would bet there will be corner case plugins where not all options are supported

Are you thinking of initial fields in `RoleSpec` or future ones? If we believe that K8sPod is the correct abstraction for overriding parts of the default pod template, then including it as part of the role spec seems appropriate. I could imagine, however, additional role-specific configuration desired over time that wouldn't fit into a common `RoleSpec` definition. Thoughts?

andrewwdye:
Given some of the outstanding questions for Spark, the pattern for multi-role plugins more broadly, and the deprecation/migration story, let me close this out until we get alignment. For now Spark plugin can use the existing TaskTemplate K8sPod field to override tolerations for both driver and executor pods, which is complimentary to the existing interface for node selectors: `spark.kubernetes.node.selector...` (same values for both roles). This is also consistent with how other plugins use pod templates for multiple roles today.

PR #4183, opened 2023-10-06
Implement Spark pod template overrides

andrewwdye:
## Tracking issue
https://github.com/flyteorg/flyte/issues/4105

## Describe your changes
Create driver, executor, and spark application spec from optional pod template. Uses [ToK8sPodSpec](https://github.com/flyteorg/flyte/blob/b3267916fd5e886805b4ec472c7e39c12b9ef397/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L258) as a shared solution for merging defaults with container or pod template target overrides. Driver and executor specs will share the same pod template values. See note below on this choice.

**Why use the same pod template for driver and executor?**
As part of https://github.com/flyteorg/flyte/pull/4179 we had a discussion about whether or not to use the task-level pod template, or introduce role-specific pod templates for driver and executor. For now this change adopts the task-level pod template pattern used by other k8s plugins. We may end up moving this and others to role-specific templates in the future.

## Test plan
**Unit tests**
I refactored, extended, and renamed the existing container based Spark unit test (`TestBuildResourceContainer`). I added a new `TestBuildResourcePodTemplate` test to cover most of the interface.

**Sanity check diff of driver and executor pods locally**
1. Setup and build [single binary](https://docs.flyte.org/en/latest/community/contribute.html#how-to-setup-dev-environment-for-flyteidl-flyteadmin-flyteplugins-flytepropeller-datacatalog-and-flytestdlib)
2. Add `spark` to enabled-plugins to the config following [this guide](https://docs.flyte.org/en/stable/deployment/plugins/k8s/index.html)
3. Add spark role, role binding, and service account template files from [this guide](https://docs.flyte.org/en/stable/deployment/plugins/k8s/index.html) to `$HOME/.flyte/cluster-resource-templates/`
4. Register and run a spark workflow following [this guide](https://docs.flyte.org/en/stable/deployment/plugins/k8s/index.html#register-the-plugin-example)
5. Compare driver and executor pods between master and this branch

**Verify behavior of specifying `@task(pod_template=...)` locally**
Modified the `my_spark` example to use a pod template
```
@task(
    ...
    pod_template=PodTemplate(
        primary_container_name="primary",
        labels={"lKeyA": "lValA", "lKeyB": "lValB"},
        annotations={"aKeyA": "aValA", "aKeyB": "aValB"},
        pod_spec=V1PodSpec(
            containers=[
                V1Container(
                    name="primary",
                    image=custom_image,
                    command="echo",
                    args=["wow"],
                ),
            ],
            tolerations=[
                V1Toleration(
                    key="x/custom",
                    operator="Equal",
                    value="foo",
                    effect="NoSchedule",
                ),
            ],
        ),
    ),
)
```
And verified that the resulting pod had tolerations set
```
❯ kubectl get pod -n flytesnacks-development a68zfbjq8r74mjb9q5tq-n0-0-driver -o yaml | grep "tolerations:" -A 12
  tolerations:
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
  - effect: NoSchedule
    key: x/custom
    operator: Equal
    value: foo
```


flyteplugins/go/tasks/pluginmachinery/flytek8s/non_interruptible.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/flyteorg/flyte/flyteidl/gen/pb-go/flyteidl/core"
+	pluginsCore "github.com/flyteorg/flyte/flyteplugins/go/tasks/pluginmachinery/core"
+)
+
+// Wraps a regular TaskExecutionMetadata and overrides the IsInterruptible method to always return false
+// This is useful as the runner and the scheduler pods should never be interruptible
+type NonInterruptibleTaskExecutionMetadata struct {
+	metadata pluginsCore.TaskExecutionMetadata
+}
+
```
hamersaw, line 17: Is there a reason that we're using this as a separate field rather than composition? Then we can just mask the `IsInterruptible` instead of everything?
```
type NonInterruptibleTaskExecutionMetadata struct {
    pluginsCore.TaskExecutionMetadata
}

func (n NonInterruptibleTaskExecutionMetadata) IsInterruptible() bool {
	return false
}
```

flyteplugins/go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -3,35 +3,29 @@ package spark
...
+	"github.com/flyteorg/flyte/flyteidl/gen/pb-go/flyteidl/plugins"
 	"github.com/flyteorg/flyte/flyteplugins/go/tasks/errors"
 	"github.com/flyteorg/flyte/flyteplugins/go/tasks/logs"
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/pluginmachinery"
 	pluginsCore "github.com/flyteorg/flyte/flyteplugins/go/tasks/pluginmachinery/core"
-
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/pluginmachinery/flytek8s"
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/pluginmachinery/flytek8s/config"
 	"github.com/flyteorg/flyte/flyteplugins/go/tasks/pluginmachinery/k8s"
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/pluginmachinery/tasklog"
```
hamersaw, line 27: +1 My secret goal is to touch every file in these repos and cleanup imports.

flyteplugins/go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -165,57 +109,158 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
...
+	spec := sparkOp.SparkPodSpec{
+		Affinity:         podSpec.Affinity,
+		Annotations:      annotations,
+		Labels:           labels,
+		EnvVars:          sparkEnvVars,
+		Image:            &container.Image,
+		SecurityContenxt: podSpec.SecurityContext.DeepCopy(),
+		DNSConfig:        podSpec.DNSConfig.DeepCopy(),
+		Tolerations:      podSpec.Tolerations,
+		SchedulerName:    &config.GetK8sPluginConfig().SchedulerName,
```
hamersaw, line 164: Can we use `podSpec.SchedulerName` here? Similarly for `HostNetwork`. These should be correctly set in `ToK8sPodSpec`.

flyteplugins/go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -165,57 +109,158 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
...
+}
 
-	execCount, err := strconv.ParseInt(sparkConfig["spark.executor.instances"], 10, 32)
-	if err == nil {
-		executorSpec.Instances = intPtr(int32(execCount))
+type driverSpec struct {
+	podSpec   *v1.PodSpec
+	container *v1.Container
+	sparkSpec *sparkOp.DriverSpec
+}
```
hamersaw, line 175: Is there a future reason to store primary container and `v1.PodSpec` in the driverSpec and executorSpec. IIUC below we just use the `sparkOp.DriverSpec` below?

hamersaw (commented):
A few nit questions, otherwise looks good. Thanks!

flyteplugins/go/tasks/plugins/k8s/spark/spark.go
andrewwdye: We should just do a big pass to clean this up and add a lint rule to enforce going forward

flyteplugins/go/tasks/plugins/k8s/spark/spark.go
andrewwdye: Just some copy/paste evolution. Let me clean up

flyteplugins/go/tasks/plugins/k8s/spark/spark_test.go
```diff
@@ -535,21 +587,20 @@ func TestBuildResourceSpark(t *testing.T) {
...
 	assert.Equal(t, 2, len(sparkApp.Spec.Executor.Tolerations))
-	assert.Equal(t, 1, len(sparkApp.Spec.Executor.NodeSelector))
-	assert.Equal(t, interruptibleNodeSelector, sparkApp.Spec.Executor.NodeSelector)
-
-	tolExecDefault := sparkApp.Spec.Executor.Tolerations[0]
-	assert.Equal(t, tolExecDefault.Key, "x/flyte")
-	assert.Equal(t, tolExecDefault.Value, "default")
-	assert.Equal(t, tolExecDefault.Operator, corev1.TolerationOperator("Equal"))
-	assert.Equal(t, tolExecDefault.Effect, corev1.TaintEffect("NoSchedule"))
+	assert.Equal(t, 2, len(sparkApp.Spec.Executor.NodeSelector))
```
andrewwdye, line 590: With this change I'm now expecting BOTH default and interruptible node selectors to be applied to the executor pod. This is consistent with how defaults and interruptible works for other plugins. @hamersaw @jeevb do you anticipate problems with this change?

jeevb: My understanding is that the executor is more important than the workers and as such has historically been scheduled on non-interruptible. This should apply to all driver-worker plugins (e.g. spark, ray). If we do want to go down this path, we should at least make it configurable. This behavior could be undesirable to users that use these plugins at scale.

andrewwdye: Ah, just to clarify here, the driver will be treated as non-interruptible and get the non-interruptible selectors *in addition* to the default ones and the default tolerations, while the executor pods will honor the task's interruptible param, and if interruptible will get the interruptible selectors *in addition* to the default ones and the interruptible tolerations in addition to the default ones.

Previously the spark plugin **[replaced](https://github.com/flyteorg/flyte/pull/4183/files#diff-5225038bf6b0b87842d5e795c8485ba63d597946970cc180637955e8baf0e38bL249)** the executor's node selector if interruptible vs. appending. This may have been a bug/oversight

jeevb: Ah. For some reason I read `executor` as `driver`, when an `executor` is just a worker. My mistake! 🤦

jeevb (commented):
LGTM! Looks like this plugin might now be primed for GPU (and GPU affinity) support. A task for a later PR. Thanks!

jeevb (approved):

hamersaw (approved):

jeevb (approved):

Issue #3965, opened 2023-08-15
[Core feature] Make Flyte work with GCP Identity Aware Proxy (IAP)


fg91:
### Motivation: Why do you think this is important?

[GCP Identity Aware Proxy (IAP)](https://cloud.google.com/iap) is a managed service that makes it easy to protect applications deployed on GCP by verifying user identity and using context to determine whether a user should be granted access.

Because requests to applications protected with IAP first have to pass IAP before they can reach the protected application, IAP provides a convenient way to implement a zero-trust access model.

(In contrast, if applications are protected using their own auth mechanism, unauthenticated requests typically first hit the application which only then redirects to e.g. a google login page. With IAP, no unauthenticated request can ever hit the application.)

Since IAP makes it very easy to implement a zero-trust model, many organizations using GCP have a security policy that any internal tool has to be protected with it.

### Goal: What should the final outcome look like, ideally?

Flyte currently does not work with IAP but there is a need in the community to enable this integration:

* > My company has a strict requirement to use IAP everywhere. We’ve been able to get authentication to work on the console but I can’t get access via the flytectl to work. Any tips here? ([source](https://flyte-org.slack.com/archives/C05A0JA1CCD/p1690825265604069))
* > We have our flyte cluster url enabled with IAP (Identity-Aware Proxy). While the UI access to flyte cluster works fine using the google account login but accessing flyte cluster using flytectl fails at authentication ([source](https://flyte-org.slack.com/archives/CP2HDHKE1/p1687944812743459))
* > we [...] currently port forward for everything except Flyte Console ([source](https://flyte-org.slack.com/archives/CP2HDHKE1/p1691621269310779?thread_ts=1691619241.183159&cid=CP2HDHKE1))

### Describe alternatives you've considered 

In organizations where there is a security policy to use IAP, workarounds typically include 1) deploying flyte itself without authentication enabled and instead with IAP in front of flyteconsole and 2) port-forwarding flyteadmin's gRPC server to localhost or interacting with it only from within the cluster/the VPC (as `pyflyte` and `flytectl` cannot reach flyteadmin through IAP).

None of this is great.

###  Propose: Link/Inline OR Additional context 

This issue tracks the integration of flyte with IAP, consisting of the following tickets:

* https://github.com/flyteorg/flytekit/pull/1795

[...]

PR #4189, opened 2023-10-09
[Monorepo] Enable proxy-authorization in admin client

eapolinario:
## Describe your changes

This PR imports https://github.com/flyteorg/flyteidl/pull/437 into the monorepo.

### Original description

Part of an effort to integrate Flyte with [GCP Identity Aware Proxy](https://cloud.google.com/iap), see https://github.com/flyteorg/flyte/issues/3965.

Allows the flyteidl admin client to sent a `"proxy-authorization"` header with every request to flyteadmin. Tokens for this header are created using an external command which is configured via the new `proxyCommand` config entry.

Same logic as was introduced in `flytekit` in https://github.com/flyteorg/flytekit/pull/1787.

I replicated the existing logic in `MaterializeCredentials` and `NewAuthInterceptor` for the second token. The external command is called a single time per `flytectl` call.

## Tracking Issue

Closes https://github.com/flyteorg/flyte/issues/3965


eapolinario:
@fg91 , can you fix the error in single-binary propeller? It repros locally, just run `make compile` from the root of the repo.

wild-endeavor (approved):

PR #4199, opened 2023-10-10
Update k8s Dependencies

EngHabu:
## Describe your changes

- [X] Update K8s dependencies and APIs that changed
- [X] Remove sagemaker plugin since:
    -  It pins old versions of K8s deps
    - I'm told is broken
    - Uses a deprecated AWS Operator
    - Should be replaced with a Flyte Agent.


.github/workflows/codespell.yml
```diff
@@ -17,7 +17,7 @@ permissions:
 jobs:
   codespell:
     name: Check for spelling errors
-    runs-on: ubuntu-latest
+    runs-on: union-k8s-runner
```
pingsutw, line 20: qq: will this GA workflows run on our cluster? If so, should we only run unit and integration test on our cluster? That said, we can use github action to run some other small workflow, such as check for spelling errors, lint, make generate, and push image.

EngHabu: Rethinking this, Union cluster need to have machines with bigger attached disks, the default ones run out of capacity really quickly and can't even run the unit tests... I'll revert this change and think about this separately.

pingsutw (approved):

eapolinario:
I'm reverting this temporarily just to get 1.10 out. The reason being that I don't want to ship these changes without (a lot) more testing.

flytepropeller/pkg/controller/garbage_collector.go
```diff
@@ -162,7 +162,7 @@ func (g *GarbageCollector) StartGC(ctx context.Context) error {
 		logger.Warningf(ctx, "Garbage collector is disabled, as ttl [%d] is <=0", g.ttlHours)
 		return nil
 	}
-	ticker := g.clk.NewTicker(g.interval)
+	ticker := g.clk.NewTimer(g.interval)
```
Tom-Newton, line 165: Why was this changed from a ticker to a timer? I'm a total golang noob but I think this means that flytepropeller will only garbage collect once each time it restarts. 

I think this is causing us some problems https://flyte-org.slack.com/archives/CP2HDHKE1/p1701118219768919?thread_ts=1700590318.336129&cid=CP2HDHKE1

PR #4200, opened 2023-10-10
Cancel a currently running workflow from the same PR when a new workflow is triggered

pingsutw:
# TL;DR
as title, always run the latest workflow in the PR.
https://stackoverflow.com/a/72408109/9574775
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
 _How did you fix the bug, make the feature etc. Link to any design docs etc_

## Tracking Issue
https://github.com/flyteorg/flyte/issues/<number>

## Follow-up issue
_NA_

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,5 +1,9 @@
 name: Components Checks
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
 on:
   pull_request:
     paths:
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,7 +1,7 @@
 name: Components Checks
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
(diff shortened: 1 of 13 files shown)


eapolinario:
This is a great feature!

.github/workflows/checks.yml
```diff
@@ -1,5 +1,9 @@
 name: Components Checks
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
```
eapolinario, line 4: We need the `|| github.ref` to handle the non-PR case.

eapolinario (commented):
That's a great find.

.github/workflows/checks.yml
pingsutw: updated it

eapolinario (approved):
Amazing!

PR #4206, opened 2023-10-12
Fix databricks plugin

pingsutw:
## Tracking issue
The databricks plugin does not check for all possible terminal states as returned from the Databricks Jobs API. As a result, if the Databricks jobs API returns a state which the plugin is not written to recognise, the plugin will think that the job is still in a running phase.

https://github.com/flyteorg/flyte/issues/4243

## Describe your changes
Transition task status to `Failure` if the lifeCycleState is `SKIPPED` or `INTERNAL_ERROR`


givanovexpe (commented):
This fixes the problems we've seen when testing the plugin.

eapolinario (approved):

PR #4216, opened 2023-10-12
Revert "[Monorepo] Enable proxy-authorization in admin client (#4189)"

andrewwdye:
## Describe your changes

Looks like this may have broken propeller -> admin events when auth is enabled

```
panic({0x25fb2e0, 0xc0024577d0})
	/usr/local/go/src/runtime/panic.go:884 +0x212
github.com/flyteorg/flyte/flyteidl/clients/go/admin/externalprocess.Execute({0x423f5f8?, 0x8?, 0x7f15bee905b8?})
	/go/src/github.com/flyteorg/flyteidl/clients/go/admin/externalprocess/command_executor.go:6 +0x68
github.com/flyteorg/flyte/flyteidl/clients/go/admin.ExternalTokenSourceProvider.GetTokenSource({{0x423f5f8?, 0x40dc87?, 0x18?}}, {0x248cec0?, 0x7f1597747501?})
	/go/src/github.com/flyteorg/flyteidl/clients/go/admin/token_source_provider.go:119 +0x25
github.com/flyteorg/flyte/flyteidl/clients/go/admin.GetProxyTokenSource({0x2d1cc90, 0xc00133c6c0}, 0xdb6d77?)
	/go/src/github.com/flyteorg/flyteidl/clients/go/admin/auth_interceptor.go:59 +0x9d
github.com/flyteorg/flyte/flyteidl/clients/go/admin.MaterializeProxyAuthCredentials({0x2d1cc90?, 0xc00133c6c0?}, 0xc0003ce280, 0x2d1cc90?)
	/go/src/github.com/flyteorg/flyteidl/clients/go/admin/auth_interceptor.go:67 +0x3a
github.com/flyteorg/flyte/flyteidl/clients/go/admin.NewProxyAuthInterceptor.func1({0x2d1cc90, 0xc00133c6c0}, {0x2833b6c, 0x32}, {0x26400e0, 0xc0015c5440}, {0x25ff2c0, 0xc0005290c0}, 0x8?, 0x28f81b8, ...)
	/go/src/github.com/flyteorg/flyteidl/clients/go/admin/auth_interceptor.go:165 +0xe6
...
github.com/flyteorg/flyte/flyteidl/gen/pb-go/flyteidl/service.(*adminServiceClient).CreateWorkflowEvent(0xc00098d228, {0x2d1ccc8, 0xc0025848d0}, 0xc00065ebd0?, {0x0, 0x0, 0x0})
[...]

PR #4217, opened 2023-10-13
Fix: Prevent unintended use of proxy authorization when no command is provided

fg91:
## Describe your changes

@andrewwdye noticed that the newly introduced proxy authorization in the admin client breaks propeller, see https://github.com/flyteorg/flyte/pull/4216#event-10638102477.

I ran flytepropeller locally in a debugger (with the flyte-core helm chart deployed in the cluster and flyteadmin and datacatalog port-forwarded).

With the debugger, I confirmed that [this](https://github.com/flyteorg/flyte/blob/68c055b7b51eebc4cf009b5caf488026328cfcc0/flyteidl/clients/go/admin/client.go#L157) if statement which adds proxy authorization is hit even if no proxy command is provided in the platform config.

The reason is that I check for `cfg.ProxyCommand != nil` but the default value of proxy command is [empty string slice  ](https://github.com/flyteorg/flyte/blob/68c055b7b51eebc4cf009b5caf488026328cfcc0/flyteidl/clients/go/admin/config.go#L77) so it is never `nil`.

With flytepropeller running locally in my IDE, I confirmed that:

1. with the fix introduced by this PR, the error reported by @andrewwdye is gone and that workflows can be executed again
2. that if one does configure a proxy command, that the proxy auth interceptor is still added to the dial options

### DIFF
--- a/flyteidl/clients/go/admin/auth_interceptor.go
+++ b/flyteidl/clients/go/admin/auth_interceptor.go
@@ -109,7 +109,7 @@ func setHTTPClientContext(ctx context.Context, cfg *Config, proxyCredentialsFutu
 		transport.Proxy = http.ProxyURL(&cfg.HTTPProxyURL.URL)
 	}
 
-	if len(cfg.ProxyCommand) > 0 {
+	if cfg.ProxyCommand != nil && len(cfg.ProxyCommand) > 0 {
 		httpClient.Transport = &proxyAuthTransport{
 			transport:              transport,
 			proxyCredentialsFuture: proxyCredentialsFuture,
--- a/flyteidl/clients/go/admin/auth_interceptor.go
+++ b/flyteidl/clients/go/admin/auth_interceptor.go
@@ -109,7 +109,7 @@ func setHTTPClientContext(ctx context.Context, cfg *Config, proxyCredentialsFutu
 		transport.Proxy = http.ProxyURL(&cfg.HTTPProxyURL.URL)
 	}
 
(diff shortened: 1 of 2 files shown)


flyteidl/clients/go/admin/auth_interceptor.go
```diff
@@ -109,7 +109,7 @@ func setHTTPClientContext(ctx context.Context, cfg *Config, proxyCredentialsFutu
 		transport.Proxy = http.ProxyURL(&cfg.HTTPProxyURL.URL)
 	}
 
-	if cfg.ProxyCommand != nil {
+	if len(cfg.ProxyCommand) > 0 {
```
wild-endeavor, line 112:
```suggestion
	if cfg.ProxyCommand != nil && len(cfg.ProxyCommand) > 0 {
```

wild-endeavor (commented):
actually can we make this

flyteidl/clients/go/admin/auth_interceptor.go
fg91: https://github.com/flyteorg/flyte/pull/4217/commits/196b71b780c088e7d50dc59160074401ec595eaf

wild-endeavor (approved):

andrewwdye (approved):
I was able to build an image and verify this addressed the original problem. Thanks!

fg91:
> I was able to build an image and verify this addressed the original problem. Thanks!

Thanks for verifying!

Issue #3864, opened 2023-07-12
[Core feature] Enable dataclass attribute access in workflows


franzigeiger:
### Motivation: Why do you think this is important?

I'd like to access dataclass attributes from within a workflow. This is already possible for `NamedTuple` but those are restricted in usage. 

### Goal: What should the final outcome look like, ideally?

I would like this to work:
```
@dataclass
@dataclass_json
class Config:
    x:int
    y:int
    z:int
    some_unused_additional_config:list[str]

@task
def add(x:int, y:int):
    return x+y
@workflow
def do_something(config:Config):
    add(config.x, config.y)
```

### Describe alternatives you've considered 

Currenlty this is only possible with workarounds. You can:
- Make the workflow dynamic
- Use `NamedTuples` only if the configuration object is an output of another workflow. They can't be a task or workflow input.
- Change the `add` task signatures to take in the full config object, which means taking in unused parameters.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

wild-endeavor:
This is something I've wanted for a while too.  I think it's possible but will require a fair bit of changes.  @franzigeiger were you hoping to contribute?

PR #4232, opened 2023-10-15
[flytepropeller] Support attribute access on promise

ByronHsu:
## Tracking issue

flyteorg/flyte#3864

## Describe your changes

See https://github.com/flyteorg/flyteidl/pull/439#issuecomment-1763223435


flytepropeller/pkg/compiler/validators/bindings.go
```diff
@@ -124,6 +124,27 @@ func validateBinding(w c.WorkflowBuilder, nodeID c.NodeID, nodeParam string, bin
 						sourceType = cType
 					}
 				}
+				var exist bool
+				var tmpType *flyte.LiteralType
```
hamersaw, line 128: these aren't used outside of the for loop are they? are they necessary?

flytepropeller/pkg/controller/nodes/attr_path_resolver_test.go
```diff
@@ -0,0 +1,329 @@
...
+	// - map {"foo": "bar"}
+	// - collection ["foo", "bar"]
+	// - struct1 {"foo": "bar"}
+	// - struct2 {"foo": ["bar1", "bar2"]}
+	// - nested list struct {"foo": [["bar1", "bar2"]]}
+	// - map+collection+struct {"foo": [{"bar": "car"}]}
+	// - exception key error with map
+	// - exception out of range with collection
+	// - exception key error with struct
+	// - exception out of range with struct
```
hamersaw, line 43: does it make sense to have these comments next to the test cases?

hamersaw (changes requested):
A few nit comments. Otherwise this looks great! This is going to be so powerful.

hamersaw (approved):
Looks great! Will wait for one more approve before merging to try and stick with old idl requirements.

flytepropeller/pkg/controller/nodes/attr_path_resolver.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// convertInterfaceToLiteral converts the protobuf struct (e.g. dataclass) to literal
+func convertInterfaceToLiteral(nodeID string, obj interface{}) (*core.Literal, error) {
+
+	literal := &core.Literal{}
+
+	switch obj := obj.(type) {
+	case map[string]interface{}:
+		newSt, err := structpb.NewStruct(obj)
```
wild-endeavor, line 94: just one question.  if you just have a typing.Dict[str, DataClass], and you bind to one of the dataclasses.  you end up in this condition right?

What is this line for?  `newSt, err := structpb.NewStruct(obj)`

In fact in this case - line 22 above:
`tmpVal, exist = currVal.GetMap().GetLiterals()[attr.GetStringValue()]`
won't tmpVal already be a Literal?  can't you just return it?  it won't get returned right now, it'll fall into the condition on line 41.

ByronHsu: i guess you are talking about the case that where `bindAttrPath[count:]` is 0 and the `currVal` on line41 is a datatclass. Currently we convert convert struct to map, and convert it back to struct without doing other things because line 63 the loop will be skipped. 

adding a check about `len(bindAttrPath[count:]) == 0` on line 41 can circumvent the unnecessary work, but I think it's ok to keep it as-is because it should not be a large overhead.

wild-endeavor (approved):

PR #4259, opened 2023-10-18
Add retry for oauth metadata

squiishyy:


flyteadmin/auth/authzserver/metadata_provider.go
```diff
@@ -72,9 +80,14 @@ func (s OAuth2MetadataProvider) GetOAuth2Metadata(ctx context.Context, r *servic
...
-		response, err := httpClient.Get(externalMetadataURL.String())
+		response, err := sendAndRetryHttpRequest(httpClient, externalMetadataURL.String(), s.cfg.AppAuth.ExternalAuthServer.RetryAttempts, s.cfg.AppAuth.ExternalAuthServer.RetryDelayMilliseconds.Duration)
 		if err != nil {
-			return nil, err
+			if response != nil {
+				logger.Errorf(ctx, "Failed to get oauth metadata. Error code: %v. Err: %v", response.StatusCode, err)
+				return nil, flyteErrors.NewFlyteAdminError(codes.Code(response.StatusCode), "Failed to get oauth metadata.")
+			}
+			logger.Errorf(ctx, "Failed to get oauth metadata. Err: %v", err)
+			return nil, flyteErrors.NewFlyteAdminError(codes.Code(500), "Failed to get oauth metadata.")
```
pmahindrakar-oss, line 90: Can we also add the err in the error which basically shows the cause

flyteadmin/auth/authzserver/metadata_provider.go
```diff
@@ -72,9 +80,14 @@ func (s OAuth2MetadataProvider) GetOAuth2Metadata(ctx context.Context, r *servic
...
 		}
 
-		response, err := httpClient.Get(externalMetadataURL.String())
+		response, err := sendAndRetryHttpRequest(httpClient, externalMetadataURL.String(), s.cfg.AppAuth.ExternalAuthServer.RetryAttempts, s.cfg.AppAuth.ExternalAuthServer.RetryDelayMilliseconds.Duration)
 		if err != nil {
-			return nil, err
+			if response != nil {
+				logger.Errorf(ctx, "Failed to get oauth metadata. Error code: %v. Err: %v", response.StatusCode, err)
+				return nil, flyteErrors.NewFlyteAdminError(codes.Code(response.StatusCode), "Failed to get oauth metadata.")
+			}
```
pmahindrakar-oss, line 88: Are we expected to see a non nil response on err, I would consider not considering this case, we have special handling around responses in error cases

flyteadmin/pkg/async/shared.go
```diff
@@ -10,6 +11,24 @@ import (
 // RetryDelay indicates how long to wait between restarting a subscriber connection in the case of network failures.
 var RetryDelay = 30 * time.Second
 
+func RetryOnSpecificErrorCodes(totalAttempts int, delay time.Duration, f func() (*http.Response, error), IsErrorCodeRetryable func(*http.Response) bool) error {
```
pmahindrakar-oss, line 14: Can we instead reuse this one which comes from K8s utils. We are using this in flyteadmin already https://github.com/flyteorg/flyte/blob/master/flyteadmin/scheduler/executor/executor_impl.go#L100

squiishyy: Definitely can, is there any specific reason as to why you're wanting to use this one? We use the other retry functions in [shared.go](https://github.com/squiishyy/flyte/blob/03fc38a814552d869f67f2a9bd2bf2b0c3a49979/flyteadmin/pkg/async/shared.go) elsewhere in FlyteAdmin as well

flyteadmin/auth/authzserver/metadata_provider.go
squiishyy: Not expected! Just wanted to add an edge case here, happy to remove it though! I did not know we had special handling around responses in error cases :)

flyteadmin/pkg/async/shared.go
pmahindrakar-oss: Oh. For consitency across our repos , would be good to use one if they provide the same functionality. 
In cloud repo too we use the k8s retry util in grafana dataplane with usage of RetryOnConflict and hence was seeing if we could use the same

flyteadmin/auth/authzserver/metadata_provider.go
pmahindrakar-oss: In this particular case, you are getting the statusCode from the response so we can keep it. 
Also on L90 , 500 is not a valid grpc error code, I would use code.Internal there

flyteadmin/pkg/async/shared.go
squiishyy: Ah gotcha! makes sense, happy to migrate over to that :) will work on that now

squiishyy:
Closed in favor of: https://github.com/flyteorg/flyte/pull/4262

PR #4266, opened 2023-10-20
Ray logs persistence

jeevb:
## Describe your changes

Injects a logging sidecar to tail and expose RayJob logs to the container's stdout, and in turn, to a cluster's logging infrastructure. This will allow job logs to be persisted to Cloudwatch, Stackdriver, or other storage backend for perusal at a later time.

Also adds support for specifying multiple containers in a pod template for Ray tasks.

## Screenshots

Got this working with the following `fluent-bit.conf` config:
```
[SERVICE]
    Log_Level error
[INPUT]
    Name tail
    Path /tmp/ray/session_latest/logs/job-driver-*
    Refresh_Interval 1
[OUTPUT]
    Name file
    Match *
    File /dev/stdout
    Format template
    Template {log}
```

Note that the `stdout` output plugin could work here as well if structured logs are desired, but I went with just exposing logs as plaintext for now.

Sidecar stdout looks as follows:
```
> kubectl logs -f r9d1584b28b7d-taskraytask-0-raycluster-kksrv-head-fhm4q -c logs
Fluent Bit v2.1.10
* Copyright (C) 2015-2022 The Fluent Bit Authors
* Fluent Bit is a CNCF sub-project under the umbrella of Fluentd
* https://fluentbit.io

tar: Removing leading `/' from member names

2023-10-31 23:17:27,711	INFO worker.py:1329 -- Using address 10.42.0.11:6379 set in the environment variable RAY_ADDRESS
2023-10-31 23:17:27,712	INFO worker.py:1458 -- Connecting to existing Ray cluster at address: 10.42.0.11:6379...
2023-10-31 23:17:27,718	INFO worker.py:1633 -- Connected to Ray cluster. View the dashboard at http://10.42.0.11:8265
(compute_squared pid=186) Computing squared of 0...
(compute_squared pid=186) Computed squared of 0: 0
(compute_squared pid=85, ip=10.42.0.12) Computing squared of 1...
(compute_squared pid=186) Computing squared of 2...
(compute_squared pid=186) Computed squared of 2: 4 [repeated 2x across cluster] (Ray deduplicates logs by default. Set RAY_DEDUP_LOGS=0 to disable log deduplication, or see https://docs.ray.io/en/master/ray-observability/ray-logging.html#log-deduplication for more options.)
(compute_squared pid=85, ip=10.42.0.12) Computing squared of 3...
(compute_squared pid=186) Computing squared of 4...
(compute_squared pid=85, ip=10.42.0.12) Computed squared of 3: 9
[2023/10/31 23:18:44] [engine] caught signal (SIGTERM)
```


flyteplugins/go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -219,7 +219,40 @@ func buildHeadPodTemplate(container *v1.Container, podSpec *v1.PodSpec, objectMe
...
+			},
+		}
+		headPodSpec.Volumes = append(headPodSpec.Volumes, vol)
+	}
+
+	writeableVolMount := v1.VolumeMount{
+		Name:      "system-ray-state",
+		MountPath: "/tmp/ray",
+	}
+	primaryContainer.VolumeMounts = append(primaryContainer.VolumeMounts, writeableVolMount)
```
EngHabu, line 244: This block should be included in the `if` block, right? otherwise it'll double add that volumemount if one exists...

flyteplugins/go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -219,7 +219,40 @@ func buildHeadPodTemplate(container *v1.Container, podSpec *v1.PodSpec, objectMe
 
 	headPodSpec := podSpec.DeepCopy()
 
-	headPodSpec.Containers = []v1.Container{*primaryContainer}
+	// Ray logs integration
+	foundTmpRayVolMount := false
+	for _, vm := range primaryContainer.VolumeMounts {
+		if vm.MountPath == "/tmp/ray" {
```
EngHabu, line 225: Is that path hardcoded in ray driver?

EngHabu (commented):
LGTM! after adding config options to control that behavior as you suggested...

flyteplugins/go/tasks/plugins/k8s/ray/ray.go
jeevb: Yea this is wrong. I think we want to use info from an existing volume / volume mount where possible. I'll clean up.

flyteplugins/go/tasks/plugins/k8s/ray/ray.go
jeevb: Yea `/tmp/ray` is a fixed path. I'll make this a const.

EngHabu (approved):

PR #4295, opened 2023-10-24
Make fetching HttpClient optional from context in PKCE auth flow

pmahindrakar-oss:
## Describe your changes
Previous PR introduced authentication with proxy http Client but it added a hard dependency to fetch this from the context .
This causes regression in cases the context is not set with this object.

Current change makes it backward compatible by using the default http client in case the context is not set with it.

Tested this change locally by pulling this commit as saw authentication working fine in the broken environments

### DIFF
--- a/flyteidl/clients/go/admin/pkce/auth_flow_orchestrator.go
+++ b/flyteidl/clients/go/admin/pkce/auth_flow_orchestrator.go
@@ -67,6 +67,7 @@ func (f TokenOrchestrator) FetchTokenFromAuthFlow(ctx context.Context) (*oauth2.
 	// Pass along http client used in oauth2
 	httpClient, ok := ctx.Value(oauth2.HTTPClient).(*http.Client)
 	if !ok {
+		logger.Debugf(ctx, "using default http client instead")
 		httpClient = http.DefaultClient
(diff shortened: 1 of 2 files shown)


flyteidl/clients/go/admin/pkce/auth_flow_orchestrator.go
```diff
@@ -68,7 +67,7 @@ func (f TokenOrchestrator) FetchTokenFromAuthFlow(ctx context.Context) (*oauth2.
 	// Pass along http client used in oauth2
 	httpClient, ok := ctx.Value(oauth2.HTTPClient).(*http.Client)
 	if !ok {
-		return nil, errors.New("Unable to retrieve httpClient used in oauth2 from context")
+		httpClient = http.DefaultClient
```
eapolinario, line 70: can we leave a log saying that the default client is going to be used?

pmahindrakar-oss: Done

eapolinario (dismissed):

eapolinario (approved):
Thanks for adding the unit test.

PR #4310, opened 2023-10-26
Fix transient secret sync error handling

### LINKED ISSUE #4309: [BUG] Flyte task permanently marked as failed after transient `failed to sync secret` error
### Describe the bug

When launching a large number of tasks (~50000 with 500 max parrallelism) we sometimes get transient errors like `Error: failed to sync secret cache: timed out waiting` during pod initialisation. As soon as flytepropeller sees this in the pod events it marks the task as failed. However if we look at pod directly we see that it recovers and runs successfully. We can even see the output that it wrote to blob storage. 

The error visible in the flyteconsole is:
```
Error: failed to sync secret cache: timed out waiting for the condition
```

Looking at the flyteworkflow CRD the relevant part is:
```
    n1:
      TaskNodeStatus:
        pState: <some long string of characters that looked a bit like a secret, maybe>
        phase: 8
        psv: 1
        updAt: "2023-10-26T09:51:56.050539301Z"
      dynamicNodeStatus: {}
      error:
        code: ContainersNotReady|CreateContainerConfigError
        kind: USER
        message: 'containers with unready status: [primary]|failed to sync secret
          cache: timed out waiting for the condition'
      laStartedAt: "2023-10-26T09:48:33Z"
      lastUpdatedAt: "2023-10-26T09:51:56Z"
      message: 'containers with unready status: [primary]|failed to sync secret cache:
        timed out waiting for the condition'
      phase: 6
      queuedAt: "2023-10-26T09:48:33Z"
      startedAt: "2023-10-26T09:48:33Z"
      stoppedAt: "2023-10-26T09:51:56Z"
```

Relevant Pod events:
```
Events:
  Type     Reason                  Age                    From                Message
  ----     ------                  ----                   ----                -------
  Normal   Pulled                  6m13s                  kubelet             Successfully pulled image <image tag> in 82.521605ms (153.070494ms including waiting)
  Warning  Failed                  6m12s (x4 over 6m44s)  kubelet             Error: failed to sync secret cache: timed out waiting for the condition
  Normal   Pulling                 6m (x5 over 8m9s)      kubelet             Pulling image <image tag>
  Normal   Pulled                  6m                     kubelet             Successfully pulled image <image tag> in 74.073486ms (152.72983ms including waiting)
  Normal   Created                 6m                     kubelet             Created container primary
  Normal   Started                 6m                     kubelet             Started container primary
```

### Expected behavior

If the k8s pod succeeds then flyte should detect it as succeeded. I don't think it should matter if there are some transient errors during pod initialisation. 

Not sure I can provide a neat way to reproduce, but I believe I have been able to debug very precisely where the error is. 

A few features of our setup that might be relevant to reproducing:
1. Running in Azure kubernetes service
2. Tasks have a pod template that adds environment variables that reference the k8s secrets. 
3. Launching lots of pods at the same time.
 
### The problem:
We can see from the snippet above from the workflow CRD that this error is considered a `CreateContainerConfigError`. The relevant code is [DemystifyPending](https://github.com/flyteorg/flyte/blob/b30cb723afc7c966516c8460f7a82246b068bea4/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L573). If we look at this logic we can see that a few error types e.g. `CreateContainerError` allow a grace period on errors before flyte considers them to mean the task has failed. However `CreateContainerConfigError` does not allow such a grace period. 

Building a custom version of flytepropeller that adds a grace period on `CreateContainerConfigError`s has vastly reduced the frquency of the problem. I'm happy to make a PR if we think this is a sensible solution. 

### Screenshots

_No response_

- [X] Yes

- [X] Yes

Tom-Newton:
## Tracking issue

Closes #4309

## Describe your changes
- Add a new grace period configuration `create-container-config-error-grace-period` for `CreateContainerConfigError`s similar to `create-container-error-grace-period` for `CreateContainerError`.
- I was a bit torn on whether to return a retry-able error and whether to enable cleanup. 
  - Without cleanup I think I observed some issues where it was trying to make an invalid state change from FAILED -> SUCCEEDED when the grace period was exceeded but the pod continued to completion. 
  - Ultimately I decided to keep it as a permanent failure without cleanup. This means non-flaky failures can still be debugged by looking at the pod events. If we enabled cleanup I think it would delete the pod and make that debugging harder. 
  - I can solve my problem just by setting a long enough timeout so that all the flaky failures recover before flyte decides they have failed. 
- I set the default grace period to 0 so the default behaviour should be unchanged. 
- Adjust the error messages so they mention the grace period if there are failures after the grace period expires. 

- [x] I updated the documentation accordingly. - The new config has the same level of documentation as `create-container-error-grace-period` I don't think there is anything else I need to do. 
- [x] All new and existing tests passed. - I think so. Presumably the CI checks will be the definitive answer

## Note to reviewers
- I've never used golang before so I'm just pattern matching with the existing code


hamersaw:
@Tom-Newton this looks great! Could you fix the unit test (should be a string match update) and then I think we're good to merge.

You can run these from the root of the repo using `make -C flyteplugins test_unit` for the flyteplugins unit tests as an example.

Tom-Newton:
> Could you fix the unit test (should be a string match update)

Done

> You can run these from the root of the repo using `make -C flyteplugins test_unit` for the flyteplugins unit tests as an example.

Thanks. I hadn't worked out how to run all tests.

hamersaw (approved):

PR #4324, opened 2023-10-30
Provide a way to ignore portions of the signature for cache key calcu…

### LINKED ISSUE #3058: [Core feature] Provide a way to ignore portions of the signature for cache key calculation purposes

It's useful to skip some inputs for cache key calculation purposes

For example, let's say we have this task:
```
@task(cache=True, cache_version="v1")
def t(log_level: int, a: str) -> str:
  ...
```

According to the [docs](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/task_cache.html#how-does-caching-work), one of the inputs to cache key calculation is the task signature, but in the case of this example, it'd be great if we could ignore `log_level` as part of the cache key calculation.

We should be able to do something along the lines of:
```
@task(cache=True, cache_version="v1", ignore_input_vars=["log_level"])
def t(log_level: int, a: str) -> str:
    ...
```

This would essentially skip some of the parameters for cache key calculation purposes.

We have the ability to override the hashing mechanism used to translate python types into Flyte types, as described in https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/task_cache.html#caching-of-non-flyte-offloaded-objects.

One could use this idea and provide constant hashes for the arguments they want to ignore, for example:
```
def constant_function(x: int) -> str:
    return "const"

@task 
def t_produce_annotated_literals() -> Annotated[int, HashMethod(constant_function)]:
   log_level = ...
   return log_level

@task(cache=True, cache_version="v1")
def t(log_level: int, a: str) -> str:
   ...

@workflow
def wf() -> str:
  log_level = t_produce_annotated_literals()
  return t(log_level=log_level, a="some string")
```

Expose `ignore_input_vars` in the `@task` decorator and ensure the new interface is used during cache key calculation in both local and remote executions.

- [X] Yes

- [X] Yes

troychiu:
## Tracking issue

Closes #3058

## Describe your changes

1. add _cache_ignore_input_vars_ as a field of _TaskMetadata_
2. change the signature of _GenerateArtifactTagName_ to GenerateArtifactTagName(ctx context.Context, inputs *core.LiteralMap, cacheIgnoreInputVars *[]string)
3. when cacheIgnoreInputVars provided, skip all input variables in it for hash calculation

## Note to reviewers

The local version is here: https://github.com/flyteorg/flytekit/pull/1914


flyteidl/protos/flyteidl/core/tasks.proto
```diff
@@ -131,6 +131,9 @@ message TaskMetadata {
     // task creates a k8s Pod. If this value is set, the specified PodTemplate will be used instead of, but applied
     // identically as, the default PodTemplate configured in FlytePropeller.
     string pod_template_name = 12;
+
+    // cache_ignore_input_vars is the input variables that should not be included when calculating hash for cache.
+    repeated string cache_ignore_input_vars  = 13;
```
eapolinario, line 136: can you run `make generate` from the `flyteidl` subdirectory to produce the stubs?

troychiu: Pushed!

pingsutw (approved):

pingsutw (approved):

flytepropeller/pkg/controller/nodes/catalog/datacatalog/transformer.go
```diff
@@ -114,9 +115,21 @@ func generateTaskSignatureHash(ctx context.Context, taskInterface core.TypedInte
...
+		inputsAfterIgnore = &core.LiteralMap{Literals: make(map[string]*core.Literal)}
+		for name, literal := range inputs.Literals {
+			if slices.Contains(*cacheIgnoreInputVars, name) {
+				continue
+			}
+			inputsAfterIgnore.Literals[name] = literal
+		}
+	} else {
+		inputsAfterIgnore = inputs
+	}
```
hamersaw, line 131: Rather than creating a new `LiteralMap` here would it make more sense to pass the `cacheIgnoreInputVars` to the `HashLiteralMap` function. Then in [this code here](https://github.com/flyteorg/flyte/blob/475e9d07644c612150f09598cf1b5eef1f1fa640/flyteplugins/go/tasks/pluginmachinery/catalog/hashing.go#L66-L68) we just need to update to:
```
	for name, literal := range literalMap.Literals {
	        if !slices.Contains(cacheIgnoreInputVars, name) {
		    hashifiedLiteralMap[name] = hashify(literal)
		}
	}
```
This achieves the same outcome in 2 LoC rather than 12. I think it makes more sense to only create a new LiteralMap in one place as well.

hamersaw:
Closing as satisfied by https://github.com/flyteorg/flyte/pull/4618

Issue #4332, opened 2023-10-30
[BUG] Flyte tasks may fail if containerStatus is not present for primary container fast enough


hamersaw:
### Describe the bug

The [DeterminePrimaryContainerPhase](https://github.com/flyteorg/flyte/blob/a7fc66481a52a3428947f6eb4e72930448d3611d/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L731) function fails if the primary container name is not found in the `containerStatuses` field on the k8s pod. However, in scenarios where the k8s api server is under heavy load the `containerStatuses` field may not be fully populated even though the container does actually exist in the Pod. Before falling back to a permanent failure of the container is not found in the `containerStatuses` field, we should check if it exists in the `container` field first.

### Expected behavior

Flyte should wait until the `containerStatuses` field is populated with the primary container name.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #4339, opened 2023-10-31
Reporting running if the primary container status is not yet reported

hamersaw:
## Tracking issue
fixes https://github.com/flyteorg/flyte/issues/4332

## Describe your changes
If the primary container status is not found we check if the pod definition contains the containers name. In this scenario the container status has not yet been reported, but will be when it begins executing.

## Screenshots
_NA_

## Note to reviewers
_NA_


flyteplugins/go/tasks/plugins/k8s/pod/sidecar_test.go
```diff
@@ -862,6 +869,34 @@ func TestDemystifiedSidecarStatus_PrimaryMissing(t *testing.T) {
 	assert.Equal(t, pluginsCore.PhasePermanentFailure, phaseInfo.Phase())
 }
 
+// TODO @hamersaw
```
eapolinario, line 872: nit: what was that about?

eapolinario (approved):

PR #4348, opened 2023-11-01
Tune sandbox readiness checks to ensure that sandbox is fully accessi…

jeevb:
…ble when marked as ready

## Describe your changes
The envoy proxy on `sandbox-bundled` was missing a healthcheck, and was incorrectly marked as ready before the listening port was even open. This causes CI to occasionally fail if using operations that depend on a ready sandbox state.


eapolinario (approved):
this is great!

PR #4389, opened 2023-11-09
Handle all ray job statuses

EngHabu:
## Describe your changes

Ray operator sets a few different possible transient statuses as well as different cluster deployment statuses. The plugin doesn't handle all of them and silently attempts to make an illegal state transition instead of erroring. This change does two things:

- [X] Update code to handle all current known statuses
- [X] Fail hard if the job moves to an unknown status

### DIFF
--- a/flyteplugins/go/tasks/plugins/k8s/ray/ray.go
+++ b/flyteplugins/go/tasks/plugins/k8s/ray/ray.go
@@ -470,7 +470,7 @@ func (plugin rayJobResourceHandler) GetTaskPhase(ctx context.Context, pluginCont
 		return pluginsCore.PhaseInfoQueued(time.Now(), pluginsCore.DefaultPhaseVersion, "Scheduling"), nil
 	}
 
-	// Kuberay creates a Ray cluster first, and then submits a Ray job to the cluster
+	// KubeRay creates a Ray cluster first, and then submits a Ray job to the cluster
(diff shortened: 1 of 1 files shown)


flyteplugins/go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -489,12 +489,18 @@ func (plugin rayJobResourceHandler) GetTaskPhase(ctx context.Context, pluginCont
 			return pluginsCore.PhaseInfoFailure(flyteerr.TaskFailedWithError, reason, info), nil
 		case rayv1alpha1.JobStatusSucceeded:
 			return pluginsCore.PhaseInfoSuccess(info), nil
-		case rayv1alpha1.JobStatusPending, rayv1alpha1.JobStatusRunning:
+		case rayv1alpha1.JobStatusPending, rayv1alpha1.JobStatusRunning, rayv1alpha1.JobStatusStopped:
```
kumare3, line 492: Stopped sounds weird to be running

EngHabu: It's not currently being used in the operator... so I took a guess it's a temporary state (Succeeded/Failed are terminal)... hence why I marked it as Running. That said, I think it's ok for now to add it to the default handler of failing as an unknown/not handled state..

jeevb (approved):

PR #4390, opened 2023-11-09
Store failed execution in flyteadmin

iaroslav-ciupin:
## Describe your changes

When creating an execution fails, store execution model in database with corresponding error.


flyteadmin/pkg/manager/impl/execution_manager.go
```diff
@@ -862,10 +863,14 @@ func (m *ExecutionManager) launchExecutionAndPrepareModel(
 	})
 
 	if err != nil {
-		m.systemMetrics.PropellerFailures.Inc()
-		logger.Infof(ctx, "Failed to execute workflow %+v with execution id %+v and inputs %+v with err %v",
-			request, workflowExecutionID, executionInputs, err)
-		return nil, nil, err
+		if strings.Contains(err.Error(), "no node can run task") {
```
iaroslav-ciupin, line 866: instead of string parsing I suggest to introduce some kind of `RuntimeError` struct that will contain error details. We can type check against this type and it will denote that we should still save execution model to database, but with Failed phase and error details.

flyteadmin/pkg/errors/errors.go
```diff
@@ -140,3 +141,13 @@ func NewWorkflowExistsIdenticalStructureError(ctx context.Context, request *admi
 	}
 	return statusErr
 }
+
+// ExecutionRuntimeError is a special error that can be returned by plugins denoting that
+// execution failed during runtime and should still be saved to database
+type ExecutionRuntimeError struct {
```
andrewwdye, line 147: Is this specific to Create calls, and if so, should it be named accordingly, i.e., `CreateExecutionRuntimeError`? Otherwise I imagine most execution errors will be runtime errors (conceptually).

iaroslav-ciupin: I don't think it has to be specific to Create calls, word "Runtime" means it happened while actually executing, no matter that execution was triggered.

hamersaw: Instead of creating a new error type here and checking for `errors.As`, does it make sense to add a code to the FlyteAdmin error code for "BuildFailure" or something and just [wrapping the build error here](https://github.com/flyteorg/flyte/blob/feature%2Fopen-telemetry/flyteadmin/pkg/workflowengine/impl/k8s_executor.go#L37-L37). It seems a little weird to me that were just checking if the error contains an `flyteidl.core.ExecutionError` and using that to indicate the we failed to correctly build the workflow.

flyteadmin/pkg/repositories/transformers/execution.go
```diff
@@ -91,11 +93,24 @@ func CreateExecutionModel(input CreateExecutionModelInput) (*models.Execution, e
 	if input.Phase == core.WorkflowExecution_RUNNING {
 		closure.StartedAt = createdAt
 	}
+	if input.Error != nil {
+		closure.Phase = core.WorkflowExecution_FAILED
+		execError := &core.ExecutionError{
+			Code:    "Unknown",
+			Message: input.Error.Error(),
+			Kind:    core.ExecutionError_UNKNOWN,
+		}
```
hamersaw, line 102: If we choose to define the error as having an `ExecutionError` s there a reason we can't maintain it through to here?

flyteadmin/pkg/errors/errors.go
iaroslav-ciupin: sorry, I'm not following, what do you actually suggest to change? can you elaborate in more details. In Union Cloud, we have our own executor plugin which flyteadmin invokes during execution creation. Its shown in [this PR](https://github.com/unionai/cloud/pull/5406). I have wrapped `flyteidl.core.ExecutionError` because this is what needed to store failed execution in ExecutionClosure in database. And I think its responsibility of plugin to set explicitly code, message and kind from this `flyteidl.core.ExecutionError` rather than flyteadmin deducing it from obtained error.

hamersaw: IIUC the goal here is that every time the build process fails we write a record to the DB indicating that the build process fails. In this scenario, we are only performing this failure maintenance when the error contains an `flyteidlcore.ExecutionError`. This covers the case in [your PR](https://github.com/unionai/cloud/pull/5406/files) because you are explicitly throwing the correct error type. Alternatively, we could wrap [this error](https://github.com/flyteorg/flyte/blob/92f6549f8723f75a9d27d58e47e3bd79eb60b162/flyteadmin/pkg/workflowengine/impl/k8s_executor.go#L41-L41) similar to how we're [doing elsewhere with a code](https://github.com/flyteorg/flyte/blob/feature%2Fopen-telemetry/flyteadmin/pkg/workflowengine/impl/k8s_executor.go#L67-L67) that indicates a build error. Then every time that the build process fails we persist it in the DB.

There are plenty of other ways to handle this as well, but just wondering if we care about all failure scenarios, or just the one where you're explicitly throwing (which may be more accurately captured using error codes rather than types). If this is not the intended behavior, then disregard my comments. Also, I don't feel strongly about this, no need to over analyze.

flyteadmin/pkg/repositories/transformers/execution.go
iaroslav-ciupin: I refactored code to pass ExecutionError explicitly. Not sure if this is what you meant.

EngHabu (dismissed):

flyteadmin/pkg/errors/errors.go
iaroslav-ciupin: I refactored this to store all failed executions in database. Please take a look and let me know if this is what you had in your mind.

EngHabu (dismissed):

EngHabu (approved):

hamersaw (approved):

PR #4427, opened 2023-11-15
Add service monitor for flyte admin and propeller service

vraiyaninv:
## Describe your changes

Service monitors are a common way to auto-discover service targets dynamically in Prometheus.  

This change allows the creation of a service monitor for the admin service as well as the creation of a service and service monitor for the propeller.  

Propeller service allows discovering multiple instances of propeller pods and exporting its metrics. 

Admin service monitor and propeller service as well as its service monitor are disabled by default

Check all the applicable boxes

- [x]  I updated the documentation accordingly.
- [x]  All new and existing tests passed.
- [x]  All commits are signed-off. 

## Related PRs


charts/flyte-core/templates/admin/service-monitor.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  name: {{ template "flyteadmin.name" . }}
+  namespace: {{ template "flyte.namespace" . }}
+  labels:
+  {{- with .Values.flyteadmin.serviceMonitor.labels }}
+  {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  endpoints:
+    - interval: 60s
```
gdabisias, line 13: NIT: Do we want to expose the interval and scrapeTimeout as values that can be set under Values.flyteadmin.serviceMonitor?

charts/flyte-core/templates/propeller/service.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if and .Values.flytepropeller.enabled .Values.flytepropeller.service.enabled }}
```
gdabisias, line 1: Should this not be only enabled when serviceMonitor is enabled?

davidmirror-ops:
>Propeller service allows discovering multiple instances of propeller pods

This is particular to multi-cluster deployments correct?

vraiyaninv:
> > Propeller service allows discovering multiple instances of propeller pods
> 
> This is particular to multi-cluster deployments correct?

It is for a single cluster, when you have multiple pods with propeller manager. We could have a pod monitor as well, but for uniformity across services, I leaned towards creating a service and having a service monitor.

charts/flyte-core/templates/propeller/service.yaml
vraiyaninv: You can still have a service without servicemonitor so there is no harm in creating just the service. This keeps it flexible to have the service for any other ports which may not be related to prometheus port.

charts/flyte-core/templates/admin/service-monitor.yaml
vraiyaninv: Added.

charts/flyte-core/templates/propeller/service.yaml
gdabisias: 👍

davidmirror-ops (approved):

Issue #4422, opened 2023-11-14
[BUG] GPU tolerations are not correctly passed to MPI workers


gdabisias:
### Describe the bug

When defining an MPITask, the tolerations that are set as default for GPU usage in the deployment charts, are not applied to the created workers.
i.e. this configuration is not applied
```
configmap:
  k8s:
    plugins:
      k8s:
        "resource-tolerations":
          "nvidia.com/gpu":
            key: "flyte/GPU"
            operator: "Exists"
            effect: "NoSchedule"
```

This has also an additional bad side effect: the MPI launcher will end up on a GPU node, even if it could just run on a cpu node, wasting resources.

### Expected behavior

The GPU tolerations should be applied automatically to MPI workers when GPUs are requested.
Currently this can be fixed by passing a flytekit.PodTemplate to the MPI Task, but this makes the code very cumbersome 
e.g.
```
flytekit.PodTemplate(
        V1PodSpec(
            tolerations=[
                V1Toleration(
                    key="flyte/GPU",
                    operator="Exists",
                    effect="NoSchedule",
                ),
            ]
        )
    )
```

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

gdabisias:
Notice that this issue is similar to
https://github.com/flyteorg/flyte/issues/4378

jeevb:
Unable to reproduce on 1.10:
```
plugins:
  k8s:
    resource-tolerations:
      nvidia.com/gpu:
      - key: "flyte/GPU"
        operator: "Exists"
        effect: "NoSchedule"
```

Pod spec:
```
> kubectl get po fff3254f454e34faa82b-n0-0-launcher -o yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    cluster-autoscaler.kubernetes.io/safe-to-evict: "false"
  creationTimestamp: "2023-11-14T23:36:04Z"
  labels:
    domain: development
    execution-id: fff3254f454e34faa82b
    group-name: kubeflow.org
    interruptible: "false"
    mpi-job-name: fff3254f454e34faa82b-n0-0
    node-id: n0
    project: flytesnacks
    shard-key: "30"
    task-name: kfmpi-plugin-mpi-mnist-horovod-train-task
    training.kubeflow.org/job-name: fff3254f454e34faa82b-n0-0
    training.kubeflow.org/job-role: master
[...]

PR #4467, opened 2023-11-21
Correctly handle resource overrides in KF plugins

jeevb:
## Tracking issue

Fixes: https://github.com/flyteorg/flyte/issues/4422

## Describe your changes
This PR:
- Adds a helper to override interruptibility and task resources in `TaskExecutionContext` before passing to `ToK8sPodSpec` where the base pod spec is constructed, and hydrated with defaults and additional configuration (e.g. affinity, tolerations, etc.).
- Updates `dask` and `spark` plugins to use this new helper for overriding interruptibility of the driver pods
- Updates KF plugins (i.e. `mpi`, `pytorch` and `tensorflow`) to correctly handle resource overrides specified in the respective task configs. This allows for correctly applying resource-specific tolerations (e.g. GPU) to the pod spec.

Config:
```
plugins:
  k8s:
    resource-tolerations:
      nvidia.com/gpu:
      - key: "flyte/GPU"
        operator: "Exists"
        effect: "NoSchedule"
```

Task decorator:
```
@task(
    task_config=MPIJob(
        launcher=Launcher(
            replicas=1,
        ),
        worker=Worker(
            replicas=1,
            limits=Resources(cpu="2", mem="1000Mi", gpu="1"),
        ),
    ),
    retries=3,
    cache=True,
    cache_version="0.1",
    limits=Resources(cpu="2", mem="1000Mi"),
)
```

Launcher tolerations:
```
  tolerations:
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
```

Worker tolerations:
```
  tolerations:
  - effect: NoSchedule
    key: flyte/GPU
    operator: Exists
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
```


flyteplugins/go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
```diff
@@ -110,36 +106,14 @@ func (mpiOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx plu
...
-				kubeflowv1.MPIJobDefaultContainerName,
-				launcherReplicaSpec.GetImage(),
-				launcherReplicaSpec.GetResources(),
-				launcherReplicaSpec.GetCommand(),
-			)
-			if err != nil {
-				return nil, err
-			}
-			launcherReplica.RestartPolicy = common.ParseRestartPolicy(launcherReplicaSpec.GetRestartPolicy())
+		launcherReplicaSpec, err = common.ToReplicaSpecWithOverrides(ctx, taskCtx, kfMPITaskExtraArgs.GetLauncherReplicas(), kubeflowv1.MPIJobDefaultContainerName, true)
```
jeevb, line 109: Need to confirm if launcher pods should be treated as special and be non-interruptible.

flyteplugins/go/tasks/plugins/k8s/kfoperators/common/common_operator.go
```diff
@@ -254,27 +251,101 @@ func ParseRestartPolicy(flyteRestartPolicy kfplugins.RestartPolicy) commonOp.Res
...
+
+type allowsCommandOverride interface {
+	GetCommand() []string
+}
+
+func ToReplicaSpecWithOverrides(ctx context.Context, taskCtx pluginsCore.TaskExecutionContext, rs kfDistributedReplicaSpec, primaryContainerName string, isMaster bool) (*commonOp.ReplicaSpec, error) {
+	taskCtxOptions := []flytek8s.PluginTaskExecutionContextOption{}
+	// Master should always run as non-interruptible
+	if isMaster {
+		taskCtxOptions = append(taskCtxOptions, flytek8s.WithInterruptible(false))
```
jeevb, line 307: Note that "master" pods will be tagged as non-interruptible.

flyteplugins/go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
```diff
@@ -151,37 +125,19 @@ func (mpiOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx plu
 			"Invalid TaskSpecification, unsupported task template version [%v] key", taskTemplate.TaskTypeVersion)
 	}
 
-	if workerReplica.ReplicaNum == 0 {
+	if *workerReplicaSpec.Replicas <= 0 {
 		return nil, fmt.Errorf("number of worker should be more then 0")
```
eapolinario, line 129: worth fixing the typos? number of worker should be more then -> number of workers should be greater than

flyteplugins/go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
```diff
@@ -151,37 +125,19 @@ func (mpiOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx plu
...
 	}
-	if launcherReplica.ReplicaNum == 0 {
+	if *launcherReplicaSpec.Replicas <= 0 {
 		return nil, fmt.Errorf("number of launch worker should be more then 0")
```
eapolinario, line 132: ditto.

eapolinario (commented):
This looks great, thank you!

Re: whether kfoperators master pods should be non-interruptible, why do we have do consider those  as special? Is it just for consistency with the other k8s plugins (dask and spark have their driver/master pods as non-interruptible?

jeevb:
> This looks great, thank you!
> 
> Re: whether kfoperators master pods should be non-interruptible, why do we have do consider those as special? Is it just for consistency with the other k8s plugins (dask and spark have their driver/master pods as non-interruptible?

If it would be detrimental to lose the master, we should run them as non-interruptible. Same logic as dask and spark, yes.

flyteplugins/go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
```diff
@@ -58,34 +55,33 @@ func (mpiOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx plu
...
 		}
 
-		workerReplica.ReplicaNum = mpiTaskExtraArgs.GetNumWorkers()
-		launcherReplica.ReplicaNum = mpiTaskExtraArgs.GetNumLauncherReplicas()
+		replicaSpec, err := common.ToReplicaSpec(ctx, taskCtx, kubeflowv1.MPIJobDefaultContainerName)
+		if err != nil {
+			return nil, flyteerr.Errorf(flyteerr.BadTaskSpecification, "Unable to create replica spec: [%v]", err.Error())
+		}
+		launcherReplicaSpec = replicaSpec.DeepCopy()
+		// TODO (jeev): Is this even a valid configuration. Can there be more than 1
```
jeevb, line 75: Maintains parity for now, but will need to follow up in a subsequent investigation.

flyteplugins/go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
jeevb: No clear evidence to suggest one way or the other. This pod is responsible for launching the jobs on the MPI workers, so will give it special treatment for now.

hamersaw (dismissed):
lgtm. would feel more confident with a few more eyes here. @fg91 @yubofredwang mind taking a quick look?

pingsutw (dismissed):

flyteplugins/go/tasks/plugins/k8s/kfoperators/common/common_operator.go
fg91: I think for PyTorch this doesn't make sense as the entire pytorch job needs to be restarted when one of the workers is preempted. This, however, is not done by the master replica but the kubeflow training operator.
I think in general for pytorch jobs the differentiation between master and worker doesn't make sense. The fact that pytorch jobs with an elastic config don't have any master replicas at all anymore bespeaks that.

fg91 (changes requested):
I don't agree with the change that a pytorchjob master replica should always be non-interruptible. The master replica doesn't start the worker replicas or restart them in case they are preempted. If one of the worker replicas is preempted, all other pods including the master replica have to stop/be restarted. So there is no value in preventing the master replica from being preempted. Or am I overlooking something here?

I think differentiating between master and worker replicas doesn't make that much sense for pytorchjobs in the first place and [in the newer elastic pytorchjobs](https://github.com/kubeflow/training-operator/blob/master/examples/pytorch/elastic/imagenet/imagenet.yaml), there are no more master replicas at all.

Could you please revert this change?

---

I appreciate that you tagged me @hamersaw @jeevb since we rely on pytorch jobs a lot :)
I ran a flytepropeller with your changes in our staging cluster and apart from the non-interruptible master replica everything works as expected.

However, for pytorchjobs everything worked as expected also before. Do I understand it correctly that the large diff in `pytorch.go` comes from the fact that now the overriding of resources and resource/interruptible tolerations happens more centrally instead of being handled separately in every single plugin?

jeevb:
I went ahead and reverted forcing non-interruptibility on KF plugin masters. This keeps parity with the current implementation for now.

@fg91: The big change here was to apply resource overrides to before `ToK8sPodSpec` is called. This ensures that things like resource-based tolerations (maybe not relevant for you guys since GKE injects GPU tolerations automatically), and other additional pod configuration work correctly - that function is doing quite a bit. The remaining changes in the PyTorch plugin are just structural to make the code less repetitive and easier to maintain - lots of commonalities between the KF plugins.

fg91:
> I went ahead and reverted forcing non-interruptibility on KF plugin masters. This keeps parity with the current implementation for now.

Parity also for MPI? Asking because I know only for pytorch that the master shouldn't always be non-interruptible - I don't know for MPI ..

> @fg91: The big change here was to apply resource overrides to before `ToK8sPodSpec` is called. This ensures that things like resource-based tolerations (maybe not relevant for you guys since GKE injects GPU tolerations automatically), and other additional pod configuration work correctly - that function is doing quite a bit. The remaining changes in the PyTorch plugin are just structural to make the code less repetitive and easier to maintain - lots of commonalities between the KF plugins.

Sounds great! Out of curiosity, does this mean that resource tolerations are automatically handled correctly if another plugin is added because this is handled before the plugin builds the resource?

jeevb:
Yes keeping parity for MPI as well for now, since no one has been able to provide a convincing argument one way or the other. I don't know MPI well enough.

> Sounds great! Out of curiosity, does this mean that resource tolerations are automatically handled correctly if another plugin is added because this is handled before the plugin builds the resource?

This is still handled when the plugin is building the resource. The overrides in this case are derived from parsing the plugin configuration. These changes basically give plugins a simple mechanism to inject overrides and benefit from the helpers in `pluginmachinery`.

Issue #4285, opened 2023-10-24
[BUG] FlyteScheduler fired launchplans on redeployment unexpectedly


leonlnj:
### Describe the bug

On FlyteScheduler restart, launchplan were fired unexpectedly. 

In one redeployment on `2023-10-11T10:29:11Z`, 41 launchplans were fired before schedule time during init of the schdeuler.

Observation
- All affected launchplan are active
- All affected launchplan have never ran before prior to redeployment
- Launchplan that was registered together with the affected ones, were not affected if they have a prior run
- 41 launchplans were catching up from `2023-10-06 07:28:13.98091`.

Using one sample of affected launchplan with version `2cf10a7`
![image](https://github.com/flyteorg/flyte/assets/30390872/c8b9f2bc-fd3c-4112-970a-c6d88bcd7d3e)
![ss 2023-10-16 at 1 36 43 PM copy](https://github.com/flyteorg/flyte/assets/30390872/084a5181-597e-48db-8930-58949c817472)

Filtered logs with startup msg and sample launchplan
```
time="2023-10-11T10:29:11Z" level=info msg="Using config file: [/etc/flyte/config/admin.yaml /etc/flyte/config/db.yaml /etc/flyte/config/logger.yaml /etc/flyte/config/server.yaml]"
{"json":{"src":"service.go:71"},"level":"info","msg":"setting metrics keys to [project domain wf task phase tasktype runtime_type runtime_version app_name]","ts":"2023-10-11T10:29:12Z"}
{"json":{"src":"server.go:97"},"level":"info","msg":"Starting profiling server on port [10254]","ts":"2023-10-11T10:29:12Z"}
{"json":{"src":"database.go:171"},"level":"info","msg":"Set connection pool values to [{MaxOpenConnections:100 OpenConnections:1 InUse:0 Idle:1 WaitCount:0 WaitDuration:0s MaxIdleClosed:0 MaxIdleTimeClosed:0 MaxLifetimeClosed:0}]","ts":"2023-10-11T10:29:12Z"}
{"json":{"src":"client.go:63"},"level":"info","msg":"Initialized Admin client","ts":"2023-10-11T10:29:12Z"}
{"json":{"src":"start.go:54"},"level":"info","msg":"Successfully initialized a native flyte scheduler","ts":"2023-10-11T10:29:12Z"}
[...]

PR #4468, opened 2023-11-22
Fix the loop variable scheduler issue

pmahindrakar-oss:
## Tracking issue
https://github.com/flyteorg/flyte/issues/4285

## Describe your changes

All kudos to @leonlnj who discovered this bug in Gojek production deployment and helped debug this nasty loop variable issue which caused unnecessary production issue where schedules ran with incorrect times

Overall the issue that was observed was when Gojek redeployed there scheduler, scheduler ran its regular catchup routine to recover from the missed schedule time during the redeploy and found schedules it needed to trigger.
This is well tested behavior in the fleet and we never encountered any issues with it.

But in case of Gojek, it mysteriously ran the schedules which shouldn't have been run and also ran with incorrect catcup from time.

There is much detail in the attached issue. 
To summarize, 
* The lastExecTime for each active schedule is computed using its updatedAt timestamp and whats stored in the lastsnapshot for that schedule
* If the schedule is available in the snapshot and has timestamp greater than the updatedAt timestamp then we use that timestamp or else we assume schedule was updated recently and hence we should use that for catching up 
* Now consider the case where schedules dont have a snapshot and they will all use updatedAt timestamp, 
* What was happening though was the lastExecTime which was pointer to loop variable s.UpdatedAt , when the loop variable moved to next schedule, the lastExecTime being pointer used the new schedules UpdatedAt
* So essentially this value will eventually have last schedules UpdatedAt timestamp and in gojek's case it was a new schedule and this schedule's updatedAt timestamp was used as lastExecTime for knowing the catupFromTime for the schedule
* All the active schedules started then firing using the last scheudes updateAt timestamp

The unit test written by Gojek team nicely capture this behavior.
Have updated those tests with assertions and added the loop variable fix.

There is ongoing open golang issue to fix the gotcha which many go developers have complained or ran into https://github.com/golang/go/discussions/56010

Hopefully they fix that soon

### DIFF
--- a/flyteadmin/run.sh
+++ b/flyteadmin/run.sh
@@ -1,0 +1,2 @@
+dlv --listen=:2345 --headless=true --api-version=2 --accept-multiclient exec bin/flyteadmin serve -- --config  flyteadmin_config.yaml 
+#--server.kube-config ~/.kube/config
--- a/flyteadmin/run.sh
+++ b/flyteadmin/run.sh
@@ -1,2 +1,0 @@
-dlv --listen=:2345 --headless=true --api-version=2 --accept-multiclient exec bin/flyteadmin serve -- --config  flyteadmin_config.yaml 
-#--server.kube-config ~/.kube/config
--- a/flyteadmin/scheduler/core/gocron_scheduler.go
+++ b/flyteadmin/scheduler/core/gocron_scheduler.go
@@ -54,24 +54,24 @@ func (g *GoCronScheduler) GetTimedFuncWithSchedule() TimedFuncWithSchedule {
 func (g *GoCronScheduler) BootStrapSchedulesFromSnapShot(ctx context.Context, schedules []models.SchedulableEntity,
 	snapshot snapshoter.Snapshot) {
 	for _, s := range schedules {
-		s := s
+		schedule := s
 		if *s.Active {
 			funcRef := g.GetTimedFuncWithSchedule()
 			nameOfSchedule := identifier.GetScheduleName(ctx, s)
 			// Initialize the lastExectime as the updatedAt time
 			// Assumption here that schedule was activated and that the 0th execution of the schedule
 			// which will be used as a reference
-			lastExecTime := &s.UpdatedAt
+			lastExecTime := &schedule.UpdatedAt
 
 			fromSnapshot := snapshot.GetLastExecutionTime(nameOfSchedule)
 			// Use the latest time if available in the snapshot
-			if fromSnapshot != nil && fromSnapshot.After(s.UpdatedAt) {
+			if fromSnapshot != nil && fromSnapshot.After(schedule.UpdatedAt) {
 				lastExecTime = fromSnapshot
 			}
-			err := g.ScheduleJob(ctx, s, funcRef, lastExecTime)
+			err := g.ScheduleJob(ctx, schedule, funcRef, lastExecTime)
 			if err != nil {
 				g.metrics.JobScheduledFailedCounter.Inc()
-				logger.Errorf(ctx, "unable to register the schedule %+v due to %v", s, err)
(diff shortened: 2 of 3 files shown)


pmahindrakar-oss:
cc : @leonlnj

flyteadmin/scheduler/core/gocron_scheduler.go
```diff
@@ -54,6 +54,7 @@ func (g *GoCronScheduler) GetTimedFuncWithSchedule() TimedFuncWithSchedule {
 func (g *GoCronScheduler) BootStrapSchedulesFromSnapShot(ctx context.Context, schedules []models.SchedulableEntity,
 	snapshot snapshoter.Snapshot) {
 	for _, s := range schedules {
+		s := s
```
kumare3, line 57: Can you please rename it?

pmahindrakar-oss: done

flyteadmin/run.sh
```diff
@@ -0,0 +1,2 @@
+dlv --listen=:2345 --headless=true --api-version=2 --accept-multiclient exec bin/flyteadmin serve -- --config  flyteadmin_config.yaml
```
pingsutw, line 1: should we remove this file?

pingsutw (approved):

wild-endeavor:
I'm actually pretty lost... can you help me understand please?

https://goplay.tools/snippet/mNyKXEvyv4q

I understand that the pointer inside the iterator (`t` here, `s` in flyte code) doesn't change, cuz it's just one object.  But why does that matter?  the values that that pointer points to do change.

can you also add a comment where you do schedule := s?  feels like something i would just delete if i came across it.

leonlnj:
> I'm actually pretty lost... can you help me understand please?
> 
> https://goplay.tools/snippet/mNyKXEvyv4q
> 
> I understand that the pointer inside the iterator (`t` here, `s` in flyte code) doesn't change, cuz it's just one object. But why does that matter? the values that that pointer points to do change.
> 
> can you also add a comment where you do schedule := s? feels like something i would just delete if i came across it.

@wild-endeavor  Because the jobStore store the `lastExecTime` as a pointer, when the iterator move, the value of it changes as well. This cause the catchup function to use an unexpected `lastExecTime` that don't belong to it and possibly fire schedule that was earlier than the original schedule's `updatedAt` time. Hope this helps.

Screenshot 2023-11-24 at 12 06 50 PM
https://goplay.tools/snippet/_jD1NzTDZC6

wild-endeavor:
ohh we're saving it, got it, didn't see that.  thank you @leonlnj

flyteadmin/scheduler/core/gocron_scheduler.go
```diff
@@ -54,23 +54,24 @@ func (g *GoCronScheduler) GetTimedFuncWithSchedule() TimedFuncWithSchedule {
 func (g *GoCronScheduler) BootStrapSchedulesFromSnapShot(ctx context.Context, schedules []models.SchedulableEntity,
 	snapshot snapshoter.Snapshot) {
 	for _, s := range schedules {
+		schedule := s
```
wild-endeavor, line 57:
```suggestion
		// Copy the object to save to a new pointer since the pointer is saved later
		schedule := s
```

wild-endeavor:
@pmahindrakar-oss could you still remove that weird file?  i added a comment.

pmahindrakar-oss:
@wild-endeavor this might be better example to showcase the issue https://goplay.tools/snippet/AKUYYafpBh7

flyteadmin/run.sh
pmahindrakar-oss: I  use this for debugging using delve remote debugger. Removed it though and keeping it only for my local

pingsutw (approved):

PR #4469, opened 2023-11-22
Fix Kubeflow TF Operator `GetTaskPhase` Bug

Future-Outlier:
## Describe your changes
- Add a function `getReplicasCount` to give replica pointer with default value.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] All commits are signed-off.

## Setup Process
```python
image_spec = ImageSpec(
            packages=["flytekit",
                    "flytekitplugins-kftensorflow",
                    "tensorflow"],
            registry="localhost:30000")

# build-essential git
task_config = TfJob(
    worker=Worker(replicas=1),
    chief=Chief(replicas=0),
    ps=PS(replicas=0),
    evaluator=Evaluator(replicas=0),
)

@task(
    task_config=task_config,
    cache=True,
    requests=Resources(cpu="1"),
    cache_version="1",
    container_image=image_spec,
)
def my_tensorflow_task(x: int = 10, y: str = "abc") -> int:
    return x


if __name__ == "__main__":
    print(my_tensorflow_task(x=10, y="hello"))
```

## Screenshots

image
image

PR #4471, opened 2023-11-23
Kf operators use `GetReplicaFunc` (Error Handling)

Future-Outlier:
## Describe your changes
- move `GetReplicaFunc` to `common` package
- make mpi, pytorch, and tensorflow use `GetReplicaFunc`, for better error handling
- add error handling about resource type in pytorch and tensorflow  (mpi has done it before)

## Related PRs
https://github.com/flyteorg/flyte/pull/4469


pingsutw (approved):

hamersaw:
What is the error here that this is fixing?

Future-Outlier:
> What is the error here that this is fixing?

Theses 2 places
https://github.com/flyteorg/flyte/pull/4471/files#diff-46ab17e48b782b6e73f152b8e3930f07a9585d25846b05feb9710f67a9820d07R221-R224
https://github.com/flyteorg/flyte/pull/4471/files#diff-06c5063166d378b704d2394410b8394f3d120ca088a5367c299f6b3a5efe4ef9R224-R227

hamersaw:
> > 
> 
> Theses 2 places

Oh, the name indicates the issue is with parsing replica counts. IIUC these should be set by default, so just trying to make sure this function is necessary.

With regards to adding the type checking. I can see the advantage of being safe here, but the `GetTaskPhase` function is only called on the resource built by this plugin. So as long as it never creates something eradic there will not be an issue.

Future-Outlier:
> > > 
> > 
> > 
> > Theses 2 places
> 
> Oh, the name indicates the issue is with parsing replica counts. IIUC these should be set by default, so just trying to make sure this function is necessary.
> 
> With regards to adding the type checking. I can see the advantage of being safe here, but the `GetTaskPhase` function is only called on the resource built by this plugin. So as long as it never creates something eradic there will not be an issue.

Yes you are right!

PR #4474, opened 2023-11-23
Artifact protos and related changes

wild-endeavor:
This change introduces the artifact protobufs and related control plane changes. The actual service backing artifacts is still being worked on, and related Helm chart changes and such are not yet available.  

Admin changes have been feature-gated.  To turn on the changes (though again, they won't work without the backend service which is coming soon), add the following to your single binary configuration.

```
flyteadmin:
  featureGates:
    enableArtifacts: true
```

To enable the new cloudevents
```
cloudEvents:
  enable: true
  cloudEventVersion: v2
```

Summary of changes
* IDL changes
  * New artifact related identifiers in core.
  * New artifact related messages and service
  * Artifact related fields in Variable and Parameter
  * Metadata field in Literal for tracking
  * New cloud events with additional information
* Admin
  * Registration of launch plans, tasks, and workflows, will send signatures to the artifact service.
  * Execution manager will resolve any artifact queries before running an execution
* Others
  * Change and combine postgres error handling logic into flytestdlib and call from admin and data catalog
  * Add a sandbox-only event processor to flytestdlib which can only be used in single binary.


flyteadmin/pkg/async/cloudevent/implementations/cloudevent_publisher.go
```diff
@@ -102,20 +113,416 @@ func (p *Publisher) shouldPublishEvent(notificationType string) bool {
...
+		inputs, _, err = util.GetInputs(ctx, c.urlData, &c.remoteDataConfig,
+			c.storageClient, rawEvent.GetInputUri())
+		logger.Infof(ctx, "DEBUGART-2: input uri %v, %v", rawEvent.GetInputUri(), inputs)
+
+		if err != nil {
+			fmt.Printf("Error fetching input literal map %v", rawEvent)
+		}
+	} else if rawEvent.GetInputData() != nil {
+		inputs = rawEvent.GetInputData()
+		logger.Infof(ctx, "DEBUGART-1: input data %v, %v", rawEvent.GetInputData(), inputs)
```
eapolinario, line 321: I'm decreasing these log statements to debug.

flyteadmin/pkg/manager/impl/node_execution_manager.go
```diff
@@ -288,7 +288,8 @@ func (m *NodeExecutionManager) CreateNodeEvent(ctx context.Context, request admi
 	}
 
 	go func() {
-		if err := m.cloudEventPublisher.Publish(ctx, proto.MessageName(&request), &request); err != nil {
+		ceCtx := context.TODO()
```
eapolinario, line 291: why do we need a separate context?

flyteadmin/pkg/manager/impl/task_execution_manager.go
```diff
@@ -205,8 +206,9 @@ func (m *TaskExecutionManager) CreateTaskExecutionEvent(ctx context.Context, req
 	}
 
 	go func() {
-		if err := m.cloudEventsPublisher.Publish(ctx, proto.MessageName(&request), &request); err != nil {
-			logger.Infof(ctx, "error publishing cloud event [%+v] with err: [%v]", request.RequestId, err)
+		ceCtx := context.TODO()
```
eapolinario, line 209: ditto.

go.mod
```diff
@@ -6,13 +6,13 @@ require (
 	github.com/flyteorg/flyte/datacatalog v0.0.0-00010101000000-000000000000
 	github.com/flyteorg/flyte/flyteadmin v0.0.0-00010101000000-000000000000
 	github.com/flyteorg/flyte/flytepropeller v0.0.0-00010101000000-000000000000
-	github.com/flyteorg/flyte/flytestdlib v0.0.0-00010101000000-000000000000
+	github.com/flyteorg/flyte/flytestdlib v1.9.5
```
eapolinario, line 9: I'm reverting this change.

flyteadmin/pkg/manager/impl/node_execution_manager.go
eapolinario: we sync'd offline. The reason is that the original context was being cancelled.

eapolinario (approved):

PR #4489, opened 2023-11-27
Implemented simple echo plugin for testing

hamersaw:
## Tracking issue
_NA_

## Docs link
_NA_

## Describe your changes
Implements a simple echo plugin which copies the task inputs to outputs. This facilitates local testing at scale without creating a large number of k8s pods. Additionally, this can support future extensions to force retries, etc.

## Setup Process
Enable to plugin in flyte configuration. Note that this plugin is set to be the default, so ALL tasks without a specific plugin assigned will use it. Additionally, two default plugins may not be enabled, so the PodPlugin cannot be simultaneously used.

## Screenshots
_NA_

## Note to reviewers
_NA_

## Related PRs
_NA_


eapolinario (commented):
thoughts on naming this plugin package name `echo` instead of `testing`?

hamersaw:
> thoughts on naming this plugin package `echo` instead of `testing`?

So it is named `echo`, just in the `testing` package assuming there will be other plugins we specifically use for testing someday.

eapolinario (approved):

PR #4500, opened 2023-11-30
Agent metadata proto

pingsutw:
## Tracking issue


flyteidl/protos/flyteidl/admin/agent.proto
```diff
@@ -45,6 +46,10 @@ message CreateTaskRequest {
   string output_prefix = 3;
   // subset of runtime task execution metadata.
   TaskExecutionMetadata task_execution_metadata = 4;
+  // Secret to be passed to the agent.
+  // Key is the name of the secret in secret manager.
+  // Value is the actual secret value.
+  map<string, string> secret = 5;
```
EngHabu, line 52: I would make the value a message:
```protobuf
message Secret {
  string value = 1;
  // In the future:
  // string encrypted_value = 2;
  // string mount_path = 3;
}
```

EngHabu, line 52:
```suggestion
  map<string, string> secrets = 5;
```

flyteidl/protos/flyteidl/admin/agent.proto
```diff
@@ -64,6 +69,9 @@ message GetTaskRequest {
 // Response to get an individual task resource.
 message GetTaskResponse {
   Resource resource = 1;
+
+  // log information for the task execution
+  repeated core.TaskLog logs = 2;
```
EngHabu, line 74:
```suggestion
  repeated core.TaskLog log_links = 2;
```

flyteidl/protos/flyteidl/admin/agent.proto
```diff
@@ -86,5 +94,35 @@ message DeleteTaskRequest {
...
+// A message containing the agent metadata.
+message Agent {
+  // The name of the agent.
+  string name = 1;
+  // A list of secret in the secret manager.
+  repeated string secret_name = 2;
+  // The type of the task that the agent can handle.
+  string supported_task_type = 3;
+  // Indicates if this agent is a sync agent.
+  bool is_sync = 4;
```
EngHabu, line 108:
```suggestion
  // Name is the developer-assigned name of the agent.
  string name = 1;
  
  // SupportedTaskTypes are the types of the tasks that the agent can handle.
  string supported_task_types = 2;
  
  // IsSync indicates whether this agent is a sync agent. Sync agents are expected to return their results synchronously when called by propeller. Given that sync agents can affect the performance of the system, it's important to enforce strict timeout policies. An Async agent, on the other hand, is required to be able to identify jobs by an identifier and query for job statuses as jobs progress.
  bool is_sync = 3;
  
  // SecretNames is a list of secrets the agent requires to execute tasks.
  repeated string secret_names = 4;
```

PR #4506, opened 2023-11-30
Bug/abort map task subtasks

pvditt:
## Tracking issue

Currently, when a single MapTask subtask fails the other subtasks are not aborted as the task has already been set to a terminal state.

It is expected behavior for subtasks to abort on a single subtask failure for maptasks.

## Describe your changes

- On legacy MapTask/Array plugin failure, abort subtasks
- Add new phase for legacy MapTask/Array plugin to abort subtasks
- Persist aborted subtask states to Flyte admin

## Setup Process

Built locally w/ flyte-single-binary-local.yaml 

## Screenshots

Screenshot 2023-12-09 at 6 27 33 PM
Screenshot 2023-12-09 at 6 25 49 PM


hamersaw:
A very high-level thought here. And apologies for steering in the wrong direction, these changes (ie. eventing) were much more involved that I anticipated. Does it make more sense to [add another state to maptask](https://github.com/flyteorg/flyte/blob/performance%2Farraynode-subnode-transition-latency/flyteplugins/go/tasks/plugins/array/core/state.go#L26-L37) for something like `FailureCleanup`. Where, while the task is still technically in the `Running` phase we abort subtasks that are still running? Then we don't need to add eventing to the abort function, add `IsSubtaskAbort` to`ExternalResourceInfo`, and update admins event handling (ie. same phase with increased `TaskPhaseVersion`) for something that will soon be phased out.

pvditt:
> A very high-level thought here. And apologies for steering in the wrong direction, these changes (ie. eventing) were much more involved that I anticipated. Does it make more sense to [add another state to maptask](https://github.com/flyteorg/flyte/blob/performance%2Farraynode-subnode-transition-latency/flyteplugins/go/tasks/plugins/array/core/state.go#L26-L37) for something like `FailureCleanup`. Where, while the task is still technically in the `Running` phase we abort subtasks that are still running? Then we don't need to add eventing to the abort function, add `IsSubtaskAbort` to`ExternalResourceInfo`, and update admins event handling (ie. same phase with increased `TaskPhaseVersion`) for something that will soon be phased out.

@hamersaw 

Good call out - definitely agree that we should minimize changes outside of the array plugin. I also think leaving terminal states as final is probably a safer design decision. Aborting the subtasks while the task is still running would remove the added eventing in the abort function and remove the change in FlyteAdmin enabling the writes on already terminal phases. 

I think we will still need to update ExternalResourceInfo to persist subtask state correctly from running to aborted. I'll try to figure another way to persist this state, but I think we should show the correct state for subtasks given they are aborted.

pvditt:
@hamersaw to persist subtask abort states when a map task is manually terminated, I think we would need to keep the added buffered task eventing in the abort function since propellor won't go through the [plugin Handle method](https://github.com/flyteorg/flyte/blob/c6a765332aaf7d7bf2ffc1a6d85aaae6f4ec3068/flyteplugins/go/tasks/plugins/array/k8s/executor.go#L74).

We can avoid the changes to flyteadmin by having the [recorded phaseinfo ](https://github.com/flyteorg/flyte/blob/bug/abort-map-task-subtasks/flyteplugins/go/tasks/plugins/array/k8s/management.go#L377) set to running. This would work when a maptask is manually terminated or when a terminating subtasks when the task is still running.

Have that working locally but need to clean up some of the eventing/phase versioning - will push those changes up tomorrow.

hamersaw (approved):

Issue #4384, opened 2023-11-08
[BUG] FAIL_AFTER_EXECUTABLE_NODES_COMPLETE tries to execute nodes which should never start when using dynamic


fellhorn:
### Describe the bug

In a specific complex workflow with dynamic tasks and caching and the `FAIL_AFTER_EXECUTABLE_NODES_COMPLETE` workflow failure policy, we observed a truly peculiar bug:

A node which should never be tried to execute as it depends on an already failed (dynamic) task, is not correctly skipped but can't be started and fails after 50 retries with:

```
Workflow[...] failed. RuntimeExecutionError: max number of system retry attempts [51/50] exhausted. Last known status message: Failed to check Catalog for previous results: failed to read inputs when trying to query catalog: [READ_FAILED] failed to read data from dataDir [.../inputs.pb]., caused by: path:gs://.../inputs.pb: not found
```
from https://github.com/flyteorg/flyte/blob/feeda193f285116a207342d17b173217650f3b0a/flytepropeller/pkg/controller/nodes/cache.go#L96

---

A minimal example:

```py
from flytekit import WorkflowFailurePolicy, dynamic, task, workflow

CACHE = "0.0.25" # Bumping cache between retries
USE_CACHE = True # Bug only occurs with active caching

@task(cache=USE_CACHE, cache_version=CACHE)
def my_task(number: int, will_fail: bool) -> int:
    if will_fail:
        raise Exception("Failure")
    return number + 1


@dynamic
def dyn_task(will_fail: bool) -> int:
    return my_task(number=42, will_fail=will_fail)


@workflow(failure_policy=WorkflowFailurePolicy.FAIL_AFTER_EXECUTABLE_NODES_COMPLETE)
def main():
    result1 = dyn_task(will_fail=False)
	# This dyn_task is never started although it should be due to the failure policy
    my_task(number=result1, will_fail=False) # <- This is stuck in `Unknown`

    result2 = dyn_task(will_fail=True)
	# This my_task should never be executed as it depends on the dyn_task above which will fail
    my_task(number=result2, will_fail=True) # <- This is stuck in `Unknown` and causes the cache check error after 50 retries
```

![image](https://github.com/flyteorg/flyte/assets/26092524/a35281ea-c8a9-4c7c-a8cb-9684cc51b9a6)

### Expected behavior

```py
@workflow(failure_policy=WorkflowFailurePolicy.FAIL_AFTER_EXECUTABLE_NODES_COMPLETE)
def main():
[...]

PR #4524, opened 2023-12-04
fastcache should not cache lookup on node skip

hamersaw:
## Tracking issue
fixes https://github.com/flyteorg/flyte/issues/4384

Currently there will be a cache lookup failure when nodes are skipped. This is because the node inputs cannot be resolved, caused by an upstream node in the `failed` phase for example.

If a node is in the `predicatePhase` `skip` after the `preexecute` function this means that the node should be evaluated. We now predicate the fastcache cache lookup on not being in the skip phase, meaning the cache lookup will not be performed.

## How was this patch tested?
Tested using the 

### Setup process
_NA_

### Screenshots
_NA_

## Related PRs
_NA_

## Docs link
_NA_


hamersaw:
@fg91 thanks for diving so deeply into this!

> So my question is, do you know whether we should also prevent fastcache lookup on EPhaseFailed or should this be handled elsewhere?
Looking through the code I think that is exactly what we should be doing. Just to be very clear, your seeing that there is a failure to resolve inputs (which creates the input files for this node) and immediately after that failure we attempt to read that inputs file to check the cache. This is clearly a bug.

Looking at [this line](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/nodes/executor.go#L980-L980) (the check you linked), the easy fix is to add `&& p.GetPhase() != handler.EPhaseFailed` as you suggest. I'm wondering if we can consolidate these to `p.GetPhase() == EPhaseQueued"? because the `preExecute` function returns Queued whenever the inputs correctly resolve. Anything else would not require a cache lookup.

fg91:
> Just to be very clear, your seeing that there is a failure to resolve inputs (which creates the input files for this node) and immediately after that failure we attempt to read that inputs file to check the cache. This is clearly a bug.

I'll summarize in my words because I don't feel 100% confident I understand the exact order in which things are happening:

When I put a breakpont [here](https://github.com/flyteorg/flyte/blob/7d53a6eb5df1c0e73a5e844b4d90318083ed84ac/flytepropeller/pkg/controller/nodes/executor.go#L980), I see `failed to check the catalog cache [...] failed to read inputs when trying to query catalog: inputs.pb: not found`. Leaving the breakpoint at this line, when I then check [`p`](https://github.com/flyteorg/flyte/blob/7d53a6eb5df1c0e73a5e844b4d90318083ed84ac/flytepropeller/pkg/controller/nodes/executor.go#L967), I see `handler.PhaseInfo {p: EPhaseFailed ... PromiseAttributeResolveError: failed to read offloaded metadata for promise"}`.

This makes sense to me: We aren't able to fetch the offloaded literal and weren't able to prepare the `inputs.pb` so the cache lookup fails.

PR #4550, opened 2023-12-07
[caching] fix issues that execution metadata with overwrite cache is not written to artifacts table

ysysys3074:
When we execute workflow with overwrite cache flag and then run the execution again, "view source execution" is still linked to the old cached execution link instead of execution with overwrite cache which is confusing, so this PR is correct this behavior.
Screenshot 2023-12-07 at 9 23 51 AM

The root cause for this issue is new metadata from execution with overwrite cache is not passed to UpdateArtifactRequest, and artifacts table is not really updated with new serializedmetadata.

## How was this patch tested?
unit tests added
E2E tests are done in Linkedin's internal flyte platform so can not share the execution link.
screeshots:
1) first run with normal cache
Screenshot 2023-12-07 at 9 31 40 AM

2) seond run, it will pick cache from first run
Screenshot 2023-12-07 at 9 32 05 AM

3) second run, i chose overwrite cache button from UI
link to first execution - fopdnzo2lxkd5a

3) fourth run, it picks the cache from second run, link to the right execution
link to the third execution - f14xkrarmxyiws. (**it will link to first run before my code changes**)


datacatalog/pkg/manager/impl/artifact_manager.go
```diff
@@ -369,6 +379,10 @@ func (m *artifactManager) UpdateArtifact(ctx context.Context, request *datacatal
 
 	// update artifact in DB, also replaces/upserts associated artifact data
 	artifactModel.ArtifactData = artifactDataModels
+	logger.Infof(ctx, "ArtifactModel is %+v", artifactModel)
+	logger.Infof(ctx, "ArtifactModel ArtifactData is %+v", artifactModel.ArtifactData)
+	logger.Infof(ctx, "ArtifactModel SerializedMetadata is %+v", artifactModel.SerializedMetadata)
```
hamersaw, line 384: Is this logging necessary or debugging information? If anything it should be `Debugf`, but I would be OK removing as well.

eapolinario:
Can you rebase? The `tests / docs` CI check failure is unrelated, fixed in https://github.com/flyteorg/flyte/pull/4556.

datacatalog/pkg/manager/impl/artifact_manager.go
ysysys3074: thanks for suggestions, i have changed it to Debugf for debug purpose.

ysysys3074:
> Thanks for the PR! Can you add DCO signoff to all commits as well? This is necessary as we are a Linux Foundation project.

sure i have added that in my commit thanks!

ysysys3074:
> Can you rebase? The `tests / docs` CI check failure is unrelated, fixed in #4556.

yeah i just rebase, thanks

datacatalog/pkg/manager/impl/artifact_manager_test.go
```diff
@@ -639,16 +640,23 @@ func TestUpdateArtifact(t *testing.T) {
 					artifactKey.DatasetVersion == expectedArtifact.Dataset.Version
 			})).Return(mockArtifactModel, nil)
 
+		metaData := &datacatalog.Metadata{
+			KeyMap: map[string]string{"key2": "value2"},
+		}
+		serializedMetadata, err := transformers.SerializedMetadata(metaData)
```
hamersaw, line 646: looks like there is a lint err here because `err` is not used. maybe add `assert.Nil(...)` on the `err` here.

datacatalog/pkg/manager/impl/artifact_manager_test.go
```diff
@@ -701,6 +713,11 @@ func TestUpdateArtifact(t *testing.T) {
 		datastore := createInmemoryDataStore(t, mockScope.NewTestScope())
 		mockArtifactModel := getExpectedArtifactModel(ctx, t, datastore, expectedArtifact)
 
+		metaData := &datacatalog.Metadata{
+			KeyMap: map[string]string{"key2": "value2"},
+		}
+		serializedMetadata, err := transformers.SerializedMetadata(metaData)
```
hamersaw, line 719: also assert on error here.

datacatalog/pkg/manager/impl/artifact_manager.go
```diff
@@ -369,6 +379,10 @@ func (m *artifactManager) UpdateArtifact(ctx context.Context, request *datacatal
 
 	// update artifact in DB, also replaces/upserts associated artifact data
 	artifactModel.ArtifactData = artifactDataModels
+	logger.Debugf(ctx, "ArtifactModel is %+v", artifactModel)
+	logger.Debugf(ctx, "ArtifactModel ArtifactData is %+v", artifactModel.ArtifactData)
+	logger.Debugf(ctx, "ArtifactModel SerializedMetadata is %+v", artifactModel.SerializedMetadata)
```
hamersaw, line 384: if we're going to leave this, can it be a single log line rather than 3? The first line should print the entire `artifactModel`, then the next two are printing fields that should already be exposed in the first.

hamersaw:
closing in favor of https://github.com/flyteorg/flyte/pull/4617

PR #4554, opened 2023-12-07
Change artifacts HTTP paths

squiishyy:

Being more declarative about talking to `artifacts` service in the http paths


wild-endeavor:
wait why was this done to `artifacts`.  the base branch here should be `artifacts-shell`

PR #4535, opened 2023-12-06
Detect subNode phase updates to reduce evaluation frequency of ArrayNode

hamersaw:
## Tracking issue
_NA_

## Why are the changes needed?
Reduces the frequency of eventing and CRD updates to improve performance.

## What changes were proposed in this pull request?
Rather than incrementing the `TaskPhaseVersion` on the `ArrayNodeState` during every loop we detect if any of the subNode phases or task phases have changed. If so, we increment the `TaskPhaseVersion`. This vastly reduces the frequency with which `ArrayNodes` are evaluated, therefore improving performance on large concurrency executions.

## How was this patch tested?
Locally with pprof and otel.

### Setup process
_NA_

### Screenshots
_NA_

## Check all the applicable boxes 

- [x] I updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] All commits are signed-off.

## Related PRs
_NA_

## Docs link
_NA_

PR #4567, opened 2023-12-09
Supporting parallized workers in ArrayNode subNodes

hamersaw:
## Tracking issue
_NA_

Hooking maptask executions (ie. ArrayNode implementation) into the core Flyte workflow evaluator is more heavy than current maptask subNode evaluations. This often results in significant performance degradation between the two implementations over similar workloads. To ensure the success and adoption of ArrayNode, performance needs to be at least on par; and with this PR it becomes significantly better.

This PR introduces a worker pool of go routines to parallelize I/O bound work during subNode evaluations. Foremost this means k8s Pod creation and blobstore operations (ex. validating outputs, etc) can be done for multiple subNodes simultaneously. The performance improvements with just 10 workers show a 4x improvement over ArrayNode without parallelization and often >3x over the current maptask implementation (while including significantly more functionality). This work is done partly as a PoC to push parallelization of all node evaluations in FlytePropeller for more widespread performance improvements over large workflows.

## How was this patch tested?
Scale tested on EKS Flyte cluster, benchmarks to follow.

## Related PRs
https://github.com/flyteorg/flyte/pull/4535

## Docs link
_NA_


pingsutw (approved):
LGTM, looking forward to parallelization of all node evaluations

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -349,56 +407,85 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
...
-				for name, literal := range outputs.GetLiterals() {
-					appendLiteral(name, literal, outputLiterals, len(arrayNodeState.SubNodePhases.GetItems()))
-				}
+			gatherOutputsRequests = append(gatherOutputsRequests, gatherOutputsRequest)
+		}
+
+		outputLiterals := make(map[string]*idlcore.Literal)
+		workerErrorCollector := errorcollector.NewErrorMessageCollector()
+		for i, gatherOutputsRequest := range gatherOutputsRequests {
+			outputResponse := <-gatherOutputsRequest.responseChannel
```
EngHabu, line 472: Reading the code I was wondering why are we creating many "small" channels instead of one channel that all workers send on, ordering can be managed through an index field within the response itself.

The benefit of a single channel is that as soon as a worker frees itself up, it can start executing other nodes... the way it's setup now the worker that picks up the last item will wait for all items ahead of it to finish before it can pickup further work...

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -460,6 +547,18 @@
...
 // Setup handles any initialization requirements for this handler
 func (a *arrayNodeHandler) Setup(_ context.Context, _ interfaces.SetupContext) error {
+	// start workers
+	for i := 0; i < config.GetConfig().NodeExecutionWorkerCount; i++ {
+		worker := worker{
+			gatherOutputsRequestChannel: a.gatherOutputsRequestChannel,
+			nodeExecutionRequestChannel: a.nodeExecutionRequestChannel,
+		}
+
+		go func() {
```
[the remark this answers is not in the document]
hamersaw: Great point. I added this in the worker execution so that we can return the error, similar to how this is [done in workflow execution](https://github.com/flyteorg/flyte/blob/performance%2Farraynode-parallel-workers/flytepropeller/pkg/controller/handler.go#L131-L149).

flytepropeller/pkg/controller/nodes/array/handler.go
hamersaw: > the way it's setup now the worker that picks up the last item will wait for all items ahead of it to finish before it can pickup further work

I'm not sure this is accurate. For example:
```
package main

import (
	"fmt"
	"time"
)

type request struct {
	index int
	responseChannel chan struct {
		int
	} 
}

func main() {
	requestCount := 20
	workerCount := 4
	requestChannel := make(chan request)
	
	for i := 0; i < workerCount; i++ {
	go func(){
		for {
			request := <- requestChannel
			
			fmt.Printf("COMPLETED %d\n", request.index)
			request.responseChannel <- struct {
				int
			}{request.index}
		}
	}()
	}
	
	requests := make([]request, 0)
	for i := 0; i < requestCount; i++ {
		request := request{
			index: i,
			responseChannel: make(chan struct{
				int
			}, 1),
		}
		
		requests = append(requests, request)
		requestChannel <- request
	}

	time.Sleep(5 * time.Second)
	
	for i, request := range requests {
		response := <-request.responseChannel
		fmt.Printf("RESPONSE %d %d\n", i, response.int)
	}
}
```

Will run all 20 items (in varying orders) and then read responses. This is because the response channel is length 1, so response writing is non-blocking, rather will be buffered. So when submitting items the request channel has no length so that is blocking, but we will not proceed (with overall execution) until all items are completed so that's not very concerning to me.

My only concern with using a single channel is then we have to size it correctly to ensure non-blocking as you suggested, where rather than using the easy `1` we need to use `max-parallelism` potentially? And frequently allocate a 5k length channel where we use a significant subset. I suppose this is all dependent on the relative overhead of creating 5k 1 length channels vs unused space.

I've seen this model used elsewhere and was excited to use it, but certainly understand a single channel approach as well. If you feel strongly, happy to update. The ultimate hope is to use a similar model for overall node evaluation parallelization and also eventing (with a separate worker pool), so it would be easier to get some consensus for uniformity now.

jeevb (approved):

Issue #4560, opened 2023-12-08
[Docs] Monodocs: add configuration to show tag/release in the monodocs build output


cosmicBboy:
### Description

The purpose of this task is to make the monodocs build process more transparent by showing the git tag and commit associated with the `flytekit`, `flytesnacks`, and `flytectl` repos when the monodocs are built.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

cosmicBboy:
fixed by: https://github.com/flyteorg/flyte/commit/d5f0002d8e4c63d4111c5eb4fd9df917422ee6b8

PR #4571, opened 2023-12-11
add repo tag and commit associated with the build

cosmicBboy:
## Tracking issue
[_https://github.com/flyteorg/flyte/issues/<number>_](https://github.com/flyteorg/flyte/issues/4560)

To make the docs more transparent. This will make it clear which version of flytekit, flytesnacks, and flytectl were used to build a particular instance of the docs. The monodocs build process will use the latest tag available in the corresponding repo.

- Adds a custom `_templates/page.html` file that adapts the furo [page.html file](https://github.com/pradyunsg/furo/blob/main/src/furo/theme/furo/page.html)
- Adds a `name` attribute to the `import_projects.Project` dataclass
- Injects the necessary metadata into `html_context`

Screenshot 2023-12-11 at 6 51 47 PM

## How was this patch tested?

- Built locally
- Preview can be found here: https://flyte-next.readthedocs.io/en/monodocs-checkout-tag/


docs/_ext/import_projects.py
```diff
@@ -19,11 +19,13 @@
 class ImportProjectsConfig:
     clone_dir: str
     flytekit_api_dir: str
+    show_repo_tags: bool
```
thomasjpfan, line 22: Can we always display the tags and commits? This way, we do not need another option to configure it.

cosmicBboy: done

docs/.readthedocs.yaml
```diff
@@ -15,7 +15,7 @@ build:
     - conda info
     - conda env list
     - cat docs/conf.py
-    - cd docs && /home/docs/monodocs-env/bin/python -m sphinx -T -E -b html -d docs/_build/doctrees -D language=en . $READTHEDOCS_OUTPUT/html
+    - cd docs && SHOW_REPO_TAGS=1 /home/docs/monodocs-env/bin/python -m sphinx -T -E -b html -d docs/_build/doctrees -D language=en . $READTHEDOCS_OUTPUT/html
```
thomasjpfan, line 18: Since this is determined dynamically now:

```suggestion
    - cd docs && /home/docs/monodocs-env/bin/python -m sphinx -T -E -b html -d docs/_build/doctrees -D language=en . $READTHEDOCS_OUTPUT/html
```

docs/_templates/page.html
```diff
@@ -0,0 +1,227 @@
+{% extends "base.html" %}
```
thomasjpfan, line 1: Should this be in https://github.com/flyteorg/furo/blob/main/src/furo/theme/furo/page.html or here?

cosmicBboy: we're no longer using that fork of furo in the monodocs

docs/.readthedocs.yaml
cosmicBboy: weird, already removed this... re-applied the change.

pingsutw (approved):

Issue #4586, opened 2023-12-12
[Docs] Use the flytesnacks {tutorials, userguide, integrations}.md pages instead of flyte monodocs doc entrypoints


cosmicBboy:
### Description

Currently the monodocs has its own pages for the tutorials, userguide, and integrations pages, but this is not great because if we add an example in flytesnacks we'll have to update the toctree in the flyte repo. We want to auto-generate the toc from the original flytesnacks pages so that the contribution experience is smoother.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

cosmicBboy:
fixed by https://github.com/flyteorg/flyte/pull/4587

PR #4587, opened 2023-12-12
auto-generate toctree from flytesnacks index.md docs

cosmicBboy:
## Tracking issue
[_https://github.com/flyteorg/flyte/issues/<number>_](https://github.com/flyteorg/flyte/issues/4586)

To make the contribution experience for flytesnacks smoother.

Updated the `import_projects` sphinx extension so that it:
- introduce a custom `list-table-toc` directive that takes the contents of `list-table` and auto-generates a toctree.
- replaces `list-table` directive with `list-table-toc` in specific pages: `flytesnacks/{userguide, integrations, tutorials}`

## How was this patch tested?

- local `make docs` build.
- readthedocs preview: https://flyte-next.readthedocs.io/en/auto-gen-toc/


neverett:
Should this also include introduction.md, since it contains the Getting Started TOC?

cosmicBboy:
> Should this also include introduction.md, since it contains the Getting Started TOC?

There's an equivalent introduction page in flyte monodocs that serves as the main index file: https://flyte-next.readthedocs.io/en/latest/introduction.html

docs/_ext/import_projects.py
```diff
@@ -32,6 +37,44 @@ class Project:
...
+
+    has_content = True
+
+    def run(self) -> list:
+        return [self.parse()]
+
+    def parse(self):
+        """Parses the list-table and adds a toctree"""
+        toc = ""
+        for doc in re.findall(r"<(/flytesnacks/.+)>", self.block_text):
```
thomasjpfan, line 66: This directive is not generic and only works for refs that is prefixed with `flytesnacks`. Can this be documented in `ListTableToc`'s docstring?

docs/conf.py
```diff
@@ -325,12 +324,19 @@
     r"{ref}`bioinformatics <bioinformatics>`": r"bioinformatics",
     PROTO_REF_PATTERN: PROTO_REF_REPLACE,
     r"/protos/docs/service/index": r"/protos/docs/service/service",
+    r"<weather_forecasting>": r"</flytesnacks/weather_forecasting>",
```
thomasjpfan, line 327: If `flytesnacks` adds another ref in `docs/tutorials.md`, we would still need to add a new entry in `flyte`. If this is true, can we document this somewhere?

An alternative is to parse `tutorials.md` and friends for anything that is `<SINGLE_WORD>` and replace it with `<flytesnacks/SINGLE_WORD>`.

neverett:
> There's an equivalent introduction page in flyte monodocs that serves as the main index file: https://flyte-next.readthedocs.io/en/latest/introduction.html

Gotcha, yeah, I was wondering if we should remove that file from Flyte and build the Getting Started TOC from the file in Flytesnacks, since we'll be making changes to other parts of the Getting Started docs in the next few weeks and months. (We will probably move the Getting Started docs entirely into Flyte, but not for some time.)

docs/_ext/import_projects.py
cosmicBboy: I'll make this configurable

docs/conf.py
cosmicBboy: no we don't, the `weather_forecasting` page is really the only exception to this (tbh it should really be an example). In general all the docs should follow the `<auto_examples/...` pattern above.

docs/_ext/import_projects.py
cosmicBboy: I made `ListTableToc` more generic

docs/conf.py
thomasjpfan: I see `weather_forecasting` and `environment_setup` having this pattern. For my understanding, if one adds a `<mlops_setup>` in `docs/tutorials.md` do we need to add the following to `REPLACE_PATTERNS`?

```python
r"<mlops_setup>": r"<flytesnacks/mlops_setup>"
```

Or we want this to be an anti-pattern and so this should not happen.

thomasjpfan (approved):
I left a question at https://github.com/flyteorg/flyte/pull/4587/files#r1425624683, but otherwise LGTM

docs/conf.py
cosmicBboy: Yeah this is an anti-pattern that we need to fix more systematically.

Eventually we can can specify the `list-table-toc` directive in these files and assume that all refs in these directives follow the pattern `<flytesnacks/examples/...>`. Any ref outside of `list-table-toc` won't be treated specially.

Basically any entries listed under `list-table-toc` in `tutorials`, `userguide`, and `integrations` ought to follow the example template described here: https://docs.flyte.org/projects/cookbook/en/latest/contribute.html

thomasjpfan (approved):

Issue #1149, opened 2021-06-15
[Core Feature] Allow tasks/config to specify max queue/wait time


EngHabu:
**Motivation: Why do you think this is important?**
In cases when the underlying execution engine (AWS Batch, K8s, Spark, Hive, AWS EMR, GCP BigQuery... etc.) is having issues scheduling flyte workloads, sometimes the workload get stuck. While Flyte has a concept of timeout, it only measures the execution timeout overall. Which doesn't allow the users to express their tolerance for how much they can wait in a queue to get a task executing.

**Goal: What should the final outcome look like, ideally?**
Expose an additional `queue_timeout` flag that can be set at a global scope through configs, or at a task scope (ideally can also be on a project/domain/WF levels). And when flytepropeller detects that a task hasn't started executing for that period of time, it should just abort it.

flixr:
While it's nice that we can now configure a `pod-pending-timeout`, we have pretty much the opposite problem:
We thought that the timeout really refers to only the actual execution time (without pending time) as the docs suggest:
https://docs.flyte.org/en/latest/api/flytekit/generated/flytekit.TaskMetadata.html?highlight=timeout#flytekit.TaskMetadata
We can only set a meaningful execution timeout and it can happen that sometimes tasks are queued for quite a long time until a GPU node is available (which is what we want).
So IMHO we should change the current timeout to mean only the execution timeout (as per current docs). Or add a new param that sets only the execution timeout.

PR #4590, opened 2023-12-13
Feature/add pod pending timeout config

pvditt:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/1149_

Users aren't able to configure timeout for pending pods. This leads for pods/tasks stuck in pending to rely upon execution timeout to fail. 

PodPendingTimeout can help in situations when pods aren't getting executed due to resources not being available or when downstream execution engines are not able to schedule work.

Introduce a "pod-pending-timeout" plugin config to enable users to set pending/queued timeouts for tasks. This value defaults to 0 such that the current pending behavior does not change unless the config is set.

## How was this patch tested?

Added unit test [here](https://github.com/flyteorg/flyte/blob/17474d7fef1a0a9e816d198e3c9a9e316c2741f1/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper_test.go#L1460)

### Setup process
Configure plugins -> k8s -> pod-pending-timeout in flyte_single_binary.yaml
### Screenshots
Screenshot 2023-12-05 at 12 21 26 PM

(should be auto updated running script/generate_config_docs.sh) 


pvditt:
accidentally ruined https://github.com/flyteorg/flyte/pulls?q=is%3Apr+is%3Aclosed with a rebase.

@hamersaw:

> My only concern with this logic is if DemystifyPendingHelper returns an error and the podPendingTimeout has elapsed the later will enforce a RetryableFailure and mask the error returned. IMO it might make more sense to return the error immediately?

The function only ever returns nil for the error as did the DemystifyPending previously. The linter for some reason flags when the function is not getting exported, so I removed the error return param from the helper. I can also do that for the DemystifyPending function as that only ever returns nil errors as well.

hamersaw (approved):

PR #4598, opened 2023-12-14
monodocs - gracefully handle case when external repo doesn't contain tags: use current commit

cosmicBboy:
This PR elegantly handles the case when an external repo (flytekit, flytesnacks, flytectl) does not have tags due to a cloning a copy of the repo with a depth of 0. This fixes a docs build issue caused by the fact that github actions `checkout` will fetch the latest commit (since `fetch-depth: 0` by default), see: https://github.com/flyteorg/flytesnacks/actions/runs/7204216790/job/19625299915?pr=1315

### DIFF
--- a/docs/_ext/import_projects.py
+++ b/docs/_ext/import_projects.py
@@ -26,7 +26,6 @@ class ImportProjectsConfig:
     flytekit_api_dir: str
     source_regex_mapping: dict = field(default_factory=dict)
     list_table_toc: List[str] = field(default_factory=list)
-    dev_build: bool = False
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


docs/_ext/import_projects.py
```diff
@@ -145,12 +145,20 @@ def import_projects(app: Sphinx, config: Config):
...
 
-        # use the latest git tag when building docs
         if repo:
             tags = sorted(repo.tags, key=lambda t: t.commit.committed_datetime)
-            tag = tags[-1]
-            update_html_context(project, str(tag), str(tag.commit)[:7], config)
-            repo.git.checkout(str(tag))
+            if not tags:
+                # If tags don't exist just use the current commit. This occurs
+                # when the git repo fetch depth is 0.
```
thomasjpfan, line 152: Based on the [actions/checkout readme](https://github.com/actions/checkout#usage) the force depth is 1 by default, which gives the one commit. I think the underlying issue is that the tags are not fetched. 

```suggestion
                # If tags don't exist just use the current commit. This occurs
                # when the git repo is a shallow clone.
```

In any case, I think setting `commit=repo.head.commit` is the correct behavior for the development builds. For dev builds, it is more informative to see the actual commits that generated the docs.

cosmicBboy: updated the comment

thomasjpfan (approved):

PR #4600, opened 2023-12-14
using Abort phase and cleaning up best-effort

hamersaw:
Updating to:
- enable use of the `Abort` phase directly in `ExternalResourceInfo` rather than a bool flag
- Remove failing w/ cleanup, because we only need to attempt abort once. While the workflow is running this happens with the `AbortSubTask` `ArrayNodePhase` and otherwise this is attempted during `Abort`
- Removed eventing as `Abort` during finalize. This function is used to handle finalizer cleanup and resource deletion (if requested), so tasks should not be phase transitioned.


pvditt (approved):
Thanks for putting this up. Introducing a new phase cleans up a lot. Didn't think of not going down the Abort handler path during failure, but since we introduced aborting while running, I agree that we shouldn't attempt aborting again. 

I'll need to make some changes to unit tests, but can do that on the other branch.

Issue #4369, opened 2023-11-06
[BUG] Outputs of a subworkflow in a remote launch plan is handled incorrectly


honnix:
### Describe the bug

https://github.com/flyteorg/flyte/blob/9fe34dbace17037f860c736ba1b8f709ee6cf773/flytepropeller/pkg/controller/nodes/subworkflow/launchplan/admin.go#L283 may not receive full outputs, depending on how `storage.limits.maxDownloadMBs` is configured in flyteadmin, and it does not fail loudly when the outputs size is larger than `storage.limits.maxDownloadMBs`. What's worse is that flytepropeller happily stores the 0-sized protobuf file to be used as the inputs of the following task.

Currently the error message is impossible to understand by users and also very hard for flyte admins to investigate:
```
failed at Node[wxetj-n1]. BindingResolutionError: Error binding Var [wf].[asi], caused by: failed at Node[wxetj-n0]. OutputsNotFoundError: Outputs not found at [gs://.../wxetj-n0/data/0/outputs.pb]
```

### Expected behavior

flytepropeller should fail loudly to indicate that it could not receive expected outputs.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

honnix:
BTW it is also very confusing having two config items with overlapping purposes:
* `remoteData.maxSizeInBytes`
* `storage.limits.maxDownloadMBs`

honnix:
I think in general having smaller `storage.limits.maxDownloadMBs` values (in flyteadmin, flytepropeller, and datacatalog) configured than `max-output-size-bytes` does not make sense it would fail at different places.

hamersaw:
@honnix this fix is rather confusing given the plethora of configuration options. Do you know if you are using [signed urls](https://github.com/flyteorg/flyte/blob/yushang%2Foverwrite-cache%2Fiteration/flyteadmin/pkg/runtime/interfaces/application_configuration.go#L431-L431)? It seems the [proto support for this flow is deprecated](https://github.com/flyteorg/flyteidl/blob/6eff9b0be4f41b717b2d89697c7187a5216a0547/protos/flyteidl/admin/execution.proto#L307-L313), but none of the configuration options are labeled as such.

[...]

PR #4602, opened 2023-12-15
[BUG] Retry fetching subworkflow output data on failure

pvditt:
## Tracking issue
Closes: _https://github.com/flyteorg/flyte/issues/4369_

- Errors are not bubbled up when fetching a subworkflow's output fails. This causes for failed output fetches to be handled later on when reading nil output literals leading to unclear/confusing error messages.

~~- When remote store reads fail in [GetOutputs](https://github.com/flyteorg/flyte/blob/1bed29a93905f3745e2618f2f32cf77adea8c835/flyteadmin/pkg/manager/impl/util/data.go#L98) or [GetInputs](https://github.com/flyteorg/flyte/blob/1bed29a93905f3745e2618f2f32cf77adea8c835/flyteadmin/pkg/manager/impl/util/data.go#L32), errors are not bubbled up - returning the URL blob to the client instead. However, the URL blob is [deprecated](https://github.com/flyteorg/flyteidl/blob/6eff9b0be4f41b717b2d89697c7187a5216a0547/protos/flyteidl/admin/execution.proto#L307,L313). Sending a clear error status improves clarity.~~ Update: moving this to another PR
- [GetLimitMegabytes](https://github.com/flyteorg/flyte/blob/32cde08cdcccc3223c6c968ba9c8c7936788c922/flytestdlib/storage/stow_store.go#L267) could potentially differ between propeller and admin given certain deployment configuration potentially causing for the GetExecutionData call to fetch subworklow output data could always fail while succeed if called on propeller. There is ~[similar](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/nodes/executor.go#L522-L534) logic utilized in recoverInputs.

- Retry fetching subworkflow output data on propeller when GetExecutionData admin call fails. 
~~- Bubble up errors when fetching Input and Output for workflow, node and task executions instead of returning empty values.~~ Update: moving this to another PR
~~- Remove UrlBlob utilization when fetching Input and Output for workflow, node and task execution~~ Update: moving this to another PR

## How was this patch tested?
- Added/updated unit tests
- Ran subworkflows.py parent_workflow in flytesnack to see if [GetExecutionData](https://github.com/flyteorg/flyte/blob/9fe34dbace17037f860c736ba1b8f709ee6cf773/flytepropeller/pkg/controller/nodes/subworkflow/launchplan/admin.go#L266) + manually set GetExecutionData to fail to test if the remote store retry in propeller would work.


Future-Outlier:
Amazing PR, I am bothered by this problem ...
I will test this.

Future-Outlier:
The code I test.
```python

@task()
def t1():
    return

@workflow()
def wf():
    sensor(path="/tmp") >> t1()

if __name__ == "__main__":
    wf()
```
Screenshots

Before
image
After
image

Future-Outlier:
@pvditt Can you make the example I gave above not an error?
I think it is okay that we don't have output for tasks like sensors.

pvditt:
@Future-Outlier 

Thank you for looking into this. I am unable to re-create the error running the code snippet you provided. I think the error displayed in your screenshots might be related to https://github.com/flyteorg/flyte/pull/4594 instead of any changes with this PR.

Future-Outlier:
> @Future-Outlier
> 
> Thank you for looking into this. I am unable to re-create the error running the code snippet you provided. I think the error displayed in your screenshots might be related to #4594 instead of any changes with this PR.

Yes, you are right XD!
I just tested it in the master branch, thank you very much!
image

hamersaw:
Please correct me here, but in the scenario where the [maxSizeInBytes](https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/runtime/interfaces/application_configuration.go#L459-L459) configuration is smaller than the file size this will still transparently return empty values because [shouldFetchData](https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/manager/impl/util/data.go#L22-L25) is never called. Am I understanding this correctly?

pvditt:
> Please correct me here, but in the scenario where the [maxSizeInBytes](https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/runtime/interfaces/application_configuration.go#L459-L459) configuration is smaller than the file size this will still transparently return empty values because [shouldFetchData](https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/manager/impl/util/data.go#L22-L25) is never called. Am I understanding this correctly?

@hamersaw 

The [urlBlob the shouldFetchOutputData](https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/manager/impl/util/data.go#L107) function is checking against is only set if remoteDataConfig.SignedURL.Enabled = true. As mentioned in the comments [here](https://github.com/flyteorg/flyte/blob/14915dcb1cbde884c6b8d514114dc48cf1342fbf/flyteadmin/pkg/manager/impl/util/data.go#L123) it is intentional to not bubble up errors on remote storage reads to instead pass the urlBlob. I believe this URL blob flow is [deprecated](https://github.com/flyteorg/flyteidl/blob/6eff9b0be4f41b717b2d89697c7187a5216a0547/protos/flyteidl/admin/execution.proto#L307,L313) as well.

Given remoteDataConfig.SignedURL.Enabled = false, storageClient.ReadProtobuf will be called unless [len(outputURI) is not greater than 0](https://github.com/flyteorg/flyte/blob/14915dcb1cbde884c6b8d514114dc48cf1342fbf/flyteadmin/pkg/manager/impl/util/data.go#L28).

However, this is confusing. I can clean up the functions to make it more clear.

pvditt:
@hamersaw 

I added an [issue](https://github.com/flyteorg/flyte/issues/4632) to cleanup the urlblob usage separately and will clean that up across propellor, admin, idl, and console. With that, we can leave the current handling of the SignedUrl.Enabled = True w/ returning a UrlBlob as is to keep. It doesn't seem that the UrlBlob is utilized in propellor nor console + it's intended usage was in console and removed in this [PR](https://github.com/flyteorg/flyteconsole/pull/391). 

As for adding retries in propellor on failed [GetExecutionData](https://github.com/flyteorg/flyte/blob/d981a96fe86feff4dec9caa0cf5b603e7dcd8648/flytepropeller/pkg/controller/nodes/subworkflow/launchplan/admin.go#L266-L283) calls, does it make sense to let the errors bubble up and just count against system retries?

pvditt:
@hamersaw I removed the FlyteAdmin changes from this PR. Will open up a house keeping PR with those changes.

PR #4604, opened 2023-12-15
Update flyte-binary values

davidmirror-ops:
Making the default values leaner, less opinionated, more informational and making sure they work OOB


charts/flyte-binary/eks-production.yaml
```diff
@@ -1,35 +1,42 @@
 configuration:
   database:
+    username: postgres
     password: <DB_PASSWORD>
     host: <RDS_HOST_DNS>
-    dbname: app
+    dbname: flyteadmin
```
jeevb, line 6: I believe datacatalog uses it too. How about just “flyte”?

charts/flyte-binary/eks-production.yaml
```diff
@@ -1,35 +1,42 @@
 configuration:
```
jeevb, line 1: Maybe we should just rename this `eks-starter.yaml`.

jeevb (approved):

PR #4606, opened 2023-12-15
Don't use experimental readthedocs build.commands config

cosmicBboy:
This PR changes the monodocs readthedocs config so that it uses the customizeable readthedocs build steps instead of the experimental `build.commands` config key. This is because there are some issues with correctly rendering the readthedocs flyout menu: https://github.com/readthedocs/readthedocs.org/issues/10955

Preview: https://flyte-next.readthedocs.io/en/monodocs-readthedocs-config/

image

vs.

https://flyte-next.readthedocs.io/en/latest/ (readthedocs inserted dropdown is missing)
image

### DIFF
--- a/docs/.readthedocs.yaml
+++ b/docs/.readthedocs.yaml
@@ -10,9 +10,9 @@ build:
     python: "mambaforge-22.9"
   jobs:
     post_install:
-      - conda install --name $READTHEDOCS_VERSION -c conda-forge readthedocs-sphinx-ext -y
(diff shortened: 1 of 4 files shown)


cosmicBboy:
@thomasjpfan I opted for [your solution](https://github.com/flyteorg/flyte/pull/4553#discussion_r1421030484) to this, because readthedocs adds additional configuration to the user-supplied `conf.py` file, and not doing this will cause issues such as not showing the readthedocs version dropdown.

Issue #4284, opened 2023-10-24
FlyIn Flytekit Plugin


ByronHsu:
### Motivation: Why do you think this is important?

When
1. The tasks fail and users want to debug them
2. Users start writing the task from scratch and want to develop in a container in k8s cluster (likely because only the cluster side has GPU)

Flyte users might `time.sleep` the task, `ssh` inside, and develop/debug. However, using raw terminal to do so is cumbersome and unintuitive. For example, in multi-level directories structure or complex code base that need heavy navigation, users may feel pretty frustrated with the terminal environment.

I propose that we can add a decorator as a flyte plugin and enable users to turn any tasks to vscode server in just one line.

### Goal: What should the final outcome look like, ideally?

`@vscode` provides an simple way for users to run the task as a interactive vscode server with **any image**. `@vscode` is a decorator which users can put within `@task` and user function.

```python
@task
@vscode
def train():
    ...
```

By doing so, the task will **install vscode dependencies** and **run a vscode server instead of the user defined functions**. Also, it starts a exit handler which keeps polling the status of the heartbeat file and terminate if the server is idle for a set duration. This is very important because the pod is most likely using several GPU's, so we want to free them for other users.

Heartbeat file can monitor [the active connection](https://coder.com/docs/code-server/latest/FAQ#what-is-the-heartbeat-file) to the code-server

> As long as there is an active browser connection, code-server touches ~/.local/share/code-server/heartbeat once a minute. If you want to shutdown code-server if there hasn't been an active connection after a predetermined amount of time, you can do so by checking continuously for the last modified time on the heartbeat file. If it is older than X minutes (or whatever amount of time you'd like), you can kill code-server.

```python

def vscode(
    _task_function: Optional[Callable] =  None,
    max_idle_seconds: Optional[int] = 3600 * 24 * 2, # 2 days
    port: Optional[int] = 8001,
    enable: Optional[bool] = True,
    code_server_remote_path: str = "XXX",
    # The untarred directory name may be different from the tarball name
    code_server_dir_name: str = "YYY",
    plugins_remote_paths: List[str] = [
        "ZZZ",
        "KKK"
    ],
    pre_execute: Optional[Callable] = None,
):
    """
    vscode decorator modifies a container to run a VSCode server:
    1. Overrides the user function with a VSCode setup function.
    2. Download vscode server and plugins from remote to local.
    3. Launches and monitors the VSCode server.
    4. Terminates if the server is idle for a set duration.

    Parameters:
[...]

PR #4610, opened 2023-12-18
Add Flyin propeller config

eapolinario:
## Tracking issue
https://github.com/flyteorg/flyte/issues/4284

We want a way to expose a log link in the case of tasks that use the flyin plugin

Expose a `flyin-enabled` pflag and ensure that only tasks that fill the task template with the appropriate config get a Flyin log.

## How was this patch tested?

Unit tests and a smoke test with single-binary running locally.

### Screenshots

Ran this workflow:
Screenshot 2023-12-18 at 12 30 54 PM

The task that's using the `@vscode` decorator:
Screenshot 2023-12-18 at 12 30 36 PM

And the other task:
Screenshot 2023-12-18 at 12 30 43 PM

### DIFF
--- a/flyteplugins/go/tasks/logs/config.go
+++ b/flyteplugins/go/tasks/logs/config.go
@@ -28,6 +28,9 @@ type LogConfig struct {
 	StackdriverLogResourceName string              `json:"stackdriver-logresourcename" pflag:",Name of the logresource in stackdriver"`
 	StackDriverTemplateURI     tasklog.TemplateURI `json:"stackdriver-template-uri" pflag:",Template Uri to use when building stackdriver log links"`
 
+	IsFlyinEnabled   bool                `json:"flyin-enabled" pflag:",Enable Log-links to flyin logs"`
+	FlyinTemplateURI tasklog.TemplateURI `json:"flyin-template-uri" pflag:",Template Uri to use when building flyin log links"`
+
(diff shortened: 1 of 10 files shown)


hamersaw (approved):

flyteplugins/go/tasks/logs/config.go
```diff
@@ -28,6 +28,9 @@ type LogConfig struct {
 	StackdriverLogResourceName string              `json:"stackdriver-logresourcename" pflag:",Name of the logresource in stackdriver"`
 	StackDriverTemplateURI     tasklog.TemplateURI `json:"stackdriver-template-uri" pflag:",Template Uri to use when building stackdriver log links"`
 
+	IsFlyinEnabled   bool                `json:"flyin-enabled" pflag:",Enable Log-links to flyin logs"`
+	FlyinTemplateURI tasklog.TemplateURI `json:"flyin-template-uri" pflag:",Template Uri to use when building flyin log links"`
+
```
EngHabu, line 33: Why is this specific to flyin? This should be a map of string & tasklog.TemplateLogPlugin... task templates should be able to enable any number of them

eapolinario: We sync'd offline and the idea is to make this more extensible by providing a way for tasks to dynamically set log links. I'll be sending a follow up PR shortly.

PR #4624, opened 2023-12-19
Clear past errors from workflow state

Tom-Newton:
## Tracking issue
https://github.com/flyteorg/flyte/issues/4569

Reduce un-needed information stored in etcd when using `failure_policy=WorkflowFailurePolicy.FAIL_AFTER_EXECUTABLE_NODES_COMPLETE`. This allows flyte to scale to larger workflows before hitting etcd size limits.

- Clears old errors whenever new ones occur 
  - This is controlled by the `node-config.enable-cr-debug-metadata` config option. Set this to true to restore the previous behaviour. 
  - Behaviour is only changed when running workflows with `FAIL_AFTER_EXECUTABLE_NODES_COMPLETE`. Without this the workflow will fail as soon as there is one failure so there can never be more than one error regardless of this PR. 
- Update docstring for `enable-cr-debug-metadata` config option. 
- Update `TestWorkflowExecutor_HandleFlyteWorkflow_Failing` to have sub test cases for combinations of `FAIL_AFTER_EXECUTABLE_NODES_COMPLETE` and `enable-cr-debug-metadata`

## How was this patch tested?
Updated unittests
I have been running something very similar to this in our prod deployment for some time.

- [x] I updated the documentation accordingly - I would like to make a follow up PR to update https://docs.flyte.org/en/latest/deployment/configuration/performance.html#improving-etcd-performance

## Related PRs
Follow up to https://github.com/flyteorg/flyte/pull/4596


Tom-Newton:
It looks like I'm missing a little bit of code coverage. I'll try to find some time to fix that.

flytepropeller/pkg/controller/nodes/executor.go
```diff
@@ -298,6 +299,10 @@ func (c *recursiveNodeExecutor) handleDownstream(ctx context.Context, execContex
 				// If the failure policy allows other nodes to continue running, do not exit the loop,
 				// Keep track of the last failed state in the loop since it'll be the one to return.
 				// TODO: If multiple nodes fail (which this mode allows), consolidate/summarize failure states in one.
+				if executableNodeStatusOnComplete != nil {
+					c.nodeExecutor.Clear(executableNodeStatusOnComplete)
```
hamersaw, line 303: I think it makes sense to add the `enableCRDebugMetadata bool` argument here to the `ClearExecutionError` function on the `MutableNodeStatus` interface. Then this call more closely reflects the `UpdatePhase` call above and we can remove adding a `Clear` function to the `nodeExecutor` struct and similarly the `NodeExecutor` interface. Thoughts?

hamersaw (commented):
If I understand this logic the goal is to delete all the errors but the last one right? I'm trying to wrap my head around the logic of iterating over downstream nodes but need to dive deeper. Is there determinism in the ordering? Or is there a scenario here where we delete all of the error messages? For example, if we have two nodes (`n0` and `n1`) if the first time we iterate over these the order is `n0`, `n1` then we clear the error from `n0` if the second time we iterate `n1`, `n0` then we clear the error from `n0` and just cleared all of our errors.

Tom-Newton:
> If I understand this logic the goal is to delete all the errors but the last one right? I'm trying to wrap my head around the logic of iterating over downstream nodes but need to dive deeper. Is there determinism in the ordering? Or is there a scenario here where we delete all of the error messages? For example, if we have two nodes (n0 and n1) if the first time we iterate over these the order is n0, n1 then we clear the error from n0 if the second time we iterate n1, n0 then we clear the error from n0 and just cleared all of our errors.

That's a good question. I guess I assumed that `stateOnComplete` would only get updated when there are new errors and not get randomly set to a different error on each round. I think that assumption is correct given that we've been using this in prod and nobody has noticed any problems, but I will need to read the code again to understand how that is achieved.

flytepropeller/pkg/controller/nodes/executor.go
Tom-Newton: I've never used golang before I started using flyte so I have little opinion on how the interfaces are organised. I will try to implement it as you suggested.

eapolinario:

Sovietaced:
Closing for now.

PR #4637, opened 2023-12-22
Feat: Add option in K8s plugin config to inject user identity as pod label

fg91:

@ByronHsu recently implemented middleware to inject the user identity into the flyte workflow's `ExecutionSpec`: https://github.com/flyteorg/flyteadmin/pull/549

Building on this, we would like to inject the user identity (not credentials, just the information who started the execution) into the task pod as an additional label in order to enable the following user stories:

* As a platform engineer, I would like to know which user started which Pod.
* As a platform engineer, I would like attribute compute cost to users (e.g. by [activating GKE cost export into big query](https://cloud.google.com/kubernetes-engine/docs/how-to/cost-allocations) which associates compute cost with pod labels).
* As a platform user/ML engineer, I would like my task executions to automatically log training progress under my account in an experiment tracking server.

* <s>I propose to add a new flag to the k8s plugin config, specifying whether the execution identity is injected as a label. The default is `false` to make this opt in.</s>
* I propose to inject the additional execution identity label in [`newTaskExecutionMetadata`](https://github.com/flyteorg/flyte/blob/853ebe4c32773b7c8815cd3c04290a6fd9caeb58/flytepropeller/pkg/controller/nodes/task/k8s/task_exec_context.go#L46) where we also inject labels and annotations for secrets.

  Because of the following reasons I believe that this is a good place but I'm looking for feedback on this:

  * The `TaskExecutionMetadata` type returned by this function has the following doc string
  
    > TaskExecutionMetadata provides a layer on top of the core TaskExecutionMetadata with customized annotations and labels for k8s plugins.
    
    I find this fitting the proposed use case.
    
  * Injecting the label here means that it is automatically injected not only for normal python tasks but also for plugins like kubeflow distributed training tasks.

## How was this patch tested?

* Added unit tests.
* Ran propeller with the proposed changes in our staging cluster and ensured that the execution identity is correctly injected for python tasks as well as pytorch tasks.
  * I also tested the case where the client authenticates with a client credentials flow instead of pkce flow as in this case the execution identity is *not* set in the flyte workflow.


fg91:
For documentation purposes:

One can set the following e.g. in the [default flyte pod template](https://docs.flyte.org/en/latest/deployment/configuration/general.html#runtime-podtemplates) in order to use the execution identity as a pod template:

```yaml
apiVersion: v1
kind: PodTemplate
template:
  spec:
    containers:
    - env:
      - name: EXECUTION_IDENTITY
        valueFrom:
          fieldRef:
            apiVersion: v1
            fieldPath: metadata.labels['execution_identity']
```

hamersaw:
@fg91 this looks great. I agree, the metadata is the correct place to inject this. My first question here is whether this should be behind a configuration flag? Is there a scenario where we would not want the execution identity available? I'm assuming that if a user has Flyte configured to inject user IDs from middleware, it would be beneficial to include this information in k8s resources, thoughts? cc. @ByronHsu

fg91:
> My first question here is whether this should be behind a configuration flag? Is there a scenario where we would not want the execution identity available?

I can't come up with such a scenario tbh. I proposed a configuration flag as the most conservative change because I wasn't sure how others feel about this. I'm happy to remove the flag though @hamersaw!


> I'm assuming that if a user has Flyte configured to inject user IDs from middleware, it would be beneficial to include this information in k8s resources, thoughts? cc. @ByronHsu

Injecting user ids from middleware in flyteadmin is not behind a config flag (see [here](https://github.com/flyteorg/flyte/blob/a726c3650bb6f88ab15d9b6f33a4ede7b03e77dd/flyteadmin/pkg/server/service.go#L84)) so users don't have to opt in.

hamersaw (approved):

flytepropeller/pkg/controller/nodes/task/k8s/task_exec_context.go
```diff
@@ -42,25 +44,28 @@ func (t TaskExecutionMetadata) GetAnnotations() map[string]string {
...
 		}
 
-		injectSecretsLabel = map[string]string{
-			secrets.PodLabel: secrets.PodLabelValue,
-		}
+		injectLabels[secrets.PodLabel] = secrets.PodLabelValue
+	}
+
+	id := tCtx.GetSecurityContext().RunAs.ExecutionIdentity
+	if id != "" {
```
hamersaw, line 62: golang typically is `if len(id) > 0 {`

hamersaw (changes requested):
Very useful, can we clean up the unit tests?

flytepropeller/pkg/controller/nodes/task/k8s/task_exec_context.go
fg91, line 62:
```suggestion
	if len(id) > 0 {
```

fg91: https://github.com/flyteorg/flyte/pull/4637/commits/2a6a3a67390b926c27bb88d069af6b7fa1d01c52 👍

fg91:
> [...] can we clean up the unit tests?

Done ✅

hamersaw (approved):

PR #4642, opened 2023-12-26
Modernize linters

eapolinario:
## Tracking issue
NA

In an effort to modernize the use of linters in the codebase we first unify the config file used across the projects.

A few changes:
- A single .golangci.yml file is shared across all repos
- The separate CI checks to check for lint errors now run in a single step


hamersaw:
Does this only work with `make -C <component> lint` from the flyte repo root? or does it still work using `make lint` from each internal project (ex. `cd ./flyte/flytepropeller; make lint`)?

PR #4646, opened 2023-12-28
Artf/cleanup post shell

wild-endeavor:
Now that the first shell pr has been merged into master, we're doing some clean up work.

* Remove the inputs/outputs logic from the cloud publisher on the admin side and move it to the event handler on the artifact side
* Remove the initialization sql bit and made hstore extension creation just a migration
* Refactor the handler to be outside of the receiver in artifact events
* Add adminclient into the servicecallhandler object
* Set gizmo sqs config to false for consuming base64 and update the sqs processor to be roughly functional
* Fixing lint issues
* Add artifact configuration to the sandbox values file
* Infra related updates
  * Remove the old python artifact service helm templates
  * update local sandbox configuration to support routing back out to local development endpoint


wild-endeavor:
@eapolinario maybe we can talk about the go mod change.  I kept running into an error with go mod tidy, ended up resolving it by setting `sigs.k8s.io/controller-runtime` to a real version rather than the 00001000 version or whatever.  but it's replaced anyways so i feel like it shouldn't matter.  not sure what i was doing wrong.

PR #4656, opened 2023-12-30
Support kuberay v1.0.0

Yicheng-Lu-llll:

Currently, KubeRay is undergoing significant refactoring of the RayJob, which includes lots of status changes that Flyte depends on. This update could potentially be released in KubeRay v1.2.0. Before refactoring the current Ray plugin to be based on KubeRay v1.2.0, it would be great to first support KubeRay v1.0.0 and all earlier versions. For KubeRay v1.2.0, maintaining backward compatibility will be almost impossible due to the extensive changes in RayJob. 

Additionally, the KubeRay maintainers have recently decided to continue maintaining version v1alpha1 (previously deprecated) until the release of KubeRay v1.2.0.

This PR enables users to use Kuberay v1.0.0 and all earlier versions.

This PR introduces several new statuses while maintaining backward compatibility.

## How was this patch tested?
I have tested this PR with Kuberay v0.5.2 and Kuberay v1.0.0.


EngHabu (approved):
Thank you for carrying this on!

PR #4657, opened 2024-01-01
[WIP] [Newbie] [RFC] Flyte Local Cluster

Future-Outlier:
## Tracking issue


hamersaw:
@Future-Outlier this is great, we have had more than a few internal discussions on this deployment model. Would love to help shepherd this as much as possible. A few things to note here:
(1) `Change runtime from Kubernetes to Process` - we already have an internal application that does this using a separate Flyte plugin to execute using local processes. Perhaps this is something we can talk about open-sourcing for this model.
(2) A large component (perhaps the largest) that is missing here is the use of etcd to store FlyteWorkflow CRs that track execution. FlyteAdmin creates these, and using the watch API FlytePropeller picks them up and then modifies them to persistently track workflow execution status'. In this scenario, it warrants some discussion about what local execution here means for this construct . For example, are executions persisted across the local cluster restart? etc.

Future-Outlier:
> @Future-Outlier this is great, we have had more than a few internal discussions on this deployment model. Would love to help shepherd this as much as possible. A few things to note here:
> 
> (1) `Change runtime from Kubernetes to Process` - we already have an internal application that does this using a separate Flyte plugin to execute using local processes. Perhaps this is something we can talk about open-sourcing for this model.
> 
> (2) A large component (perhaps the largest) that is missing here is the use of etcd to store FlyteWorkflow CRs that track execution. FlyteAdmin creates these, and using the watch API FlytePropeller picks them up and then modifies them to persistently track workflow execution status'. In this scenario, it warrants some discussion about what local execution here means for this construct . For example, are executions persisted across the local cluster restart? etc.

Thank you really much, I will discuss more with @pingsutw and update the rfc here!

Future-Outlier:
etcd information cache -> local file system

Issue #4666, opened 2024-01-03
[BUG] Unable to validate certificates using ray plugin


andresgomezfrr:
### Describe the bug

```
failed to load plugin - ray: [PluginInitializationFailed] Error getting informer for %!s(\u003cnil\u003e), caused by: failed to get API group resources: unable to retrieve the complete list of server APIs: ray.io/v1alpha1: Get \"https://X.X.X.X/apis/ray.io/v1alpha1\": x509: certificate signed by unknown authority
```

Unable to initialize the ray plugin because of a problem validating the certificate even with the caCertificate configured.

### Expected behavior

Be able to validate the certificate using the caCertificate provided by the config.

### Additional context to reproduce

Update to the last flyte version `v1.10.6`

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #4667, opened 2024-01-03
Avoid to use the http.DefaultClient

andresgomezfrr:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/4666

We need to avoid using the `http.DefaultClient` because it doesn't have the proper transport configured with the TLS configurations.

Instead of using the `http.DefaultClient` to configure the read-only requests (cache) and the dynamic rest mapper. We are going to create a new client using the config to ensure that the TLS configuration is loaded.


flyteplugins/go/tasks/pluginmachinery/k8s/client.go
```diff
@@ -40,9 +38,14 @@ type Options struct {
 // NewKubeClient creates a new KubeClient that caches reads and falls back to
 // make API calls on failure. Write calls are not cached.
 func NewKubeClient(config *rest.Config, options Options) (core.KubeClient, error) {
+	httpClient, err := rest.HTTPClientFor(config)
```
honnix, line 41: A question. We used to have three different http clients for different purposes, and now we will only have one. Do you know the implication, or do you prefer we still have three? @eapolinario

flyteplugins/go/tasks/pluginmachinery/k8s/client.go
```diff
@@ -53,7 +56,7 @@ func NewKubeClient(config *rest.Config, options Options) (core.KubeClient, error
 
 	if options.CacheOptions == nil {
 		options.CacheOptions = &cache.Options{
-			HTTPClient: http.DefaultClient,
+			HTTPClient: httpClient,
```
honnix, line 59: This should be set to line 70 as well.

flyteplugins/go/tasks/pluginmachinery/k8s/client.go
honnix: To be clear, I don't see why we would need three different struct instances given the config is the same.

honnix:
```
// The Client's Transport typically has internal state (cached TCP
// connections), so Clients should be reused instead of created as
// needed. Clients are safe for concurrent use by multiple goroutines.
```

honnix: I think this was the PR https://github.com/kubernetes-sigs/controller-runtime/pull/2122, which carried exactly the same rationale.

honnix (approved):

PR #4689, opened 2024-01-08
GetDynamicNodeWorkflow endpoint

iaroslav-ciupin:
## Tracking issue

This new endpoint will be called by Flyte console on execution page. When workflow contains dynamic nodes, for each dynamic node, console will call this endpoint with corresponding node execution id and backend will respond with dynamic workflow closure. Its a replacement API for `GetNodeExecutionData` which can only be called when use clicks on task execution tab to see task Inputs/Outputs. 

## How was this patch tested?

Unit & integration tests, tested end-to-end by deploying this branch on real admin deployment.


flyteadmin/pkg/manager/impl/workflow_manager.go
```diff
@@ -361,6 +361,33 @@ func (w *WorkflowManager) ListWorkflowIdentifiers(ctx context.Context, request a
...
+	ctx = getNodeExecutionContext(ctx, request.Id)
+	nodeExecutionModel, err := util.GetNodeExecutionModel(ctx, w.db, request.Id)
+	if err != nil {
+		logger.Debugf(ctx, "Failed to get node execution with id [%+v] with err %v",
+			request.Id, err)
+		return nil, err
+	}
+
+	if nodeExecutionModel.DynamicWorkflowRemoteClosureReference == "" {
+		return &admin.DynamicNodeWorkflowResponse{}, nil
```
hamersaw, line 378: Should this return a `NotFound` error instead or relying on `nil` closure to indicate it doesn't exist.

hamersaw (commented):
Only top-level comment is this seems kind of odd to expose through the `workflow_manager`. I do not see a `node_manager` in flyteadmin and I'm not sure that this small update merits adding one.

iaroslav-ciupin:
@hamersaw I was thinking between `node_executions_manager`(since dynamic workflow is stored inside a node execuction) and `workflow_manager`(since it's still a workflow even if its a dynamic one). Do you want to introduce a new manager just for this? do we even have a `node` entity in Flyte API? I see that each manager corresponds to a Flyte entity.

hamersaw:
> @hamersaw I was thinking between `node_executions_manager`(since dynamic workflow is stored inside a node execuction) and `workflow_manager`(since it's still a workflow even if its a dynamic one). Do you want to introduce a new manager just for this? do we even have a `node` entity in Flyte API? I see that each manager corresponds to a Flyte entity.

To me, it makes more sense to put this in the `node_execution_manager`. Lets not worry about adding anything new here. @katrogan do you have any concerns with this route?

katrogan:
+1 to node execution manager, this is execution data produced by the node at runtime and isn't a workflow in the traditional sense (we can't launch invocations of it, it's never formally registered, etc)

hamersaw (dismissed):

flyteidl/protos/flyteidl/admin/workflow.proto
```diff
@@ -90,3 +90,11 @@ message CreateWorkflowFailureReason {
         WorkflowErrorExistsIdenticalStructure exists_identical_structure = 2;
     }
 }
+
+message GetDynamicNodeWorkflowRequest {
```
katrogan, line 94: can we move this definition to  https://github.com/flyteorg/flyte/blob/master/flyteidl/protos/flyteidl/admin/node_execution.proto

flyteadmin/pkg/manager/impl/node_execution_manager.go
```diff
@@ -297,6 +297,35 @@ func (m *NodeExecutionManager) CreateNodeEvent(ctx context.Context, request admi
...
+
+	ctx = getNodeExecutionContext(ctx, request.Id)
+	nodeExecutionModel, err := util.GetNodeExecutionModel(ctx, m.db, request.Id)
+	if err != nil {
+		logger.Errorf(ctx, "failed to get node execution with id [%+v] with err %v",
+			request.Id, err)
+		return nil, err
+	}
+
+	if nodeExecutionModel.DynamicWorkflowRemoteClosureReference == "" {
```
katrogan, line 313: since this re-uses a lot of code with https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/manager/impl/node_execution_manager.go#L518-L537 any chance we can refactor to share the implementation?

katrogan (changes requested):

katrogan (approved):

PR #4715, opened 2024-01-10
Remove dockerfiles from subfolder

pingsutw:
All the docker files in the subfolder are unused and outdated.

Remove them


pingsutw:
cc @pmahindrakar-oss

pmahindrakar-oss:
Can we also have make docker_build work for each component so that it picks the right Dockerfile based on the component
Eg : If you are inside flytepropellerdirectory and you do make docker_build then it picks Dockerfile.propeller and builds that.

Cloud repo has this convention. Happy to have this as separate ticket too. Thanks for taking care of this

pmahindrakar-oss (approved):

pingsutw:
That's a good idea. I'll add that in a separate PR.

PR #4720, opened 2024-01-11
monodocs uses flytekit/flytectl index rst file

cosmicBboy:
The purpose of this PR is to use docs toctree files that are in the `flytectl` and `flytekit` repos respectively. This is so that content changes can occur exclusively in these repos without requiring changes to the `flyte` repo.


docs/conf.py
```diff
@@ -337,6 +337,7 @@
         "flytesnacks/tutorials",
         "flytesnacks/integrations",
     ],
+    "dev_build": bool(int(os.environ.get("MONODOCS_DEV_BUILD", 1))),
```
cosmicBboy, line 340: will set this to 0 by default when the next flyte release happens

kumare3 (approved):

PR #4737, opened 2024-01-17
Update artifact IDL with new time partition

wild-endeavor:
The current partition information coerces the time value into a string and uses the special `ds` key to indicate that it's a time.  This is awkward so let's pull it out into a separate field that's correctly typed.

### Admin
* Remove tag handling when running artifact queries.
* Update the query handling to fill in the new time partition if needed.
* Left the time parsing bit in there for now, but updated to ensure UTC.
* Some tests.

### IDL
* Update `ArtifactBindingData` to be able to bind to the time partition, and also move the old and new field together into a `oneof`.
* Add a new `TimePartition` message which has a timestamp field.
* Add a new timestamp type field `time_value` to the `LabelValue` value oneof.  This value should only be used for the new time partition message. 
* Unrelated, but made the existing `partitions` field in the `ArtifactID` not a oneof.  Even if we one day need existence checking, the fact that it's a separate message already gives us that (the `HasField` call in python and the nil check in go).


flyteidl/protos/flyteidl/core/artifact_id.proto
```diff
@@ -19,8 +20,13 @@ message ArtifactBindingData {
   uint32 index = 1;
 
   // These two fields are only relevant in the partition value case
-  string partition_key = 2;
-  string transform = 3;
+  oneof partition_data {
```
squiishyy, line 23: above this, `ArtifactBindingData` is listed as being only valid for triggers, why is that?

squiishyy (dismissed):

squiishyy (approved):

PR #4738, opened 2024-01-18
Deprecated Agent State to Agent Phase

Future-Outlier:
## Tracking issue
Fixes https://github.com/flyteorg/flyte/issues/3936

We want to deprecate `State` in webapi and use `core.TaskExecution.Phase` instead to make users less confused.

- add `core.TaskExecution.Phase` phase in `agent.proto`
- add `DEPRECATED` annotation in `State` in `agent.proto`
- add switch case statements to route the phase in `agent/plugin.go`
- add tests for above

Note:
The reason why we need to:
1. check the phase
2. check the state

is because the default value of phase is `UNDEFINED`, and the default value of state is `RETRYABLE FAILURE`.
We can check the state when the phase is `UNDEFINED`.

## How was this patch tested?
- single binary and flytekit [here](https://github.com/flyteorg/flytekit/pull/2123)
- unit test
### Setup process
0. start agent server by flytekit
1. start flyte components with `--dev`
2. compile and start the flyte cluster
3. run a remote task with sensor agent

```python
pyflyte serve agent
```
```bash
flytectl demo start --dev
```
```bash
cd flyte
make compile;
flyte start --config ../flyte-single-binary-local-dev.yaml
```
```bash
pyflyte run --remote sensor_example.py wf
```

```yaml
tasks:
  task-plugins:
    enabled-plugins:
      - agent-service

plugins:
  agent-service:
    supportedTaskTypes:
      - sensor
    defaultAgent:
      insecure: true
      timeouts:
        GetTask: 100s
      defaultTimeout: 100s
```
```python

@task()
def t1():
    return

@workflow()
def wf():
    sensor(path="/tmp") >> t1()

if __name__ == "__main__":
    wf()
```

### Screenshots
image

## Related PRs
https://github.com/flyteorg/flytekit/pull/2123


Future-Outlier:
image

flyteidl/protos/flyteidl/admin/agent.proto
```diff
@@ -78,7 +78,7 @@ message GetTaskResponse {
 }
 
 message Resource {
-  // The state of the execution is used to control its visibility in the UI/CLI.
+  // DEPRECATED. The state of the execution is used to control its visibility in the UI/CLI.
```
honnix, line 81: There is a `deprecated` keyword to use: https://protobuf.dev/programming-guides/proto3/#options

```
State state = 1 [deprecated = true];
```

flyteplugins/go/tasks/plugins/webapi/agent/plugin.go
```diff
@@ -190,6 +193,31 @@ func (p Plugin) Status(ctx context.Context, taskCtx webapi.StatusContext) (phase
...
+			return core.PhaseInfoUndefined, err
+		}
+		return core.PhaseInfoSuccess(taskInfo), nil
+	case flyteIdl.TaskExecution_ABORTED:
+		return core.PhaseInfoFailure(pluginErrors.TaskFailedWithError, "failed to run the job with aborted phase", taskInfo), nil
+	case flyteIdl.TaskExecution_FAILED:
+		return core.PhaseInfoFailure(pluginErrors.TaskFailedWithError, "failed to run the job", taskInfo), nil
+	}
+
+	logger.Infof(ctx, "Agent returned an undefined phase [%v]. Checking state.", resource.Phase)
```
honnix, line 219: I'm not sure I like this type of fallback logic. Suppose we add a new phase later, the previous switch would not match that and then we fallback to state, which is obviously not what we want.

I prefer explicitly checking whether `resource.Phase` is `UNDEFINED` and then use state.

Future-Outlier: This is a great idea! Thank you so much!

flyteidl/protos/flyteidl/admin/agent.proto
Future-Outlier: Thank you so much! I learned a lot

Future-Outlier: Fixed! Thank you again!

flyteplugins/go/tasks/plugins/webapi/agent/plugin.go
Future-Outlier: Fixed! Thank you again!

pingsutw (approved):

PR #4741, opened 2024-01-18
Docs/Clarify propeller scaling

wild-endeavor:
The enumer generated code needs the type to be `Hash` not `hash`.  Also clarify that it's not autoscaling in the traditional sense.


davidmirror-ops (approved):
Thanks!

PR #4733, opened 2024-01-17
Reintroduce k8s client fallback to cache lookups

hamersaw:
## Tracking issue
fixes https://github.com/flyteorg/flyte/issues/4730

## Why are the changes needed?
[This PR](https://github.com/flyteorg/flyte/pull/4199) updated k8s and removed the logic to fallback to the client `Reader` when the cache `Reader` was a miss. In high traffic scenarios the k8s informer can not populate the cache fast enough, so the fallback API call is required to ensure correctness.

## What changes were proposed in this pull request?
Creating the k8s client with a fallback to API call when an object does not exist in the informer cache.

## How was this patch tested?
Faked an object `NotFound` error because of cache miss.

### Setup process
Updated k8s [client](https://github.com/flyteorg/flyte/blob/0c8dc61833683c7d2482156527c8f45616ee4a20/flytestdlib/otelutils/k8s.go#L43-L47) and [cache](https://github.com/flyteorg/flyte/blob/master/flytestdlib/otelutils/k8s.go#L21-L25) through the otel wrapper to fail retrievals if the object was created less than 3 seconds ago. Example code:
```
    func (c *K8sCacheWrapper) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
        ctx, span := NewSpan(ctx, K8sClientTracer, fmt.Sprintf("%s.Cache/Get", k8sSpanPathPrefix))
        defer span.End()

        err := c.Cache.Get(ctx, key, obj, opts...)
        if time.Since(obj.GetCreationTimestamp().Time) < (time.Second * 3) {
            obj = nil
            return k8serrors.NewNotFound(schema.GroupResource{Group: "foo", Resource: "bar"}, key.Name)
        }

        return err
        //return c.Cache.Get(ctx, key, obj, opts...)
    }
```

Then validated that this causes Pods to be stuck in `Terminating` state with finalizers still intact:
```
    hamersaw@ragnarok:~$ kubectl -n flytesnacks-development get pods
    NAME                        READY   STATUS        RESTARTS   AGE
    f5329181f1364487b8a9-n0-0   0/1     Terminating   0          95s
    f5329181f1364487b8a9-n0-1   0/1     Terminating   0          86s
    f5329181f1364487b8a9-n0-2   0/1     Terminating   0          76s
```
with log
```
[...]

PR #4748, opened 2024-01-19
Fix read cache by delegating to uncached client

honnix:
This PR tries to solve a problem in #4733 where cache lookup would be delegated forever in case of a cache miss.

Note that this PR does not try to address the missing writethrough cache which got removed by #4199. If that feature is still needed, we should bring it back (tried that in https://github.com/honnix/flyte/pull/2).

Use an uncached client as a fallback to avoid endless cache lookup.


honnix:
Closing in favour of #4752

Issue #4730, opened 2024-01-16
[BUG] flytepropeller fails trying to get pod resource using the kubeClient


andresgomezfrr:
### Describe the bug

Yesterday we upgraded the propeller to the version `1.10.6` and today we noticed that we have a lot of pods with the `Terminating` state in our cluster. After some checks we detected that we started to see this log many times 
https://github.com/flyteorg/flyte/blob/38d1833181ff44cedaccbc5c775341b7be3f354e/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go#L269 before the upgraded we didn't have any logs like this.

We deleted all the pods manually and downgraded the version and the logs disappeared, now the pods are deleted again without issues.

### Expected behavior

The pods should be deleted properly and not stuck in the Terminating state.

### Additional context to reproduce

Upgrade to the latest version of the propeller and execute pods.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #4752, opened 2024-01-19
Wrapping k8s client with write filter and cache reader

hamersaw:
## Tracking issue
fixes https://github.com/flyteorg/flyte/issues/4730

The previous attempt (https://github.com/flyteorg/flyte/pull/4733) mistakenly put he cache lookup into a loop until it was populated. This caused large CPU spikes.

This PR bypasses the k8sClient internal caching logic and instead just wraps an uncached k8s client with a cache lookup for writes and a write through filter on `create` / `delete` operations.

## How was this patch tested?
local

## Related PRs
_NA_

## Docs link
_NA_


cmd/single/start.go
```diff
@@ -122,7 +122,7 @@ func startPropeller(ctx context.Context, cfg Propeller) error {
 			DefaultNamespaces: namespaceConfigs,
 		},
 		NewCache: executors.NewCache,
-		NewClient: executors.NewClient,
+		NewClient: executors.BuildNewClientFunc(50000, propellerScope),
```
honnix, line 125: I think if we don't see the `writeFilterSize` would be different, maybe for now we can just inline it. Wdyt?

flytepropeller/cmd/controller/cmd/webhook.go
```diff
@@ -109,7 +109,7 @@ func runWebhook(origContext context.Context, propellerCfg *config.Config, cfg *w
 			DefaultNamespaces: namespaceConfigs,
 		},
 		NewCache:  executors.NewCache,
-		NewClient: executors.NewClient,
+		NewClient: executors.BuildNewClientFunc(50000, propellerScope),
```
honnix, line 112:
```suggestion
		NewClient: executors.BuildNewClientFunc(50000, webhookScope),
```

flytepropeller/pkg/controller/executors/kube.go
```diff
@@ -33,107 +33,91 @@ var NewCache = func(config *rest.Config, options cache.Options) (cache.Cache, er
...
+		if options.Cache != nil && options.Cache.Reader != nil {
+			cacheReader = options.Cache.Reader
+			cachelessOptions.Cache = nil
 		}
 
-		options.Cache.Reader = reader
-	}
+		k8sClient, err := client.New(config, cachelessOptions)
+		if err != nil {
+			return k8sClient, err
```
honnix, line 47: It doesn't seem to be reasonable to return `k8sClient`, which could potentially be something built partially if `client.New` does something weird.

```suggestion
			return nil, err
```

cmd/single/start.go
hamersaw: sounds good to me!

flytepropeller/cmd/controller/cmd/webhook.go
hamersaw: good catch, thanks.

flytepropeller/pkg/controller/executors/kube.go
hamersaw: done.

honnix:
This looks good to me. I was struggling with unit test for this and I guess you might be the same. Not sure how to properly test it without some refactoring in place.

pingsutw (approved):

Issue #4349, opened 2023-11-02
[BUG] Etcdserver: request is too large error causing never-ending tasks and log spamming


deadlycoconuts:
### Describe the bug

We’ve observed this issue whereby the FlytePropeller isn’t failing workflows that cannot be written to the etcd store. These workflows get stuck in an enqueue/resync/fail loop which repeats itself every minute or so indefinitely, even though they are shown as ‘failed’ on the FlyteConsole. This also floods our logging services with redundant logs.

We tried toggling the `useOffloadedWorkflowClosure` [flag](https://docs.flyte.org/en/latest/deployment/configuration/performance.html#offloading-static-workflow-information-from-crd) but unfortunately it didn’t help resolve the etcd errors that we’ve been seeing from the FlytePropeller. We’ve restarted both the FlyteAdmin and the FlytePropeller deployments after updating the configs but that didn’t seem to stop the existing executions from throwing the etcd errors continuously. In the end we had to delete the affected executions one by one manually to stop them from being scheduled endlessly.

There are some related issues and fixes that we found from 2-3 years ago which do not seem to work anymore, so we’re wondering what might’ve changed since:

Related Issues
- https://github.com/flyteorg/flyte/issues/363
- https://github.com/flyteorg/flyte/issues/511

Related Fixes
- https://github.com/flyteorg/flytepropeller/pull/240

In particular, it seems like this if condition is no longer working as expected: https://github.com/flyteorg/flyte/blob/c6476cc178e576a4b7853d5d2adcb54081848974/flytepropeller/pkg/controller/workflowstore/passthrough.go#L99

### Expected behavior

Tasks should never end up in a state where they are retried endlessly and should be terminated with an error if there is an issue updating their state. 

### Screenshots

Here are some images to illustrate the problem that I’ve described:
1. Affected workflow - this was first executed at 2.17 AM (SGT/HKT) on the 7th of October and failed at around 2.34 AM; it was however being restarted endlessly ever since
![image](https://github.com/flyteorg/flyte/assets/36802364/91bab5e3-ab96-4d5a-a822-98acf1c4b80d)
2. FlytePropeller logs when the workflow first failed
![image](https://github.com/flyteorg/flyte/assets/36802364/b775a9ec-d471-4e5d-9a44-94b2883f0658)
3. FlytePropeller logs 11 days after the initial failure
[...]

PR #4755, opened 2024-01-22
[BUG] Handle Potential Indefinite Propeller Update Loops

pvditt:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/4349_

A user pointed out that they had workflows stuck indefinitely in a update loop with updates failing to write to etcd.

This can happen when:
- updating the mutated workflow fails w/ ErrWorkflowToLarge propeller falls back to [updating the phase of the pre-mutated wf to failing ](https://github.com/flyteorg/flyte/blob/ace09677050eb6fd8cc8b712c9282019e1052439/flytepropeller/pkg/controller/handler.go#L388)such that the wf can transition to failing on the next loop
- on the next loop, [HandleFlyteWorkflow](https://github.com/flyteorg/flyte/blob/ace09677050eb6fd8cc8b712c9282019e1052439/flytepropeller/pkg/controller/handler.go#L142) will then transition/update the wf to WorkflowPhaseFailed. This doesn't matter if updating the CRD w/ the mutated workflow fails again w/ ErrWorkflowToLarge as it will fall back to [updating the phase of the pre-mutated wf to failing ](https://github.com/flyteorg/flyte/blob/ace09677050eb6fd8cc8b712c9282019e1052439/flytepropeller/pkg/controller/handler.go#L388). Workflow then gets suck in a loop of getting updated in non-terminal WorkflowPhaseFailing

When encountering ErrWorkflowToLarge on an update to etcd, set the phase to Failed if the mutated workflow is already in a terminal state. This should prevent the indefinite loop described in https://github.com/flyteorg/flyte/issues/4349.

## How was this patch tested?

- Added a unit test
- Manually updated TryMutate to always cause etcd writes to fail on updates. Then ensured that propeller would stop evaluating the workflow.

### DIFF
--- a/flytepropeller/pkg/controller/handler.go
+++ b/flytepropeller/pkg/controller/handler.go
@@ -385,12 +385,13 @@ func (p *Propeller) streak(ctx context.Context, w *v1alpha1.FlyteWorkflow, wfClo
 			// Workflow is too large, we will mark the workflow as failing and record it. This will automatically
 			// propagate the failure in the next round.
 			mutableW := w.DeepCopy()
-			// if workflow is already in a terminal state then cleanup is already handled
+			// catch potential indefinite update loop
 			if mutatedWf.GetExecutionStatus().IsTerminated() {
 				SetFinalizerIfEmpty(mutableW, FinalizerKey)
 				SetDefinitionVersionIfEmpty(mutableW, v1alpha1.LatestWorkflowDefinitionVersion)
 				SetCompletedLabel(mutableW, time.Now())
-				mutableW.Status.UpdatePhase(v1alpha1.WorkflowPhaseFailed, "Workflow size has breached threshold, aborted", &core.ExecutionError{
+				msg := fmt.Sprintf("Workflow size has breached threshold. Finalized with status: %v", mutatedWf.GetExecutionStatus().GetPhase())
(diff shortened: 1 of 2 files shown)


flytepropeller/pkg/controller/handler.go
```diff
@@ -385,11 +385,24 @@ func (p *Propeller) streak(ctx context.Context, w *v1alpha1.FlyteWorkflow, wfClo
...
 			// propagate the failure in the next round.
 			mutableW := w.DeepCopy()
-			mutableW.Status.UpdatePhase(v1alpha1.WorkflowPhaseFailing, "Workflow size has breached threshold, aborting", &core.ExecutionError{
-				Kind:    core.ExecutionError_SYSTEM,
-				Code:    "WorkflowTooLarge",
-				Message: "Workflow execution state is too large for Flyte to handle.",
-			})
+			// catch potential indefinite update loop
+			if mutatedWf.GetExecutionStatus().IsTerminated() {
+				SetFinalizerIfEmpty(mutableW, FinalizerKey)
```
hamersaw, line 390: This should be `ResetFinalizers(mutatedWf)` I believe. The existence of finalizers stop k8s from garbage collecting the resource, so we need to remove them to allow the CR to be cleaned up.

pvditt: thanks for catching this

hamersaw (approved):

PR #4756, opened 2024-01-22
Fix webhook typo, add podLabels, add podEnv to flyte-core Helm chart

ddl-ebrown:
## Tracking issue

Hopefully this isn't too controversial a change. To deploy in our environments, we've standardized on the use of `podAnnotations`, `podLabels` and `podEnv` helm values. Globally, this allows for setting values across all deployment, statefulset and daemonset resources.

Flyte already has `podAnnotations`, but not `podLabels` or `podEnv`. This PR adds both to `flyte-core`.

Of not, `flyteadmin` has a `env: []` value set already, but it's the only one. I didn't want to break anything so left it as-is. Ideally I think  that would become `podEnv: {}` and would render in the same way as the other charts.

## How was this patch tested?

`make helm` was used locally to generate charts / docs. I set values explicitly in the `values.yaml` to verify correct rendering

i.e.

```helm
podEnv:
  - name: TEST
    value: value
```

### DIFF
--- a/charts/flyte-core/README.md
+++ b/charts/flyte-core/README.md
@@ -60,13 +60,14 @@ helm install gateway bitnami/contour -n flyte
 | cloud_events.eventsPublisher.eventTypes[0] | string | `"all"` |  |
 | cloud_events.eventsPublisher.topicName | string | `"arn:aws:sns:us-east-2:123456:123-my-topic"` |  |
 | cloud_events.type | string | `"aws"` |  |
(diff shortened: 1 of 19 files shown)


davidmirror-ops (approved):
Thanks @ddl-ebrown!
This is very helpful not only for consistent config but eventually even for observability use cases I think. I'm also curious about the actual use case that led your org to standardize on these keys.

BTW, please consider signing up for the weekly [community raffle](https://www.union.ai/blog-post/flyte-weekly-community-raffle-lets-soar-to-new-heights), this contribution would give you points!

Issue #4291, opened 2023-10-24
[Core feature] Allow to specify userdata bucket in the storage configuration in flyte-core charts


gdabisias:
### Motivation: Why do you think this is important?

Currently it is only possible to specify the metadata container in the storage object in the flyte core chart, e.g.
```
storage: {
			type: "custom"
			custom: {
				type: "stow"
				stow: {
					kind: "s3"
					config: {
						auth_type:     "accesskey"
						region:        "xxx"
						access_key_id: string | *""
						secret_key:    string | *""
						endpoint:      xxx
						disable_ssl:   false
						v2_signing:    false
					}
				}
            container: helm.values.userSettings.metadataBucket
```
while in the single binary deployment it is possible to specify both user and metadata containers in the storage configuration, e.g.
```
 storage:
    # metadataContainer Bucket to store Flyte metadata
    metadataContainer: "my-organization-flyte-container"
    # userDataContainer Bucket to store Flyte user data
    userDataContainer: "my-organization-flyte-container"
```
from [here](https://github.com/flyteorg/flyte/blob/4fc4988400b9f448f61fe796d5252c1aa075fe6f/charts/flyte-binary/values.yaml#L54)

The current fix for this lack of specification is to define the following field in the propeller configuration
```
core: propeller: {
				"rawoutput-prefix":     "s3://" + helm.values.userSettings.userdataBucket + "/"
...
```

### Goal: What should the final outcome look like, ideally?

We should use the same storage configuration for the flyte deployment charts.

I think it would be good for 2 reasons:
- it unifies configuration across deployments 
- it allows the users to specify all storage configurations just in one place

### Describe alternatives you've considered 

None

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

eapolinario:
@gdabisias , would you be willing to contribute with this change?

gdabisias:
hi @eapolinario, possibly in the future, but not right now:)

PR #4759, opened 2024-01-22
Introduce rawDataBucket

davidmirror-ops:
## Tracking issue
#4291 

Enables users to specify a separate (from metadata) Raw Data bucket in the form of a userDataBucket parameter in flyte-core. 


charts/flyte-core/values-eks.yaml
```diff
@@ -5,6 +5,7 @@ userSettings:
   dbPassword: <DB_PASSWORD>
   rdsHost: <RDS_HOST>
   bucketName: <BUCKET_NAME>
+  userBucketName: <RAW_DATA_BUCKET_NAME>
```
pingsutw, line 8: would it be better to call it `rawDataBucketName`?

deployment/eks/flyte_aws_scheduler_helm_generated.yaml
```diff
@@ -179,6 +179,8 @@ data:
       signedUrls:
         durationMinutes: 3
   storage.yaml: | 
+    propeller:
+      rawoutput-prefix: s3://<RAW_DATA_BUCKET_NAME>/data
```
pingsutw, line 183: why do we need `/data` here?

deployment/sandbox/flyte_helm_generated.yaml
```diff
@@ -498,6 +500,8 @@ data:
       profiler-port: 10254
       storage-prefix: metadata/datacatalog
   storage.yaml: | 
+    propeller:
+      rawoutput-prefix:
```
pingsutw, line 504: IIRC, `rawoutput-prefix` can't be empty. does it work in the sandbox?

deployment/eks/flyte_aws_scheduler_helm_generated.yaml
davidmirror-ops: Not needed, it's the way binary handles it currently. I thought it'd be nice to align them

Issue #4765, opened 2024-01-23
[BUG] Persisting task on PhaseTimeout


pvditt:
### Describe the bug

When a task fails due to timing out with respect to the node-active-deadline, the node is set to NodePhaseTimingOut -> node is aborted -> NodePhaseTimedOut which is a terminal phase.

Since the abort is not set as a final transition, the state does not get persisted to admin causing for the task state to show up as "UNKNOWN" instead of "ABORTED"

### Expected behavior

When a task is aborted, the aborted state should get persisted to admin. 

### Additional context to reproduce

Set node-active-deadline to small value in Flyte config:

propeller:
  node-config:
    default-deadlines:
      node-active-deadline: 10s
      
 run workflow that runs longer than that deadline + don't set an execution timeout:
 
 ```
 @task(
    retries=1,
)
def demo_task():
    time.sleep(3000)


@workflow
def demo_workflow():
    demo_task()
 ```

### Screenshots

![image](https://github.com/flyteorg/flyte/assets/37558497/55d9420b-eea8-4492-9fe8-5893d199765f)

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

pvditt:
this issue should be closed by https://github.com/flyteorg/flyte/pull/4766

Issue #4087, opened 2023-09-27
[BUG] Timed out task in subworkflow does not fail workflow


bstadlbauer:
### Describe the bug

In the following example, a task in a subworkflow fails due to a timeout. When that task fails, it goes into an `Unkonwn` state in the dashboard, whilst the toplevel workflow is still runnning instead of failing as expected.

I've tested this in a dev sendbox with version `v1.9.1`.

```python
import time
from datetime import timedelta

from flytekit import task, workflow, Resources


@task(
    timeout=timedelta(seconds=1),
    # Just making sure that things run on my local machine; not required to reproduce
    limits=Resources(cpu="0.5", mem="500Mi"),
)
def demo_task():
    time.sleep(10)


@workflow
def subworkflow():
    demo_task()


@workflow
def demo_workflow():
    subworkflow()
```

![image](https://github.com/flyteorg/flyte/assets/11799671/b87dec50-57c3-4ad7-8b22-6508567dd40c)

### Expected behavior

The workflow should fail as the task timed out.

### Additional context to reproduce

Here's the output of `flytectl` for that execution: 

```yaml
- node_exec:
    closure:
      createdAt: "2023-09-27T08:14:27.730147Z"
      outputUri: s3://my-s3-bucket/metadata/propeller/dask-testing-development-f43853b3ae7cc4a469ee/start-node/data/0/outputs.pb
      phase: SUCCEEDED
      updatedAt: "2023-09-27T08:14:27.730373Z"
    id:
      executionId:
        domain: development
        name: f43853b3ae7cc4a469ee
        project: dask-testing
      nodeId: start-node
    inputUri: s3://my-s3-bucket/metadata/propeller/dask-testing-development-f43853b3ae7cc4a469ee/start-node/data/inputs.pb
    metadata:
      specNodeId: start-node
- child_nodes:
  - node_exec:
      closure:
        createdAt: "2023-09-27T08:14:30.079744Z"
        phase: SKIPPED
        updatedAt: "2023-09-27T08:14:30.079926Z"
      id:
        executionId:
          domain: development
          name: f43853b3ae7cc4a469ee
          project: dask-testing
        nodeId: n0-0-end-node
      inputUri: s3://my-s3-bucket/metadata/propeller/dask-testing-development-f43853b3ae7cc4a469ee/n0/data/0/end-node/inputs.pb
      metadata:
        retryGroup: "0"
        specNodeId: end-node
  - node_exec:
      closure:
[...]

PR #4766, opened 2024-01-23
[BUG] subworkflow timeout propagation

pvditt:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/4087_
_https://github.com/flyteorg/flyte/issues/4765_

- When a subworkflow's task fails due to reaching node-active-deadline, the workflow does not fail. The workflow continues to run after the timed out task gets aborted and getting set to "UNKNOWN" state

- When a task time's out due to reaching node-active-deadline, the task's status gets set to "UNKNOWN"

- Have subworkflows handle TimedOut state 

- Transform EPhaseTimedout -> NodeExecution_TIMED_OUT when creating node execution events

## How was this patch tested?

Ran a couple workflows. Would be a decent effort to add unit testing - can add if needed.

### Setup process

Test subworkflow timeout propagation (will hit node-active-deadline on retry)

```
@task(
    timeout=timedelta(seconds=10),
    retries=1,
)
def demo_task():
    time.sleep(3000)



@task
def demo_task_1():
    time.sleep(3000)


@workflow
def subworkflow():
    demo_task()
    demo_task_1()


@workflow
def demo_workflow():
    subworkflow()
```

To isolate test to persisting aborted node on TimeOut:

Set node-active-deadline to small value in Flyte config:

```
propeller:
    node-config:
        default-deadlines:
            node-active-deadline: 10s
```
```
@task(
   retries=1,
)
def demo_task():
   time.sleep(3000)


@workflow
def demo_workflow():
   demo_task()
```

### Screenshots

image

### DIFF
--- a/flytepropeller/pkg/controller/nodes/executor.go
+++ b/flytepropeller/pkg/controller/nodes/executor.go
@@ -1341,7 +1341,7 @@ func (c *nodeExecutor) HandleNode(ctx context.Context, dag executors.DAGStructur
 
 	if currentPhase == v1alpha1.NodePhaseTimingOut {
 		logger.Debugf(ctx, "node timing out")
-		if err := c.Abort(ctx, h, nCtx, "node timed out", true); err != nil {
+		if err := c.Abort(ctx, h, nCtx, "node timed out", false); err != nil {
 			return interfaces.NodeStatusUndefined, err
 		}
 
--- a/flytepropeller/pkg/controller/nodes/executor.go
+++ b/flytepropeller/pkg/controller/nodes/executor.go
@@ -1341,7 +1341,7 @@ func (c *nodeExecutor) HandleNode(ctx context.Context, dag executors.DAGStructur
 
 	if currentPhase == v1alpha1.NodePhaseTimingOut {
 		logger.Debugf(ctx, "node timing out")
-		if err := c.Abort(ctx, h, nCtx, "node timed out", false); err != nil {
+		if err := c.Abort(ctx, h, nCtx, "node timed out", true); err != nil {
 			return interfaces.NodeStatusUndefined, err
 		}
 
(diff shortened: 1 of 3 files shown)


flytepropeller/pkg/controller/nodes/executor.go
```diff
@@ -1341,7 +1341,7 @@ func (c *nodeExecutor) HandleNode(ctx context.Context, dag executors.DAGStructur
 
 	if currentPhase == v1alpha1.NodePhaseTimingOut {
 		logger.Debugf(ctx, "node timing out")
-		if err := c.Abort(ctx, h, nCtx, "node timed out", false); err != nil {
+		if err := c.Abort(ctx, h, nCtx, "node timed out", true); err != nil {
```
hamersaw, line 1344: Does this mean that in the UI it shows up as `ABORTED` for the node phase? I believe there is a `TIMED_OUT` node phase in idl too, I'm wondering if there was a regression along the lines where we lost that. If the UI supports timed out, it might be better to send that?

flytepropeller/pkg/controller/nodes/subworkflow/subworkflow.go
```diff
@@ -88,6 +88,25 @@ func (s *subworkflowHandler) handleSubWorkflow(ctx context.Context, nCtx interfa
 		return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoRunning(nil)), nil
 	}
 
+	if state.HasTimedOut() {
```
hamersaw, line 91: Can you confirm this works with the workflow failure policy set to execute as many nodes as possible? Just want to make sure we're covering this if we weren't previously.

flytepropeller/pkg/controller/nodes/executor.go
pvditt: great call out. There's an issue with transforming node exec events. Updating this

flytepropeller/pkg/controller/nodes/subworkflow/subworkflow.go
pvditt: Just tested setting `failure_policy=WorkflowFailurePolicy.FAIL_AFTER_EXECUTABLE_NODES_COMPLETE`
and it works as expected. Updated the screenshot above showing another task in the subworkflow completing after a task times out.

hamersaw: perfect, thanks!

hamersaw (approved):

PR #4774, opened 2024-01-24
Dynamic log links

### LINKED ISSUE #4773: [Core feature] Dynamic log links

We need a way to indicate from a task that additional log links are required. 

This is useful, for example, in the case of [flyin](https://docs.flyte.org/en/latest/flytesnacks/examples/flyin_plugin/vscode.html), where users can specify the port used in the VSCode server spawn in the task. Allowing for the task to emit a log link that exposes the VScode server on a specific port.

Other plugins and extensions can make use of this feature as well, like extensions that can benefit from other information present in the task template to generate log links. 

We should be able to configure dynamic log links in the logs configuration, like so:
```
plugins:
  logs:
    dynamic-log-links-enabled: true
    dynamic-log-links:
      flyin: http://flyin.vs:{{ .taskConfig.port }}/{{ .podName }}
```

Notice the way the the configuration passed from the task is accessed, in this case assume that the task is setting `port` in its task template, so in order to access that in the template we use the prefix `.taskConfig` in the template.

The `dynamic-log-links` is a map of string to `TemplateLogPlugin`, where the key is the name of the dynamic log link, which is used to identify which logs should be applied in the case of a specific task. For example, consider the case of 2 dynamic logs links being enabled in the configuration, i.e.:
```
plugins:
  logs:
    dynamic-log-links-enabled: true
    dynamic-log-links:
      abc: http://foo.bar/{{ .taskConfig.V1 }}
      def: http://baz/{{ .taskConfig.V2 }}
```

and the task only specifies `def` in its list of dynamic log links:
```
@task(dynamic_log_link="def")
def t():
  ...
```

(I'm abusing the notation a little bit, since `dynamic_log_link` is not a field of the `task` decorator at the moment) Assume that by setting `dynamic_log_link="def"` I'm setting a field in `t`'s task template to indicate that I want `def` to be generated as a dynamic log link. In the backend, we use that information to ensure that only `def` is applied to executions of `t`.

The idea is that at registration time tasks are going to use a comma-separated property called `link_type` set in the task template metadata and we're going to use that to control which dynamic log links are going to produced for that task.

I implemented a version of this idea in https://github.com/flyteorg/flyte/pull/4610, but that failed short when it came to extensibility.

_No response_

- [X] Yes

- [X] Yes

eapolinario:
## Tracking issue
Closes #4773 

We need a way to specify dynamic log links from tasks as defined in the linked tracking issue. This PR implements this idea. Essentially operators are going to add a section to their config files for specific dynamic log links. For example, we have flyin as a plugin that would benefit from this feature.

We add a new subsection to the plugin section in the propeller config file that lets users specify dynamic plugins. The template uris for these dynamic plugins can be configured similarly to the usual plugins, the only difference that specific fields can be plucked from the task template to be used in the template uri. Take this example:
```
plugins:
  logs:
    dynamic-log-links-enabled: true
    dynamic-log-links:
      vscode: http://flyteinteractve.vs:{{ .taskConfig.port }}/{{ .podName }}
```

Notice how in the template uri for the `vscode` dynamic plugin we access the port (which is defined in a task template metadata) using the prefix `.taskConfig`, whereas the usual template variables are also available.

## How was this patch tested?

Unit tests and tested locally in a sandbox.


flyteplugins/go/tasks/logs/config.go
```diff
@@ -28,8 +28,8 @@ type LogConfig struct {
 	StackdriverLogResourceName string              `json:"stackdriver-logresourcename" pflag:",Name of the logresource in stackdriver"`
 	StackDriverTemplateURI     tasklog.TemplateURI `json:"stackdriver-template-uri" pflag:",Template Uri to use when building stackdriver log links"`
 
-	IsFlyinEnabled   bool                `json:"flyin-enabled" pflag:",Enable Log-links to flyin logs"`
-	FlyinTemplateURI tasklog.TemplateURI `json:"flyin-template-uri" pflag:",Template Uri to use when building flyin log links"`
+	IsDynamicLogLinksEnabled bool                           `json:"dynamic-log-links-enabled" pflag:",Enable dynamic log links"`
```
EngHabu, line 31: Why is this needed?

flyteplugins/go/tasks/pluginmachinery/tasklog/plugin.go
```diff
@@ -65,8 +69,9 @@ type Plugin interface {
 }
 
 type TemplateLogPlugin struct {
-	DisplayName   string                     `json:"displayName" pflag:",Display name for the generated log when displayed in the console."`
-	TemplateURIs  []TemplateURI              `json:"templateUris" pflag:",URI Templates for generating task log links."`
-	MessageFormat core.TaskLog_MessageFormat `json:"messageFormat" pflag:",Log Message Format."`
-	Scheme        TemplateScheme             `json:"scheme" pflag:",Templating scheme to use. Supported values are Pod and TaskExecution."`
+	DisplayName         string                     `json:"displayName" pflag:",Display name for the generated log when displayed in the console."`
+	TemplateURIs        []TemplateURI              `json:"templateUris" pflag:",URI Templates for generating task log links."`
+	DynamicTemplateURIs []DynamicTemplateURI       `json:"dynamicTemplateUris" pflag:",Dynamic URI Templates for generating task log links."`
```
EngHabu, line 74: Why do these need to be different?

flyteplugins/go/tasks/pluginmachinery/tasklog/template.go
```diff
@@ -194,31 +200,53 @@ func (input Input) templateVarsForScheme(scheme TemplateScheme) TemplateVars {
...
+	}
+	linkType := config["link_type"]
+	if linkType == "" {
+		return nil
+	}
+	dynamicLogLinkTypes := []string{}
+	for _, linkType := range strings.Split(linkType, ",") {
+		// NB: we are going to grandfather in flyin.
+		if linkType == "vscode" {
+			dynamicLogLinkTypes = append(dynamicLogLinkTypes, "flyin")
```
EngHabu, line 219: Why special case flyin? is `vscode` not enough?

flyteplugins/go/tasks/logs/config.go
eapolinario: I was just following the pattern we use for the other log types (for example k8s). 

I can see a case of someone having complex template uris that they want to temporarily disable. Do you think we should abandon that convention and rely on the configured log links instead?

flyteplugins/go/tasks/pluginmachinery/tasklog/plugin.go
eapolinario: A `DynamicTemplateURI` has a kind that we use to [match](https://github.com/flyteorg/flyte/pull/4774/files#diff-a7959cb7781e8c67a136b74e9051e5f64bf9e18eb236043de3fd4c35b9188407R241) with the values passed via task template. The use case is if we have multiple dynamic log links configured, but the task only enables a subset of them.

flyteplugins/go/tasks/pluginmachinery/tasklog/template.go
eapolinario: Setting the link type to `vscode` is a hack. We should standardize on the name of the plugins (including `flyin`).

EngHabu: Can we get rid of vscode then?

flyteplugins/go/tasks/logs/config.go
eapolinario: good call. I'll drop them (so they will only show up if configured).

flyteplugins/go/tasks/pluginmachinery/tasklog/template.go
eapolinario: `vscode` is already being used in [flyin](https://github.com/flyteorg/flytekit/blob/master/plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/decorator.py#L487). We can make the change in flytekit (and I think we should), but this is going to break existing uses.

eapolinario:
As we talked about yesterday, this next batch of updates implements:
- remove enabled flag
- do not hardcode `port`
  - In fact, we now let users match any value set in the task template config
- Use `vscode` as the name of the log
- Add name to TemplateLogPlugin to help distinguish which plugin is used
  -Remove =DynamicTemplateURI=
- Get rid of Schemes

flyteplugins/go/tasks/pluginmachinery/tasklog/template.go
kumare3: @eapolinario do not use `flyin` as LinkedIn prefers FlyteInteractive

flyteplugins/go/tasks/pluginmachinery/tasklog/template.go
```diff
@@ -76,48 +80,36 @@ func replaceAll(template string, vars TemplateVars) string {
...
+		TemplateVar{defaultRegexes.PodName, input.PodName},
+		TemplateVar{defaultRegexes.PodUID, input.PodUID},
+		TemplateVar{defaultRegexes.Namespace, input.Namespace},
+		TemplateVar{defaultRegexes.ContainerName, input.ContainerName},
+		TemplateVar{defaultRegexes.ContainerID, containerID},
+		TemplateVar{defaultRegexes.Hostname, input.HostName},
+	)
+	if gotExtraTemplateVars {
+		vars = append(vars, input.ExtraTemplateVars...)
+	}
```
EngHabu, line 111: We already did this in line 89, right?

flyteplugins/go/tasks/pluginmachinery/tasklog/template.go
```diff
@@ -172,9 +164,9 @@ func (input Input) templateVarsForScheme(scheme TemplateScheme) TemplateVars {
 				},
 			)
 		}
-		if gotExtraTemplateVars {
-			vars = append(vars, input.ExtraTemplateVarsByScheme.TaskExecution...)
-		}
+	}
+	if gotExtraTemplateVars {
+		vars = append(vars, input.ExtraTemplateVars...)
```
EngHabu, line 169: ditto

flyteplugins/go/tasks/pluginmachinery/tasklog/template.go
```diff
@@ -191,34 +183,53 @@ func (input Input) templateVarsForScheme(scheme TemplateScheme) TemplateVars {
 		},
 	)
 
+	// Add values from task template config as dynamic regexes (i.e. templates prefixed by .taskConfig)
+	for key, value := range input.TaskTemplate.GetConfig() {
+		if value != "" {
+			vars = append(vars, TemplateVar{MustCreateDynamicLogRegex(key), value})
+		}
+	}
```
EngHabu, line 191: That looks expensive. Are we not able to use capture groups as we discussed?

flyteplugins/go/tasks/logs/logging_utils.go
```diff
@@ -84,39 +85,45 @@ func (t templateLogPluginCollection) GetTaskLogs(input tasklog.Input) (tasklog.O
...
-		} else {
-			plugins = append(plugins, tasklog.TemplateLogPlugin{DisplayName: "Flyin Logs", Scheme: tasklog.TemplateSchemeFlyin, TemplateURIs: []tasklog.TemplateURI{fmt.Sprintf("https://flyin.%s/logs/{{ .namespace }}/{{ .podName }}/{{ .containerName }}/{{ .containerId }}", cfg.GCPProjectName)}, MessageFormat: core.TaskLog_JSON})
-		}
+	for logLinkType, dynamicLogLink := range cfg.DynamicLogLinks {
+		dynamicPlugins = append(
+			dynamicPlugins,
+			tasklog.TemplateLogPlugin{
+				Name:        logLinkType,
+				DisplayName: fmt.Sprintf("%s logs", logLinkType),
+				DynamicTemplateURIs: []tasklog.TemplateURI{
```
EngHabu, line 120: Is this field still needed? can't we use `TemplateURIs`...

eapolinario: This simplifies the distinction between task logs produced by dynamic log links vs regular log links.

For example, let's say we enable one dynamic log link in propeller. How do we make sure that dynamic log links only show up for tasks that have the right bits set in their task templates?

flyteplugins/go/tasks/logs/logging_utils.go
```diff
@@ -84,39 +85,45 @@ func (t templateLogPluginCollection) GetTaskLogs(input tasklog.Input) (tasklog.O
...
-			plugins = append(plugins, tasklog.TemplateLogPlugin{DisplayName: "Flyin Logs", Scheme: tasklog.TemplateSchemeFlyin, TemplateURIs: []tasklog.TemplateURI{cfg.FlyinTemplateURI}, MessageFormat: core.TaskLog_JSON})
-		} else {
-			plugins = append(plugins, tasklog.TemplateLogPlugin{DisplayName: "Flyin Logs", Scheme: tasklog.TemplateSchemeFlyin, TemplateURIs: []tasklog.TemplateURI{fmt.Sprintf("https://flyin.%s/logs/{{ .namespace }}/{{ .podName }}/{{ .containerName }}/{{ .containerId }}", cfg.GCPProjectName)}, MessageFormat: core.TaskLog_JSON})
-		}
+	for logLinkType, dynamicLogLink := range cfg.DynamicLogLinks {
+		dynamicPlugins = append(
+			dynamicPlugins,
+			tasklog.TemplateLogPlugin{
+				Name:        logLinkType,
+				DisplayName: fmt.Sprintf("%s logs", logLinkType),
```
EngHabu, line 119: nit: It won't always be "logs", like `vscode logs` isn't the right description, `wandb logs` maybe... right? I think just `logLinkType` is fine... alternatively, maybe the config should just be the logplugin?
so propeller config would be:
```
plugins:
  logs:
    dynamic-log-links:
      vscode: 
        displayName: "vscode debugger"
        templateURIs:
          - http://flyteinteractve.vs:{{ .taskConfig.port }}/{{ .podName }}
```

flyteplugins/go/tasks/pluginmachinery/tasklog/template.go
```diff
@@ -194,31 +182,51 @@ func (input Input) templateVarsForScheme(scheme TemplateScheme) TemplateVars {
...
 		})
 	}
 
+	for _, dynamicLogLinkType := range getDynamicLogLinkTypes(input.TaskTemplate) {
+		for _, dynamicTemplateURI := range p.DynamicTemplateURIs {
+			if p.Name == dynamicLogLinkType {
+				for _, match := range dynamicLogRegex.FindAllStringSubmatch(dynamicTemplateURI, -1) {
+					if len(match) > 1 {
+						value := input.TaskTemplate.GetConfig()[match[1]]
+						if value != "" {
```
EngHabu, line 217: nit:
```suggestion
						if value, found := input.TaskTemplate.GetConfig()[match[1]]; found && len(value) > 0 {
```

EngHabu: or maybe empty value should be supported, and you only need to check for `found` ?

flyteplugins/go/tasks/logs/logging_utils.go
eapolinario: pflags is not able to generate code for this. (I'm not familiar with the code, but from a cursory look it only parses [these](https://github.com/flyteorg/flyte/blob/master/flytestdlib/cli/pflags/api/generator.go#L30-L39) types).

If I define `LogConfig` like this:
```
type DynamicLogLink struct {
	TemplateURI tasklog.TemplateURI `json:"template-uri" pflag:",Template Uri to use when building log links"`
	DisplayName string              `json:"display-name" pflag:",Display name for the generated log when displayed in the console."`
}

// LogConfig encapsulates plugins' log configs
type LogConfig struct {
	...

	DynamicLogLinks map[string]*DynamicLogLink `json:"dynamic-log-links" pflag:",Map of dynamic log links"`

	Templates []tasklog.TemplateLogPlugin `json:"templates" pflag:"-,"`
}
```

I get [this](https://github.com/flyteorg/flyte/blob/master/flytestdlib/cli/pflags/api/generator.go#L102) error:
```
...
INFO[0000] [dynamic-log-links] is of a map type with default value []. 
Error: map of type [*github.com/flyteorg/flyte/flyteplugins/go/tasks/logs.DynamicLogLink] is not supported. Only basic slices or slices of json-unmarshalable types are supported
```

EngHabu: 1. Why not use TemplateLogPlugin :-D? 
2. Just exclude it from pflags:
    ```
	DynamicLogLinks map[string]*DynamicLogLink `json:"dynamic-log-links" pflag:"-,Map of dynamic log links"`
    ```

flyteplugins/go/tasks/pluginmachinery/tasklog/template.go
eapolinario: Empty string is now valid.

flyteplugins/go/tasks/logs/logging_utils.go
```diff
@@ -84,39 +85,43 @@ func (t templateLogPluginCollection) GetTaskLogs(input tasklog.Input) (tasklog.O
...
-			plugins = append(plugins, tasklog.TemplateLogPlugin{DisplayName: "Flyin Logs", Scheme: tasklog.TemplateSchemeFlyin, TemplateURIs: []tasklog.TemplateURI{fmt.Sprintf("https://flyin.%s/logs/{{ .namespace }}/{{ .podName }}/{{ .containerName }}/{{ .containerId }}", cfg.GCPProjectName)}, MessageFormat: core.TaskLog_JSON})
-		}
+	for logLinkType, dynamicLogLink := range cfg.DynamicLogLinks {
+		dynamicPlugins = append(
+			dynamicPlugins,
+			tasklog.TemplateLogPlugin{
+				Name:                logLinkType,
+				DisplayName:         dynamicLogLink.DisplayName,
+				DynamicTemplateURIs: dynamicLogLink.TemplateURIs,
+				MessageFormat:       core.TaskLog_JSON,
```
EngHabu, line 121: Does it not have a `MessageFormat` field?

flyteplugins/go/tasks/pluginmachinery/tasklog/template.go
```diff
@@ -10,9 +10,17 @@ import (
 )
 
 func MustCreateRegex(varName string) *regexp.Regexp {
+	// TODO: Is this regex correct? Why are we allowing $? Shouldn't it be {{\s*\.%s\s*}}?
+	// Also, case insensitive matching is enabled. Is this correct?
 	return regexp.MustCompile(fmt.Sprintf(`(?i){{\s*[\.$]%s\s*}}`, varName))
 }
 
+var dynamicLogRegex = regexp.MustCompile(`(?i){{\s*.taskConfig[\.$]([a-zA-Z_]+)\s*}}`)
```
EngHabu, line 18: nit: Name it more specifically... something like `taskConfigVarRegex` or something...

flyteplugins/go/tasks/pluginmachinery/tasklog/template.go
```diff
@@ -194,31 +182,50 @@ func (input Input) templateVarsForScheme(scheme TemplateScheme) TemplateVars {
...
 			Uri:           replaceAll(templateURI, templateVars),
 			Name:          p.DisplayName + input.LogName,
 			MessageFormat: p.MessageFormat,
 		})
 	}
 
+	for _, dynamicLogLinkType := range getDynamicLogLinkTypes(input.TaskTemplate) {
+		for _, dynamicTemplateURI := range p.DynamicTemplateURIs {
+			if p.Name == dynamicLogLinkType {
+				for _, match := range dynamicLogRegex.FindAllStringSubmatch(dynamicTemplateURI, -1) {
```
EngHabu, line 214: nit: Would be nice to support this for the "regular" logs as well...

EngHabu (dismissed):
Thank you ❤️ ❤️

flyteplugins/go/tasks/pluginmachinery/tasklog/template.go
eapolinario: I modified all uses of `flyin` to `FlyteInteractive`.

flyteplugins/go/tasks/pluginmachinery/tasklog/template.go
eapolinario: It will be easy to rearrange this logic when we find a good use-case for regular logs.

EngHabu (approved):

PR #4808, opened 2024-01-31
Flyte-core add missing nodeSelector values

ddl-ebrown:
- The syncresources (cluster resource sync) deployment never exposed a way to configure a standard nodeSelector. Set one up.

 - The propeller webhook should use the defined propeller nodeSelector to configure itself

## Tracking issue


davidmirror-ops (approved):
Thank you. I see this change to be useful in multi-cluster deployments, among other scenarios.

PR #4809, opened 2024-01-31
Flyte-core define pod and container securityContext

ddl-ebrown:
Like #4785 does for the agent, this defines the security context for flyte-core pods / containers.

 - Define secure defaults for all pod / container specs:

   * runAsNonRoot: true
   * capabilities: drop: ['ALL']
   * allowPrivilegeEscalation: false
   * seLinuxOptions: type: spc_t

   This is required in many locations where policy enforcement agents
   may be installed (like OPA Gatekeeper) which may otherwise prevent
   deployments from launching.

   The hard work of making sure the containers run as non-0 uids seems
   to have already been done given all containers are already specifying
   a runAsUser value of 1000 or 1001, so this should hopefully just be a
   little more hardening around restricting kernel permissions /
   enforcement within the container runtime.

   These are generally considered standard / secure default settings and
   are not currently made configurable given these services are all
   owned by Flyte

## Tracking issue

## How was this patch tested?

In local testing, all pods seem to be behaving correctly with these additional restrictions 

### DIFF
--- a/charts/flyte-core/templates/admin/deployment.yaml
+++ b/charts/flyte-core/templates/admin/deployment.yaml
@@ -19,9 +19,12 @@ spec:
       labels: {{ include "flyteadmin.podLabels" . | nindent 8 }}
     spec:
       securityContext:
+        runAsNonRoot: true
         fsGroup: 65534
         runAsUser: 1001
         fsGroupChangePolicy: "Always"
+        seLinuxOptions:
+          type: spc_t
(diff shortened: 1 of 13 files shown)


davidmirror-ops (commented):
This one looks very interesting. @ddl-ebrown have you already deployed the chart with these settings?

davidmirror-ops (approved):
Thank you!

PR #4823, opened 2024-02-02
change propeller compiler test

wild-endeavor:
Update how the propeller compile test works.

* Removed all yaml files.
* Commented out the `branch` directory test since walking through it, it didn't actually do anything.  Leaving it there as a comment for now.
* Some of the tests were literally testing for what https://github.com/golang/protobuf/issues/1121 advises against testing for.  That is, we were serializing proto objects to yaml and checking that they string matched the saved yaml files.  Removed these tests (which is also why the yaml files were deleted.)
* However, since it's possible to unmarshal from json, we _can_ use the existing json files to reconstruct a proto object and do a proto equals.
* The last check at the bottom of the loop however is not a proto file, it's the Flyte workflow crd, so convert to a new json file, and then make the tester compare strings where multiple spaces have all been converted to one space.


flyteadmin/pkg/manager/impl/validation/execution_validator_test.go
```diff
@@ -98,7 +98,8 @@ func TestValidateExecInputsWrongType(t *testing.T) {
 		lpRequest.Spec.FixedInputs,
 		lpRequest.Spec.DefaultInputs,
 	)
-	assert.EqualError(t, err, "invalid foo input wrong type. Expected simple:STRING , but got simple:INTEGER ")
+	expected := "invalid foo input wrong type. Expected simple:STRING , but got simple:INTEGER"
+	assert.Equal(t, regex(err.Error()), regex(expected))
```
eapolinario, line 102: where is `regex` defined?

eapolinario (approved):

PR #4836, opened 2024-02-05
Fix asterisk in cron table being rendered as list item

neverett:

The asterisks in the cron table in the [Schedules doc](https://docs.flyte.org/en/latest/concepts/schedules.html#cron-expression) are rendering as list items instead of asterisks.

Simply escaping the asterisk didn't work, since it's in a table, so I turned the rst table into an HTML table and that seemed to fix the issue.

## Docs link

https://flyte--4836.org.readthedocs.build/en/4836/concepts/schedules.html

### DIFF
--- a/docs/concepts/schedules.rst
+++ b/docs/concepts/schedules.rst
@@ -24,37 +24,116 @@ Let's now look at how schedules can be defined through cron_expression_ or rate_
 
 Cron Expression
 ---------------
-Cron expression strings use :ref:`this <cron_expression_table>` syntax. They are validated at launch plan registration time.
-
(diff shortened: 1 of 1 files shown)


docs/concepts/schedules.rst
```diff
@@ -95,6 +176,7 @@ Below is an example with duration in `days`.
 
 )
 
+.. _rate_unit:
 
 Rate Unit
```
neverett, line 181: This is outside the scope of this PR, but I don't understand the difference between this section and the "Fixed rate schedules" section above.

ppiegaze (approved):

Issue #4487, opened 2023-11-27
[Housekeeping] Use single configuration option for maximum blobstore size


hamersaw:
### Describe the issue

Currently there are multiple configuration values for the maximum size for a blobstore file between I/O writing, dynamic closure offloading, etc.

### What if we do not do this?

Users frequently run into issues with increasing the maximum size with unclear instructions on what values need to change.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #4852, opened 2024-02-07
[House Keeping] deprecate MaxDatasetSizeBytes propeller config in favor of GetLimitMegabytes storage config

pvditt:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/4487_

- decrease confusion when adjusting the maximum size for storing a blob file

- deprecate max-output-size-bytes propeller config in favor of maxDownloadMBs storage config 


neverett (approved):
Docs changes look good to me, thanks!

hamersaw (approved):
Does it make sense to go a level deeper here and strip out the [MaxDatasetSizeBytes in NodeExecutionContext](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/nodes/node_exec_context.go#L202-L204)? This is just used to restrict file sizes and now that we are setting it to the thing as blobstore limits we are doing this check twice: (1) blobstore read and (2) `maxPayloadSize` in `RemoteFileOutputReader`. OK either way, but should at least mention that there is some dead code here.

pvditt:
@hamersaw yup that makes sense. I'll make a follow up PR to handle that.

PR #4872, opened 2024-02-09
Add composite index for executions

pablocasares:

We are experiencing high load on query:
```SELECT * FROM "executions" WHERE executions.execution_project = $1 AND executions.execution_domain = $2 AND executions.phase in ($3) AND executions.execution_created_at >= $4 AND executions.execution_created_at <= $5 AND executions.state = $6 LIMIT $7```

It takes 95% of the db load. In our setup it is roughly called ~800k times per hour. The average execution time for the query is 13ms. It sometimes hits the pkey index but some other times it has to do a scan taking around 150ms. 

We added a composite index on `execution_project, execution_domain, phase,  execution_created_at and state` and the avg execution time went down to 0.01ms. 

That query was also taking most of our memory as most shared cache hits were done by it. After adding the change the query went from ~180k/s shared block accessed to ~200/s (roughly a factor of 1000). This is also helping to use the memory for other queries.

This change adds a composite index in executions table to avoid doing scans.

## How was this patch tested?

The index was manually created in the db.

### Screenshots
image

Load by time (first row is the mentioned query)
image

### DIFF
--- a/flyteadmin/pkg/repositories/config/migrations.go
+++ b/flyteadmin/pkg/repositories/config/migrations.go
@@ -1180,6 +1180,16 @@ var NoopMigrations = []*gormigrate.Migration{
 			return tx.AutoMigrate(&Execution{})
 		},
 	},
+	//Add composite index for executions table on execution_project, execution_domain, phase, created_at and state
+	{
+		ID: "2024-02-09-executions-composite-idx",
+		Migrate: func(tx *gorm.DB) error {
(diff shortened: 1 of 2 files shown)


katrogan:
hi @pablocasares thanks for this change, I'm curious if you're running an out of band process that generates these queries? from inspecting the console I don't see may queries issued to the execution table that filter on phase (but do see that they query on created_at DESC)

eapolinario:

PR #4907, opened 2024-02-16
Increase default memory and CPU resource

MortalHappiness:
## Tracking issue
None

When running using `pyflyte run`, the tasks are killed due to out of memory.

Increase the default memory and CPU in `flyte-single-binary-local.yaml`

### DIFF
--- a/flyte-single-binary-local.yaml
+++ b/flyte-single-binary-local.yaml
@@ -85,3 +85,11 @@ storage:
       access_key_id: minio
       secret_key: miniostorage
   container: my-s3-bucket
+
+task_resources:
+  defaults:
+    cpu: 500m
+    memory: 500Mi
+  limits:
+    cpu: 800m
+    memory: 800Mi


pingsutw (approved):

Future-Outlier (approved):
before (with no settings in `flyte-single-binary-local-dev.yaml`
image
after with settings in `flyte-single-binary-local-dev.yaml`
image
Failure case (request more than limit memory)
```python
from flytekit.loggers import logger
from flytekit import Resources

@task(requests=Resources(cpu="800m", mem="900Mi"))
def say_hello() -> str:
    return "hello"

@workflow
def wf() -> str:
    return say_hello()
```
```yaml
task_resources:
  defaults:
    cpu: 500m
    memory: 500Mi
  limits:
    cpu: 800m
    memory: 800Mi
```
image


It works well, LGTM, thank you both.

PR #4874, opened 2024-02-09
Agents protocol refactor

pingsutw:
## Tracking issue
https://github.com/flyteorg/flyte/issues/3936

## Why are the changes needed?
Add ExecuteTaskSync to support running tasks synchronously

## What changes were proposed in this pull request?
Proto updates
- Add `rpc ExecuteTaskSync`, which sends the streaming request and gets the streaming response
- `GetTaskLogs` returns a streaming response
- Add more fields to TaskExecutionMetadata for container task
- Deprecate `string deprecated_task_type`, and use `TaskType task_type` instead

Add syncAgentClient to the agent plugin.

## How was this patch tested?
unit tests/sandbox

### Screenshots
image

## Check all the applicable boxes 

- [x] I updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] All commits are signed-off.

## Related PRs
https://github.com/flyteorg/flytekit/pull/2146

## Docs link
NA

PR #4922, opened 2024-02-20
Fix flyte-core rendering when flyteagent enabled

ddl-ebrown:
- Flyte agent now tries to render a specific set of podLabels, so assumes that there is an empty `podLabels: {}` set. This change to values.yaml for flyte-core ensures empty values are there.

 - Fixes:

   Error: template: flyte-core/charts/flyteagent/templates/agent/deployment.yaml:17:17: executing "flyte-core/charts/flyteagent/templates/agent/deployment.yaml" at <include "flyteagent.podLabels" .>: error calling include: template: flyte-core/templates/_helpers.tpl:122:16: executing "flyteagent.podLabels" at <.Values.flyteagent.podLabels>: nil pointer evaluating interface {}.podLabels

## Tracking issue

### DIFF
--- a/charts/flyte-core/values.yaml
+++ b/charts/flyte-core/values.yaml
@@ -279,6 +279,8 @@ flyteagent:
         defaultAgent:
           endpoint: "dns:///flyteagent.flyte.svc.cluster.local:8000"
           insecure: true
+  # -- Labels for flyteagent pods
+  podLabels: {}
 
 #
 # FLYTEPROPELLER SETTINGS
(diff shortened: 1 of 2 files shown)


charts/flyte-core/values.yaml
```diff
@@ -279,6 +279,8 @@ flyteagent:
         defaultAgent:
           endpoint: "dns:///flyteagent.flyte.svc.cluster.local:8000"
           insecure: true
+  # -- Labels for flyteagent pods
+  podLabels: {}
```
pingsutw, line 283: cc @ddl-ebrown  I remove `podLabels` from the helper in this [PR](https://github.com/flyteorg/flyte/pull/4874/files#diff-e2eb39162dcb726f53d87b4c2ef9d5af3b2450ac5f30508deef6bf0cded88a2bL105-L118).

Also, would you mind taking a look at the new agent API? We want to make it as simple as possible, so make some change to it. https://github.com/flyteorg/flytekit/pull/2146

pingsutw: > just as a workaround for the chart install issue
yes

> I added it recently in https://github.com/flyteorg/flyte/pull/4756)

I see, got it. Let's merge your PR first.

Issue #4762, opened 2024-01-23
[Core feature] Pyflyte 'task with different structure' should show delta


katrogan:
### Motivation: Why do you think this is important?

It's difficult as a user when I try to register Flyte inventory and get back 

```
details: task with different structure already exists with id resource_type:TASK project:"system" domain:"production" name:"task-name" version:"1234" 
```

but I can't tell what's changed in my local version of the task.

Since we already store the registered task template and have the current serialized version of the task template we're trying to register, it would be super slick if flytekit (pyflyte) could compute what parameters in the template changed across registration and highlight that to the end user. (same for workflows and launch plans)

### Goal: What should the final outcome look like, ideally?

Should be straightforward to identify what has changed across registration versions when the back-end blocks a re-registration attempt

### Describe alternatives you've considered 

Current error response code user experience

###  Propose: Link/Inline OR Additional context 

Much like we use json marshalling to compute the hash, we could just json serialization to compute the delta here.

_No response_

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

hampusrosvall:
You can assign this to me!

austin362667:
Hi, @katrogan

Wanna making sure of few things before starting a PR. Thank you~

### 1. Delta what?

> compute what parameters in the template changed across registration

By *parameters*, did you mean like the `task` / `workflow` / `launch plan`  source code itself (different code structure but with the same id)  **OR** in `container.args` dict from  **registration request** bad rpc response e.g.,

Input Request:
```json
{ 
  "id": { 
    "resourceType": "TASK", 
    "project": "flytesnacks",
    "domain": "development",
    "name": "workflow.say_hello",
    "version": "1234"
  },
  "spec": {
    "template": {
      "id": {
        "resourceType": "TASK",
        "project": "flytesnacks",
        "domain": "development",
        "name": "workflow.say_hello",
        "version": "1234"
      },
      "type": "python-task",
      "metadata": {
        "runtime": {
          "type": "FLYTE_SDK",
          "version": "1.10.7",
          "flavor": "python"
        },
        "retries": {}
      },
      "interface": {
        "inputs": {},
        "outputs": {
          "variables": {
            "o0": {
              "type": {
                "simple": "STRING"
              },
              "description": "o0"
            }
          }
        }
      },
      "container": {
        "image": "basics:v1",
        "args": [
          "pyflyte-fast-execute",
[...]

PR #4924, opened 2024-02-21
[flyteadmin] Show diff structure when re-registration two different task with same ids

austin362667:
## Tracking issue

https://github.com/flyteorg/flyte/issues/4762

To highlight diff(delta part) error message for `NewTaskExistsDifferentStructureError()`

#### 1. Modify `task.proto` to contain error reasoning info.

```proto
// The task id is already used and the structure is different
message TaskErrorExistsDifferentStructure {
    core.Identifier id = 1;
    core.TaskTemplate old_spec = 2;
    core.TaskTemplate new_spec = 3;
}
```

```proto
// The task id is already used with an identical sctructure
message TaskErrorExistsIdenticalStructure {
    core.Identifier id = 1;
}

// When a CreateTaskRequest fails due to matching id
message CreateTaskFailureReason {
    oneof reason {
        TaskErrorExistsDifferentStructure exists_different_structure = 1;
        TaskErrorExistsIdenticalStructure exists_identical_structure = 2;
    }
}
```

> Actually it may not be necessary here, if we only return the diff structure as error msg string.

#### 2. Refactor `NewFlyteAdminErrorf()` in `task_manager.go` to `errors.go`
 - `NewTaskExistsDifferentStructureError()`
 - `NewTaskExistsIdenticalStructureError()`

###
### Setup process

1. register two different task code file, but with the same `project`, `domain`, `name`, `version`.

### Screenshots


austin362667:
- successful register task
- re-register same task with same id (pass)
  Screenshot 2024-02-21 at 3 15 37 PM
- re-register different(modified) task with same id (failed)
  Screenshot 2024-02-21 at 3 16 12 PM
- re-register different(modified) task with same id (failed), with diff error message
Screenshot 2024-02-21 at 3 18 31 PM

austin362667:
What about `Workflow`, `Laucnplan`?

e.g.,
Modify a task name(function name) used by another workflow

austin362667:
> 2\. extend the error [message](https://github.com/flyteorg/flyte/blob/962d45827646f106265630a0fb1d9ae74968bc32/flyteidl/protos/flyteidl/admin/workflow.proto#L77-L79) to include structured data about what in the spec has changed

What if we just show `diff` as string, in order to prevent modify too much `task` proto buff?

Future-Outlier (commented):
It looks well to me.
## first time registration
image

## second time registration
image

flyteadmin/pkg/manager/impl/task_manager.go
```diff
@@ -92,11 +92,13 @@ func (t *TaskManager) CreateTask(
 	existingTask, err := util.GetTaskModel(ctx, t.db, request.Spec.Template.Id)
 	if err == nil {
 		if bytes.Equal(taskDigest, existingTask.Digest) {
-			return nil, errors.NewFlyteAdminErrorf(codes.AlreadyExists,
-				"identical task already exists with id %s", request.Id)
+			return nil, errors.NewTaskExistsIdenticalStructureError(ctx, &request)
 		}
-		return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument,
-			"task with different structure already exists with id %v", request.Id)
+		t2, err2 := util.GetTask(ctx, t.db, *request.Id)
```
Future-Outlier, line 97:
```suggestion
		request_task, err := util.GetTask(ctx, t.db, *request.Id)
```

flyteadmin/pkg/manager/impl/task_manager.go
```diff
@@ -92,11 +92,13 @@ func (t *TaskManager) CreateTask(
...
+		if err2 != nil {
+			return nil, errors.NewTaskExistsDifferentStructureError(ctx, &request, t2.Closure.CompiledTask.Template, compiledTask.Template)
```
Future-Outlier, line 99:
```suggestion
			return nil, errors.NewTaskExistsDifferentStructureError(ctx, &request, request_task.Closure.CompiledTask.Template, compiledTask.Template)
```

flyteadmin/pkg/manager/impl/task_manager.go
```diff
@@ -92,11 +92,13 @@ func (t *TaskManager) CreateTask(
...
+		}
+		return nil, errors.NewTaskExistsDifferentStructureError(ctx, &request, t2.Closure.CompiledTask.Template, compiledTask.Template)
```
Future-Outlier, line 101:
```suggestion
		return nil, errors.NewTaskExistsDifferentStructureError(ctx, &request, request_task.Closure.CompiledTask.Template, compiledTask.Template)
```

Future-Outlier:
I think overall looks well, however I am not familiar with flyteadmin's codebase.
cc @katrogan, can you help review? thank you.

austin362667:
1. 
```python
import time

@task(container_image="image1")
def say_hello() -> str:
    time.sleep(1)
    return "Hello, World!"

@workflow
def wf() -> str:
    res = say_hello()
    return res

if __name__ == "__main__":
    print(f"Running hello_world_wf() {wf()}")
```

2. 
```python
import time

@task(container_image="image2")
def say_hello() -> str:
    time.sleep(1)
    return "Hello, World!"

@workflow
def wf() -> str:
    res = say_hello()
    return res

if __name__ == "__main__":
    print(f"Running hello_world_wf() {wf()}")

```

Screenshot 2024-02-24 at 4 23 51 AM

flyteadmin/pkg/errors/errors.go
```diff
@@ -109,8 +111,68 @@ func NewIncompatibleClusterError(ctx context.Context, errorMsg, curCluster strin
...
+			result := fmt.Sprintf("%s: %s -> %s\t", obj.Path, obj.Value, val.Value)
+			results = append(results, result)
+		}
+	}
+	return results
+}
+
+func NewTaskExistsDifferentStructureError(ctx context.Context, request *admin.TaskCreateRequest, oldSpec *core.TaskTemplate, newSpec *core.TaskTemplate) FlyteAdminError {
+	errorMsg := "task with different structure already exists:\n"
+	// omit source code file object storage path
```
katrogan, line 132: args/2 is still used for the digest computation, right? any reason to omit it here?

flyteadmin/pkg/errors/errors.go
```diff
@@ -109,8 +111,68 @@ func NewIncompatibleClusterError(ctx context.Context, errorMsg, curCluster strin
...
+	diff, _ := jsondiff.Compare(oldSpec, newSpec, jsondiff.Ignores("/Target/Container/args/2"))
```
katrogan, line 133: why do we need two Compare calls here?

it looks like according to https://pkg.go.dev/github.com/wI2L/jsondiff#Differ.Compare this should output the diff between src & tgt

flyteadmin/pkg/errors/errors.go
```diff
@@ -109,8 +111,68 @@ func NewIncompatibleClusterError(ctx context.Context, errorMsg, curCluster strin
...
+
+	statusErr, transformationErr := NewFlyteAdminError(codes.InvalidArgument, errorMsg).WithDetails(&admin.CreateTaskFailureReason{
+		Reason: &admin.CreateTaskFailureReason_ExistsDifferentStructure{
+			ExistsDifferentStructure: &admin.TaskErrorExistsDifferentStructure{
+				Id: request.Id,
+			},
+		},
+	})
+	if transformationErr != nil {
+		logger.Panicf(ctx, "Failed to wrap grpc status in type 'Error': %v", transformationErr)
```
katrogan, line 146: nit, I know that the existing code panics but can we instead dump an error log?

flyteadmin/pkg/errors/errors.go
```diff
@@ -109,8 +111,68 @@ func NewIncompatibleClusterError(ctx context.Context, errorMsg, curCluster strin
...
+	for _, obj := range jsonArray2 {
+		if val, ok := map1[obj.Path]; ok {
+			result := fmt.Sprintf("%s: %s -> %s\t", obj.Path, obj.Value, val.Value)
+			results = append(results, result)
+		}
+	}
+	return results
+}
+
+func NewTaskExistsDifferentStructureError(ctx context.Context, request *admin.TaskCreateRequest, oldSpec *core.TaskTemplate, newSpec *core.TaskTemplate) FlyteAdminError {
```
katrogan, line 130: mind adding tests for these new functions? https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/errors/errors_test.go

flyteadmin/pkg/manager/impl/task_manager.go
```diff
@@ -89,14 +89,16 @@ func (t *TaskManager) CreateTask(
...
-		if bytes.Equal(taskDigest, existingTask.Digest) {
-			return nil, errors.NewFlyteAdminErrorf(codes.AlreadyExists,
-				"identical task already exists with id %s", request.Id)
+		if bytes.Equal(taskDigest, existingTaskModel.Digest) {
+			return nil, errors.NewTaskExistsIdenticalStructureError(ctx, &request)
 		}
-		return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument,
-			"task with different structure already exists with id %v", request.Id)
+		existingTask, err2 := util.GetTask(ctx, t.db, *request.Id)
+		if err2 != nil {
```
katrogan, line 98: we already have `existingTaskModel` fetched above, can we just call https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/repositories/transformers/task.go#L41 here instead?

katrogan (commented):
this looks so great! thank you for updating workflow validation as part of this task too

flyteadmin/pkg/errors/errors.go
austin362667: Yes, we still hash the whole task for digest.
The reason to omit file path shown in error message is that it already trapped in "different structure with same id" error, saying code file must be different. So it's trivial to hide the path changed information. cc @pingsutw

austin362667: Another concern of mine is what if the file path value (`"s3://my-s3-bucket/flytesnacks/..."`) being stored with different `key path` in the future, for example, `args/3`.

flyteadmin/pkg/errors/errors.go
austin362667: Because [jsondiff.Compare](https://pkg.go.dev/github.com/wI2L/jsondiff#Differ.Compare) returns a json patch file meaning a series of operations of that json.

It only show the modified `target` json other than `source` json. Since we're talking about `diff` message. It must show the difference between changes from `source` to `target`.

`diff = jsondiff.Compare(oldTemplate, newTemplate)`
`rdiff = jsondiff.Compare(newTemplate, oldTemplate`
which prefix `r` means `reverse`.

Finally `compareJsons()` zip the `diff` and its complements `rdiff`.

flyteadmin/pkg/errors/errors.go
austin362667: Sure thing!

flyteadmin/pkg/errors/errors.go
austin362667: Sure thing!

flyteadmin/pkg/manager/impl/task_manager.go
austin362667: Great! Thanks for the heads-up.

flyteidl/protos/flyteidl/admin/task.proto
```diff
@@ -69,3 +69,21 @@ message TaskClosure {
     // Time at which the task was created.
     google.protobuf.Timestamp created_at = 2;
 }
+
+// The task id is already used and the structure is different
+message TaskErrorExistsDifferentStructure {
+    core.Identifier id = 1;
```
pingsutw, line 75: I think we can remove it

austin362667: Sure! thx

austin362667:
1. `Task`: `macos` to `ubuntu`
```python
import time
from flytekit.image_spec import ImageSpec

image_spec = "macos"

@task(container_image=image_spec)
def say_hi() -> str:
    time.sleep(1)
    return "Hello, World!"


@workflow
def wf() -> str:
    res = say_hi()
    return res

if __name__ == "__main__":
    print(f"Running hello_world_wf() {wf()}")
```
```python
import time
from flytekit.image_spec import ImageSpec

image_spec = "ubuntu"

@task(container_image=image_spec)
def say_hi() -> str:
    time.sleep(1)
    return "Hello, World!"
    

@workflow
def wf() -> str:
    res = say_hi()
    return res

if __name__ == "__main__":
    print(f"Running hello_world_wf() {wf()}")
 ```
 Screenshot 2024-02-27 at 7 53 15 PM


2. `Workflow`: `say_hi()` to `say_hey()`
```python
import time
from flytekit.image_spec import ImageSpec

image_spec = "ubuntu"

@task(container_image=image_spec)
def say_hi() -> str:
    time.sleep(1)
    return "Hello, World!"


@workflow
def wf() -> str:
    res = say_hi()
    return res

if __name__ == "__main__":
    print(f"Running hello_world_wf() {wf()}")
```
```python
import time
from flytekit.image_spec import ImageSpec

image_spec = "ubuntu"

@task(container_image=image_spec)
def say_hey() -> str:
    time.sleep(1)
    return "Hello, World!"
    

@workflow
def wf() -> str:
    res = say_hey()
    return res

if __name__ == "__main__":
    print(f"Running hello_world_wf() {wf()}")
 ```
Screenshot 2024-02-27 at 7 53 46 PM

File: flyteadmin/pkg/errors/errors.go [hunk 10]
```diff
@@ -109,8 +111,47 @@ func NewIncompatibleClusterError(ctx context.Context, errorMsg, curCluster strin
...
+			result := fmt.Sprintf("\t\t- %v: %v -> %v", obj.Path, obj.Value, val.Value)
+			results = append(results, result)
+		}
+	}
+	return results
+}
+
+func NewTaskExistsDifferentStructureError(ctx context.Context, request *admin.TaskCreateRequest, oldSpec *core.CompiledTask, newSpec *core.CompiledTask) FlyteAdminError {
+	errorMsg := "task with different structure already exists:\n"
+	diff, _ := jsondiff.Compare(oldSpec, newSpec)
```
katrogan, line 132: since in golang map ordering is random/never guaranteed, I'm worried that converting the specs to bytes here might lead to non deterministic byte arrays

when we compute the digest, we use [pbhash](https://github.com/flyteorg/flyte/blob/2adaf612348c9b3d83f4640b0eedba6e2ab33f4c/flyteadmin/pkg/manager/impl/util/digests.go#L30) to marshal the protos to json 

should we do the same here before calling Compare()?

alternatively, do you mind double checking that jsondiff.Compare() returns no diff for multiple calls comparing 2 identical pb objects that contain maps in the message?

flyteadmin/pkg/manager/impl/launch_plan_manager.go
```diff
@@ -88,12 +88,14 @@ func (m *LaunchPlanManager) CreateLaunchPlan(
...
 	if err == nil {
 		if bytes.Equal(existingLaunchPlanModel.Digest, launchPlanDigest) {
-			return nil, errors.NewFlyteAdminErrorf(codes.AlreadyExists,
-				"identical launch plan already exists with id %s", request.Id)
+			return nil, errors.NewLaunchPlanExistsIdenticalStructureError(ctx, &request)
 		}
-
-		return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument,
-			"launch plan with different structure already exists with id %v", request.Id)
+		existingLaunchPlan, err2 := transformers.FromLaunchPlanModel(existingLaunchPlanModel)
```
katrogan, line 93:
```suggestion
		existingLaunchPlan, transformerErr := transformers.FromLaunchPlanModel(existingLaunchPlanModel)
```

flyteadmin/pkg/manager/impl/launch_plan_manager.go
```diff
@@ -88,12 +88,14 @@ func (m *LaunchPlanManager) CreateLaunchPlan(
...
+		if err2 != nil {
+			return nil, err2
```
katrogan, line 95: nit, can we error log in this case?

flyteadmin/pkg/manager/impl/task_manager.go
```diff
@@ -89,14 +89,16 @@ func (t *TaskManager) CreateTask(
...
 	if err == nil {
-		if bytes.Equal(taskDigest, existingTask.Digest) {
-			return nil, errors.NewFlyteAdminErrorf(codes.AlreadyExists,
-				"identical task already exists with id %s", request.Id)
+		if bytes.Equal(taskDigest, existingTaskModel.Digest) {
+			return nil, errors.NewTaskExistsIdenticalStructureError(ctx, &request)
 		}
-		return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument,
-			"task with different structure already exists with id %v", request.Id)
+		existingTask, err2 := transformers.FromTaskModel(*existingTaskModel)
```
katrogan, line 97:
```suggestion
		existingTask, transformerErr := transformers.FromTaskModel(*existingTaskModel)
```

katrogan (commented):
looks great, just one question about map comparison determinism

flyteadmin/pkg/errors/errors.go
austin362667: can!
I've carefully double checked that `jsondiff.Compare()` is a identical operation. Also added the test for this function.

katrogan: thanks @austin362667 for the super fast turnaround!

in this case, i was more concerned about identical maps not showing a false diff - would you mind double checking that? otherwise PR looks great!!

austin362667: done

katrogan (dismissed):
this so awesome, thank you!

Future-Outlier (approved):
LGTM, Let's merge it when all tests passed

PR #4925, opened 2024-02-21
Fix broken gpu resource override when using pod templates

fg91:

```py
@task(
    requests=Resources(cpu="11", mem="11Gi", gpu="11"),
)
def foo():
    ...

@workflow
def test_wf():
    foo().with_overrides(requests=Resources(cpu="13", mem="13Gi", gpu="13"), limits=Resources(cpu="13", mem="13Gi", gpu="13"))
```

The resulting task pod - as expected - uses the following resources:

```yaml
resources:
  limits:
    cpu: "13"
    memory: 13Gi
    nvidia.com/gpu: "13"
  requests:
    cpu: "13"
    ephemeral-storage: 2Gi
    memory: 13Gi
    nvidia.com/gpu: "13"
```

If I now add `pod_template=PodTemplate(labels={"foo": "bar"})` to the task decorator, I would expect the same pod resources but the creation of the pod actually fails:

```
E0221 13:50:14.805583  427772 workers.go:103] error syncing 'development/f454d2131ec484221b70': failed at Node[n0]. RuntimeExecutionError: failed during plugin execution, caused by: failed to execute handle for plugin [container]: [Invalid] failed to create resource, caused by: Pod "f454d2131ec484221b70-n0-0" is invalid: spec.containers[0].resources.requests: Invalid value: "11": must be equal to nvidia.com/gpu limit
```

This PR fixes this.

The bug happens as follows:

* When not using the pod template, [here](https://github.com/flyteorg/flyte/blob/22b0005b4d889fce7ac9f201e227f8654f3bbeac/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L275), from our example above, we create a container with `cpu: 11` , `memory: 11Gi`, and `nvidia.com/gpu: 11` (both for requests and limits).
* However, when using a pod template, [here](https://github.com/flyteorg/flyte/blob/22b0005b4d889fce7ac9f201e227f8654f3bbeac/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L293), we retrieve a pod spec from the task proto which sets `cpu: 11` , `memory: 11Gi`, and `gpu: 11` (not `nvidia.com/gpu`!) as requests and which doesn't have limits.  This pod spec was built on the flytekit side [here](https://github.com/flyteorg/flytekit/blob/d1d1bd911cacb36d732ecb57977ece7931076b50/flytekit/core/utils.py#L97).

In the latter case, when merging the container resources with the resource overrides [here](https://github.com/flyteorg/flyte/blob/22b0005b4d889fce7ac9f201e227f8654f3bbeac/flyteplugins/go/tasks/pluginmachinery/flytek8s/container_helper.go#L322), we end up with the following requests:

```yaml
cpu: 13
memory: 13Gi
nvidia.com/gpu: 13
gpu: 11  # This value should have been overridden but wasn't because of the wrong resource name
```

Finally, [here](https://github.com/flyteorg/flyte/blob/22b0005b4d889fce7ac9f201e227f8654f3bbeac/flyteplugins/go/tasks/pluginmachinery/flytek8s/container_helper.go#L173), we override `nvidia.com/gpu: 13` with the wrong value `11` and delete the `gpu` key from the resource requirements.

(The actual error message above states that the gpu request and limit are not the same. The root cause for this is what I described here.)

**Proposed fix:**

I propose to fix this by cleaning up the wrong gpu resource name *before* merging the resource overrides.

## How was this patch tested?

I ran propeller with my proposed change. The resource overriding in the example above works as expected, also when using the pod template.

**I will still adapt the tests**

### DIFF
--- a/flyteplugins/go/tasks/pluginmachinery/flytek8s/container_helper_test.go
+++ b/flyteplugins/go/tasks/pluginmachinery/flytek8s/container_helper_test.go
@@ -237,17 +237,35 @@ func TestApplyResourceOverrides_OverrideGpu(t *testing.T) {
 	gpuRequest := resource.MustParse("1")
 	overrides := ApplyResourceOverrides(v1.ResourceRequirements{
 		Requests: v1.ResourceList{
-			resourceGPU: gpuRequest,
+			ResourceNvidiaGPU: gpuRequest,
 		},
 	}, v1.ResourceRequirements{}, assignIfUnset)
 	assert.EqualValues(t, gpuRequest, overrides.Requests[ResourceNvidiaGPU])
 
 	overrides = ApplyResourceOverrides(v1.ResourceRequirements{
 		Limits: v1.ResourceList{
-			resourceGPU: gpuRequest,
+			ResourceNvidiaGPU: gpuRequest,
 		},
 	}, v1.ResourceRequirements{}, assignIfUnset)
 	assert.EqualValues(t, gpuRequest, overrides.Limits[ResourceNvidiaGPU])
+}
+
+func TestSanitizeGPUResourceRequirements(t *testing.T) {
+	gpuRequest := resource.MustParse("4")
+	requirements := v1.ResourceRequirements{
+		Requests: v1.ResourceList{
+			resourceGPU: gpuRequest,
+		},
+	}
+
+	expectedRequirements := v1.ResourceRequirements{
+		Requests: v1.ResourceList{
+			ResourceNvidiaGPU: gpuRequest,
+		},
+	}
+
+	SanitizeGPUResourceRequirements(&requirements)
+	assert.EqualValues(t, expectedRequirements, requirements)
 }
 
 func TestMergeResources_EmptyIn(t *testing.T) {
@@ -601,6 +619,42 @@ func TestAddFlyteCustomizationsToContainer_Resources(t *testing.T) {
 		assert.True(t, container.Resources.Limits.Cpu().Equal(resource.MustParse("10")))
 		assert.True(t, container.Resources.Requests.Memory().Equal(resource.MustParse("2")))
 		assert.True(t, container.Resources.Limits.Memory().Equal(resource.MustParse("2")))
+	})
+	t.Run("ensure gpu resource overriding works for tasks with pod templates", func(t *testing.T) {
+		container := &v1.Container{
+			Command: []string{
+				"{{ .Input }}",
+			},
+			Args: []string{
+				"{{ .OutputPrefix }}",
+			},
+			Resources: v1.ResourceRequirements{
+				Requests: v1.ResourceList{
+					resourceGPU: resource.MustParse("2"), // Tasks with pod templates request resource via the "gpu" key
+				},
+				Limits: v1.ResourceList{
+					resourceGPU: resource.MustParse("2"),
+				},
+			},
+		}
+
+		overrideRequests := v1.ResourceList{
+			ResourceNvidiaGPU: resource.MustParse("4"), // Resource overrides specify the "nvidia.com/gpu" key
+		}
+
+		overrideLimits := v1.ResourceList{
+			ResourceNvidiaGPU: resource.MustParse("4"),
+		}
+
+		templateParameters := getTemplateParametersForTest(&v1.ResourceRequirements{
+			Requests: overrideRequests,
+			Limits:   overrideLimits,
+		}, &v1.ResourceRequirements{})
+
+		err := AddFlyteCustomizationsToContainer(context.TODO(), templateParameters, ResourceCustomizationModeMergeExistingResources, container)
+		assert.NoError(t, err)
+		assert.Equal(t, container.Resources.Requests[ResourceNvidiaGPU], overrideRequests[ResourceNvidiaGPU])
+		assert.Equal(t, container.Resources.Limits[ResourceNvidiaGPU], overrideLimits[ResourceNvidiaGPU])
 	})
 }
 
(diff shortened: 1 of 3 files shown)


flyteplugins/go/tasks/pluginmachinery/flytek8s/container_helper_test.go
```diff
@@ -237,19 +237,37 @@ func TestApplyResourceOverrides_OverrideGpu(t *testing.T) {
 	gpuRequest := resource.MustParse("1")
 	overrides := ApplyResourceOverrides(v1.ResourceRequirements{
 		Requests: v1.ResourceList{
-			resourceGPU: gpuRequest,
+			ResourceNvidiaGPU: gpuRequest,
```
hamersaw, line 240: Is it safe to make this change? IIUC the arguments to `ApplyResourceOverrides` will have "gpu" rather than "nvidia/gpu" currently?

flyteplugins/go/tasks/plugins/array/awsbatch/transformer.go
```diff
@@ -95,6 +95,8 @@ func FlyteTaskToBatchInput(ctx context.Context, tCtx pluginCore.TaskExecutionCon
 	if platformResources == nil {
 		platformResources = &v1.ResourceRequirements{}
 	}
+
+	flytek8s.SanitizeGPUResourceRequirements(res)
```
hamersaw, line 99: This function is called internally in the `ApplyResourceOverrides` below I think. Are we then sanitizing in both locations?

hamersaw (commented):
tbh I'm having some difficulty following this through, which is probably an indicator that it is far too complex. Obviously this is a bug, but I'm concerned about breaking existing functionality. Do you think there are cases where moving the sanitization after adjusting would break? I'm going to have to dive much deeper and ask to get a few eyes here.

flyteplugins/go/tasks/plugins/array/awsbatch/transformer.go
fg91: In the current state of the PR, the sanitation doesn't happen in `ApplyResourceOverrides` anymore but is moved to the new `SanitizeGPUResourceRequirements` which is called in `AddFlyteCustomizationsToContainer` directly before calling `ApplyResourceOverrides`.
This is why I think `SanitizeGPUResourceRequirements` would have to be called here as well.

fg91:
> tbh I'm having some difficulty following this through, which is probably an indicator that it is far too complex. Obviously this is a bug, but I'm concerned about breaking existing functionality. Do you think there are cases where moving the sanitization after adjusting would break? I'm going to have to dive much deeper and ask to get a few eyes here.

Hey @hamersaw,
thank you for taking a look. I understand your concerns and difficulty to follow this through, it definitely took me a while to figure out what is happening here ...

I first considered sanitizing the wrong `"gpu"` resouce name in the pod template spec [here](https://github.com/flyteorg/flyte/blob/22b0005b4d889fce7ac9f201e227f8654f3bbeac/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L293) directly after unmarshalling it from the protobuf. But if I'm not overlooking something, the logic [here](https://github.com/flyteorg/flyte/blob/22b0005b4d889fce7ac9f201e227f8654f3bbeac/flyteplugins/go/tasks/pluginmachinery/flytek8s/container_helper.go#L172) was supposed to do exactly that - only that it is done too late, causing the GPU override issue. 
If we additionally patched the resource name directly after unmarshalling the pod template spec, without touching the old logic, we'd leave the old logic as dead code if I'm not mistaken. This is why I feel it would be nicer to confirm whether this change is safe.

We've been running a propeller version with this change in prod for a week now and haven't observed any issues but I certainly don't claim that we have 100% coverage of all scenarios. For instance I cannot test aws batch.

Do you think there is somebody else who can judge this change as well?
I'd also be happy to jump on a quick call with you or sb else to step through the master and modified code with a debugger while screensharing. This might make it easier to understand what is happening.

hamersaw:
Really appreciate the patience here. Dove through this, and think I understand everything now. What I'm not seeing is how just the container works - neither case successfully overrides the gpu resource in my dev environment. Here's what I'm seeing:

When we call the [ApplyResourceOverrides function](https://github.com/flyteorg/flyte/blob/c77e2c32ef83dbc8015e771d27def194eba5a36c/flyteplugins/go/tasks/pluginmachinery/flytek8s/container_helper.go#L137) the container example has "nvidia/gpu" for the resource type because the [ToK8sResourceRequirements function ](https://github.com/flyteorg/flyte/blob/c77e2c32ef83dbc8015e771d27def194eba5a36c/flyteplugins/go/tasks/pluginmachinery/flytek8s/container_helper.go#L216)converts this internally. Alternatively the PodTemplate means that it will be "gpu".

Regardless, both are being updated to "nvidia/gpu" as you [mentioned because of this logic](https://github.com/flyteorg/flyte/blob/c77e2c32ef83dbc8015e771d27def194eba5a36c/flyteplugins/go/tasks/pluginmachinery/flytek8s/container_helper.go#L173-L183).

I'm seeing neither being updated by the overrides because in the [call to adjustResourceRequirement](https://github.com/flyteorg/flyte/blob/c77e2c32ef83dbc8015e771d27def194eba5a36c/flyteplugins/go/tasks/pluginmachinery/flytek8s/container_helper.go#L186) the `gpuResourceName` is "nvidia/gpu" and the resources in `platformResources` are "gpu" - so they do not apply the override.

It would be good to understand where I'm diverging from what you saw here.

hamersaw:
Disregard everything ^^, I had my GPU platform limits set to 1 in the configuration so it was lowering the value each time. Diving back in here.

hamersaw (approved):
I think I've convinced myself that this is correct. The only surprising thing here is how long it took me to come to this conclusion.

Basically, leaving dangling "gpu" resource names caused things to override because sometimes we worked on the sanitized names (ie. "nvidia/gpu") and others on non-sanitized (ie. "gpu"). So this PR sanitizes everything before any processing. It looks like we can guarantee platform resources to be sanitized as well.

fg91:
> The only surprising thing here is how long it took me to come to this conclusion. Great PR!

Definitely took me a while to come to this fix, the logic is just very complicated :)

> So this PR sanitizes everything before any processing. It looks like we can guarantee platform resources to be sanitized as well.

Yes, exactly 👌

PR #4782, opened 2024-01-26
Replace `Storage` To `Ephemeral Storage` in Helm Chart

Future-Outlier:
## Tracking issue
Fixes https://github.com/flyteorg/flyte/issues/3728

## Why are the changes needed?
Make users will not request storage when using flyte.

## What changes were proposed in this pull request?
replace `storage` to `ephemeralStorage` in [charts/flyte-core/values.yaml](https://github.com/flyteorg/flyte/pull/4782/files#diff-638f077159e277b61cb799423fd14a2c496b05799d05fb998823984bf979344a)

## How was this patch tested?
you can check the request name from flyteadmin here, that's why the key is `ephemeralStorage`.
https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/runtime/interfaces/task_resource_configuration.go#L5-L11

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] All commits are signed-off.

## Related PRs
https://github.com/flyteorg/flytekit/pull/2078

PR #4929, opened 2024-02-21
Remove eph storage

wild-endeavor:
## Tracking issue
Delete the ephemeral storage settings [introduced](https://github.com/flyteorg/flyte/pull/4782/files) a while back.  These numbers don't make sense, but more broadly, this is not a setting that we want to set.

Note that this change (and the original PR) only affects the `flyte` and `flyte-core` Helm charts, not the single binary Helm chart.


pingsutw (approved):

eapolinario (approved):
Thank you.

PR #4938, opened 2024-02-23
[BUG] Datacatalog put errors don't get persisted

pvditt:
## Tracking issue

Noticed this while checking handling parity across catalog vs the new cacheservice. 

1) When non-serialized identical cached tasks run in parallel the tasks that finish after the first one will fail writing to datacatalog but will show up as "The result of this execution was written to cache" in console. This occurs since [this](https://github.com/flyteorg/flyte/blob/fdaa0216dd849e461e6a493ea2f99e99a30e3447/flytepropeller/pkg/controller/nodes/catalog/datacatalog/datacatalog.go#L254) doesn't bubble up the error
2) When  cache update/put errors out, the status only shows up as cache disabled in console. 
The PutFailureStatus is set [here](https://github.com/flyteorg/flyte/blob/fdaa0216dd849e461e6a493ea2f99e99a30e3447/flytepropeller/pkg/controller/nodes/cache.go#L229)
This cache status isn't updated as admin needs [catalog metadata to be set](https://github.com/flyteorg/flyte/blob/fdaa0216dd849e461e6a493ea2f99e99a30e3447/flyteadmin/pkg/repositories/transformers/node_execution.go#L157) 
Metadata isn't set on failures in the client: [ex](https://github.com/flyteorg/flyte/blob/fdaa0216dd849e461e6a493ea2f99e99a30e3447/flytepropeller/pkg/controller/nodes/catalog/datacatalog/datacatalog.go#L234) 

- add CatalogMetadata to catalog status on update/put errors

Note: additional context such as "Put failed due to AlreadyExist" isn't shown in console

### Screenshots
New:
image
Old:
image


flytepropeller/pkg/controller/nodes/catalog/datacatalog/datacatalog.go
```diff
@@ -251,20 +251,16 @@ func (m *CatalogClient) CreateArtifact(ctx context.Context, key catalog.Key, dat
...
 	_, err = m.client.AddTag(ctx, &datacatalog.AddTagRequest{Tag: tag})
 	if err != nil {
-		if status.Code(err) == codes.AlreadyExists {
-			logger.Warnf(ctx, "Tag %v already exists for Artifact %v (idempotent)", tagName, cachedArtifact.Id)
-		} else {
-			logger.Errorf(ctx, "Failed to add tag %+v for artifact %+v, err: %+v", tagName, cachedArtifact.Id, err)
-			return catalog.Status{}, err
-		}
+		logger.Errorf(ctx, "Failed to add tag %+v for artifact %+v, err: %+v", tagName, cachedArtifact.Id, err)
+		return catalog.Status{}, err
```
hamersaw, line 255: Love the rest of it. Can we leave this as an idempotent operation and not return an error (ie. `PUT_FAILURE`) on already exists? Unless we add a `reason` that says `Artifact already exists` I feel like users are going to see a failure here and think something broke.

pvditt: yup sounds good. I think the easiest approach would be similar to adding a new cache_status ~ALREADY_EXISTS. 

I'll make note to add that when I get back into the console, IDL sync issue for the evict cache status.

hamersaw (approved):

PR #4956, opened 2024-02-26
[RFC] Select cluster execution when triggering an execution

RRap0so:
RFC to propose a new Cluster Execution implementation.


davidmirror-ops:
This is amazing, I can think of several users who'd like to see a better mechanism to handle executions in multi-cluster scenarios.
Thanks!

rfc/core language/4956-cluster-pools.md
```diff
@@ -0,0 +1,100 @@
...
+
+## 5 Drawbacks
+
+Other than the usual drawbacks of implementing a new feature (the introduction of new bugs/problems) we don’t see any potential drawbacks with this implementation. The fields are already exposed but they are only missing the implementation. 
+
+We don’t expect any additional overhead for execution or impact in terms of API for flyteadmin.
+
+## 6 Alternatives
+
+As mentioned above, the only alternative we have implemented today is using the [Execution Cluster Labels](https://docs.flyte.org/en/latest/deployment/deployment/multicluster.html#configure-execution-cluster-labels). Even though we could make this work for our use case it’s not ideal and can bring further problems when we want to route executions to different clusters from the same Project/Domain.
```
sonjaer, line 86:
```suggestion
As mentioned above, the only alternative we have implemented today is using the [Execution Cluster Labels](https://docs.flyte.org/en/latest/deployment/deployment/multicluster.html#configure-execution-cluster-labels). This can bring further problems when we want to route executions to different clusters from the same Project/Domain.
```

rfc/core language/4956-cluster-pools.md
```diff
@@ -0,0 +1,100 @@
...
+## 2 Motivation
+
+The primary motivation for this proposal is to select the clusters where an execution is to be performed dynamically at execution time, without the constant need to overwrite the Execution Cluster Label.
+
+Additionally, we would like to have the possibility of different executions targeting different clusters, even if they are published under the same Project/Domain. This is not entirely feasible using the current Execution Cluster Label.
+
+
+## 3 Proposed Implementation
+
+We plan to use the `cluster_assignement` field (refer to [proto](https://github.com/flyteorg/flyte/blob/master/flyteidl/protos/flyteidl/admin/execution.proto#L312)) in each execution, this is a ClusterAssignment object with a `cluster_pool_name` which would be the field that contains a string with the corresponding ClusterPool ID.
```
fg91, line 24:
```suggestion
We plan to use the `cluster_assignement` field (refer to [proto](https://github.com/flyteorg/flyte/blob/a2a63eb893bdff3c2d1a01242d0d25b73f287b1f/flyteidl/protos/flyteidl/admin/execution.proto#L315)) in each execution, this is a ClusterAssignment object with a `cluster_pool_name` which would be the field that contains a string with the corresponding ClusterPool ID.
```

rfc/core language/4956-cluster-pools.md
```diff
@@ -0,0 +1,100 @@
...
+As mentioned above, the only alternative we have implemented today is using the [Execution Cluster Labels](https://docs.flyte.org/en/latest/deployment/deployment/multicluster.html#configure-execution-cluster-labels). This can bring further problems when we want to route executions to different clusters from the same Project/Domain.
+
+## 7 Potential Impact and Dependencies
+
+We do not see any.
+
+
+## 8 Unresolved questions
+
+What takes priority Labels or Pools?
```
bstadlbauer, line 95: I would vote for labels to take priority as those seem to be more execution specific. I might be missing something though

bstadlbauer (dismissed):
This seems like a great usecase and the required changes shouldn't introduce too much extra complexity 👍

rfc/core language/4956-cluster-pools.md
sonjaer: I think since pools are specified dynamically at execution time I think they should override labels as that is set more "statically"?

RRap0so: Yes the label usage seems more related to  "I need this X type of cluster for this <PROJECT/Domain>".

So to be sure, we first check if there's a label set, then under the clusters in that label if there's a pool. If not then default using the weight. (Similar to what is done today.)

rfc/core language/4956-cluster-pools.md
```diff
@@ -0,0 +1,100 @@
...
+
+We plan to use the `cluster_assignement` field (refer to [proto](https://github.com/flyteorg/flyte/blob/a2a63eb893bdff3c2d1a01242d0d25b73f287b1f/flyteidl/protos/flyteidl/admin/execution.proto#L315)) in each execution, this is a ClusterAssignment object with a `cluster_pool_name` which would be the field that contains a string with the corresponding ClusterPool ID.
+
+In the FlyteAdmin configuration we have a list of `clusterpool_config` with all the different available pools and corresponding clusters.
+
+**Note**: Today there are configurations to assign domains to specific pools, see [PR](https://github.com/flyteorg/flyte/pull/4208) and `cluster_pool` [interface](https://github.com/flyteorg/flyte/blob/962d45827646f106265630a0fb1d9ae74968bc32/flyteadmin/pkg/runtime/interfaces/cluster_pools.go). We propose another configuration to assign Clusters to a clusterPoolName by having another map under cluster object with weights and cluster assignments.
+
+Proposed configuration:
+```
+    clusters:
```
fg91, line 32: As said in the contributors' sync:

Love the proposal, my only wish would be to figure out whether there is a way to simplify this config here by not having to configure a cluster twice, once for the existing project/domain mechanism, once for the new mechanism. Would be nice if the cluster configuration could be shared.

RRap0so: @fg91 I think it should be possible by expanding the `ExecutionParameters` struct. By adding the `ExecutionClusterLabel` to this [struct](https://github.com/flyteorg/flyte/blob/a2a63eb893bdff3c2d1a01242d0d25b73f287b1f/flyteadmin/pkg/workflowengine/interfaces/executor.go#L20) and making sure we pass it when executing it should make things simpler and avoid having double configs.

This will mean a bit of re-writing, since it's late I'll also cross-check my thinking and get back with some certainties next week.

katrogan (commented):
a couple of edge cases come to mind which would be great to clarify
- are cluster pools ever reserved or restricted? can any execution target any cluster pool?
- what happens when a cluster pool AND an execution cluster are specified in an execution spec? 
- what about when a CreateExecution request targets a specific cluster but the hierarchical overrides include a (conflicting) cluster pool (or vice versa) do we always read from the execution spec

also a general question on validation, what happens if flyteadmin is brought up with a config where cluster pools reference non-existent execution clusters? what if a [matchable attribute](https://github.com/flyteorg/flyte/blob/master/flyteidl/protos/flyteidl/admin/matchable_resource.proto#L153) is added with a cluster pool referencing a non-existent cluster pool?

RRap0so:
@fg91 I've gave it a shot in [this PR](https://github.com/flyteorg/flyte/pull/4998). Instead of adding a new field in the `ExecutionTargetSpec` struct, we're reusing the `ClusterAssignment` and adding it there.

This is an alternative to what we propose with the RFC above but keeping the same idea, this was much of the code will be reused and minimal changes are required.

@katrogan  please take a look at the alternative here. This should solve any question with pools vs labels. We will first honor whatever label was first set in the execution and then project.

wild-endeavor:
hey @RRap0so,

late to the party on this one, but i'm not sure I understand.  Could you help clear up some confusion for me please?

Today admin already has the ability to do:

```
      labelClusterMap:
        production:
        - id: flyte-1-k8s
          weight: 0.3
        - id: flyte-2-k8s
          weight: 0.7
```

(And looking at how we do the [weighting](https://github.com/flyteorg/flyte/blob/3ff007a69b02b489d77b83856dd92d55b9808616/flytestdlib/random/weighted_random_list.go#L76), you can do 3 and 7 also.)

And then you can set a project or domain to use the `production` label, and have it randomly pick between them.  Does the behavior you're proposing differ from this?  If so could you explain more please, I'm missing something.

RRap0so:
> hey @RRap0so,
> 
> late to the party on this one, but i'm not sure I understand. Could you help clear up some confusion for me please?
> 
> Today admin already has the ability to do:
> 
> ```
>       labelClusterMap:
>         production:
>         - id: flyte-1-k8s
>           weight: 0.3
>         - id: flyte-2-k8s
>           weight: 0.7
> ```
> 
> (And looking at how we do the [weighting](https://github.com/flyteorg/flyte/blob/3ff007a69b02b489d77b83856dd92d55b9808616/flytestdlib/random/weighted_random_list.go#L76), you can do 3 and 7 also.)
> 
> And then you can set a project or domain to use the `production` label, and have it randomly pick between them. Does the behavior you're proposing differ from this? If so could you explain more please, I'm missing something.

Sure thing @wild-endeavor . This RFC is not changing what is done in terms of weighs and picking the cluster, just introducing another way to select a workflow to run in another cluster at execution time. The problem today is that we can only set a project or domain and we also want to define per execution.

We started this RFC by using the [cluster_assignment](https://github.com/flyteorg/flyte/blob/a2a63eb893bdff3c2d1a01242d0d25b73f287b1f/flyteidl/protos/flyteidl/admin/execution.proto#L315) since it was a field that already existed but wasn't fully implemented.

There's an alternative PR that probably explains in a simpler way what exactly we intend to do and if this is the way forward, happy to update the RFC accordingly.

https://github.com/flyteorg/flyte/pull/4998

wild-endeavor:
oh @RRap0so are you just saying there's no way to pass an execution label at execution kick-off time?

cuz matchable overrides at the workflow level is also possible already, but i agree i'm not seeing the label at execution kickoff time.

RRap0so:
@wild-endeavor exactly! We spotted the cluster assignment and noticed some artifacts about clusterPools so we did the RFC to fully implement the concept but probably the alternative PR is the quickest way without duplicating configs and enough for the use-case.

wild-endeavor:
Sorry for the delay.  In that case, can we just add the [ExecutionClusterLabel](https://github.com/flyteorg/flyte/blob/e07084c16e7dce06c0c88978d0994a2b6fa5a5a7/flyteidl/protos/flyteidl/admin/matchable_resource.proto#L72) object to the [ExecutionSpec](https://github.com/flyteorg/flyte/blob/db1182142e473836f47fa3a8b376a10fc6ef8386/flyteidl/protos/flyteidl/admin/execution.proto#L267) message?  And then have admin respect that as an override?

Will you be at the contributor meeting today?  Maybe we can discuss there.

RRap0so:
> Sorry for the delay. In that case, can we just add the [ExecutionClusterLabel](https://github.com/flyteorg/flyte/blob/e07084c16e7dce06c0c88978d0994a2b6fa5a5a7/flyteidl/protos/flyteidl/admin/matchable_resource.proto#L72) object to the [ExecutionSpec](https://github.com/flyteorg/flyte/blob/db1182142e473836f47fa3a8b376a10fc6ef8386/flyteidl/protos/flyteidl/admin/execution.proto#L267) message? And then have admin respect that as an override?
> 
> Will you be at the contributor meeting today? Maybe we can discuss there.

Sorry it was a bit late for me to attend, so we've done that already if you take a look at this alternative. https://github.com/flyteorg/flyte/pull/4956  if this is the way forward, happy to close this RFC and continue there.

wild-endeavor:
This is the way forward yes.  Can you point me to where the alternative is?  The link to #4956 just points back to this pull request.  Did you already make a PR that adds `ExecutionClusterLabel` to `ExecutionSpec`?

RRap0so:
> This is the way forward yes. Can you point me to where the alternative is? The link to #4956 just points back to this pull request. Did you already make a PR that adds `ExecutionClusterLabel` to `ExecutionSpec`?

@wild-endeavor  🤦  here's the correct link. We've added it in the cluster_assignment to make this a bit more clean but happy to change.

wild-endeavor:
hey @RRap0so this is the [ExecutionClusterLabel message](https://github.com/flyteorg/flyte/blob/3fe45336245f125d832a94b1c5eef5e3b3ba2491/flyteidl/protos/flyteidl/admin/matchable_resource.proto#L72) that we talked about

RRap0so:
We've decided to take the alternative approach of adding the [ExecutionClusterLabel](https://github.com/flyteorg/flyte/blob/3fe45336245f125d832a94b1c5eef5e3b3ba2491/flyteidl/protos/flyteidl/admin/matchable_resource.proto#L72) into the execution spec.

For more details see Issue https://github.com/flyteorg/flyte/issues/5081 and PR https://github.com/flyteorg/flyte/pull/4998

fg91:
@RRap0so would appreciate your input on this 🙏 

Based on [your PR ](https://github.com/flyteorg/flyte/issues/5081) which exposes `ExecutionClusterLabel` when creating an execution, [this](https://github.com/flyteorg/flytekit/pull/2422) PR is exposing this option in `pyflyte run`. Because we really need this functionality as well, I just tested this and can confirm it works - all good here. (Thank you for your work in the backend!)

My question is this:

This PR https://github.com/flyteorg/flytekit/pull/1208 a few months ago exposed `--cluster-pool` in `pyflyte` which is what you @RRap0so proposed in this RFC, correct?.

If I understood you correctly in the discussions in the contrib sync, the cluster pool logic was never fully implemented in the backend, is this correct or did I misunderstand you?

I'm thinking whether all this "cluster pool" logic is dead code that never worked and should be cleaned up.
@davidmirror-ops FYI would be good to revisit this in the contrib sync.

RRap0so:
This is correct as it stands right now there's a "placeholder" but not implemented in the flyteadmin backend.

Even tough we don't use it in the backend, it is possible other plugins (outside the flyteadmin backend) are actually using it? Maybe we can start by opening an issue, remove that code and bring it up the next contrib sync.

Great that it's growing into pyflyte, I'm also looking at potentially adding it to flytectl so might grab some inspiration :)

davidmirror-ops:
Thanks for bringing this up!
We can discuss this further. That code is being used by Union, so should not be deleted. We're looking for a way to better label/signal this type of work.

PR #4969, opened 2024-02-27
Add k8s env froms

### LINKED ISSUE #4968: [Core feature] Allow k8s plugin to pull env from default configMaps / secrets

We are able to specify default environment variables through the k8s plugin, but only those hard-coded into the configuration. Allowing default environment variables to be set via configMap or secret would allow automated systems in our environment to access secrets without embedding them in multiple locations.

The k8s plugin should have an option where additional environment variable sources can be provided, like:
```
k8s:
  default-cpus: 100m
  default-env-vars:
    - MY_FOO: "bar"
  default-env-vars-from-configmaps:
    - "config-map-a"
    - "config-map-b"
  default-env-vars-from-secrets:
    - "secret-a"
    - "secret-b"
```

Inserting environment variables via mutatingWebhooks is the only other method I could see as a basic option, but then I would need a centralized repo of secret values like vault in order to properly store secrets on a per-namespace basis.

_No response_

- [X] Yes

- [X] Yes

neilisaur:
## Tracking issue

Closes #4968 

We are able to specify default environment variables through the k8s plugin, but only those hard-coded into the configuration. Allowing default environment variables to be set via configMap or secret would allow automated systems in our environment to access secrets without embedding them in multiple locations.

Add ability to specify default configMap and secrets to mount on k8s plugin tasks.

## How was this patch tested?

Unit tested, but local builds could not emulate the propeller build process currently.


flyteplugins/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
```diff
@@ -127,6 +127,24 @@ func DecorateEnvVars(ctx context.Context, envVars []v1.EnvVar, taskEnvironmentVa
...
 	}
+	for _, configMapName := range config.GetK8sPluginConfig().DefaultEnvVarsFromConfigMaps {
+		envVars = append(envVars, v1.EnvVar{
+			ValueFrom: &v1.EnvVarSource{
+				ConfigMapKeyRef: &v1.ConfigMapKeySelector{
+					Key: configMapName,
+				},
+			},
+		})
+	}
```
eapolinario, line 138: actually, I gave this a try and the code is not quite correct. In order to load config maps as environment variables you need to specify the keys you're going to load explicitly.

So the config file would look like:
```
default-env-vars-from-configmaps:
  test-config-map:
    - key1
    - key2
  config-map-2:
    - another-key-1
    - another-key-2
```

Which would cause the 4 env vars `key1`, `key2`, `another-key-1`, and `another-key-2` in the pod.

This has the added benefit that you can guarantee that the names do not conflict by visually inspecting the config file (and we could even add some validation there afterwards).

In order to load the config map we need to modify the type of `DefaultEnvVarsFromConfigMaps` to be an dict of list, such that we can do something like:
```
for configMapName, keys := range config.GetK8sPluginConfig().DefaultEnvVarsFromConfigMaps {
    for key := range keys {
        envVars = append(envVars, v1.EnvVar{
            Name: key,
            ValueFrom: &v1.EnvVarSource{
                ConfigMapKeyRef: &v1.ConfigMapKeySelector{
                    Key: key,
                    LocalObjectReference: v1.LocalObjectReference{
                        Name: configMapName,
                    },
                },
            },
        })
    }
}
```

We might need to revisit the namespacing of the env vars (maybe use the config map name as the prefix?).

I didn't investigate the support for secrets, but I'm assuming that it should be pretty similar.

eapolinario (changes requested):
We need a little bit more work to load configmaps and secrets.

flyteplugins/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
neilisaur: Interesting. It's supported in that manner via raw k8s, so I was assuming we'd be able to: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables

eapolinario: That's definitely interesting. The use is slightly different, right? In the case of https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables we have to fill in the `envFrom` field in the container, whereas the support we have today is for the `env` field. Using `envFrom` should be pretty simple though.

The docs don't mention, but I'm not sure if we can the mix `envFrom` and `env`. What happens if there's a clash (i.e. two config maps that define the same key)?

neilisaur: Yeah, envFrom should be the target to the container; I'll see what I can do to dig through this deeper.

For the conflicts, `env` will override anything else specified, so the ultimate priority would be: `env` > `envFrom` > `Dockerfile` builtin ENV for priorities

neilisaur: @eapolinario I've pushed up a change that allows direct `envFrom[]` into k8s pods as originally intended, but I see your point on pulling specific keys on configMaps/secrets into the env array using the pick-and-choose method.

If you're ok with it I've renamed the config keys to be more specific that this is for the envFrom section in k8s to leave room for future `env[].valueFrom.configMapRef` / `env[].valueFrom.secretRef` to be implemented as you described

eapolinario:
@neilisaur , can you fix the lint error and merge master while you're at it?

neilisaur:
> @neilisaur , can you fix the [lint error](https://github.com/flyteorg/flyte/actions/runs/8287460098/job/22682775676?pr=4969#step:4:289) and merge master while you're at it?

@eapolinario done and done! Re-ran make helm as well

neilisaur:
@eapolinario for the Chart version update that's failing is there a specific method of changing it, or should I just update everything under charts/ to v0.1.11?

eapolinario (approved):
@neilisaur , don't worry about that CI check failure. We've started to [validate](https://github.com/flyteorg/flyte/pull/5027) changes to the helm charts only recently, which means that  we shouldn't be blocking PRs using those checks yet.

eapolinario:
Thanks you for your contribution!

flixr:
@neilisaur @eapolinario seems this broke tasks which already have `env_from` set in the pod template!
My secrets from the pod template are not there anymore... had to roll back to flyte 1.11.

I have a pod template something like:
```python
primary_container = V1Container(name="primary")
primary_container.env_from = [V1EnvFromSource(secret_ref=V1SecretEnvSource(name="my-secret))]
ps = V1PodSpec(containers=[primary_container])
pod_template = PodTemplate(pod_spec=ps, primary_container_name="primary")
```

PR #4970, opened 2024-02-28
Report subNode output and error using ExternalResourceInfo

hamersaw:
## Tracking issue
_NA_

Maximizing the utility of maptasks requires proper visibility.

This PR adds reporting of subNode error and output data from ArrayNode executions. This means if a subNode fails, the error will be available independently in the UI and output data will be available before all subNodes complete.

## How was this patch tested?
Ran a series of ArrayNode executions and validated correct `ExternalResourceInfo` for subNode metadata within flyteadmin.

## Related PRs
_NA_

## Docs link
_NA_


flyteadmin/pkg/repositories/transformers/task_execution.go
```diff
@@ -308,16 +310,44 @@ func mergeExternalResource(existing, latest *event.ExternalResourceInfo) *event.
...
 		existing.CacheStatus = latest.CacheStatus
 	}
+	if latest.CatalogKey != nil {
+		existing.CatalogKey = latest.CatalogKey
+	}
 	existing.Logs = mergeLogs(existing.Logs, latest.Logs)
 
-	return existing
+	// set output metadata if exists
+	if len(latest.GetOutputUri()) > 0 || latest.GetError() != nil {
```
pvditt, line 319: would we want to add a check if existing.OutputResult is not set to avoid making the uri, err := common.OffloadLiteralMap call multiple times?

I know above there's checks if existing.Value != latest.Value, but that wouldn't be applicable due to offloaded outputs needing to be fetched to check values. If the outputResult is a permanent/non-changing state, then the existing.OutputResult != nil should be equivalent.

hamersaw: Right, I was relying on the fact this is only ever reached once e2e. But probably better to be explicit about this as you mentioned. Thanks!

pvditt (approved):

pvditt:
looks good after cleaning up the admin tests with that no reset change

PR #4971, opened 2024-02-28
Add extra header to signed url

pingsutw:
- Dataproxy fails to generate a signed URL if a file exists since `metadata.etag` does [not equal contentMD5](https://github.com/flyteorg/flyte/blob/2b709f1c4788fe598f9159acd31f8ae769da1849/flyteadmin/dataproxy/service.go#L83).
- Flytekit fails to upload the same file twice since dataproxy fails to generate signed URL

- Add extra header (`x-amz-meta-flyteContentMD5`, `x-goog-meta-flyteContentMD5`, and `x-ms-meta-flyteContentMD5`) to signed URL 
- The client-side has to add that header to the request as well. (Dataproxy will return a header for the client to use)
- Dataproxy will check if the metadata `flyteContentMD5` is equal to ContentMD5

## How was this patch tested?

```
pyflyte --verbose run --remote test2.py wf --f test/test.csv --d test
```

```python
import pandas as pd

@task()
def print_data_directory(f: FlyteFile, d: FlyteDirectory):
    df = pd.read_csv(f)
    print(d)
    print(df.head())


@workflow
def wf(f: FlyteFile, d: FlyteDirectory):
    print_data_directory(f=f, d=d)
```

### Screenshots
- AWS 
image

- Azure
image

- GCP
image

- minio
image

## Related PRs
Blocked by https://github.com/flyteorg/stow/pull/13

## Docs link
NA


wild-endeavor (approved):

flyteadmin/go.mod
```diff
@@ -229,3 +229,5 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20230905202853-d090da108d2f
 	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.16.2
 )
+
+replace github.com/flyteorg/stow => github.com/pingsutw/stow v0.3.6-0.20240228091138-cafd670a5c8d
```
EngHabu, line 233: just a reminder not to merge with this

flytestdlib/storage/stow_store.go
```diff
@@ -221,12 +227,16 @@ func (s *StowStore) Head(ctx context.Context, reference DataReference) (Metadata
 			// Err will be caught below
 		} else if etag, err := item.ETag(); err != nil {
 			// Err will be caught below
+		} else if metadata, err := item.Metadata(); err != nil {
+			// Err will be caught below
 		} else {
 			t.Stop()
+			contentMD5, _ := metadata[strings.ToLower(stow.FlyteContentMD5)].(string)
```
EngHabu, line 234:
```suggestion
			contentMD5 := metadata[strings.ToLower(stow.FlyteContentMD5)].(string)
```

EngHabu: Can you confirm we get the metadata without the x-amz-... whatever prefix?

flyteadmin/dataproxy/service.go
```diff
@@ -117,23 +117,28 @@ func (s Service) CreateUploadLocation(ctx context.Context, req *service.CreateUp
...
 	}
 
+	logger.Infof(ctx, "CreateSignedURL for", storagePath)
 	resp, err := s.dataStore.CreateSignedURL(ctx, storagePath, storage.SignedURLProperties{
-		Scope:      stow.ClientMethodPut,
-		ExpiresIn:  req.ExpiresIn.AsDuration(),
-		ContentMD5: md5,
+		Scope:       stow.ClientMethodPut,
+		ExpiresIn:   req.ExpiresIn.AsDuration(),
+		ContentMD5:  md5,
```
EngHabu, line 128: nit: can we rename this just for a consistent naming with the read code path? base64digestMD5... or something

flyteadmin/dataproxy/service.go
```diff
@@ -49,11 +48,13 @@ func (s Service) CreateUploadLocation(ctx context.Context, req *service.CreateUp
 	// If it doesn't exist, then proceed as normal.
 
 	if len(req.Project) == 0 || len(req.Domain) == 0 {
+		logger.Errorf(ctx, "project and domain are required parameters")
 		return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument, "project and domain are required parameters")
 	}
 
 	// At least one of the hash or manually given prefix must be provided.
 	if len(req.FilenameRoot) == 0 && len(req.ContentMd5) == 0 {
+		logger.Errorf(ctx, "content_md5 or filename_root is a required parameter")
```
EngHabu, line 57: This shouldn't be an error. `Infof`

flyteadmin/dataproxy/service.go
```diff
@@ -49,11 +48,13 @@ func (s Service) CreateUploadLocation(ctx context.Context, req *service.CreateUp
 	// If it doesn't exist, then proceed as normal.
 
 	if len(req.Project) == 0 || len(req.Domain) == 0 {
+		logger.Errorf(ctx, "project and domain are required parameters")
```
EngHabu, line 51: This shouldn't be an error. `Infof`

pingsutw: sorry, I was debugging

flyteadmin/dataproxy/service.go
pingsutw: done

flytestdlib/storage/stow_store.go
```diff
@@ -221,12 +228,16 @@ func (s *StowStore) Head(ctx context.Context, reference DataReference) (Metadata
 			// Err will be caught below
 		} else if etag, err := item.ETag(); err != nil {
 			// Err will be caught below
+		} else if metadata, err := item.Metadata(); err != nil {
+			// Err will be caught below
 		} else {
 			t.Stop()
+			contentMD5, _ := metadata[strings.ToLower(FlyteContentMD5)].(string)
```
EngHabu, line 235: Can we log the value found here if casting fails...

go.mod
```diff
@@ -229,6 +229,7 @@ replace (
 	github.com/flyteorg/flyte/flyteplugins => ./flyteplugins
 	github.com/flyteorg/flyte/flytepropeller => ./flytepropeller
 	github.com/flyteorg/flyte/flytestdlib => ./flytestdlib
+	github.com/flyteorg/stow => github.com/pingsutw/stow v0.3.6-0.20240229093623-86cb63d6001a
```
EngHabu, line 232: Not to update with the released version

flytestdlib/storage/storage.go
```diff
@@ -34,6 +34,7 @@ type Metadata interface {
 	Exists() bool
 	Size() int64
 	Etag() string
+	ContentMD5() string
```
EngHabu, line 37:
```suggestion
	// ContentMD5 retrieves the value of a special metadata tag added by the system that contains the MD5 of the uploaded file. If there is no metadata attached or that `FlyteContentMD5` key isn't set, ContentMD5 will return empty.
	ContentMD5() string
```

flyteadmin/dataproxy/service.go
```diff
@@ -117,23 +128,33 @@ func (s Service) CreateUploadLocation(ctx context.Context, req *service.CreateUp
...
 	if err != nil {
+		logger.Errorf(ctx, "failed to create shardedStorageLocation. Error %v", err)
 		return nil, errors.NewFlyteAdminErrorf(codes.Internal, "failed to create shardedStorageLocation, Error: %v", err)
 	}
 
+	metadata := make(map[string]string)
+
+	if req.AddMetadata {
+		metadata[storage.FlyteContentMD5] = base64digestMD5
+	}
```
EngHabu, line 139: 1. I think `dataStore.CreateSignedURL()` should be the one that adds this metadata... ideally you want to isolate assumptions here. If the storage package assumes this tag is set and its value is the contentMD5, then it should be the one that adds it as well.

EngHabu: 2. Is the reason for having `req.AddMetadata` to allow backward compatibility? Added more notes in the overall review for how to handle backward compatibility. Discussed this with Eduardo earlier.

flyteadmin/dataproxy/service.go
```diff
@@ -268,6 +289,21 @@ func (s Service) validateCreateDownloadLinkRequest(req *service.CreateDownloadLi
...
+	headers := map[string]string{"Content-Length": strconv.Itoa(len(contentMd5)), "Content-MD5": contentMd5}
+	if addMetadata {
+		if strings.HasPrefix(reference.String(), "s3://") {
+			headers[fmt.Sprintf("x-amz-meta-%s", storage.FlyteContentMD5)] = contentMd5
+		} else if strings.HasPrefix(reference.String(), "gs://") {
+			headers[fmt.Sprintf("x-goog-meta-%s", storage.FlyteContentMD5)] = contentMd5
+		} else if strings.HasPrefix(reference.String(), "abfs://") {
+			headers[fmt.Sprintf("x-ms-meta-%s", storage.FlyteContentMD5)] = contentMd5
+			headers["x-ms-blob-type"] = "BlockBlob" // https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob?tabs=microsoft-entra-id#remarks
+		}
```
EngHabu, line 302: Let's not leak storage layer abstraction into outside layers... no other layers should reason about s3 vs gcs... etc.

EngHabu (commented):
Overall I've two comments:
1. Maintain the right abstractions, `storage` package should be the one handling the different behaviors between different storage backends. If you find yourself adding half the logic within one package and the other half in another package, something smells...
2. Backward compatibility:
  a. Overwriting files in GCS and Azure just never worked... no need to attempt to maintain backward compatibility here.
  b. We should always add the ContentMD5 tag in all storage providers when CreateSignedURL is called
  c. When the SDK calls CreateSignedURL, we should check if the target exists and has that tag set, we should compare the request's MD5 to that value.
  d. When the SDK calls CreateSignedURL, if the target doesn't have MD5 tag set, We should compare it against ETag for S3 and just fail for all others (existing behavior)

go.mod
```diff
@@ -229,6 +229,7 @@ replace (
 	github.com/flyteorg/flyte/flyteplugins => ./flyteplugins
 	github.com/flyteorg/flyte/flytepropeller => ./flytepropeller
 	github.com/flyteorg/flyte/flytestdlib => ./flytestdlib
+	github.com/flyteorg/stow => github.com/pingsutw/stow v0.3.6-0.20240305225123-f9af3a975dd4
```
EngHabu, line 232: Are you updating this?

pingsutw: updated

EngHabu (approved):

Issue #4593, opened 2023-12-13
[Housekeeping] Ensure accurate ActiveNodeExecutions and ActiveTaskExecutions metrics


hamersaw:
### Describe the issue

Currently the only `ActiveNodeExecutions` and `ActiveTaskExecutions` metrics exported are from flyteadmin. This can result in inaccuracies if, for example, flyteadmin restarts or there is an uncommon inconsistency in data and control plane state (ex. FlyteWorkflow CR manually deleted).

### What if we do not do this?

If we do nothing, then these metrics are often inaccurate and therefore useless.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

hamersaw:
One proposal, is to move these metrics to FlytePropeller so they can more accurately reflect actual execution status. A relatively high-level design is as follows:
- Add `NodeExecutionCount` and `TaskExecutionCount` values to the `ExecutionContext` through the [ControlFlow struct](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/executors/execution_context.go#L33-L33) and increment these accordingly as FlytePropeller progresses through DAG execution ([entrypoint here](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/nodes/executor.go#L177-L177)).
- Maintain an in-memory mapping of `execution_id` to node and task execution counts that is updated as FlytePropeller evaluations executions. This update will in-turn update the prometheus gauge metric.
- Use a separate go routine to periodically iterate over the in-memory mapping and remove executions that have been manually deleted (decrementing the gauge metric).

With this solution the metric under FlytePropeller restarts will be eventually consistent. Although, as the workflow reeval duration is, by default, `30s` it should converge within that timeframe.

Future-Outlier:
@hamersaw Are you working on this issue?
If not, can I try it?

hamersaw:
@Future-Outlier this is already being addressed, will assign accordingly. Thanks for the initiative!

PR #4986, opened 2024-02-29
Add tracking for active node and task execution counts in propeller

sshardool:
## Tracking issue
https://github.com/flyteorg/flyte/issues/4593

Fixes this from the issue description 
```
Currently the only ActiveNodeExecutions and ActiveTaskExecutions metrics exported are from flyteadmin. 
This can result in inaccuracies if, for example, flyteadmin restarts or there is an uncommon inconsistency in data and control plane state (ex. FlyteWorkflow CR manually deleted).

```

The metrics for active node and task executions are added to flytepropeller with the expectation that the workflow re-evaluation duration is an acceptable delay for updating the metrics (eventually consistent)

## How was this patch tested?
Tested on a dev flyte setup (single node).
Tested is in progress in larger scale internal deployments.

Sample of the new metrics:
```
flyte:propeller:flyte_dev:execstats:active_node_executions 9
flyte:propeller:flyte_dev:execstats:active_task_executions 7
flyte:propeller:flyte_dev:execstats:active_workflow_executions 3
```


pvditt:
@sshardool this PR is great. 

Have you noticed any increases in latency for workflow executions with this running? How many [workers/threads](https://github.com/flyteorg/flyte/blob/524b473b832918f7477c6ec098e2ca190f914b5e/flytepropeller/pkg/controller/config/config.go#L129) are running for your propeller deployments? I figure this should only add in the order of nanoseconds for propeller loops with lock contention with ExecutionStatsHolder updates.

We might want to run this on an internal cluster for a little bit prior to merging, but this PR LGTM.

sshardool:
> @sshardool this PR is great.
> 
> Have you noticed any increases in latency for workflow executions with this running? How many [workers/threads](https://github.com/flyteorg/flyte/blob/524b473b832918f7477c6ec098e2ca190f914b5e/flytepropeller/pkg/controller/config/config.go#L129) are running for your propeller deployments? I figure this should only add in the order of nanoseconds for propeller loops with lock contention with ExecutionStatsHolder updates so it shouldn't be a concern.
> 
> We might want to run this on an internal cluster for a little bit prior to merging, but this PR LGTM.

Thanks for the review @pvditt . We currently have this running with 4 worker threads and have not seen any increase in execution latencies. You're right, critical section for all operations which are inline with the work-queue tasks is very small (single map entry update). Let me know if an additional metric tracking the wait duration (i.e time taken for mutex acquisition) in `ExecutionStatsHolder:AddOrUpdateEntry()` will help.

pvditt (approved):
thank you for adding this

PR #4988, opened 2024-03-01
Allow specifying flyte pod webhook deployment resources

katrogan:
Various resources recommend setting resources and limits for k8s deployments, e.g: https://www.cncf.io/blog/2022/10/20/kubernetes-best-practice-how-to-correctly-set-resource-requests-and-limits/

Optionally allows specifying flyte pod deployment resource requests and limits.

## How was this patch tested?

In values.yaml
```
webhook:
...
  resources:
    limits:
      cpu: 200m
      ephemeral-storage: 100Mi
      memory: 200Mi
    requests:
      cpu: 10m
      ephemeral-storage: 50Mi
      memory: 100Mi
```

verified the rendered webhook deployment included

```
          ...
          securityContext:
            allowPrivilegeEscalation: false
            capabilities:
              drop: ["ALL"]
          resources:
            limits:
              cpu: 200m
              ephemeral-storage: 100Mi
              memory: 200Mi
            requests:
              cpu: 10m
              ephemeral-storage: 50Mi
              memory: 100Mi
          volumeMounts:
            - name: config-volume
              mountPath: /etc/flyte/config
              readOnly: true
            - name: webhook-certs
              mountPath: /etc/webhook/certs
              readOnly: true
          ...
```

### DIFF
--- a/charts/flyte-core/templates/propeller/webhook.yaml
+++ b/charts/flyte-core/templates/propeller/webhook.yaml
@@ -103,6 +103,7 @@ spec:
             allowPrivilegeEscalation: false
             capabilities:
               drop: ["ALL"]
+          resources: {{- toYaml .Values.webhook.resources | nindent 12 }}
           volumeMounts:
             - name: config-volume
               mountPath: /etc/flyte/config
--- a/charts/flyte-core/templates/propeller/webhook.yaml
+++ b/charts/flyte-core/templates/propeller/webhook.yaml
@@ -103,7 +103,9 @@ spec:
             allowPrivilegeEscalation: false
             capabilities:
               drop: ["ALL"]
+          {{- if .Values.webhook.resources -}}
           resources: {{- toYaml .Values.webhook.resources | nindent 12 }}
+          {{- end }}
           volumeMounts:
             - name: config-volume
               mountPath: /etc/flyte/config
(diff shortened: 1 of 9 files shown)


charts/flyte-core/templates/propeller/webhook.yaml
```diff
@@ -103,6 +103,9 @@ spec:
             allowPrivilegeEscalation: false
             capabilities:
               drop: ["ALL"]
+          {{- if .Values.webhook.resources -}}
```
mhotan, line 106: nit: Not important in this case but we could also do the following. 

```yaml
{{- with .Values.webhook.resources }}
resources: {{- toYaml . | nindent 12 }}
{{- end }}
```

This is not really that important as it's just syntactic difference, only really benefits for more complex config. We can always update later.

mhotan (approved):
LGTM, we could add reasonable `resources.requests` but omit `resources.limits`. Requests defaults but are not immediately necessary. It may make more sense to have a single issue/PR set reasonable defaults missing for all workloads so folks can track what's changed more easily.

charts/flyte-core/templates/propeller/webhook.yaml
katrogan: thanks @mhotan updated to address both your comments, although having a hard time reasoning about what sensible defaults would be here esp given the recent ephemeral storage incident. figure it might be safer to err on overprovisioning for now

pingsutw (approved):

pingsutw (approved):

PR #4993, opened 2024-03-02
Enable agent service by default

pingsutw:
## Tracking issue

Note: This PR only enables the agent-service plugin by default in the helm chart and won't deploy an agent server.
The reason to do that because people won't need to modify the propeller configmap after they install the agent (`helm install flyteagent ...`), and also give people with more visibility into flyte agent.

## How was this patch tested?
Local sandbox

## Related PRs
NA

## Docs link
NA


Future-Outlier (dismissed):
LGTM, thank you

EngHabu (dismissed):
What happens if you try to use an agent that isn't deployed now?

Future-Outlier (approved):
LGTM, thank you

Future-Outlier:
> What happens if you try to use an agent that isn't deployed now?

flytepropeller will send error to flyteadmin, and flyteadmin will send error message to flyteconsole.
We will have a watcher to dynamically check avaialbe agent deployments and update to flytepropeller.

PR #4998, opened 2024-03-04
Allow setting a ExecutionClusterLabel when triggering a Launchplan/Workflow/Task

RRap0so:

This is an alternative implementation to the https://github.com/flyteorg/flyte/pull/4956. Instead of trying to implement the `clusterPool` functionality we reuse the ClusterAssignment struct and Execution Label.

This was changes are kept to a minimal and would cover the needs of the linked RFC

Closes https://github.com/flyteorg/flyte/issues/5081

## How was this patch tested?

This was tested with two additional unit tests.


davidmirror-ops:
cc @gdabisias

wild-endeavor:
Won't we also need to add the message to the [incoming request](https://github.com/flyteorg/flyte/blob/1caa86da1152ad3639f05625470968eb28992a2c/flyteidl/protos/flyteidl/admin/execution.proto#L267)?  It'll need to start from there no?

wild-endeavor:
hey @RRap0so do you have a real multi-cluster setup available to help test this?  It's a fair bit of work for us to set one up to test.  We can fake it, but it'd be nice to have a real setup to test against.  I pulled this pr, built a new sandbox image for it, and ran a local/normal test, and it works, but i'm obviously not testing any of the new functionality.

RRap0so:
@wild-endeavor I think we might have something that we can use. I'll report back.

RRap0so:
Hi @wild-endeavor.

All done and tested. I've created executions via grpc and manage to reproduce everything perfectly:
- if a label is found, goes through the proper cluster
- if not found then default behaviours applies (in our case the execution would hit one of the two clusters we have)

I've also added a Debug line to help out and done another comparison fix.

flyteadmin/pkg/executioncluster/execution_target.go
```diff
@@ -1,6 +1,7 @@
 package executioncluster
 
 import (
+	"github.com/flyteorg/flyte/flyteidl/gen/pb-go/flyteidl/admin"
```
katrogan, line 4: nit: mind running `make goimports` to order these imports?

katrogan (commented):
thank you for testing this out end to end!

do you mind also updating https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/executioncluster/impl/in_cluster.go#L25 to fail fast when a `ExecutionTargetSpec.ExecutionClusterLabel` is specified (like we do for TargetID)?

flyteadmin/pkg/executioncluster/execution_target.go
RRap0so: done!

katrogan (approved):

PR #5013, opened 2024-03-06
Set flyteadmin grpc port correctly in config / Flyte-core flyteadmin / datacatalog expose ports

ddl-ebrown:
## Tracking issue

 - Flyteadmin setting for the grpc port is nested and grpcPort should
   not be used. Address incorrect config / documentation

 - Both flyteadmin and datacatalog have configurable ports to use for
   HTTP and GRPC, but the k8s spec for the containers doesn't expose
   matching ports.

   Fix that!

Correctly use the values from values.yaml inside the k8s specs for listening container ports


deployment/eks/flyte_aws_scheduler_helm_generated.yaml
```diff
@@ -1208,7 +1208,7 @@ spec:
         imagePullPolicy: "IfNotPresent"
         name: datacatalog
         ports:
-        - containerPort: 8088
+        - containerPort: 8080
```
pingsutw, line 1211: could we change the default port back to 8088?

pingsutw: yes, I know. I mean, we should change this [port](https://github.com/flyteorg/flyte/blob/ba5925868f9ece84c834a9d5a1e09f6287394bce/charts/flyte-core/values.yaml#L692) to 8088

pingsutw: ah, I see, so 8088 never works. we use 8080 in the [configmap](https://github.com/flyteorg/flyte/blob/caed58456f3039840081bf42f7775500b542a36b/charts/flyte-core/values.yaml#L692), but use 8088 for container port. 
@ddl-ebrown Thanks for investigating and fixing it.

pingsutw (dismissed):

davidmirror-ops:
@ddl-ebrown with the merge of #5072 you should rebase and merge master to pass the lint test.
Thank you

davidmirror-ops (approved):
Thank you!

PR #5018, opened 2024-03-06
Fix minio config on helpers.tpl

### LINKED ISSUE #5004: [Housekeeping] Fix hardcoded minio config on flyte-core
### Describe the issue

As a user, when deploying the flyte-core chart to an on-prem environment that uses minio, I'd like to have control over the minio service configuration.

When I set `sandbox` as the storage type, the following key is added (with no possibility of overriding it):

```
storage:
   signedUrl:
      stowConfigOverride:
        endpoint: http://localhost:30084 
```

This is the nodePort that is accessible inside the K8s cluster but not from a client. It causes pyflyte executions to fail:

`Failed to establish a new connection: [Errno 61] Connection refused'))
HTTPConnectionPool(host='localhost', port=30084)`

Presumably, the solution would be to either make it configurable by taking it out from [`_helpers.tpl`](https://github.com/flyteorg/flyte/blob/2adaf612348c9b3d83f4640b0eedba6e2ab33f4c/charts/flyte-core/templates/_helpers.tpl#L248-L250) to `values` or to set it to the value of the service that the client can connect to via port-forward (`minio.flyte.svc.cluster.local:9000`)

Installations of flyte-core with minio would not be usable, or would require users to edit the configmap post-installation, which is not ideal.

### Related component(s)

_No response_

- [X] Yes

- [X] Yes

davidmirror-ops:
## Tracking issue
Closes #5004 

## How was this patch tested?
Deployed to an on-prem K8s cluster with minio and it successfully executes workflows.

I used `storage.type: sandbox`

- [X] I updated the documentation accordingly. (It will be part of the Flyte the Hard Way repo.)

### DIFF
--- a/charts/flyte-core/templates/_helpers.tpl
+++ b/charts/flyte-core/templates/_helpers.tpl
@@ -247,7 +247,7 @@ storage:
       region: us-east-1
   signedUrl:
     stowConfigOverride:
-      endpoint: http://localhost:30084
+      endpoint: http://minio.{{ .Release.Namespace }}.svc.cluster.local:9000
 {{- else if eq .Values.storage.type "custom" }}
 {{- with .Values.storage.custom -}}
   {{ tpl (toYaml .) $ | nindent 2 }}
--- a/deployment/sandbox/flyte_helm_generated.yaml
+++ b/deployment/sandbox/flyte_helm_generated.yaml
@@ -326,7 +326,7 @@ data:
           region: us-east-1
       signedUrl:
         stowConfigOverride:
-          endpoint: http://localhost:30084
+          endpoint: http://minio.flyte.svc.cluster.local:9000
       enable-multicontainer: false
       limits:
         maxDownloadMBs: 10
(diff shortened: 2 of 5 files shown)


eapolinario (approved):
This was put in place [initially](https://github.com/flyteorg/flyte/commit/edf529687c6815a09bced1ee9c2fa560f1439f0d#diff-e2eb39162dcb726f53d87b4c2ef9d5af3b2450ac5f30508deef6bf0cded88a2b) to accommodate fast registration in the legacy sandbox. In the demo cluster (i.e. the current sandbox) we [override](https://github.com/flyteorg/flyte/blob/f448a0358d8706a09b65b96543134f629327d755/charts/flyte-sandbox/values.yaml#L51-L53) that value separately.

We're in the process of removing the legacy sandbox, so I'm wondering if we shouldn't go through the effort of setting up a separate helm chart for the on-prem scenario. Not for this PR obviously.

davidmirror-ops:
@eapolinario I've considered a chart for on-prem too. The problem is that it'd be hard to design it following Helm conventions, especially considering how many options are available to achieve the same thing in the on prem world. For example, what default values to set for Ingress? for what controller? Also, what if someone is not using minio? (they can go with custom, which is not greatly documented but that's a different problem). 
So, I'd rather try to make our two main charts more flexible.

PR #5058, opened 2024-03-14
integration test config

troychiu:
## Tracking issue
NA

Local integration test for flyteadmin will fail because of the out-of-dated db config.

Set sandbox db as default and add a environment variable in github action.
With these changes, contributors can easily run integration test locally and the github action can keep running


flyteadmin/tests/bootstrap.go
```diff
@@ -19,8 +20,18 @@ const insertExecutionQueryStr = `INSERT INTO "executions" ` +
 	`("execution_project","execution_domain","execution_name","phase","launch_plan_id","workflow_id") ` +
 	`VALUES ('%s', '%s', '%s', '%s', '%d', '%d')`
 
+const ENV_DB_CONFIG = "ADMIN_DB_CONFIG"
+
 var adminScope = promutils.NewScope("flyteadmin")
 
+func getDbConfigWithEnv() *database.DbConfig {
+	if os.Getenv(ENV_DB_CONFIG) == "CI" {
```
pingsutw, line 28:
```suggestion
	if os.Getenv(USE_LOCAL_SANDBOX) == "True" {
```

flyteadmin/tests/bootstrap.go
```diff
@@ -19,8 +20,18 @@ const insertExecutionQueryStr = `INSERT INTO "executions" ` +
...
+		return getDbConfig()
```
pingsutw, line 29:
```suggestion
		return getSandboxDbConfig()
```

flyteadmin/tests/bootstrap.go
```diff
@@ -32,13 +44,15 @@ func getDbConfig() *database.DbConfig {
 	}
 }
 
-func getLocalDbConfig() *database.DbConfig {
+// Run `flytectl demo start` to start the sandbox
+func getSandboxDbConfig() *database.DbConfig {
```
pingsutw, line 48: cc @ysysys3074 you can try this config to run integration tests

flyteadmin/tests/bootstrap.go
```diff
@@ -19,8 +20,18 @@ const insertExecutionQueryStr = `INSERT INTO "executions" ` +
 	`("execution_project","execution_domain","execution_name","phase","launch_plan_id","workflow_id") ` +
 	`VALUES ('%s', '%s', '%s', '%s', '%d', '%d')`
 
+const sandboxEnv = "USE_LOCAL_SANDBOX"
+
 var adminScope = promutils.NewScope("flyteadmin")
 
+func getDbConfigWithEnv() *database.DbConfig {
+	if os.Getenv(sandboxEnv) == "False" {
```
katrogan, line 28: What if we use sandbox config by default and only use the k8s integration test config if explicitly instructed to from the env var?

troychiu: Yes this is the current behavior. It will use sandbox config unless you set USE_LOCAL_SANDBOX as false

katrogan (approved):

flyteadmin/tests/bootstrap.go
katrogan: Actually thinking about it some more, I think it might be more intuitive to assume a default bool value where possible (false). So rather than checking for Falsiness can we by default check to see if "USE_INTEGRATION_TEST_CONFIG" is set and _true_ and then use that specific config?

katrogan (approved):

flyteadmin/tests/bootstrap.go
```diff
@@ -19,8 +20,18 @@ const insertExecutionQueryStr = `INSERT INTO "executions" ` +
 	`("execution_project","execution_domain","execution_name","phase","launch_plan_id","workflow_id") ` +
 	`VALUES ('%s', '%s', '%s', '%s', '%d', '%d')`
 
+const integrationTestConfigEnvVar = "USE_INTEGRATION_TEST_CONFIG"
+
 var adminScope = promutils.NewScope("flyteadmin")
 
+func getDbConfigWithEnv() *database.DbConfig {
+	if os.Getenv(integrationTestConfigEnvVar) == "True" {
+		return getDbConfig()
```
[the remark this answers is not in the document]
troychiu: Good point. Resolved.

troychiu:
Catch some places that I didn't modify correspondingly. (don't know why IDE didn't catch that)

katrogan (approved):
:shipit:

pingsutw (approved):

Issue #5053, opened 2024-03-14
[BUG] flyte-secret-auth secret not mounted properly


yini7777:
### Describe the bug

When setting `Values.secrets.adminOauthClientCredentials.enabled` as false and managing the `flyte-secret-auth` with secret manager ourselves (reference: [flyte doc on auth setup](https://docs.flyte.org/en/latest/deployment/configuration/auth_setup.html)), it seems the auth is not mounted as expected due to [this condition](https://github.com/flyteorg/flyte/commit/0ec0c25a50124025b240fc2e340ac3666d3497f0).

We propose enhancing the configuration in [values.yaml](https://github.com/flyteorg/flyte/blob/master/charts/flyte-core/values.yaml#L431-L438) by introducing an additional key to indicate whether `flyte-secret-auth` to create. This enhancement would enable conditional mounting of flyte-secret-auth based on its creation status.

### Expected behavior

Able to mount the `flyte-secret-auth` secret properly when `Values.secrets.adminOauthClientCredentials.enabled` as false.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

yini7777:
@lowc1012 Thanks for your prompt response!
Error log looks like: 
```
│ Error: cannot patch "flytescheduler" with kind Deployment: Deployment.apps "flytescheduler" is invalid: spec.template.spec.initContainers[0].volumeMounts[2].name: Not found: "auth"
│ 
│   with helm_release.flyte-unstable-core,
│   on flyte-unstable.tf line 1, in resource "helm_release" "flyte-unstable-core":
│    1: resource "helm_release" "flyte-unstable-core" {
```
The context is I'm upgrading from v1.10.6 to v1.11.0 and nothing has been changed on our end. I suspect the reason of not able to find the 'auth' is that we set `Values.secrets.adminOauthClientCredentials.enabled` as false, in which case will prevent mounting 'auth' since the condition.

PR #5054, opened 2024-03-14
Fix: flyte-secret-auth secret not mounted properly in flyte-core

lowc1012:
## Tracking issue
Closes #5053 

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

PR #4914, opened 2024-02-17
fix: mount flyte-secret-auth secret conditionally

lowc1012:
## Tracking issue
Closes #4908

## Why are the changes needed?

If .Values.secrets.adminOauthClientCredentials.enabled is false in flyte-core chart, some deployments will mount flyte-secret-auth secret even though the secret isn't created.

## What changes were proposed in this pull request?

Add conditional brach volume and volumeMount to make creation of secret and volume mount are consistent

## How was this patch tested?

Can test it locally via `helm template charts/flyte-core`

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

Issue #4908, opened 2024-02-16
[BUG] Do not mount `flyte-secret-auth` secret if `.Values.secrets.adminOauthClientCredentials.enabled` is false.


debajyoti-truefoundry:
### Describe the bug

https://github.com/flyteorg/flyte/blob/1ae5a8f46d5fa77955f4b88e135ebd2222d947f4/charts/flyte-core/templates/common/secret-auth.yaml#L1

https://github.com/flyteorg/flyte/blob/1ae5a8f46d5fa77955f4b88e135ebd2222d947f4/charts/flyte-core/templates/flytescheduler/deployment.yaml#L96
https://github.com/flyteorg/flyte/blob/1ae5a8f46d5fa77955f4b88e135ebd2222d947f4/charts/flyte-core/templates/propeller/deployment.yaml#L103

As you can see in the above links, even though the `flyte-secret-auth` secret is optionally created, we always mount the corresponding secret in the propeller and the scheduler deployments.

### Expected behavior

Explained above.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

lowc1012:
Hi there, I'd like to take this. Can anyone assign this to me? thanks

lowc1012:
#self-assign

PR #5063, opened 2024-03-15
Fix mounting secrets

yini7777:
## Tracking issue
Closes: https://github.com/flyteorg/flyte/issues/5053

**Context** --> we're upgrading from v1.10.6 to v1.11.0 and nothing has been changed on our end. But got the following error:
```
│ Error: cannot patch "flytescheduler" with kind Deployment: Deployment.apps "flytescheduler" is invalid: spec.template.spec.initContainers[0].volumeMounts[2].name: Not found: "auth"
│ 
│   with helm_release.flyte-unstable-core,
│   on flyte-unstable.tf line 1, in resource "helm_release" "flyte-unstable-core":
│    1: resource "helm_release" "flyte-unstable-core" {
```
we set `Values.secrets.adminOauthClientCredentials.enabled` to false because we don't want Flyte to create the secret, as we have created it via an ExternalSecret . In this case Flyte does not mount the secret to the flytescheduler.

**Solution** --> Add a new value as `Values.secrets.adminOauthClientCredentials.create` . We will set this to false and enabled to true.


davidmirror-ops:
@yini7777 I think a missing conditional for `flytescheduler` was added in #5054. Nevertheless I guess you're still facing the same issue as the conditional logic is not met when you set the `adminOauthCredentials.enabled` to `false` which not only doesn't create the secret but doesn't mount it. 
Is my understanding correct?

yini7777:
> Nevertheless I guess you're still facing the same issue as the conditional logic is not met when you set the `adminOauthCredentials.enabled` to `false` which not only doesn't create the secret but doesn't mount it. Is my understanding correct?

Yes exactly!

docs/deployment/configuration/auth_setup.rst
```diff
@@ -597,7 +601,8 @@ Follow the steps in this section to configure `flyteadmin` to use an external au
    
          secrets:
            adminOauthClientCredentials:
-             enabled: true # see the section "Disable Helm secret management" if you require to do so
+             enabled: true
+             create: true # see the section "Disable Helm secret management" if you require to do so
```
davidmirror-ops, line 605: Thanks for your contribution and sorry for the inconvenience so far.

To make it as clear as possible to users, the challenge is that right now `create` here means both `create` AND `mount`.

What we should try to do is to decouple those two actions. So `enabled` should become `mount` and be used as a conditional for the volumeMount, with `create` used as conditional for the `secret-auth` template as you're proposing.

There would be an unsupported combination tho: `mount.false` and `create.true`.

The use case of an external Secrets Manager would be handled with `mount.create` and `create.false`

The default for auth would be both `true` and defaults without `auth` both `false`.

What do you think?

yini7777: No worries, glad that I got chance to contribute! 
I totally agree that using `create` and `mount` as key would be more intuitive, will modify it! 
BTW I’m wondering if it’s necessary to warn the user if they use the unreasonable combination, the most low-hanging way of doing it would be adding comments in the `value.yaml`, any other suggestions?

davidmirror-ops: I'd go with the comment for now to warn users

charts/flyte-core/values-keycloak-idp-flyteclients-without-browser.yaml
```diff
@@ -298,10 +298,13 @@ deployRedoc: false
...
-    # -- If enabled is true, helm will create and manage `flyte-secret-auth` and populate it with `clientSecret`.
-    # If enabled is false, it's up to the user to create `flyte-secret-auth` as described in
+    # If mount is true, helm will mount `flyte-secret-auth`.
+    # If mount is false, helm will not mount `flyte-secret-auth`.
+    # If create is true, helm will create the `flyte-secret-auth`.
+    # If create is false, it's up to the user to create `flyte-secret-auth` as described in
     # https://docs.flyte.org/en/latest/deployment/cluster_config/auth_setup.html#oauth2-authorization-server
-    enabled: true
+    mount: true
+    create: true
```
davidmirror-ops, line 307: nit: switch the ordering? `create` first and `mount` next

charts/flyte-core/templates/_helpers.tpl
```diff
@@ -261,3 +261,12 @@ storage:
   limits:
     maxDownloadMBs: {{ .Values.storage.limits.maxDownloadMBs }}
 {{- end }}
+
+{{/*
+Check for deprecated .Values.secrets.adminOauthClientCredentials.enabled and warn users
+*/}}
+{{- define "chart.checkDeprecatedValue" -}}
+{{- if  .Values.secrets.adminOauthClientCredentials.enabled }}
+WARNING: You are using a deprecated value: .Values.secrets.adminOauthClientCredentials.enabled. Please update your values.yaml to use the new preferred value.
```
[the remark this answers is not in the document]
yini7777: @davidmirror-ops Done reverting!

yini7777:
@davidmirror-ops I'm wondering if there is anything needs to be tested or modified in this PR.

davidmirror-ops:
@yini7777 Yesterday we merged #5072 to fix the lind-and-test-charts test. Also I see the generate Helm test failing

Could you please

1. Rebase to master and merge
2. Run make helm and push the changes

Let me know if you have any question

Thanks!

davidmirror-ops:
@yini7777 Thanks for your patience so far. A number of PRs right now changing the Helm charts so I hope one last rebase should fix the conflicts and get us ready to merge

Thanks!

yini7777:
@davidmirror-ops Regarding this PR, Yee reached out to me and he mentioned he will handle this issue together with some necessary cleanup and refactoring to align with the Flyte binary setup, so we agreed on closing this PR ATM.

wild-endeavor:
So @eapolinario and I had a chat, and after discussing it more, I think we're going to continue with a variation of what you proposed here.

Some background on this, just to get everyone on the same page.  Also, this is from the perspective of the flyte-core helm chart - the flyte-binary helm chart behaves differently in at least some of the following points.

* If authentication is turned on, then propeller typically needs a way to talk to Admin.
* The way this is done typically is through the oauth2 client credentials authentication method (basically a client id and secret).
* In the past, we just put the client secret in the [helm chart](https://github.com/flyteorg/flyte/blob/c61d976c6f1d934d38afc3097a64f45f36b7c7ad/charts/flyte-core/values.yaml#L437) as a value, and then helm would [create a K8s secret](https://github.com/flyteorg/flyte/blob/master/charts/flyte-core/templates/common/secret-auth.yaml) using it.
* Users (rightfully) complained that this was not secure.
* So in the [auth docs](https://docs.flyte.org/en/latest/deployment/configuration/auth_setup.html#disable-helm-secret-management) it now states that you can set `enabled: false` in that section of your values file, and create the `flyte-secret-auth` secret yourself.

This is fine and mostly worked and supported the two normal use cases: 1) create the secret and mount it where needed.  2) don't create the secret, assume it's there, mount it anyways.

Somehow when we first made the helm chart, I think we missed the third possibility - auth is not turned on, don't mount anything, don't create anything.  To address that, the [issue 4908](https://github.com/flyteorg/flyte/issues/4908) and [pr 4914](https://github.com/flyteorg/flyte/pull/4914/files) were merged.

Unfortunately 4914 broke the recommended use-case, where users want to mount the secret, but have the enabled flag set to false, because they don't want it specified in the values file.  After 4914 if you manually created the secret, it just wasn't getting mounted.

I had recommended to @yini7777 that we close this PR earlier because I felt like we should resolve it like how we do in the [flyte-binary helm chart](https://github.com/flyteorg/flyte/blob/cf7c638497224aa60912967df529182b4c41b5d5/charts/flyte-binary/templates/deployment.yaml#L260-L286), by letting users just specify arbitrary external/inline secrets, that are just consumed by the helm chart and projected into the config map.  But after more thinking, it seems like too much work for now, esp. if we wanted to make it in a backwards compatible way.  For now, it's probably enough to just

* change the creation of the secret to key off of whether or not the secret value is provided.

Then, to...
* create & mount: specify `clientSecret` in the values file and `enabled: true`
* don't create, but do mount: just set `enabled: true`
* don't do anything: set `enabled: false`

yini7777:
@wild-endeavor @davidmirror-ops Thank you for your update and explanation. I've made some modifications to align with your suggestions. Could you please review them?

davidmirror-ops (dismissed):
I deployed the changes proposed here and validated the behavior is as expected:

>create & mount: specify clientSecret in the values file and enabled: true
don't create, but do mount: just set enabled: true
don't do anything: set enabled: false

yini7777:
@davidmirror-ops Thanks a lot for testing it out! Just curious about when validating `don't create, but do mount: just set enabled: true`, did u also overwrite the default `clientSecret: foobar` with `clientSecret: null`?

davidmirror-ops:
@yini7777 I commented out that line entirely tbh

yini7777:
@davidmirror-ops Got it! I think it’s also crucial to test if overwriting with `null` works as expected, since from a user perspective it’s not practical to comment it out, I am also open to carry out this test myself. 

The reason I’m so cautious about this is due to [this issue](https://github.com/helm/helm/issues/12637). If you have any guidance on how to test it effectively, I’m eager to learn and contribute. Thanks!

davidmirror-ops:
@yini7777 you nailed it. when I set
```
secrets:
  adminOauthClientCredentials:
    enabled: true
    clientSecret: null
```

propeller fails to start:

`MountVolume.SetUp failed for volume "auth" : secret "flyte-secret-auth" not found`

yini7777:
Awesome! Thanks a lot! Then I'll do a rebase and merge it!

wild-endeavor (dismissed):

wild-endeavor (dismissed):

wild-endeavor (approved):

PR #5067, opened 2024-03-15
Adapt ray flyteplugin to Kuberay 1.1.0

ByronHsu:

From Kuberay 1.0.0 to 1.1.0, few changes is made in rayjob CRD, including:
1. Rayjob status, which flyteplugin uses to determine flyte status
2. RuntimeEnv is deprecated, which flyteplugin uses to create rayjob
3. Few new fields introduced

image
 

In general, the change is to support kuberay 1.1.0 but ensure backward compatibility at the same time. I will first walk through the change i made, then enumerate user scenerio.

### Flyteidl
1. Deprecate but not remove runtime_env
2. Add runtime_env_yaml

### Flytekit
1. Serialize runtime_env to yaml and pass to runtime_env_yaml

### Flyteplugin
1. Remove all v1apha1 logics
2. If runtime_env_yaml is not present, convert runtime_env to yaml to ensure backward comp
3. Re-map rayjob status to flyte task status. As suggested by @Yicheng-Lu-llll , only rely on jobdeployment status

### User scenario

| flytekit | flyte | kuberay | Result |
|--------|--------|--------|-------|
| Old | Old | 1.0.0 | just works |
| New | New | 1.1.0 | works |
| Old | New | 1.1.0 | works because new flyte also handle base64 `runtime_env`|
| New | Old | 1.0.0 | works because new flytekit also passes `runtime_env` |

I have run end-to-end tests on local sandbox with the following tests.

1. rayjob succeeds

```python
import typing

from flytekit import ImageSpec, Resources, task, workflow

commit_hash = "5a57f299c1ae789c9e2f269302becf13ca3b21e8"
ray_plugin = f"git+https://github.com/flyteorg/flytekit.git@{commit_hash}#subdirectory=plugins/flytekit-ray"

custom_image = ImageSpec(
    name="ray-flyte-plugin",
    registry="localhost:30000",
    packages=[ray_plugin],
    apt_packages=['git', 'wget', 'vim'],
)

if custom_image.is_container():
    import ray
    from flytekitplugins.ray import HeadNodeConfig, RayJobConfig, WorkerNodeConfig

@ray.remote
def f(x):
    print(a)
    print(f"ray task get {x}")
    return x * x

ray_config = RayJobConfig(
    head_node_config=HeadNodeConfig(ray_start_params={"log-color": "True"}),
    worker_node_config=[WorkerNodeConfig(group_name="ray-group", replicas=1)],
    runtime_env={"pip": []},
)

@task(
    task_config=ray_config,
    requests=Resources(mem="4Gi", cpu="1"),
    container_image=custom_image,
)
def ray_task(n: int) -> typing.List[int]:
    futures = [f.remote(i) for i in range(n)]
    l = ray.get(futures)
    print(l)
    return l


@workflow
def ray_workflow(n: int) -> typing.List[int]:
    return ray_task(n=n)

if __name__ == "__main__":
    f.remote(x=2)
```

image

2. rayjob fails

```python
import typing

from flytekit import ImageSpec, Resources, task, workflow

commit_hash = "5a57f299c1ae789c9e2f269302becf13ca3b21e8"
ray_plugin = f"git+https://github.com/flyteorg/flytekit.git@{commit_hash}#subdirectory=plugins/flytekit-ray"

custom_image = ImageSpec(
    name="ray-flyte-plugin",
    registry="localhost:30000",
    packages=[ray_plugin],
    apt_packages=['git', 'wget', 'vim'],
)

if custom_image.is_container():
    import ray
    from flytekitplugins.ray import HeadNodeConfig, RayJobConfig, WorkerNodeConfig

@ray.remote
def f(x):
    a = 1 // 0
    print(f"ray task get {x}")
    return x * x

ray_config = RayJobConfig(
    head_node_config=HeadNodeConfig(ray_start_params={"log-color": "True"}),
    worker_node_config=[WorkerNodeConfig(group_name="ray-group", replicas=1)]
)

@task(
    task_config=ray_config,
    requests=Resources(mem="4Gi", cpu="1"),
    container_image=custom_image,
)
def ray_task(n: int) -> typing.List[int]:
    futures = [f.remote(i) for i in range(n)]
    l = ray.get(futures)
    print(l)
    return l


@workflow
def ray_workflow(n: int) -> typing.List[int]:
    return ray_task(n=n)

if __name__ == "__main__":
    f.remote(x=2)
```

image

## Related PRs

https://github.com/flyteorg/flytekit/pull/2274

## Notes

If they upgrade flyte, they also have to upgrade kuberay to 1.1.0. Have to add doc somewhere


ByronHsu:
@pingsutw @Yicheng-Lu-llll could you provide guidance on how to resolve the `make generate` CI errors? Thanks!

flyte-single-binary-local.yaml
```diff
@@ -41,9 +41,11 @@ tasks:
       - container
       - sidecar
       - K8S-ARRAY
+      - ray
```
pingsutw, line 44: Could we remove it? Otherwise, people always have to install Kuberay in the sandbox.

flyteidl/protos/flyteidl/plugins/ray.proto
```diff
@@ -8,13 +8,18 @@ option go_package = "github.com/flyteorg/flyte/flyteidl/gen/pb-go/flyteidl/plugi
 message RayJob {
   // RayClusterSpec is the cluster template to run the job
   RayCluster ray_cluster = 1;
+  
+  // DEPRECATED
   // runtime_env is base64 encoded.
   // Ray runtime environments: https://docs.ray.io/en/latest/ray-core/handling-dependencies.html#runtime-environments
   string runtime_env = 2;
```
pingsutw, line 15:
```suggestion
  string runtime_env = 2 [deprecated = true];
```

flyteidl/protos/flyteidl/plugins/ray.proto
```diff
@@ -8,13 +8,18 @@ option go_package = "github.com/flyteorg/flyte/flyteidl/gen/pb-go/flyteidl/plugi
...
   // shutdown_after_job_finishes specifies whether the RayCluster should be deleted after the RayJob finishes.
   bool shutdown_after_job_finishes = 3;
   // ttl_seconds_after_finished specifies the number of seconds after which the RayCluster will be deleted after the RayJob finishes.
   int32 ttl_seconds_after_finished = 4;
+	// RuntimeEnvYAML represents the runtime environment configuration
+	// provided as a multi-line YAML string.
+  string runtime_env_yaml = 5;
```
pingsutw, line 22:
```suggestion
  // RuntimeEnvYAML represents the runtime environment configuration
  // provided as a multi-line YAML string.
  string runtime_env_yaml = 5;
```

flyteplugins/go/tasks/plugins/k8s/ray/config.go
```diff
@@ -22,7 +22,6 @@ var (
 		IncludeDashboard:         true,
 		DashboardHost:            "0.0.0.0",
 		EnableUsageStats:         false,
-		KubeRayCrdVersion:        "v1alpha1",
```
pingsutw, line 25: cc @peterghaddad

EngHabu:
Thank you for the due diligence in the testing matrix. I've one additional question related to backward compatibility. Will this handle running jobs in a graceful way? or will it fail to read the old CRD completely? I imagine if somebody has Ray jobs running and an upgrade happened at the same time.

cc @pingsutw

pingsutw:
cc @kevin85421 Will kuberay handle running jobs in a graceful way

kevin85421:
>  Will this handle running jobs in a graceful way? or will it fail to read the old CRD completely?

KubeRay does not provide a conversion webhook to handle CRD upgrades gracefully. I haven't tried upgrading the CRD while RayJob custom resources were running, but I suspect there will be some issues, especially if the running RayJob uses `runtimeEnv`, which was deprecated in KubeRay v1.1.0.

ByronHsu:
Even though there might be issues as @kevin85421 mentioned. Since we only use rayjob (which is a batch job) in flyte, it shall be tolerable for users to rerun the job if the old one failed due to incompatibility.

flyte-single-binary-local.yaml
ByronHsu: Thanks for the catch. Reverted it back

flyteplugins/go/tasks/plugins/k8s/ray/config.go
peterghaddad: @pingsutw after this MR is complete is there no going back to v1alpha1?

peterghaddad: Unsure if Flyte has decided to do a hard switchover from v1alpha1 to v1. If not, then I would recommend with staying on v1alpha1 as the default value for the time being.

ByronHsu: v1alpha1 in kuberay 1.0.0 and kuberay 1.1.0 are different. That being said, if you upgrade flyte but keep kuberay at 1.0.0. The new v1alpha1 imported in flyte cannot be run on the old kuberay. Also, there is no purpose to keep both v1alpha1 and v1, so we decide to discard v1alpha1.

If users really want to stay at v1alpha1, they can feel free to use the old flyte and kuberay 1.0.0. But if they want to upgrade kuberay to 1.1.0, it is required for them to upgrade flyte as well.

ByronHsu: Users can stills stay at v1.0.0 with old flyte server. See (old, old) row
image

pingsutw (approved):

Issue #5076, opened 2024-03-19
[Core feature] Add new config to disable fallback to container task handler


MortalHappiness:
### Motivation: Why do you think this is important?

If a task does not have a registered plugin handler, it will fallback to the container task handler. We usually do not want this behavior and expect the task without the corresponding task handlers to fail directly.

### Goal: What should the final outcome look like, ideally?

Add a new config to allow users to prevent the default fallback task handler.

### Describe alternatives you've considered 

N/A

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

MortalHappiness:
#take

eapolinario:
@MortalHappiness , before we consider this more seriously, can you talk about your use case a bit more?

MortalHappiness:
@eapolinario For example, if we create a Ray task and we forget to install the Ray operator, then it will fallback to use the container handler to run the task, which seems to be fine and the result is correct. However, no matter how many replicas you specified in your task, there will be only one container because the task is run as a container task instead of a Ray task.

PR #5077, opened 2024-03-19
feat(handler): Add new config to prevent default container task handler

MortalHappiness:
## Tracking issue

Resolves: flyteorg/flyte#5076

If a task does not have a default plugin handler, it will fall back to the container task handler. We usually do not want this behavior and expect the task without corresponding task handlers to fail.

Add a new config `fallback-to-container-handler` under the `task-plugins` section.


MortalHappiness:
@pingsutw Can you help me review this PR? Thanks.

PR #5097, opened 2024-03-22
Update container builds from go 1.21.5 to 1.21.latest

ddl-ebrown:
## Tracking issue

The container images include a number of critical and high CVEs that should be addressed.

 - go 1.21.5 was released on 12-5-2023 and go 1.21.8 was released on 3-5-2024. During that time a number of high and critical vulns have been addressed

 - This PR changes the build process to float with the latest go release to automatically pick up remediated vulnerabilities rather than pin to a specific go release

 - Ideally there would be a renovate or dependabot process that automatically puts up PRs to this repo to update the pinning, which is safer from a build / test perspective.  Without that process in place, it's probably better to float with the latest patch release of go 1.21

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ ARG FLYTECONSOLE_VERSION=latest
 FROM ghcr.io/flyteorg/flyteconsole:${FLYTECONSOLE_VERSION} AS flyteconsole
 
 
-FROM --platform=${BUILDPLATFORM} golang:1.21.5-bookworm AS flytebuilder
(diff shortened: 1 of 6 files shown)


pingsutw (approved):

eapolinario (approved):

eapolinario:
> * Ideally there would be a renovate or dependabot process that automatically puts up PRs to this repo to update the pinning, which is safer from a build / test perspective.

This is a great suggestion. I'm going to investigate how to enable that in dependabot.

PR #5127, opened 2024-03-27
Fix execution phase

troychiu:
The execution phase should be FAILED because of the error in input pass to `CreateExecutionModel()`. However, we only update the phase in closure but not in `execution.phase`.

As a follow up, we don't even need `Phase` in the `CreateExecutionModelInput`

1. Fix the bug
2. Add assertion in test.
3. Clean up redundant phase in `CreateExecutionModelInput{`

### DIFF
--- a/flyteadmin/pkg/repositories/transformers/execution.go
+++ b/flyteadmin/pkg/repositories/transformers/execution.go
@@ -34,7 +34,6 @@ type CreateExecutionModelInput struct {
 	LaunchPlanID          uint
 	WorkflowID            uint
 	TaskID                uint
-	Phase                 core.WorkflowExecution_Phase
 	CreatedAt             time.Time
 	Notifications         []*admin.Notification
 	WorkflowIdentifier    *core.Identifier
@@ -76,7 +75,7 @@ func CreateExecutionModel(input CreateExecutionModelInput) (*models.Execution, e
 	}
 	createdAt := timestamppb.New(input.CreatedAt)
 	closure := admin.ExecutionClosure{
-		Phase:         input.Phase,
+		Phase:         core.WorkflowExecution_UNDEFINED,
 		CreatedAt:     createdAt,
 		UpdatedAt:     createdAt,
 		Notifications: input.Notifications,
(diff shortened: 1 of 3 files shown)


flyteadmin/pkg/repositories/transformers/execution.go
```diff
@@ -128,7 +128,7 @@ func CreateExecutionModel(input CreateExecutionModelInput) (*models.Execution, e
 			Name:    input.WorkflowExecutionID.Name,
 		},
 		Spec:                  spec,
-		Phase:                 input.Phase.String(),
+		Phase:                 closure.Phase.String(),
```
katrogan, line 131: should we get rid of Phase in the Input altogether now that we always read it from the closure?

troychiu: Looks like we need phase in execution model to do the filter db query?

katrogan: oh yes we should still set the Phase in the model, I think we just don't need to have it as a field in `CreateExecutionModelInput` if we read it from the closure now

troychiu: Sounds good

flyteadmin/pkg/repositories/transformers/execution.go
```diff
@@ -76,7 +75,7 @@ func CreateExecutionModel(input CreateExecutionModelInput) (*models.Execution, e
 	}
 	createdAt := timestamppb.New(input.CreatedAt)
 	closure := admin.ExecutionClosure{
-		Phase:         input.Phase,
+		Phase:         core.WorkflowExecution_UNDEFINED,
```
katrogan, line 78: nit, no need to explicitly set this, this is the default enum value

katrogan (approved):

flyteadmin/pkg/repositories/transformers/execution.go
troychiu: good catch

katrogan (approved):

PR #5128, opened 2024-03-27
Change retry error from RuntimeError to FlyteRecoverableException

dansola:
## Tracking issue
https://unionai.atlassian.net/browse/DOCS-385

It turns out that all user exceptions are considered non-recoverable unless the exception is a subclass of FlyteRecoverableException ([Tasks - Flyte](https://docs.flyte.org/en/latest/concepts/tasks.html#fault-tolerance)). 

This means that the RuntimeError in the retry example [here](https://docs.flyte.org/en/latest/flyte_fundamentals/optimizing_tasks.html#retries) will not actually retry. 

We want to change the error in the example to one which will trigger a retry of the task if the decorator is used.

Slack convo: https://unionai.slack.com/archives/C06RE4ZR5QQ/p1711576283580859

As only user exceptions that are subclasses of `FlyteRecoverableException` are compatible with retries, `FlyteRecoverableException` is used rather than a `RuntimeError`.

## How was this patch tested?

[This workflow](https://dogfood.cloud-staging.union.ai/console/projects/daniel/domains/development/workflows/workflows.retry_example.wf) was ran various times with `RuntimeError` (version CT3eWXJm_xOmb73ko1ww1w) and with `FlyteRecoverableException` (version d7nmtvMO6afNxNs4WFbTqA). With `RuntimeError` no retires were triggered, while with `FlyteRecoverableException` retries were triggered.

### Screenshots

image

### DIFF
--- a/docs/concepts/tasks.rst
+++ b/docs/concepts/tasks.rst
@@ -72,6 +72,7 @@ Inherent Features
 -----------------
 
 .. _fault-tolerance:
+
 Fault tolerance
 ^^^^^^^^^^^^^^^
 
--- a/docs/concepts/tasks.rst
+++ b/docs/concepts/tasks.rst
@@ -71,6 +71,7 @@ types in the system. Flyte has a set of defined, battle-tested task types. It al
 Inherent Features
 -----------------
 
+.. _fault-tolerance:
 Fault tolerance
 ^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


neverett:
Thanks for opening this! For your ref link to work, I think you'll need to add `.. _fault-tolerance:` above the "Fault tolerance" header in [this doc](https://raw.githubusercontent.com/flyteorg/flyte/master/docs/concepts/tasks.rst) to create that reference, similar to the first header in that doc.

(Docs are not in great shape right now, so updating them is not straightforward, but we're working on a big project to overhaul them and make them easier to contribute to.)

neverett (approved):

Issue #5087, opened 2024-03-21
[BUG] Flytepropeller error and crash when "limit-namespace" is set


timheb:
### Describe the bug

Flytepropeller error and crash of the pod (version 1.11.0, same for version 1.10.6/1.10.7)

```
I0321 12:57:51.491581       1 leaderelection.go:260] successfully acquired lease peach2-flyte/propeller-leader
{"json":{},"level":"panic","msg":"failed to load plugin - container: [PluginInitializationFailed] wrong type. Actual: *cache.multiNamespaceInformer","ts":"2024-03-21T12:57:51Z"}
panic: (*logrus.Entry) 0xc000703880

goroutine 366 [running]:
github.com/sirupsen/logrus.(*Entry).log(0xc000703810, 0x0, {0xc00013fd00, 0x71})
	/go/pkg/mod/github.com/sirupsen/logrus@v1.9.3/entry.go:260 +0x491
github.com/sirupsen/logrus.(*Entry).Log(0xc000703810, 0x0, {0xc000f813d0?, 0x4521720?, 0xc000f813d0?})
	/go/pkg/mod/github.com/sirupsen/logrus@v1.9.3/entry.go:304 +0x48
github.com/sirupsen/logrus.(*Entry).Panic(0x2f37ff0?, {0xc000f813d0?, 0x23ac240?, 0x1?})
	/go/pkg/mod/github.com/sirupsen/logrus@v1.9.3/entry.go:342 +0x25
github.com/flyteorg/flyte/flytestdlib/logger.Panic({0x2f37ff0?, 0xc0009fc5f0?}, {0xc000f813d0, 0x1, 0x1})
	/go/src/github.com/flyteorg/flytestdlib/logger/logger.go:144 +0x47
github.com/flyteorg/flyte/flytepropeller/pkg/controller.(*Controller).onStartedLeading.func1()
	/go/src/github.com/flyteorg/flytepropeller/pkg/controller/controller.go:132 +0x99
created by github.com/flyteorg/flyte/flytepropeller/pkg/controller.(*Controller).onStartedLeading in goroutine 395
	/go/src/github.com/flyteorg/flytepropeller/pkg/controller/controller.go:130 +0xc5
```

### Expected behavior

no crash

### Additional context to reproduce

1. Version 1.9.0 not crash
2. Version 1.10.0 not crash
3. Version 1.10.6 crash
4. Version 1.10.7 crash
5. Version 1.11.0 crash

[...]

PR #5129, opened 2024-03-28
fix(propeller): Replace SharedIndexInformer with Informer

MortalHappiness:
## Tracking issue
Resolves: flyteorg/flyte#5087

See the issue for detailed information.

- Replace `SharedIndexInformer` (`client-go`) with `Informer` (`controller-runtime`)
- Add `core.KubeClient` to `ResourceLevelMonitor`
- Replace `objects` with `PodList`


flytepropeller/pkg/controller/nodes/task/k8s/plugin_collector.go
```diff
@@ -70,8 +68,14 @@ func (r *ResourceLevelMonitor) countList(ctx context.Context, objects []interfac
 func (r *ResourceLevelMonitor) collect(ctx context.Context) {
 	// Emit gauges at the namespace layer - since these are just K8s resources, we cannot be guaranteed to have the necessary
 	// information to derive project/domain
-	objects := r.sharedInformer.GetStore().List()
-	counts := r.countList(ctx, objects)
+
+	pods := &corev1.PodList{}
+	if err := r.kubeClient.GetClient().List(ctx, pods); err != nil {
```
ByronHsu, line 73: I am uncertain about the change here. Originally, sharedInformer fetches the resources it listens to, but now it changes to only fetches pods. Also, the code seems be metrics collecting and is from four years ago. I am not sure if the code is still relevant. cc @wild-endeavor

ByronHsu (commented):
The fix looks good overall! Thanks for the quick help! But it would be better if you can add the logics behind why you made this change. The reader only knows "what is the issue" and "the solution", but there is a gap in between

flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go
```diff
@@ -668,16 +668,10 @@ func getPluginGvk(resourceToWatch runtime.Object) (schema.GroupVersionKind, erro
 	return kinds[0], nil
 }
 
-func getPluginSharedInformer(ctx context.Context, kubeClient pluginsCore.KubeClient, resourceToWatch client.Object) (cache.SharedIndexInformer, error) {
+func getPluginSharedInformer(ctx context.Context, kubeClient pluginsCore.KubeClient, resourceToWatch client.Object) (cache.Informer, error) {
```
ByronHsu, line 671: nit: 
```suggestion
func getPluginInformer(ctx context.Context, kubeClient pluginsCore.KubeClient, resourceToWatch client.Object) (cache.Informer, error) {
```

MortalHappiness: Done

pingsutw:
@EngHabu @hamersaw mind taking a look

hamersaw (commented):
@MortalHappiness thanks for the PR here! Great insight, but I'm actually wondering if we have gone too far here. The original code just [retrieves the informer](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go#L672-L672) and uses the `cache.sharedIndexInformer` to periodically [retrieve the list of k8s resources](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/nodes/task/k8s/plugin_collector.go#L73-L73) so that it can emit a metric. This obviously pulls from a cache.

Does it make more sense to just use the cache (ie. `kubeClient.GetCache()` ) within the `ResourceLevelMonitor` and periodically retrieve directly from there depending on object type? So [this line](https://github.com/flyteorg/flyte/blob/44c701ebe6dcf8469b910069d50fdd38bb6ec80d/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go#L648-L648) could instead be:
```
rm := monitorIndex.GetOrCreateResourceLevelMonitor(ctx, metricsScope, kubeClient.GetCache(), entry.ResourceToWatch, gvk)
```

flytepropeller/pkg/controller/nodes/task/k8s/plugin_collector.go
hamersaw: Agreed, it is not safe to update this to a `PodList` because it will not capture other k8s resource types for other Flyte plugins (ex. Spark, Ray, TF, PyTorch, etc).

eapolinario:
Superseded by https://github.com/flyteorg/flyte/pull/5238.

Issue #3966, opened 2023-08-15
[BUG] run-migrations fails with RDS Postgres 14.8


davidmirror-ops:
### Describe the bug

Trying to deploy `flyte-core` to an EKS cluster and using an RDS DB with the following configuration:

* Aurora engine (Postgres compatible)
* Postgres 14.8
* Single master
* Password authentication

Renders an unusable deployment. `flyteadmin` pods remain on `CrashLoopBackOff`

Logs from the `run-migrations` init container show the following:

`Error: database migration failed: ERROR: relation "description_entities" does not exist (SQLSTATE 42P01)`

The same behavior has been reported previously (see [here](https://discuss.flyte.org/t/12652002/hey-guys-sorry-if-this-is-a-dumb-question-but-recently-been-))

### Expected behavior

DB version, especially under the same major release, should not matter. Otherwise, supported DB versions should be clearly documented

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

eapolinario:
Let's try to repro this, first on a sandbox and only then on an EKS cluster. David mentioned that this doesn't happen on version 14.6.

eapolinario:
fixed by https://github.com/flyteorg/flyte/pull/5157.

PR #5157, opened 2024-03-30
Fix Id bigint conversation for not yet created table

ongkong:
## Tracking issue
https://github.com/flyteorg/flyte/issues/3966

This commit fixes a bug with migrations. The previous code was modifying a table (`description_entities`) that does not exists yet when the migration is ran in order. 

Only modify id column of tables that exist already.

## How was this patch tested?
N/A


eapolinario (approved):
Thank you!

PR #5166, opened 2024-04-03
[Feature] add retries and backoffs for propeller sending events to admin

pvditt:
## Tracking issue

Propeller does not have retries or backoffs for the eventsink adminclient (but does for datacatalog and cacheservice). Instead on failure it will consume a system retry and retry (near immediately) to process the workflow, continuing to hammer admin.

- add retries to sending events to admin from propeller + add exponential backoff
- add config to make this configurable

## How was this patch tested?

- ran locally w/ admin down to ensure backoffs were being respected.

### DIFF
--- a/docs/deployment/configuration/generated/flytepropeller_config.rst
+++ b/docs/deployment/configuration/generated/flytepropeller_config.rst
@@ -671,6 +671,42 @@ The max bucket size for event recording tokens.
   "1000"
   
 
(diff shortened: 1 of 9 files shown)


hamersaw (changes requested):
looks great, should just need to `make generate` to create pflags. do we want to go ahead and add jitter configuration to the other places for uniformity?

Issue #4962, opened 2024-02-27
[BUG] flyte-binary chart: GRPC routes with traefik not working as expected


ldunkum:
### Describe the bug

We're using traefik as an Ingress controller in our EKS cluster and are deploying the flyte-binary chart.
The flyte-binary-grpc service has the annotation `traefik.ingress.kubernetes.io/service.serversscheme: h2c`, which according to traefik docs should be fine to serve GRPC traffic.

Testing the GRPC endpoints with curl works fine, e.g.:   
`curl -v -X POST --http2 'https://flyte.example.com/grpc.health.v1.Health' -d "" -H 'Content-Type: application/grpc' -H 'Accept: application/grpc'`    

```shell
< HTTP/2 200
< content-type: application/grpc
```

Trying to run a workflow fails with the following error message:

```shell
> pyflyte -v run -r  test2/workflows/example.py wf

E0227 10:25:36.676138107  567282 ssl_transport_security.cc:1511]       Handshake failed with fatal error SSL_ERROR_SSL: error:10000410:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE.
E0227 10:25:36.703445362  567282 ssl_transport_security.cc:1511]       Handshake failed with fatal error SSL_ERROR_SSL: error:10000410:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE.
E0227 10:25:36.728929063  567282 ssl_transport_security.cc:1511]       Handshake failed with fatal error SSL_ERROR_SSL: error:10000410:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE.
Verbose mode on
╭───────────────────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────────────────────────────────────────╮
│ /home/user/.local/pipx/venvs/flytekit/lib/python3.11/site-packages/grpc/_interceptor.py:315 in continuation                                                                                                        │
│                                                                                                                                                                                                                     │
│ ❱ 315 │   │   │   │   response, call = self._thunk(new_method).with_call(                                                                                                                                           │
│                                                                                                                                                                                                                     │
│ /home/user/.local/pipx/venvs/flytekit/lib/python3.11/site-packages/grpc/_interceptor.py:343 in with_call                                                                                                           │
│                                                                                                                                                                                                                     │
[...]

PR #5168, opened 2024-04-03
[flyte-core] Create separate grpc service for flyteadmin

Jeinhaus:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/4962_

This allows setting annotations that are required for some ingress controllers for grpc communication only on parts that actually use grpc.
Without this separation, either the console or the grpc endpoints did not work properly with some ingress controllers, e.g. traefik.

This PR splits the existing `flyteadmin` service of the `flyte-core` chart into two separate services (like in the `flyte-binary` chart).
One service for grpc communication and another one for all other communication.
Accordingly, the `annotations` field of the `values.yaml` was extended to also use `httpAnnotations` and `grpcAnnotations`.
Technically, this is a breaking change for contour users, because the `values.yaml` contained a default `annotation` (that I think should not really be there anyway):

```yaml
    annotations:
      projectcontour.io/upstream-protocol.h2c: grpc
```
I removed this default, but we could also keep this for backwards compatibility.

## Docs link

I'm not sure how to regenerate the docs for this chart from the `values.yaml`. Is there some command I can run?

### DIFF
--- a/charts/flyte-core/templates/_helpers.tpl
+++ b/charts/flyte-core/templates/_helpers.tpl
@@ -35,6 +35,45 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end }}
 {{- end -}}
 
+{{- define "flyteadmin.service.grpc.name" -}}
+{{- if .Values.common.ingress.separateGrpcIngress }}
+{{- printf "%s-grpc" ( include "flyteadmin.name" . ) -}}
+{{- else }}
+{{- template "flyteadmin.name" . -}}
+{{- end}}
+{{- end -}}
+
+{{- define "flyteadmin.service.grpc.port" -}}
+{{- if eq .Values.configmap.adminServer.server.security.secure true -}}
+80
+{{- else -}}
+81
+{{- end }}
+{{- end -}}
+
+{{- define "flyteadmin.service.http.port" -}}
+80
+{{- end -}}
+
+
+{{/*
+Get the Flyte service GRPC paths.
+*/}}
+{{- define "flyteadmin.ingress.grpcPaths" -}}
+- /flyteidl.service.AdminService
+- /flyteidl.service.AdminService/*
+- /flyteidl.service.AuthMetadataService
+- /flyteidl.service.AuthMetadataService/*
+- /flyteidl.service.DataProxyService
+- /flyteidl.service.DataProxyService/*
+- /flyteidl.service.IdentityService
+- /flyteidl.service.IdentityService/*
+- /flyteidl.service.SignalService
+- /flyteidl.service.SignalService/*
+- /grpc.health.v1.Health
+- /grpc.health.v1.Health/*
+{{- end -}}
+
 {{- define "flytescheduler.name" -}}
 flytescheduler
 {{- end -}}
--- a/charts/flyte-core/templates/admin/service-grpc.yaml
+++ b/charts/flyte-core/templates/admin/service-grpc.yaml
@@ -1,17 +1,17 @@
-{{- if .Values.flyteadmin.enabled }}
+{{- if and .Values.flyteadmin.enabled .Values.common.ingress.separateGrpcIngress }}
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ template "flyteadmin.name" . }}
+  name: {{ include "flyteadmin.service.grpc.name" .}}
   namespace: {{ template "flyte.namespace" . }}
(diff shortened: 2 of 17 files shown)


davidmirror-ops (commented):
I think we should make this optional (just as in flyte-binary) and probably leave the default behavior to be single service. wdyt?

davidmirror-ops:
> I'm not sure how to regenerate the docs for this chart from the values.yaml. Is there some command I can run?

I think `make helm` should do the trick

Jeinhaus:
> I think we should make this optional (just as in flyte-binary) and probably leave the default behavior to be single service. wdyt?

Sounds good. I'll give it a shot.

Jeinhaus:
> I think we should make this optional (just as in flyte-binary) and probably leave the default behavior to be single service. wdyt?

This got quite big, because I tried to make it kinda similar to the `flyte-binary` chart. wdyt?

charts/flyte-core/templates/admin/service-grpc.yaml
```diff
@@ -0,0 +1,31 @@
...
+  {{- end }}
+  {{- with .Values.flyteadmin.service.loadBalancerSourceRanges }}
+  loadBalancerSourceRanges:
+    {{ . }}
+  {{- end }}
+  ports:
+    - name: grpc
+      port: {{ include "flyteadmin.service.grpc.port" . }}
+      protocol: TCP
+      targetPort: grpc
```
davidmirror-ops, line 29: Shouldn't this one be `8088` if `Values.configmap.adminServer.server.security.secure`: `true` and `8089` otherwise?

charts/flyte-core/templates/_helpers.tpl
```diff
@@ -35,6 +35,45 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
...
+{{- define "flyteadmin.service.grpc.port" -}}
+{{- if eq .Values.configmap.adminServer.server.security.secure true -}}
+80
+{{- else -}}
+81
+{{- end }}
+{{- end -}}
+
+{{- define "flyteadmin.service.http.port" -}}
+80
```
davidmirror-ops, line 55: What if someone needs to override this? I guess putting it into a template would complicate it

charts/flyte-core/templates/admin/service.yaml
```diff
@@ -18,19 +24,20 @@ spec:
...
     - name: http
-      port: 80
+      port: {{ include "flyteadmin.service.http.port" . }}
       protocol: TCP
-      targetPort: 8088
-    - name: grpc
-      port: 81
-      protocol: TCP
-      targetPort: 8089
+      targetPort: http
```
davidmirror-ops, line 29: Not sure I follow, shouldn't this one be `8088` instead of `80`?

charts/flyte-core/templates/common/ingress.yaml
```diff
@@ -1,95 +1,24 @@
...
-  backend:
-    service:
-      name: flyteadmin
-      port:
-        number: {{ $grpcPort }}
-- path: /grpc.health.v1.Health/*
+{{- $paths := (include "flyteadmin.ingress.grpcPaths" .) | fromYamlArray }}
+{{- range $path := $paths }}
+- path: {{ $path }}
+  {{- if semverCompare ">=1.18-0" $.Capabilities.KubeVersion.GitVersion }}
```
davidmirror-ops, line 6: I'm not quite sure `Capabilities.KubeVersion.GitVersion` is still supported on Helm 3 ([ref](https://helm.sh/docs/chart_template_guide/builtin_objects/))

charts/flyte-core/values.yaml
```diff
@@ -50,8 +50,11 @@ flyteadmin:
     - flyteexamples
   # -- Service settings for Flyteadmin
   service:
```
davidmirror-ops, line 52: I didn't see the Values.common.ingress.separateGrpcIngress key added to values

davidmirror-ops (commented):
It's a big one for sure! Thanks for putting so much effort into this and sorry for the delay.
I think this also needs some local testing that I plan to complete once you review the suggestions.

Thanks

davidmirror-ops:
@Jeinhaus any chance you can take a look at the comments? Thanks for putting so much effort into this!

mruoss:
This seems quite far and would be great to have! @Jeinhaus any chance you get to work on this still?

eapolinario:

PR #5172, opened 2024-04-03
Fix streak length metric reporting

### LINKED ISSUE #5170: [BUG] Streak length metric reporting is broken
### Describe the bug

The streak length is always reported to be 0, but I can see from the logs that its actually non-zero. 

### Expected behavior

The metric should accurately report streak length. 

Deploy flytepropeller and look at the `flyte:propeller:all:round:streak_length_unlabeled` metric. It will always report 0. 

## Root cause
Code looks like this:
```
	streak := 0
	defer p.metrics.StreakLength.Add(ctx, float64(streak))

	maxLength := p.cfg.MaxStreakLength
	if maxLength <= 0 {
		maxLength = 1
	}

	for streak = 0; streak < maxLength; streak++ {
		w, err = p.streak(ctx, w, wfClosureCrdFields)
		if err != nil {
			return err
		} else if w == nil {
			break
		}

		logger.Infof(ctx, "FastFollow Enabled. Detected State change, we will try another round. StreakLength [%d]", streak)
	}
	logger.Infof(ctx, "Streak ended at [%d]/Max: [%d]", streak, maxLength)
	return nil
}
```
The problem is that `defer` evaluates `float64(streak) = 0` immediately and just waits until the end to report a zero. I can make a PR to fix this. 

### Screenshots

_No response_

- [X] Yes

- [X] Yes

Tom-Newton:
## Tracking issue
Closes #5170

Accurate metrics are useful.

Fix the streak length metric reporting by avoiding use of `defer`. Included slightly simplifying the branching logic. 

## How was this patch tested?
I've been using it in production. 
image

- [x] I updated the documentation accordingly - not applicable
- [x] All new and existing tests passed

### DIFF
--- a/flytepropeller/pkg/controller/handler.go
+++ b/flytepropeller/pkg/controller/handler.go
@@ -230,7 +230,6 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
 	}
 
 	streak := 0
-	defer p.metrics.StreakLength.Add(ctx, float64(streak))
 
 	maxLength := p.cfg.MaxStreakLength
 	if maxLength <= 0 {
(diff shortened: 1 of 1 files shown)


flytepropeller/pkg/controller/handler.go
```diff
@@ -230,7 +230,6 @@ func (p *Propeller) Handle(ctx context.Context, namespace, name string) error {
 	}
 
 	streak := 0
-	defer p.metrics.StreakLength.Add(ctx, float64(streak))
```
eapolinario, line 233: Golang's defer semantics allows us to use this idiom with a slight change. From the [spec](https://go.dev/ref/spec#Defer_statements):
> Each time a "defer" statement executes, the function value and parameters to the call are [evaluated as usual](https://go.dev/ref/spec#Calls) and saved anew but the actual function is not invoked.

This means that we can define a function that closes over `streak` like so:
```
defer func() { p.metrics.StreakLength.Add(ctx, float64(streak)) }()
```

Tom-Newton: Done. As you can probably tell I don't really know what I'm doing with golang :sweat_smile:

Tom-Newton:
I rebased on latest master, which I'm hoping will fix the docs build error

Tom-Newton:
Pretty sure I fixed the linter error - I struggled a bit to run linting locally.

eapolinario (approved):

PR #5183, opened 2024-04-04
fix: renaming standaloneDeployment variable to be consistent

TresTristesTrigues:
This PR fixes 2 issues:

1. There was a mismatch with variable naming in the Flyte-Core chart, where standalone_deploy and standaloneDeployment were being used interchangeably
2. cluster_resource_manager could be deployed twice as part of flyteadmin and standalone

### DIFF
--- a/charts/flyte-core/templates/admin/deployment.yaml
+++ b/charts/flyte-core/templates/admin/deployment.yaml
@@ -74,7 +74,7 @@ spec:
             {{- end }}
             {{- end }}
         {{- end }}
-        {{- if and (.Values.cluster_resource_manager.enabled) (not .Values.cluster_resource_manager.standalone_deploy) }}
+        {{- if and (.Values.cluster_resource_manager.enabled) (not .Values.cluster_resource_manager.standaloneDeployment) }}
         - command:
           - flyteadmin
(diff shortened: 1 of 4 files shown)


neverett:
cc @davidmirror-ops to review the charts

davidmirror-ops (commented):
> cluster_resource_manager could be deployed twice as part of flyteadmin and standalone

Do you think this is still the case with the changes you're proposing?

Thanks!

TresTristesTrigues:
@davidmirror-ops 
I am not sure if it should be possible to deploy it has both standalone and flyte-admin, I assume not.
In that case, should this [line](https://github.com/flyteorg/flyte/blob/master/charts/flyte-core/templates/clusterresourcesync/deployment.yaml#L1) be changed to:
```
{{- if and (.Values.cluster_resource_manager.enabled) (.Values.cluster_resource_manager.standaloneDeployment) }}
```
? (in contrast to this [line](https://github.com/flyteorg/flyte/blob/master/charts/flyte-core/templates/admin/deployment.yaml#L77))

davidmirror-ops (approved):
Sorry for the delays. Thank you!

PR #5188, opened 2024-04-05
Change phase to queue on job submit for webapi plugins

pingsutw:
## Tracking issue

log links are not shown up on FlyteConsole when the task status is running.
It will also fix other webAPI plugins.

propeller only sends the event when the current event is not equal to the previous event, so if `ResourceCache` sets the phase to running here without info, the log link will never show up on flyteconsole when the status is running.

The phase should be `PhaseQueue` when there is no resource in the [cacheItem](https://github.com/flyteorg/flyte/blob/328de82f2d81fb4b8fc848d2c178858b49f2c2db/flyteplugins/go/tasks/pluginmachinery/internal/webapi/monitor.go#L33).

The async ResourceCache is empty if the `Get` [request](https://github.com/flyteorg/flyte/blob/ea02a22134acb6e548ed034829a2b17361a65c86/flyteplugins/go/tasks/pluginmachinery/internal/webapi/cache.go#L108) is not yet sent to the agent.

## How was this patch tested?
Local

### Screenshots
image

## Related PRs
NA

## Docs link
NA

### DIFF
--- a/flyteplugins/go/tasks/pluginmachinery/internal/webapi/monitor.go
+++ b/flyteplugins/go/tasks/pluginmachinery/internal/webapi/monitor.go
@@ -39,7 +39,7 @@ func monitor(ctx context.Context, tCtx core.TaskExecutionContext, p Client, cach
 			}
 			return state, core.PhaseInfoFailure(errors.CacheFailed, cacheItem.ErrorMessage, nil), nil
 		}
-		return state, core.PhaseInfoInitializing(time.Now(), core.DefaultPhaseVersion, "job submitted", nil), nil
+		return state, core.PhaseInfoQueued(time.Now(), core.DefaultPhaseVersion, "job submitted"), nil
 	}
 
 	newPhase, err := p.Status(ctx, newPluginContext(cacheItem.ResourceMeta, cacheItem.Resource, "", tCtx))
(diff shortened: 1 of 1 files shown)


Future-Outlier (commented):
It works, and it can also show the message now.
```go
return core.PhaseInfoRunningWithReason(core.DefaultPhaseVersion, resource.Message, taskInfo), nil
```
image
(I'm not sure what's happened to the error in the screenshot.)


However, this is still great, I've tested it in the sensor agent.
```python
return Resource(phase=TaskExecution.RUNNING, outputs=None, log_links=[TaskLog(uri="sensor_test_uri", name="Sensor Console").to_flyte_idl()])
```

Future-Outlier (commented):
I'm also wondering that cases for databricks agent, 
it will return `phase=TaskExecution.INITIALIZING` with loglinks when running agents.
If we implement this change, it will not show loglinks when `phase=TaskExecution.INITIALIZING`.

image

Future-Outlier:
> I'm also wondering that cases for databricks agent, it will return `phase=TaskExecution.INITIALIZING` with loglinks when running agents. If we implement this change, it will not show loglinks when `phase=TaskExecution.INITIALIZING`.
> 
> image

One possible solution is that we change the code in flytekit, and this will solve the problem.

pingsutw:
I changed the initial phase to queue; it will fix the Databricks issue as well.

PR #5192, opened 2024-04-05
fix: Modify the callback URL string in auth flow, to support custom base URLs in deployments

ddl-rliu:

In the auth flow, it seems that Flyteadmin always generates a callback URL, that looks like `foo.tech/callback`

However, when Flyte is configured on a custom subpath, this causes an issue, and we actually want the callback to look like `foo.tech/custom-subpath/callback`

By changing the `callbackURL` variable to "callback", the logic is fixed. Note that this also may require a change to the flyte config, since e.g. `foo.tech/custom-subpath/login` should be set as an AuthorizedURI in the `config.yaml`


flyteadmin/auth/auth_context.go
```diff
@@ -207,8 +207,10 @@ func GetOAuth2ClientConfig(ctx context.Context, options config.OpenIDOptions, pr
 
 	secret = strings.TrimSuffix(secret, "\n")
 
+	logger.Infof(ctx, "FOO %v %v", URLFromContext(ctx), URLFromContext(ctx).ResolveReference(callbackRelativeURL).String())
+
 	return oauth2.Config{
-		RedirectURL:  callbackRelativeURL.String(),
+		RedirectURL:  URLFromContext(ctx).ResolveReference(callbackRelativeURL).String(),
```
ddl-rliu, line 213: I believe the previous logic `RedirectURL: callbackRelativeURL.String()` causes issues when a deployment is using a custom base URL like `/custom-flyte/console` rather than `/console`. This is because the generated redirectURL for callbacks in the auth flow then becomes `redirect=host/callback` rather than `redirect=host/custom-flyte/console` which causes the callback to fail.

The fix here is to mirror how its done here in OAuth2ClientConfig https://github.com/ddl-rliu/flyte/blob/7953d627adf73dc66089f8d30e23acb594792bee/flyteadmin/auth/auth_context.go#L63

Here it resolves the callback using the request URL, to get an absolute URL that can be happily used in the auth flow.

ddl-rliu: Update: my reasoning here is likely incorrect, still investigating

flyteadmin/auth/auth_context.go
```diff
@@ -29,7 +29,7 @@ const (
 )
 
 var (
-	callbackRelativeURL = config.MustParseURL("/callback")
+	callbackRelativeURL = config.MustParseURL("callback")
```
ddl-rliu, line 32: I believe this is necessary, because when Flyteadmin generates the callback URL, the logic looks like `URL("foo.tech").ResolveReference(URL("/callback")) -> "foo.tech/callback"`

In fact, the base URL can have any path after it, but the result is the same. `URL("foo.tech/any/path").ResolveReference(URL("/callback")) -> "foo.tech/callback"`

This causes issues when flyte is configured on a custom subpath. To fix this...

By adding `foo.tech/custom-subpath/login` as an AuthorizedURI, and changing the reference to "callback", the logic becomes `URL("foo.tech/custom-subpath/login").ResolveReference(URL("callback")) -> "foo.tech/custom-subpath/callback"`

ddl-rliu: Added the test here https://github.com/flyteorg/flyte/pull/5192/files#diff-aa5ac252b2ec45c95649a3a874e727e436be5c932e3d93b5b2aadcfcc83d57feR12

Although I could also merge this into handlers_test.go, no strong preference.

eapolinario (approved):

eapolinario (approved):

PR #5195, opened 2024-04-07
Change phase to WaitingForResources when quota exceeded

pingsutw:
## Tracking issue
https://flyte-org.slack.com/archives/CP2HDHKE1/p1708818106936209

Tasks are stuck in the queue phase when the resource quota is exceeded since the propeller won't send the event to Flyteadmin if the current phase is equal to the [previous](https://github.com/flyteorg/flyte/blob/ea02a22134acb6e548ed034829a2b17361a65c86/flyteplugins/go/tasks/pluginmachinery/internal/webapi/allocation_token.go#L54) phase. Therefore, propeller won't update the state (`PhaseNotStarted` ->`PhaseAllocationTokenAcquired`) as well, and it will always assume that there is no resource available to execute the task.

Change phase to WaitingForResources when the quota is exceeded, so propeller will send the event to flyteadmin.

## How was this patch tested?
local sandbox and enable the resource manager

### Screenshots

Before:

Screenshot 2024-04-06 at 11 20 03 PM

After:

Screenshot 2024-04-06 at 11 03 40 PM

## Related PRs
NA

## Docs link
NA

### DIFF
--- a/flyteplugins/go/tasks/pluginmachinery/internal/webapi/allocation_token.go
+++ b/flyteplugins/go/tasks/pluginmachinery/internal/webapi/allocation_token.go
@@ -64,8 +64,8 @@ func (a tokenAllocator) allocateToken(ctx context.Context, p webapi.AsyncPlugin,
 		return &State{
 				AllocationTokenRequestStartTime: startTime,
 				Phase:                           PhaseNotStarted,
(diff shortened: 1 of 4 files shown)


EngHabu (approved):
Thank you, @pingsutw!

Issue #5189, opened 2024-04-05
Flyte Project isolation


robert-ulbrich-mercedes-benz:
### Discussed in https://github.com/flyteorg/flyte/discussions/5084

<div type='discussions-op-text'>

<sup>Originally posted by **robert-ulbrich-mercedes-benz** March 21, 2024</sup>
**Motivation**

Flyte allows having multiple tenants by using projects and domains. Flyte also comes with an authentication that only 
allows access to Flyte if the user is successfully authenticated. Once the user is successfully authenticated, she has full access to all projcets and domains in Flyte e.g. submit a workflow in any Flyte project. 

**Functional description**

In order to protect data from unauthorized access, it makes sense to isolate the Flyte projects from each other.
That way a user can only submit workflows in the project he or she is assigned to. If trying to register a launchplan, 
submitting a Flyte workflow or a task in a project the user is not assigned to, will cause an error.

**Implementation idea** 

The projects a user is assigned to, will be identified by reading a claim from an identity token that is issued in the course
of the authentication process. The claim to be read for this purpose is configured in Flyte's configuration file. Also a mapping
between the claim's value and the actual project to be assigned to is maintained in the Flyte configuration. 

**Limits**

So this approach only aims on isolating single Flyte tenants from each other and there is no intention to introduce a full blown
IAM concept including user roles. So any member of a Flyte project still has full access to their project's launch plans data etc.

</div>

tkent:
:wave: @robert-ulbrich-mercedes-benz 

Glad to see you picking this up. This has long been a need of ours as well, but we have no capacity in the foreseeable future to tackle the issue. We are handling the limitation by having separate Flyte environments (and clusters) per access boundary, which is not great.

I did want to bring up one facet around enforcing project-level access boundaries, which is how Flyte handles getting data to/from tasks via the data plane. Today, a sidecar on each Flyte task is responsible for communicating with the user & metadata storage containers. In most configurations, the entire task _pod_  will have access to the credentials used to access the storage bucket. This would need to be considered to make any access boundary enforced so that a clever task author doesn't just pickup the necessary credentials for the shared containers and access another project's data.

PR #5204, opened 2024-04-09
[RFC] - Project isolation

robert-ulbrich-mercedes-benz:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/5189_

It proposes a way how to isolate single tenants in Flyte from each other. 

Suggesting an RFC.

## How was this patch tested?

No tests applied.


rfc/system/5204-project-isolation.md
```diff
@@ -0,0 +1,110 @@
...
+## 3 Proposed Implementation
+
+The Flyte operators can enable or disable the project isolation feature. When enabled, Flyte will evaluate the identity token issued by the IDP and extract the values of a configured claim. The user can then 
+access a project if the claim value matches the project. If the project isolation feature is enabled and the claim value does not match any project, the user is denied access to all projects. So it is a whitelist mechanism.
+
+The only external interfaces of Flyte is the API that is being used by the Flyte console, the flytectl cli and pyflyte. So it is sufficient to add a check if the user interacting with the API for a certain project has the required permission.
+
+Since the API endpoints are very different from one another, the check needs to be implemented in each endpoint. Because of the different formats of the endpoint holding the project it will be hard to define a middleware centrally, which performs the check.
+
+The mapping between claim value and project to authorize can be configured in the following way and put in the Helm chart of the Flyte deployment:
```
katrogan, line 27: how does this work for dynamically registered projects via the CLI? is there a catch-all or default mapping or do deployments now need to be updated when a new project is registered?

rfc/system/5204-project-isolation.md
```diff
@@ -0,0 +1,110 @@
...
+
+```go
+func (m *AdminService) GetWorkflow(ctx context.Context, request *admin.ObjectGetRequest) (*admin.Workflow, error) {
+	defer m.interceptPanic(ctx, request)
+
+    if request == nil {
+        return nil, status.Errorf(codes.InvalidArgument, "Incorrect request, nil requests not allowed")
+    }
+	
+	if ProjectAccessPermitted(ctx, request.GetId().Project) {
```
katrogan, line 82: this is an opt-in mechanism, right? how do you ensure that new endpoints also get the same treatment?

rfc/system/5204-project-isolation.md
robert-ulbrich-mercedes-benz: For newly registered projects per RFC we will currently need to adapt the configuration in order to map the project to a certain claim value. We can also introduce a default mapping - if you think it is a good idea, I will adapt the RFC

rfc/system/5204-project-isolation.md
robert-ulbrich-mercedes-benz: Yes, it is an opt-in mechanism. We have meanwhile implemented it in a private fork of the project. We moved away from changing every single endpoint, but added a central middleware. But since there are no common fields for all of the request objects, actually each and every endpoint and its request object need to be explicitly mentioned, which I consider a weakness because if newly added endpoints are forgotten to be added in that middleware, then those endpoints will not benefit from the project isolation. Unfortunately we could not find a better solution to this issue.

robert-ulbrich-mercedes-benz: If you like to, I can also show you how we actually implemented it in the fork.

rfc/system/5204-project-isolation.md
katrogan: I like the idea of a default in that it reduces the friction for creating a new project, but then I guess it breaks from your proposed paradigm of explicitly white-listing access to projects.

My only concern is that since all it currently takes to create a project is a simple api call that any user can make, the change as it's currently proposed introduces more friction for self-service project registration (which is maybe undesirable in a context where you're already managing per-project access!)

I'm curious, in practice what has the overhead been like for your organization when it comes to provisioning new projects? Is that an infrequent occurrence?

rfc/system/5204-project-isolation.md
katrogan: @robert-ulbrich-mercedes-benz that all makes sense, I can see where maybe the middleware rejects unhandled endpoints by default but then that still requires manual code changes per new endpoint (and for newer or infrequent endpoints, it may not be obvious that the authorization check was missed)

How do you keep up with new service definition changes in your fork? Or is the process manual currently?

rfc/system/5204-project-isolation.md
robert-ulbrich-mercedes-benz: The Flyte instance in our organization is centrally managed. So it is not desired to have users creating projects on their own. In our organization project creation is put in the hands of the team managing the Flyte instance. So the Flyte projects we have are well defined and there are not too many projects being added or removed. 

In companies I assume the Flyte instances will usually be managed by a dedicated team.

rfc/system/5204-project-isolation.md
robert-ulbrich-mercedes-benz: We manually went over the GRPC service definitions that are available. We found out, that only Admin Services are of concern. But I share your concerns regarding new endpoints that might be missed in that central middleware.

davidmirror-ops:
Contributors meetup notes: Eduardo o have a sync with Robert to better review this.

davidmirror-ops:
06/06/2024 Contributors meeting notes: the core challenge is how to add an opt-in middleware to flyteadmin. We couldn't find a great solution to solve this, and just relying on developers being careful with listing potential new endpoints is not enough, it's too risky, with the risks outweighing the potential benefits. We tried to use reflection on flyteadmin but it failed completely so, if someone more versed in Go wants to try it, we'd be happy to support them. While we're glad it solves the author's use case on their fork, there are multiple technical objections that lead us to not accept this proposal as an upstream change to Flyte. cc @robert-ulbrich-mercedes-benz Thanks and please let us know your thoughts.

PR #5212, opened 2024-04-10
Remove extra EKS values file from flyte-binary

davidmirror-ops:
This PR removes a values file that doesn't contain anything new compared to the `eks-starter.yaml`  file provided.


pingsutw (approved):
Should we remove `eks-starter.yaml` instead? `eks-production.yaml` sounds like production ready 😆

davidmirror-ops:
@pingsutw I think the idea here is aligning with the other charts where a single `starter` values file is provided.  What's considered "production-grade"  varies a lot per organizations. Even this "starter" file is opinionated (eg. is ingress considered "starter"?) So I guess having the most common overrides on a values file is what we could look for :)

pingsutw:
Ah, I see; that makes sense to me.

PR #5216, opened 2024-04-10
helm: Add support for passing env variables to flyteadmin using envFrom

ctso:
## Tracking issue
Related to #3970 

Users may want to pass environment variables from Secrets or ConfigMaps instead of relying on the Helm chart's `flyteadmin.secrets` Helm value.  This is convenient when using a secrets manager such as External Secrets, for example.

Add a new Helm value, `flyteadmin.envFrom` which you can pass through directly to the underlying Deployment.

## How was this patch tested?
Tested locally using `helm template`, it works.


charts/flyte-core/templates/admin/deployment.yaml
```diff
@@ -47,6 +47,12 @@ spec:
             {{- tpl (toYaml .) $ | nindent 12 }}
             {{- end }}
             {{- end }}
+          {{- if .Values.flyteadmin.envFrom }}
+          envFrom:
+            {{- with .Values.flyteadmin.envFrom -}}
+            {{- tpl (toYaml .) $ | nindent 12 }}
+            {{- end }}
+            {{- end }}
```
pmahindrakar-oss, line 55: nit: can be indented with outer if

pmahindrakar-oss (approved):

charts/flyte-core/templates/admin/deployment.yaml
ctso: fae3f595

pmahindrakar-oss (approved):

wild-endeavor:
Should we match https://github.com/flyteorg/flyte/blob/ab95f7e7241afd9c020b1fd1dedd69f9ff613baa/charts/flyte-binary/templates/deployment.yaml#L169?

pingsutw (dismissed):

wild-endeavor:
@ctso I think it would feel nicer if at least the value var names matched between single binary and flyte-core.  yah?

davidmirror-ops:
@ctso what are we missing to ship this? There are users dealing with limitations where this feature would be helpful

ctso:
I think this could be merged as-is, though it sounds like maybe we want to update the single binary chart to have the same behavior as well?

davidmirror-ops:
@wild-endeavor not sure I follow, what you mean is matching in flyte-core the way [this section](url) enables envVars from different sources?
https://github.com/flyteorg/flyte/blob/0eaa20e8b92245555be8fd725f97e52fbed54e78/charts/flyte-binary/templates/deployment.yaml#L169-L177

davidmirror-ops (approved):

PR #5221, opened 2024-04-11
Clear finalizer on abort for arrays

### LINKED ISSUE #5222: [BUG] MapTasks don't cleanup pods when aborting an execution
### Describe the bug

When a workflow is triggered with map tasks and then cleaned up, pods are deleted but the finalizer never removed so pods stay in `Terminating` State

### Expected behavior

Propeller should not only delete the pod but also cleanup the finalizer so pod are properly removed.

Create a Workflow with a MapTask with many items, abort execution and check the pods in the kubernetes namespace

### Screenshots

_No response_

- [X] Yes

- [X] Yes

RRap0so:
## Tracking issue

Closes #5222 

This adds the ability for propeller to remove finalizer when the Phase of the plugin execution is aborted. This somewhat mimics what was done in the Happy path phase with small differences.

**Note:** We've tested this PR in our test cluster and noticed some slowness in cleaning the pods but they eventually are cleaned up (about 500 pods) not sure if expected or not and can test this further.

Before the deletion of pod we remove the finalizer of the aborted MapTask if set

## How was this patch tested?

In our test cluster, no unit tests added since there's no unit tests for `subtask.go`

### DIFF
--- a/flyteplugins/go/tasks/plugins/array/k8s/subtask.go
+++ b/flyteplugins/go/tasks/plugins/array/k8s/subtask.go
@@ -105,37 +105,65 @@ func abortSubtask(ctx context.Context, stCtx SubTaskExecutionContext, cfg *Confi
 	deleteResource := true
 	abortOverride, hasAbortOverride := plugin.(k8s.PluginAbortOverride)
 
-	resourceToFinalize := o
 	var behavior k8s.AbortBehavior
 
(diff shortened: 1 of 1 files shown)


RRap0so:
We're still investigating this, we found some odd behaviour when an execution is terminated too fast.

PR #5240, opened 2024-04-17
Add appProtocol to agent service to allow agent to work with istio

noahjax:
If you try to use an agent on a cluster with istio, flytepropeller is unable to connect to the agent. You will either need to remove the agent from flytepropeller's configmap or flytepropeller will fail on startup.

Connecting to flyteconsole is also broken in istio because missing appProtocol prevents some http ports from working correctly.

Specify an `appProtocol` that istio can use to determine how to proxy requests (see [here](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/) for more details on how this works). 

Update: I also modified the flyteadmin and flyteconsole services so that they will work with istio as well. More extensive changes are necessary to get everything working, but I will split those into a separate PR as they are likely more controversial

## How was this patch tested?
Tested on my flyte deployment with istio enabled for the whole cluster.


pingsutw (dismissed):

pingsutw:
@noahjax could you run `make helm` locally, and push it again?

Future-Outlier:
> @noahjax could you run `make helm` locally, and push it again?

@noahjax Maybe we need to try `make helm;make helm_update;`

pingsutw:
@ddl-ebrown could you run `make helm` to resolve the test failure? thanks!

pingsutw (approved):

pingsutw:
Thank you @noahjax @ddl-ebrown

fg91:
@noahjax I have a question about this change in `charts/flyte-core/templates/admin/service.yaml`:

```
    - name: grpc
      port: 81
      protocol: TCP
      # intentionally set to TCP instead of grpc
      appProtocol: TCP
```

Could you please explain why the `appProtocol` needs to be set to `TCP` for you instead of `grpc`?

When upgrading flyte-core from `1.12.0` (which does not have this change) to `1.13.0` my istio setup (described [here](https://github.com/flyteorg/flytekit/tree/master/plugins/flytekit-identity-aware-proxy)) stops working. It does work again if I remove the app protocol again or set it to `grpc` instead of `TCP`. I've tried for istio version 1.17.2 as well as the newest 1.23.2.

Could you please share which istio version you use and how your virtual service for flyteadmin looks like? Thank you 🙏

noahjax:
@fg91 Apologies in advance for the unsatisfying answer. I submitted these changes as part of my work to get Flyte + Istio running at my previous employer, but I have since left the company and as a result I no longer have access to any of the specific configuration details.

Maybe @ddl-ebrown would be able to share some additional information here

Issue #2333, opened 2022-04-08
[Core feature] Exponential backoff retry


mkao006:
### Motivation: Why do you think this is important?

Sometimes there are downtimes for an upstream system which causes the workflow to fail, the workflow would succeed after retrying at a later time. 

The limitation with the current retry is the maximum count and lacking the flexibility to set the interval thus the retry is exhausted before the upstream system recover from downtime.

### Goal: What should the final outcome look like, ideally?

The retry will be performed on an increased interval [1m, 3m, 10m, 30m etc] up to the number of retries.

### Describe alternatives you've considered 

No alternatives are available at the moment.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

hamersaw:
Commenting to keep open.

kumare3:
i really like this idea. 
I think from a flytekit pov, it should be implemented like so
```
# retries is simply implemented using a retry class, similar to container_image=ImageSpec or str
@task(retries=BackoffRetry(interval=timedelta(minutes=2), count=3))
def foo():
   ...
```

In flyteidl, we can add it to [RetryStrategy](https://github.com/flyteorg/flyte/blob/b6f35add1227c930e9208103235d89fe4b864b8e/flyteidl/protos/flyteidl/core/literals.proto#L179) which is already a class. 

In Flytepropeller we can add it to
[Propeller's struct](https://github.com/flyteorg/flyte/blob/b6f35add1227c930e9208103235d89fe4b864b8e/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/nodes.go#L67)

and then to the computation logic [here](https://github.com/flyteorg/flyte/blob/b6f35add1227c930e9208103235d89fe4b864b8e/flytepropeller/pkg/compiler/transformers/k8s/utils.go#L21)

We can use the last_updated_at from the node_status to wait for the retry. 

This should not be too hard to do.
We can also implement exponential backoff

kdhingra307:
@kumare3 will it be possible to add a feature where a user can provide the delay for retry? 

It is useful in case upstream system provides the delay mechanism

kumare3:
Yup absolutely- this is indeed the idea with retry strategy and implementation should be easy. But someone has to be contribute

kumare3:
@kdhingra307 join slack.flyte.org and we can collaborate

kdhingra307:
@kumare3 sure let's connect on slack once

[...]

PR #5263, opened 2024-04-22
Implement task retry with delay

mucahitkantepe:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/2333
Related to https://github.com/flyteorg/flytekit/pull/2368


eapolinario:
@mucahitkantepe , can we help you move this PR out of draft so that we can get it properly review?

mucahitkantepe:
@eapolinario sure, this is not the complete implementation as it's lacking the tests/docs etc but as the codebase is new for me, I wanted to make sure the logic I added is at the right place

flytepropeller/pkg/controller/nodes/executor.go
```diff
@@ -201,6 +201,16 @@ func (c *recursiveNodeExecutor) RecursiveNodeHandler(ctx context.Context, execCo
 		// 6. Thus we can get rid of SetInputs for StartNode as well
 		logger.Debugf(currentNodeCtx, "Handling node Status [%v]", nodeStatus.GetPhase().String())
 
+		// if it's configured to retry with delay, we should sleep before retrying
+		if nodeStatus.GetPhase() == v1alpha1.NodePhaseRetryableFailure {
+			if currentNode.GetRetryStrategy().RetryDelay != nil {
+				sleepDuration := time.Until(nodeStatus.GetTaskNodeStatus().GetLastPhaseUpdatedAt().Add(currentNode.GetRetryStrategy().RetryDelay.Duration))
+				if sleepDuration > 0 {
+					logger.Infof(currentNodeCtx, "Sleeping for [%v] before retrying", sleepDuration)
+					time.Sleep(sleepDuration)
```
kumare3, line 210: oooh we do not want to add sleep to the event loop. We just need to return and not actually do a retry till the time has elapsed.

flyteidl/protos/flyteidl/core/literals.proto
```diff
@@ -180,4 +180,7 @@ message RetryStrategy {
     // Number of retries. Retries will be consumed when the job fails with a recoverable error.
     // The number of retries must be less than or equals to 10.
     uint32 retries = 5;
+
+    // Delay between retries.
+    google.protobuf.Duration retry_delay = 6;
```
kumare3, line 185: shall we make this more elaborate - 
```
min_delay
max_delay
exponent
```

we can simply implement exponent of 1 for now and we can add other exponents later? we can help

flytepropeller/pkg/controller/nodes/executor.go
mucahitkantepe: ah ok, I thought this was part of a green thread.

If we just return and the delay is configured to be let's say 1 hour. Will propeller receive the same event in a loop for an hour that will overload it?

kumare3: it wont overload, it wont check it for next refresh interval. Another way to do this is to implement a timer.wheel that can check for active timers and fire the necessary workflow event. We can do that on a separate goroutine

eapolinario:

PR #5268, opened 2024-04-23
added configuration for arraynode default parallelism behavior

hamersaw:
## Tracking issue
_NA_

Defaulting to use the workflow parallleism in ArrayNode rather than unlimited may confuse some users and result in unexpected behavior.

This PR introduces ArrayNode default parallelism configuration, with three options:
- `unlimited`: works as the legacy maptask parallelism where all subNodes are evaluated
- `workflow`: applies the parent workflow parallelism to ArrayNode subNode evaluations
- `hybrid`: the previous changes allowed differentiation withing flytekit, where `nil` / `None` parallelism indicates using the parent workflow and `0` indicates using unlimited.

## How was this patch tested?
Extensive unit testing and deployed locally in multiple scenarios / configurations.

## Related PRs
https://github.com/flyteorg/flyte/pull/5214
https://github.com/flyteorg/flytekit/pull/2268

## Docs link
_NA_


flytepropeller/pkg/controller/config/config.go
```diff
@@ -258,6 +261,29 @@ type EventConfig struct {
...
+type ParallelismBehavior = string
+
+const (
+	// ParallelismBehaviorHybrid means that ArrayNode will adhere to the parallelism defined in the
+	// ArrayNode exactly. This means `nil` will use the workflow parallelism, and 0 will have
+	// unlimited parallelism.
+	ParallelismBehaviorHybrid ParallelismBehavior = "hybrid"
+
+	// ParallelismBehaviorUnlimited means that ArrayNode subNodes will be evaluated with unlimited
+	// parallelism for both nil and 0.
```
pvditt, line 274: nit: would it provide more clarity to update this comment with something like:

"If a non-nil and non-zero parallelism is defined, then ArrayNode will adhere to the parallelism defined."

to both unlimited and workflow descriptions

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -418,6 +406,12 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 			// wait until all tasks have completed before declaring success
 			arrayNodeState.Phase = v1alpha1.ArrayNodePhaseSucceeding
 		}
+
+		// if incrementWorkflowParallelism is not set then we need to increment the parallelism by one
+		// to indicate that the overall ArrayNode is still running
+		if !incrementWorkflowParallelism && arrayNodeState.Phase == v1alpha1.ArrayNodePhaseExecuting {
```
pvditt, line 412: ah - good catch with this

pvditt (dismissed):
definitely a lot clearer implementation. Left a nit/possible update to the config description.

flytepropeller/pkg/controller/nodes/array/handler.go
hamersaw: yeah, it's actually why i started diving in here. not sure how this slipped by for so long.

flytepropeller/pkg/controller/config/config.go
hamersaw: good call, just added.

pvditt (dismissed):

pvditt (approved):

PR #5294, opened 2024-04-29
Upgrade actions/checkout

pingsutw:
## Tracking issue
NA

Upgrade actions/checkout and set fetch-depth to 0 to fix the CI error.

https://github.com/flyteorg/flyte/actions/runs/8876065707/job/24366874983?pr=5126


.github/workflows/codespell.yml
```diff
@@ -21,7 +21,9 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
```
pingsutw, line 24: v3 and v2 is deprecated

thomasjpfan (commented):
Normally, I try to use the default `force-depth: 1` and only use `force-depth: 0` when the GitHub Action requires it. For this PR, can we only set `force-depth: 0` for the jobs that failed on the [CI logs](https://github.com/flyteorg/flyte/actions/runs/8876065707/job/24366874983?pr=5126)?

eapolinario:
Can we not use `force-depth: 0`? Essentially just move to v4, since the original failure in the PR description was fixed by https://github.com/flyteorg/flyte/pull/5298.

eapolinario (approved):

Issue #5316, opened 2024-05-02
[BUG] flytectl unit tests failing


eapolinario:
### Describe the bug

flytectl unit tests are failing pretty consistently in CI. 

Tests run locally and only fail occasionally.

Here's the full output:

```
2024-05-02T22:40:45.9508476Z Current runner version: '2.316.0'
2024-05-02T22:40:45.9536123Z ##[group]Operating System
2024-05-02T22:40:45.9537001Z Ubuntu
2024-05-02T22:40:45.9537475Z 22.04.4
2024-05-02T22:40:45.9538049Z LTS
2024-05-02T22:40:45.9538421Z ##[endgroup]
2024-05-02T22:40:45.9538763Z ##[group]Runner Image
2024-05-02T22:40:45.9539283Z Image: ubuntu-22.04
2024-05-02T22:40:45.9539698Z Version: 20240422.1.0
2024-05-02T22:40:45.9540848Z Included Software: https://github.com/actions/runner-images/blob/ubuntu22/20240422.1/images/ubuntu/Ubuntu2204-Readme.md
2024-05-02T22:40:45.9543007Z Image Release: https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20240422.1
2024-05-02T22:40:45.9544051Z ##[endgroup]
2024-05-02T22:40:45.9544475Z ##[group]Runner Image Provisioner
2024-05-02T22:40:45.9545063Z 2.0.369.1
2024-05-02T22:40:45.9545366Z ##[endgroup]
2024-05-02T22:40:45.9548339Z ##[group]GITHUB_TOKEN Permissions
2024-05-02T22:40:45.9550363Z Actions: write
2024-05-02T22:40:45.9551340Z Attestations: write
2024-05-02T22:40:45.9552019Z Checks: write
2024-05-02T22:40:45.9552577Z Contents: write
2024-05-02T22:40:45.9553315Z Deployments: write
2024-05-02T22:40:45.9553906Z Discussions: write
2024-05-02T22:40:45.9554549Z Issues: write
2024-05-02T22:40:45.9555065Z Metadata: read
2024-05-02T22:40:45.9555603Z Packages: write
2024-05-02T22:40:45.9556330Z Pages: write
2024-05-02T22:40:45.9556914Z PullRequests: write
2024-05-02T22:40:45.9557345Z RepositoryProjects: write
2024-05-02T22:40:45.9558253Z SecurityEvents: write
[...]

PR #5309, opened 2024-04-30
Fixup flytectl

eapolinario:
## Why are the changes needed?
This PR brings the necessary changes to run flytectl CI changes in accordance with the model defined for the other components in the monorepo, including bumping the version of go (from 1.19 to 1.21).

This PR also renames the flytectl module and all its internal imports.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

PR #5325, opened 2024-05-03
Fix flytectlt tests

eapolinario:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5316

We fixed a few of the flytectl issues in https://github.com/flyteorg/flyte/pull/5309, but decided to merge that PR with a few unit tests failing. This PR fixes them.

We have a few flytectl unit tests that rely on a [TestStruct](https://github.com/flyteorg/flyte/blob/master/flytectl/cmd/testutils/test_utils.go#L29-L43) to capture what's written to standard output and error, but if those file descriptors are not restored then the respective test suites would fail every now and then. In this PR we add a defer statement to ensure that the original file descriptors are restored to the os package.

## How was this patch tested?
Unit tests running in a loop locally (e.g. `seq 100 | xargs -n1 make test_unit`).

### DIFF
--- a/flytectl/cmd/create/execution_util_test.go
+++ b/flytectl/cmd/create/execution_util_test.go
@@ -48,6 +48,8 @@ func createExecutionUtilSetup() {
 
 func TestCreateExecutionForRelaunch(t *testing.T) {
 	s := setup()
+	defer s.RestoreStandardFileDescriptors()
+
 	createExecutionUtilSetup()
 	s.MockAdminClient.OnRelaunchExecutionMatch(s.Ctx, relaunchRequest).Return(executionCreateResponse, nil)
 	err := relaunchExecution(s.Ctx, "execName", config.GetConfig().Project, config.GetConfig().Domain, s.CmdCtx, executionConfig, "")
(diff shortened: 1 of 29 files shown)


flytectl/cmd/testutils/test_utils.go
```diff
@@ -70,32 +71,9 @@ func Setup() (s TestStruct) {
...
-	s.UpdaterExt.OnAdminServiceClient().Return(s.MockClient.AdminClient())
-	s.DeleterExt.OnAdminServiceClient().Return(s.MockClient.AdminClient())
-	s.MockAdminClient = s.MockClient.AdminClient().(*mocks.AdminServiceClient)
-	s.MockOutStream = s.Writer
-	s.CmdCtx = cmdCore.NewCommandContextWithExt(s.MockClient, s.FetcherExt, s.UpdaterExt, s.DeleterExt, s.MockOutStream)
-	config.GetConfig().Project = projectValue
-	config.GetConfig().Domain = domainValue
-	config.GetConfig().Output = output
-
-	return s
```
eapolinario, line 98: This function is an exact copy of `Setup`, so I'm dropping it.

flytectl/cmd/create/execution_util_test.go
```diff
@@ -48,6 +48,8 @@ func createExecutionUtilSetup() {
 
 func TestCreateExecutionForRelaunch(t *testing.T) {
 	s := setup()
+	defer s.RestoreStandardFileDescriptors()
```
katrogan, line 51: nit: maybe call this `teardown()` to match with go test name conventions, so users don't have to think about whether or not to call this function (and potentially miss doing so?)

pmahindrakar-oss (dismissed):

katrogan (dismissed):
nice find

katrogan (approved):

PR #5337, opened 2024-05-08
Idl/json type

wild-endeavor:


flyteidl/protos/flyteidl/core/literals.proto
```diff
@@ -46,6 +46,10 @@ message Binary {
     string tag = 2;
 }
 
+message Json {
+    string value = 1;
```
pingsutw, line 50: Should we use byte instead here?

PR #5340, opened 2024-05-08
add maptask cache clarification

dansola:

Confusion regarding map task caching and regular task caching.

Docs change 

## How was this patch tested?

No tests for docs

### DIFF
--- a/docs/user_guide/advanced_composition/map_tasks.md
+++ b/docs/user_guide/advanced_composition/map_tasks.md
@@ -58,7 +58,7 @@ def map_workflow_with_metadata(data: list[int] = [10, 12, 11, 10, 13, 12, 100, 1
         data_point=data
     )
 ```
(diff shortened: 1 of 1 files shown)


docs/user_guide/advanced_composition/map_tasks.md
```diff
@@ -58,6 +58,7 @@ def map_workflow_with_metadata(data: list[int] = [10, 12, 11, 10, 13, 12, 100, 1
         data_point=data
     )
 ```
+When `cache` and `cache_version` are used in `TaskMetadata` for a map task, the cache hits occur on individual tasks being mapped over, rather than the parent map task operation. This means that if one input item in a list changes, each previously executed tasks are read from cache and only the task for the changed item is actually executed, rather than re-executing the task for every item.  Note that this has the same effect as adding `cache` and `cache_version` in the `@task` decorator for a task being mapped over.
```
neverett, line 61:
```suggestion
When `cache` and `cache_version` are used in `TaskMetadata` for a map task, the cache hits occur on individual tasks being mapped over, rather than the parent map task operation. This means that if one input item in a list changes, each previously executed task is read from cache and only the task for the changed item is actually executed, rather than the task being re-executed for every item.  Note that this has the same effect as adding `cache` and `cache_version` in the `@task` decorator for a task being mapped over.
```

neverett (approved):
Two small suggestions, otherwise LGTM, thanks!

neverett (approved):

PR #5341, opened 2024-05-09
Add Google Tag Manager

EngHabu:

Instead of using scarf pixel directly, this adds Google GTM to allow us to experiment with different code snippets from the GTM side.

- [X] Remove scarf pixel
- [X] Add GTM to the base page

## How was this patch tested?

Used the deployed change from the PR.
Looked at the source code to confirm GTM is installed
Installed scarf through GTM and noticed it get injected.

### DIFF
--- a/docs/_templates/page.html
+++ b/docs/_templates/page.html
@@ -203,9 +203,7 @@
             {%- endif %}
           </div>
         </div>
-        {% endblock footer %}    
-        <div class="extra_footer">
-        </div>
+        {% endblock footer %}
       </footer>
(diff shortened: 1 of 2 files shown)


docs/_templates/page.html
```diff
@@ -205,7 +205,6 @@
         </div>
         {% endblock footer %}    
         <div class="extra_footer">
```
neverett, line 207: This entire div can be removed

neverett (dismissed):
One comment, otherwise LGTM

PR #5362, opened 2024-05-14
Improve flyte-core helm chart

### LINKED ISSUE #5361: [BUG] flyte-core helm chart missing ServiceAccounts and chart has warning messages
### Describe the bug

flyte-core helm chart is missing some configuration functionality

- It is missing ServiceAccounts on several Deployments
- It has a misconfiguration in the defaults in the values.yaml causing warning messages

```
coalesce.go:286: warning: cannot overwrite table with non table for flyte-core.flyteagent.podEnv (map[])
```
- It has naming convention inconsistencies

### Expected behavior

All Deployments should have the ability to create a ServiceAccount
Helm charts should not throw errors
Should have the ability to overwrite the deployment names (same as other charts, e.g., flyteagent)

_No response_

### Screenshots

_No response_

- [X] Yes

- [X] Yes

mvaalexp:
Fixed flyte-core helm chart where ServiceAccounts were missing, incorrect typing, and some naming inconsistencies

## Tracking issue
Closes #5361 

Removes error messages in when running the chart
```
coalesce.go:286: warning: cannot overwrite table with non table for flyte-core.flyteagent.podEnv (map[])
```

Adds missing ServiceAccounts to services. 
* This default service account has limited permissions and is generally not recommended for use in production environments
In our production clusters, we have governance that requires all Deployments must have an associated SA, this is currently preventing us from using the chart.

Adds the ability to override the service names.
* This makes it consistent with other charts like flyteagent
  * https://github.com/flyteorg/flyte/blob/master/charts/flyteagent/templates/_helpers.tpl#L17
* Our cluster governance has naming restrictions and this will allow us to use this chart

Changed FlyteClusterResourceSync Deployment name
* The names were hard coded
* There existed a template already in the _helpers file
* Labels did not match the resource names
* This allows for consistency across the services
* Note: Will delete and recreate the Deployment (not backwards compatible)

Helper methods refactored to allow `nameOverride`.
Service Accounts added and applied to Deployments.
values.yaml default values updated to array from map as that is what is expected

## How was this patch tested?
Hand tested, I didn't see unit tests for the helm chart.

### DIFF
--- a/charts/flyte-core/templates/clusterresourcesync/deployment.yaml
+++ b/charts/flyte-core/templates/clusterresourcesync/deployment.yaml
@@ -2,7 +2,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: syncresources
+  name: {{ template "flyteclusterresourcesync.name" . }}
   namespace: {{ template "flyte.namespace" . }}
   labels: {{ include "flyteclusterresourcesync.labels" . | nindent 4 }}
 spec:
@@ -55,7 +55,7 @@ spec:
           - mountPath: /var/run/credentials
             name: cluster-secrets
           {{- end }}
-      serviceAccountName: {{ .Values.cluster_resource_manager.service_account_name }}
+      serviceAccountName: {{ .Values.cluster_resource_manager.serviceAccount.create | ternary (include "flyteclusterresourcesync.name" .) .Values.cluster_resource_manager.service_account_name }}
       volumes:  {{- include "databaseSecret.volume" . | nindent 8 }}
         - configMap:
             name: clusterresource-template
--- a/charts/flyte-core/values.yaml
+++ b/charts/flyte-core/values.yaml
@@ -171,7 +171,7 @@ flytescheduler:
   # -- Annotations for Flytescheduler pods
   podAnnotations: {}
   # -- Additional Flytescheduler container environment variables
-  podEnv: {}
+  podEnv: []
   # -- Labels for Flytescheduler pods
   podLabels: {}
   # -- nodeSelector for Flytescheduler deployment
@@ -243,7 +243,7 @@ datacatalog:
   # -- Annotations for Datacatalog pods
   podAnnotations: {}
   # -- Additional Datacatalog container environment variables
-  podEnv: {}
+  podEnv: []
   # -- Labels for Datacatalog pods
   podLabels: {}
   # -- nodeSelector for Datacatalog deployment
@@ -329,7 +329,7 @@ flytepropeller:
   # -- Annotations for Flytepropeller pods
   podAnnotations: {}
   # -- Additional Flytepropeller container environment variables
-  podEnv: {}
+  podEnv: []
   # -- Labels for Flytepropeller pods
   podLabels: {}
   # -- nodeSelector for Flytepropeller deployment
@@ -926,7 +926,7 @@ cluster_resource_manager:
   # -- Annotations for ClusterResource pods
   podAnnotations: {}
   # -- Additional ClusterResource container environment variables
-  podEnv: {}
+  podEnv: []
   # -- Labels for ClusterResource pods
   podLabels: {}
   # -- nodeSelector for ClusterResource deployment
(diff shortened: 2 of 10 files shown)


charts/flyte-core/templates/clusterresourcesync/deployment.yaml
```diff
@@ -55,7 +55,7 @@ spec:
           - mountPath: /var/run/credentials
             name: cluster-secrets
           {{- end }}
-      serviceAccountName: {{ .Values.cluster_resource_manager.service_account_name }}
+      serviceAccountName: {{ .Values.cluster_resource_manager.serviceAccount.create | ternary (include "flyteclusterresourcesync.name" .) .Values.cluster_resource_manager.service_account_name }}
```
mvaalexp, line 58: Attempted to keep backwards compatibility here.

charts/flyte-core/templates/clusterresourcesync/deployment.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: syncresources
+  name: {{ template "flyteclusterresourcesync.name" . }}
```
mvaalexp, line 5: This will technically recreate the deployment when it renames, not sure if that is an issue.

charts/flyte-core/values.yaml
```diff
@@ -171,7 +171,7 @@ flytescheduler:
   # -- Annotations for Flytescheduler pods
   podAnnotations: {}
   # -- Additional Flytescheduler container environment variables
-  podEnv: {}
+  podEnv: []
```
mvaalexp, line 174: These are mapped to arrays, so it gives a warning when you set them in the values.yaml.

charts/flyte-core/templates/clusterresourcesync/deployment.yaml
wild-endeavor: does this re-create the deployment?  or does this create a new deployment?

wild-endeavor:
sorry it's taken a long time to get to this @mvaalexp - would you mind taking one more look and regenerating the generated files to fix the conflicts please?

Only flyte console and the cluster resource controller are getting new service accounts correct?  The others already have them?  Also the cluster controller deployment will get recreated?  Or will there be a duplicate one?

If you think it would be helpful to the community, could you also write a tiny blurb please that we can include on the release notes when this goes out?  Just so there are no surprises for anyone using this helm chart.  Would be most helpful.

charts/flyte-core/templates/clusterresourcesync/deployment.yaml
mvaalexp: new, it will tear down the old deployment and make a new one. it renames this to `flyteclusterresourcesync`

https://github.com/flyteorg/flyte/blob/master/charts/flyte-core/templates/_helpers.tpl#L62

mvaalexp:
No problem. I will take a look tomorrow and see what I am missing and write the blurb.

As I mentioned in the comment it will be new as a rename is basically a delete and create (as k8s can't rename a resource as far as I know). it renames this to flyteclusterresourcesync which is what was defined in the helpers file. If we can't afford to risk a rename here, we can probably work around it with a ternary or a new helper function, but it seems reasonable to keep it consistent if its not going to be a problem.

https://github.com/flyteorg/flyte/blob/master/charts/flyte-core/templates/_helpers.tpl#L62
```

mvaalexp:
### :pencil: Description
- Made the resource naming more configurable in case there is governance in place in the cluster that requires certain naming conventions on resources
- Added ServiceAccounts to Deployments that were missing them (this is in general best practice)
  - flyteclusterresourcesync, flyteconsole and flyte-core
- Renamed a few files to follow the standards from other charts (mostly sa.yaml -> rbac.yaml)
- Changed default helm values that were supposed to be arrays from objects to arrays (this was causing warnings in helm)

### :pencil: User Notes
- `syncresources` was renamed to `flyteclusterresourcesync` to follow the values found in the helper class rather than it being hard coded
  - This will delete and recreate the deployment
- clusterresourcesync now uses its own ServiceAccount instead of sharing flyteadmin's
  - If users changed `cluster_resource_manager`, this may cause some backwards compatibility issue

PR #5370, opened 2024-05-15
Fix flytectl demo start

eapolinario:
`flytectl demo start` fails with the newly released flytectl from the monorepo.

Filter out the tags used to release flytectl in the command used to start the demo cluster.

Over the course of this I noticed that the self-upgrade command is no longer working, but this will be fixed in a separate PR as the github API we're using to list tags is not returning the prefixed tags (i.e. https://api.github.com/repos/flyteorg/flyte/tags doesn't return `flytectl/vX.Y.Z`).

## How was this patch tested?
I built a local binary of flytectl and was able to start sandbox.

### DIFF
--- a/flytectl/pkg/github/githubutil.go
+++ b/flytectl/pkg/github/githubutil.go
@@ -27,14 +27,13 @@ const (
 	flyte                   = "flyte"
 	flytectl                = "flytectl"
 	sandboxSupportedVersion = "v0.10.0"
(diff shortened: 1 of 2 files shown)


eapolinario:
Issue to track the failures in the `upgrade` command: https://github.com/flyteorg/flyte/issues/5372

wild-endeavor (approved):

PR #5402, opened 2024-05-21
Update core helm chart for propeller configuration of agent service

Sovietaced:
The existing configuration for enabling the agent service was out of date and would result in unhandled tasks in propeller. This updates the helm chart to include default values for the agent service `supportedTaskTypes` which are required to be configured for tasks to be handled correctly by propeller.

### DIFF
--- a/charts/flyte-core/values.yaml
+++ b/charts/flyte-core/values.yaml
@@ -275,10 +275,19 @@ flyteagent:
   enabled: false
   plugin_config:
     plugins:
-      agentService:
+      # -- Agent service configuration for propeller.
+      agent-service:
+        # -- The default agent service to use for plugin tasks.
         defaultAgent:
+          # -- The agent service endpoint propeller should connect to.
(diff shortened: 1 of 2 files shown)


charts/flyte-core/values.yaml
```diff
@@ -277,12 +277,6 @@ datacatalog:
 
 flyteagent:
   enabled: false
-  plugin_config:
-    plugins:
-      agentService:
-        defaultAgent:
-          endpoint: "dns:///flyteagent.flyte.svc.cluster.local:8000"
-          insecure: true
```
pingsutw, line 285: Are we able to add `supportedTaskTypes` to this section? since we should only add this config when agent is enabled.

Sovietaced: Yeah! I can restructure the templates to do the right thing.

Sovietaced:
I just realized that the chart READMEs are templates and automatically generated so pushed an update for that.

pingsutw (approved):

eapolinario (approved):

Issue #5335, opened 2024-05-08
[BUG] Subworkflow status is not reported to the parent workflow


pablocasares:
### Describe the bug

After upgrading our cluster to v1.12.0 some workflow executions got stuck in "running" state when a subworkflow node failed. It seems that the failed state from the subworkflow is not being reported to the parent workflow so the workflow gets stuck in running state.

We rolled back flytepropeller to v1.11.0 and the issue disappeared. Looks like a race condition issue because it happens from time to time and restarting the workflow seems to mitigate the problem.

### Expected behavior

The subworkflow status should be reported back to the parent workflow.

### Screenshots

![image](https://github.com/flyteorg/flyte/assets/4056828/1fc2f64b-6c9b-4093-935b-57c23d25acd0)

Logs keeps saying `LaunchPlan running, parallelism is now set to [X]` - Node Event [failing] seems missing.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

kumare3:
This is a pretty serious bug? cc @pvditt

pvditt:
@pablocasares thank you for pointing this out. Were you able to determine if the workflow was still getting executed by propeller or was it just the console showing it as running? (Looking to see if this is just an eventing/persisting of state to admin bug)

pvditt:
@pablocasares My initial thoughts were this had to do with the [cache](https://github.com/flyteorg/flyte/blob/43c0f1a61b288ce150968630de3aef0fd0aa980e/flytepropeller/pkg/controller/nodes/subworkflow/launchplan/admin.go#L172) not getting updated but I'm not noticing anything while stepping through the code + looking at the updates.

Are you able to provide a workflow that could reproduce the error?

pvditt:
@pablocasares thank you for the added info. And just to circle back/confirm, 

- Were you able to determine if the parent workflow was still getting executed by propeller or was it just the console showing it as running?

- When you mention it happens time to time, does that mean you had cases of external/child workflows failing and then the parent workflow correctly handling that state while running propeller on 1.12.0?

pablocasares:
- Were you able to determine if the parent workflow was still getting executed by propeller or was it just the console showing it as running?

The parent workflow was still getting executed but stuck because it thought that the subworkflow node was still running (you can check the yaml I sent in our internal Slack channel)

[...]

PR #5406, opened 2024-05-22
[BUG] Handle auto refresh cache race condition

pvditt:
## Tracking issue

Potentially closes: #5335 

Propeller v1.12.0 introduced a bug in which child/external workflow status was not propagated back up to the parent workflow.  

Not able to repro exactly. Current theory is that there's a race condition in which an item be in the processing set (which was introduced in new Flyte release) while not being in the workqueue. Due to `if item, ok := value.(Item); !ok || (ok && !item.IsTerminal() && !w.processing.Contains(k)) {` (in [enqueueBatches](https://github.com/flyteorg/flyte/blob/master/flytestdlib/cache/auto_refresh.go)), this would cause an item to no longer get added to the workqueue to then be re-synced. 

Why we think this happens:
- the item (workflow) is still in the LruCache as we keep getting status for it in [GetStatus](https://github.com/flyteorg/flyte/blob/cfaedce95aa5327cc272682655eaa779a2f03f6d/flytepropeller/pkg/controller/nodes/subworkflow/launchplan/admin.go#L167).
- If the item were not in the cache, then the item would get re-added to the workqueue. If an item were in the workqueue, then it'd be included as part of the [syncItem](https://github.com/flyteorg/flyte/blob/cfaedce95aa5327cc272682655eaa779a2f03f6d/flytepropeller/pkg/controller/nodes/subworkflow/launchplan/admin.go#L227) process that's trigged in the auto_refresh's [sync](https://github.com/flyteorg/flyte/blob/cfaedce95aa5327cc272682655eaa779a2f03f6d/flytestdlib/cache/auto_refresh.go#L336). Sync grabs batches off the workqueue.
- [enqueueBatches](https://github.com/flyteorg/flyte/blob/cfaedce95aa5327cc272682655eaa779a2f03f6d/flytestdlib/cache/auto_refresh.go#L236) adds items to the workqueue. An item only gets added to the workqueue if it's [not in processing among other conditions](https://github.com/flyteorg/flyte/blob/cfaedce95aa5327cc272682655eaa779a2f03f6d/flytestdlib/cache/auto_refresh.go#L248).
- [gorm logs](https://github.com/flyteorg/flyte/issues/5335#issuecomment-2114752004) indicate that admin is not getting GetExecution requests for the child workflow that's status is not updating. 

- the addition of the processing sync.set was the only change that stood out in between flyte 1.11 and 1.12.

We want to keep the processing optimization to reduce to overhead of adding duplicate items to the workqueue. 

We swap out the processing set in favor of a map  in which they keys are the same set and the values are a timestamp of when the item was added to processing. We then check for how long the item has been in processing - if an item has been in processing for 10 sync periods we "evict" it from processing such that the item will get re-added to the workqueue.

## How was this patch tested?

- added a simple unit test for the inProcessing expiration check
- ran a workflow launching external wf -> ensured that status was propagated to the parent.


flytestdlib/cache/auto_refresh.go
```diff
@@ -359,6 +359,20 @@ func (w *autoRefresh) sync(ctx context.Context) (err error) {
 	}
 }
 
+// Checks if the item is currently being processed and returns false if the item has been in processing for too long
+func (w *autoRefresh) inProcessing(key interface{}) bool {
+	item, found := w.processing.Load(key)
+	if found {
+		// handle potential race conditions where the item is in processing but not in the workqueue
+		if timeItem, ok := item.(time.Time); ok && time.Since(timeItem) > (w.syncPeriod*10) {
```
pvditt, line 367: lean to not make this configurable

flytestdlib/cache/auto_refresh.go
```diff
@@ -359,6 +359,20 @@ func (w *autoRefresh) sync(ctx context.Context) (err error) {
 	}
 }
 
+// Checks if the item is currently being processed and returns false if the item has been in processing for too long
+func (w *autoRefresh) inProcessing(key interface{}) bool {
+	item, found := w.processing.Load(key)
+	if found {
+		// handle potential race conditions where the item is in processing but not in the workqueue
```
pingsutw, line 366: is that because one of the workers shuts down while processing the batch items?

flytestdlib/cache/auto_refresh.go
pingsutw: Should we increase the timeout (>30s) here? Since we enqueue batch items every 30 seconds by default.

pingsutw (commented):
LGTM, left some minor comments

flytestdlib/cache/auto_refresh.go
pingsutw: `w.syncPeriod*10` => 300 seconds by default. I feel we can reduce the timeout a bit. Maybe 5. wdyt

pingsutw (approved):

andresgomezfrr:
Is there any plan to create a fix release with this fix?

pablocasares:
Thank you for working on this. Once the next release is available I can test this and I'll report back if the issue is solved.

pvditt:
> Is there any plan to create a fix release with this fix?

@andresgomezfrr Yes, we are validating a new release end of this week and barring any issues will get an official release out next week.

pvditt:
@andresgomezfrr @pablocasares there's a [RC](https://github.com/flyteorg/flyte/tree/v1.12.1-rc0) that contains this fix. I'm unsure of when a final release containing this change will be made. I'll ping when that happens.

Issue #5387, opened 2024-05-17
[BUG]  Fix non thread safe token cache behavior


pmahindrakar-oss:
### Describe the bug

Currently the client auth interceptor used by multiple flyte services faces an issue when multiple parallel attempts are made to get token as the underlying token source implementation is not thread safe

https://github.com/flyteorg/flyte/blob/master/flyteidl/clients/go/admin/auth_interceptor.go#L145

Eg : https://github.com/flyteorg/flyte/blob/master/flyteidl/clients/go/admin/cache/token_cache_inmemory.go

### Expected behavior

The implementation of token cache should be thread safe

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

pingsutw:
@pmahindrakar-oss, does #5407 fix it? If so, could we close the issue?

pmahindrakar-oss:
Yes we can close this issue

PR #5407, opened 2024-05-22
Upstream revert revert auth token fix

pmahindrakar-oss:
Fixes race condition in token auth cache https://github.com/flyteorg/flyte/issues/5387

Fixes the racy behavior in auth token cache using mutexes and condition variables.

In a nutshell, the change does the following.

* Allows all parallel token requesters to try to acquire a lock
* One of the requester gets the lock and rest get waitlisted and wait for notification from the requester who acquired the lock
* The requester who acquired the lock refreshes the token and save to the token cache
* He also notifies the waitlisted requesters to now go and read the saved token cache and proceed with the request

The test plan below summarizes how this was tested

## How was this patch tested?

This was tested locally using integration test and was ran using 20 parallel routines
```
go test -v -tags=integration -run '^TestAPISuite$/^Test_RandomNameWorkflow$'
```

* All routines get unauthenticated in the first attempt of using inmemory cached token provider
```
cat t |grep "If it's an unauthenticated error, we will attempt to establish an authenticated context" |wc -l
      20
```
*  All routines try to lock
```
cat t |grep "try lock"  |wc -l                                                         python
      20
```
*  One of the routine is able to acquire the lock
```
cat t |grep "Locked : true"  |wc -l                                                    python
       1
```

* Rest fail to lock
```
cat t |grep "Locked : false"  |wc -l                                                   python
      19
```
* The ones who fail to lock go into waiting state . i.e they get added to condition variables waitlist

```
cat t |grep "Waiting" |grep -v "Coming" |wc -l                                         python
      19
```

eg : log for ref
```
{"json":{"src":"token_cache_inmemory.go:77"},"level":"info","msg":"Coming out of Waiting","ts":"2024-05-22T23:52:47-07:00"}
```

* The one who has locked goes and saves the token
```
cat t |grep "Saved"  |wc -l                                                            python
       1
```

* Broadcasts to all the others waiting using single notification which notifies all the waiters.
```
cat t |grep "Broadcasted"  |wc -l                                                      python
       1
```

* Post which they come out of waiting
```
 cat t |grep "Coming out of Waiting"|wc -l                                              python
      19
```

And go ahead and read the token

Log file for reference

[t.txt](https://github.com/unionai/flyte/files/15413364/t.txt)


flyteidl/clients/go/admin/cache/token_cache_inmemory.go
```diff
@@ -2,23 +2,93 @@ package cache
...
+func (t *TokenCacheInMemoryProvider) Unlock() {
+	t.mu.Unlock()
+}
+
+// CondWait  adds the current go routine to the condition waitlist and waits for another go routine to notify using CondBroadcast
+// The current usage is that one who was able to acquire the lock using TryLock is the one who gets a valid token and notifies all the waitlist requesters so that they can use the new valid token.
+// It also locks the Locker in the condition variable as the semantics of Wait is that it unlocks the Locker after adding
+// the consumer to the waitlist and before blocking on notification.
+// We use the condLocker which is noOp locker to get added to waitlist for notifications.
+// The underlying notifcationList doesn't need to be guarded as it implmentation is atomic and is thread safe
```
eapolinario, line 53: please, fix the typo.

pmahindrakar-union-02: done

eapolinario (approved):

eapolinario (dismissed):

flyteidl/clients/go/admin/cache/token_cache_inmemory.go
pmahindrakar-oss: done

eapolinario (approved):

Issue #5366, opened 2024-05-15
[Core feature] Flytekit should support using output with `Non-Any` type as the input with `Any` type.


Mecoli1219:
### Motivation: Why do you think this is important?

Intuitively, `Any` type should take every possible type as the input. Flytekit today supports using all Python types as the input, but it doesn't support 
1. the non-Any output from another `task`.
2. the non-Any input from the `workflow`.

### Goal: What should the final outcome look like, ideally?

The following scenarios should succeed:

Scenario 1:
```
@task()
def foo(a: Any) -> int:
    if type(a) == int:
        return a + 1
    return 0

@workflow
def wf(a: int) -> int:
    return foo(a=a)
```

Scenario2:
```
@workflow
def wf1(a: Any) -> int:
    a1 = foo(a=a)
    return foo(a=a1)
```

### Describe alternatives you've considered 

Adding some tags as type hints for the transformer?

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

eapolinario:
Related to https://github.com/flyteorg/flyte/issues/5319.

Future-Outlier:
This is important to my LangChain Agent PR, so I will take over it.

Mecoli1219:
NP

PR #5408, opened 2024-05-22
[flytepropeller] [compiler] Support non-Any Python types as Any input in workflows

Future-Outlier:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5366

We want to allow the Any type to accept all kinds of inputs ~~and outputs~~, not just the types we currently can't handle.

For example, workflow like this.
```python
@task
def foo(a: Any) -> int:
    if type(a) == int:
        return a + 1
    return 0

@workflow
def wf(a: int) -> int:
    return foo(a=a)
```

## Potential Issues
Note: This PR only supports `Any` in python, if we want to support java, we will need to add more code.

1. Added a function `isTypeAny` to check if ~~`upstreamType` or~~ `downstreamType` in the workflow is `Any` in Python.
2. Updated the AreTypesCastable function to use the `isTypeAny` function. If ~~`upstreamType`~~ or `downstreamType` is `Any` in Python, the function should return `true`, indicating that the types are castable and the compilation passes.

## How was this patch tested?
unit test and remote cluster

```python
from typing import Any

@task
def foo(a: Any) -> int:
    if type(a) == int:
        return a + 1
    return 0

@workflow
def wf(a: int) -> int:
    return foo(a=a)
```
```Dockerfile
FROM python:3.9-slim-buster
USER root
WORKDIR /root
ENV PYTHONPATH /root
RUN apt-get update && apt-get install build-essential -y
RUN apt-get install git -y

RUN pip install -U git+https://github.com/flyteorg/flytekit.git@3804a5155a523e2028ece8e7e581f794ebabd788
```

### Setup process
```python
pyflyte run --remote --image localhost:30000/any:0522 any_task.py wf --a 1
```
### Screenshots
image
image

## Related PRs
https://github.com/flyteorg/flytekit/pull/2432


Future-Outlier:
Should we try to add a new type scalar Any here?
https://github.com/flyteorg/flyte/blob/master/flyteidl/protos/flyteidl/core/literals.proto#L80-L92

Future-Outlier:
Not sure if we should treat `Any` as `enum SimpleType` or create a field for Any like
```proto
message AnyType {
    LiteralType variant = 1;
}
```

https://github.com/flyteorg/flyte/blob/362bea3428d75c3dfa02405d0f3d6440971566e3/flyteidl/protos/flyteidl/core/types.proto#L10

cosmicBboy:
@kumare3 isn't this necessary for gradual typing that serializes as pickle (at least for Python)?

kumare3:
My concern is - how do we coerce the type into another type? I think once something is any it has to be any all the way through.
If we pickle we cannot move to int. We can move from into to any maybe - that too if we support it

fg91:
> Hold on - I am not sure we should do this? Cc @EngHabu / @fg91 / @cosmicBboy Can you folks tal

I'm personally also not convinced we need to support `Any`, type safety is a good thing and Flyte workflows are a [DSL after all that uses a subset of python syntax](https://docs.flyte.org/en/latest/flyte_fundamentals/tasks_workflows_and_launch_plans.html#workflows-versus-tasks-under-the-hood) but doesn't support all of python.
I do, however, feel we should catch when users want to use `Any` and raise an error mentioning the DSL (like e.g. [this](https://github.com/flyteorg/flytekit/blob/72af7c6d8982892047e7051e1af10d628f9f9b39/flytekit/core/promise.py#L1088) one) and explain that workflows are type safe.

Future-Outlier:
I will list why we need this and how we can do it today.
Thank you all!

Future-Outlier:
@kumare3 @pingsutw @cosmicBboy @fg91 @EngHabu 

# Why do we need this PR?

In the upcoming agent, LangChain, we will use `Interface(inputs={"input": Any}, outputs={"o0": Any})` 
for our `LangChainTask`.
Since LangChain supports many types,
it would be very difficult to write numerous type transformers for all types.

However, our ideal LangChain workflow might look like this:

```python
import os
from typing import Any, Union

from flytekit import workflow
from langchain_core.output_parsers import StrOutputParser
from langchain_core.prompts.prompt import PromptTemplate
from langchain_openai import ChatOpenAI

api_key = os.environ.get("OPENAI_API_KEY")

model = ChatOpenAI(
    model="gpt-3.5-turbo",
    openai_api_key=api_key,
    openai_organization="org-NayNG68kGnVXMJ8Ak4PMgQv7",
)

prompt = PromptTemplate(
    input_variables=["question"],
    template="Question: {question}?",
)

output_parser = StrOutputParser()

@workflow
def wf(input: str) -> Union[str, Any]:
    message = prompt(input=input)
    o0 = model(input=message)
    o1 = output_parser(input=o0)
    return o1
```

It's necessary for users to input `str` as the prompt input in a LangChain task.

Reference for Flytekit LangChain task implementation: https://github.com/flyteorg/flytekit/pull/2436/files#diff-76c7b754bdfc3be1caaba940a66376edd7f443185fe492e7ac16a142cf63c70dR54-R55
Reference for LangChain (input schema and output schema): https://python.langchain.com/v0.1/docs/expression_language/interface/#input-schema

# How can we support this in Flytekit?
- For default inputs, we can iterate over all type transformers to check if it's possible to convert the literal to a Python value.
- For others, we will use metadata to get the python_dotted_path and convert it to the right type we need.

```python
class TypeEngine(typing.Generic[T]):

    @classmethod
    def to_literal(cls, ctx: FlyteContext, python_val: typing.Any, python_type: Type, expected: LiteralType) -> Literal:
        # The `metadata` field is added because of the artifact feature
        try:
            metadata.update({"python_dotted_path": f"{python_type.__module__}.{python_type.__qualname__}"})
            lv.set_metadata(metadata=metadata)
        except AttributeError as e:
            logger.warning(f"Attribute error occurred: {e}")

class FlytePickleTransformer(TypeTransformer[FlytePickle]):

    def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[T]) -> T:
        try:
            uri = lv.scalar.blob.uri
            return FlytePickle.from_pickle(uri)
        except Exception as e:
            from pydoc import locate

            metadata = lv.metadata
            if metadata and metadata.get("python_dotted_path"):
                python_dotted_path = metadata.get("python_dotted_path")
                py_type = locate(python_dotted_path)
                if py_type != typing.Any:
                    return TypeEngine.to_python_value(ctx, lv, py_type)
            raise e
```
PR reference: https://github.com/flyteorg/flytekit/pull/2432

# Pros
1. We can now support non-Any types in a workflow, which is more intuitive.
2. LangChain agent can benefit from this.

# Cons
1. Slightly breaks type safety.

davidmirror-ops:
06/06/2024 Contributors meetup notes: while this change is not considered completely necessary, more follow-up discussion to come.

Future-Outlier:
Hi, folks.
@kumare3 @cosmicBboy @fg91 @EngHabu 
Let me clarify the details of this PR and some discussion I've finished with @pingsutw.

1. We only want to support Non-Any to Any in workflow bindings.

for example:
(O) int -> Any 
(X) Any -> int

So the example @cosmicboys created should fail.
Cause this is int -> Any

```python
@task
def foo(a: Any) -> Any:
    if type(a) == int:
        return a + 1
    return 0

@task
def bar(a: int) -> int:
    return a * 2

@workflow
def wf(a: int) -> int:
    x = foo(a=a)  # here x is Any
    return bar(a=x)  # here should show compile error
```

2. How can we support it in flytekit?

We can use the field metadata in our literal value.

reference: https://github.com/flyteorg/flyte/blob/master/flyteidl/protos/flyteidl/core/literals.proto#L95-L114

For example:
TypeEngine 

```python
def to_literal:
    # lv is LiteralValue
    lv.metadata.update({"python_type": str(python_type)})
```

FlytePickleTransformer
```python
def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[T]) -> T:
    metadata = lv.metadata

	try:
	    uri = lv.scalar.blob.uri
            if lv.scalar.blob.metadata.type.format == self.PYTHON_PICKLE_FORMAT:
                return FlytePickle.from_pickle(uri)
            elif lv.scalar.blob.metadata.type.dimensionality == BlobType.BlobDimensionality.MULTIPART:
                return TypeEngine.to_python_value(ctx, lv, FlyteDirectory)
            elif lv.scalar.blob.metadata.type.dimensionality == BlobType.BlobDimensionality.SINGLE:
                return TypeEngine.to_python_value(ctx, lv, FlyteFile)

         except Exception as e:
             metadata = lv.metadata
             if metadata and metadata.get("python_type"):
                 python_type = metadata.get("python_type")
                 py_type = eval(python_type)              # turn 'list[int]' to list[int]
                 print("@@@ pickle -> py_type:", py_type)
             if py_type != typing.Any:
                 return TypeEngine.to_python_value(ctx, lv, py_type)

         raise e

def get_literal_type(self, t: Type[T]) -> LiteralType:
    lt = LiteralType(simple=SimpleType.ANY)
    lt.metadata = {"python_class_name": str(t)}
    lt.metadata = {"isAny": str(t == typing.Any)} # check downstream type is any or not in propeller's compiler 

    return lt
```

Does this implementation look good to you all?
Please take a look, thank you!

PR #5425, opened 2024-05-28
[monodocs] Fix build failure

eapolinario:
We're seeing build failures (e.g. ) caused by https://github.com/flyteorg/flytesnacks/pull/1660.

Update line numbers to match new lines.

### DIFF
--- a/docs/user_guide/data_types_and_io/flytedirectory.md
+++ b/docs/user_guide/data_types_and_io/flytedirectory.md
@@ -83,7 +83,7 @@ You can run the workflow locally as follows:
 
 ```{rli} https://raw.githubusercontent.com/flyteorg/flytesnacks/master/examples/data_types_and_io/data_types_and_io/folder.py
 :caption: data_types_and_io/folder.py
-:lines: 98-118
+:lines: 94-114
 ```
 
 [flytesnacks]: https://github.com/flyteorg/flytesnacks/tree/master/examples/data_types_and_io/


docs/user_guide/data_types_and_io/flytedirectory.md
```diff
@@ -83,7 +83,7 @@ You can run the workflow locally as follows:
 
 ```{rli} https://raw.githubusercontent.com/flyteorg/flytesnacks/master/examples/data_types_and_io/data_types_and_io/folder.py
```
eapolinario, line 84: @neverett , this could have been caught in https://github.com/flyteorg/flytesnacks/pull/1660 if this reference didn't mention `master` as the branch, but the PR branch. This seems to be a [pattern](https://gist.github.com/eapolinario/3ac39ce4cf0e08e6128174be245f9cef) in our docs, but I'm wondering if we shouldn't lean on a bit of automation to catch these at PR time.

pingsutw (approved):

docs/user_guide/data_types_and_io/flytedirectory.md
neverett: Yeah, we do reference `master` in all the places where we use the `rli` directive. I'm in favor of automation to catch these at PR time, although I'm not sure what that would look like -- I can make a Linear ticket to investigate.

eapolinario: It turns out that we had another failure (fixed in https://github.com/flyteorg/flyte/pull/5428/). Notice how in the aforementioned PR we're trading staleness (as in changes to the original file will not be reflected to the docs since we're pointing to a specific SHA) for consistency.

@neverett , can you open a Linear ticket to discuss which approach we should take going forward?

neverett: @eapolinario in the interest of avoiding build failures, I'll switch all `rli` URLs to SHAs for now, plus open a ticket so we can discuss how we want to approach this long term. When I started the work of separating docs from example code, the user guide examples were fairly stable, but they've seen a lot of activity in the last few weeks, and I assume will continue to be updated regularly going forward.

PR #5431, opened 2024-05-29
Inherit execution cluster label from source execution

va6996:
## Tracking issue

Today if a user creates an execution with the `execution-cluster-label`, then the execution launches in the specified cluster. However, if the workflow has a reference workflow in the workflow, that reference workflow launches in the default cluster since this information isn't passed along. This change fixes that issue by inheriting the cluster label from the source.

Without this change, part of the workflow will run in cluster 1 and part in cluster 2.

## How was this patch tested?
In a local cluster, a workflow was launched with a reference workflow and the reference workflow launched in the correct cluster.


RRap0so (approved):
Nice find!

PR #5444, opened 2024-06-03
Return non-0 exit code from copilot upload failure

ddl-ebrown:

 - The current copilot behavior is to:

   * try to upload the specified file
   * if an error occurs, try to write an error file instead
     * if the error file write succeeds, return success -- otherwise error

   This behavior is not ideal, because it suppresses real issues like
   the inability to find the expected output file. While a case can be
   made that the sidecar succeeded in handling its responsibility,
   important information is lost in the process that is useful to other
   containers in the job.

   Generallyk, it's better to consider the inability to
   find the original file a failure of the job to produce the expected
   file in the expected location (useful especially for container tasks).
   That means that a failure to upload the file should also return a
   non-0 exit code.

 - The new behavior is therefore:

   * try to upload the specified file
   * if an error occurs, try to write an error file instead
     * if the error file write succeeds, return the original error
     * if the error file write fails, return a joined version of the errors

### DIFF
--- a/flytecopilot/cmd/sidecar.go
+++ b/flytecopilot/cmd/sidecar.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	stderrors "errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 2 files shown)


wild-endeavor:
I like the joining of the errors but personally I think the original behavior of "if the error file write succeeds, return success -- otherwise error" more if only because this is the pattern that flytekit follows for the vast majority of its tasks.  I think to change this we'd also have to understand how the plugin would handle the case.  An error file implies a clean exit, implying that a user error occurred rather than a system error.  The case where copilot is not able to find the expected output file is an okay scenario for making it a non-0 exit code, since that's kind of like a system level error, but in the normal case I think it should still be 0.

If you really need it, maybe you can opt in via an environment variable or something... interpret all errors as exit code 1.  However I'm not sure if propeller checks for an errors.pb file if that happens though, so if there's information there then it might get lost.  This is kinda of an unspoken API, which I don't like.  There's a contract between task and engine that's not formally spelled out.  I think this is why i'm averse to changing it.

eapolinario:

Issue #5201, opened 2024-04-08
[BUG] ArrayNode maptask subnode/task eventing


pvditt:
### Describe the bug

- Map task succeeds with respect to min_success_ratio but the last running tasks fail. The last running tasks should get set to phase failed but are stuck in running according to console. This is just an eventing issue.

- subnodes that are cache hits and evaluated on the last loop before success appear in state "UNKOWN"

### Expected behavior

Parent task succeeds + subtask has phase failed

### Additional context to reproduce

```
@task(
    retries=0,
)
def hello(name: str) -> None:
    if name == "b":
        time.sleep(10)
        raise Exception("This is a test exception")
    return


@workflow
def map_workflow():
    map_task(hello, min_success_ratio=0.5, concurrency=0)(name=["a", "b", "c"])
```

Run this multiple times:
```
@task(
    cache_version="1.0",
    cache=True,
)
def square(val: int) -> int:
    return val ** 2


@workflow
def map_square() -> list[int]:
    temp = list(range(1, 51))
    a = map_task(square, concurrency=30)(val=temp)
    return a
```

### Screenshots

![image](https://github.com/user-attachments/assets/7b53e071-f189-40e7-bed5-8823bc894c1c)

![image](https://github.com/user-attachments/assets/bed54258-4d24-4a46-bdf9-6860c828fae0)

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #5451, opened 2024-06-05
[Bug] fix ArrayNode state's TaskPhase reset

pvditt:
## Tracking issue

Resolves: https://github.com/flyteorg/flyte/issues/5201

When a NodePhase change is detected, the current implementation resets the TaskPhaseVersion to 0 prior to emitting the event. This would create a duplicate event  since the TaskPhase isn't also updated causing for the event to not get emitted/persisted to admin.

- Bump the task phase version after emitting the event when a NodePhase change is detected in the ArrayNode handler.

---

We opt to do this instead of bumping the task phase as that can cause issues. 
Example issue:
- updating the task phase to a terminal phase and then emitting that event to admin would set that task execution to a terminal state 
- admin does not persist follow up event when a task execution is already in a terminal state
- the next propeller loop that handles cleanup wouldn't be able to persist new task events such as aborting subnodes on faliure cleanup.

Also this seems to be consistent with letting the proceeding propeller loop handling the next state. 

## How was this patch tested?

Ran through different failing and succeeding scenarios and ensured that the subnodes phases transitioned to the correct terminal phase.

### Setup process

Repro issues:
```
@task(
    retries=0,
)
def hello(name: str) -> None:
    if name == "b":
        time.sleep(10)
        raise Exception("This is a test exception")
    return


@workflow
def map_workflow():
    map_task(hello, min_success_ratio=0.5, concurrency=0)(name=["a", "b", "c"])
```

Run this multiple times:
```
@task(
    cache_version="1.0",
    cache=True,
)
def square(val: int) -> int:
    return val ** 2


@workflow
def map_square() -> list[int]:
    temp = list(range(1, 51))
    a = map_task(square, concurrency=30)(val=temp)
    return a
```

### Screenshots

image

image


flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -576,18 +576,20 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
...
 
 		if err := eventRecorder.finalize(ctx, nCtx, taskPhase, arrayNodeState.TaskPhaseVersion, a.eventConfig); err != nil {
 			logger.Errorf(ctx, "ArrayNode event recording failed: [%s]", err.Error())
 			return handler.UnknownTransition, err
 		}
+
+		// if the ArrayNode phase has changed we need to reset the taskPhaseVersion to 0
+		if currentArrayNodePhase != arrayNodeState.Phase {
+			arrayNodeState.TaskPhaseVersion = 0
+		}
```
hamersaw, line 592: Doesn't setting this afterwards cause problems with admin requiring incremental values?

pvditt: The scenario of interest would be if we would lose some eventing data if we have incrementPhase = False and currentArrayNodePhase != arrayNodeState.Phase + have a previous event emitted with the same TaskPhase and TaskPhaseVersion.

We set incrementTaskPhaseVersion = True if there's a subnode phase change. The arrayNodeState.Phase is updated in 3 different places: [arrayNodeState.Phase = v1alpha1.ArrayNodePhaseFailing](https://github.com/flyteorg/flyte/blob/25c35961f7d39b75868c0cad312984d73aa245ae/flytepropeller/pkg/controller/nodes/array/handler.go#L404), [arrayNodeState.Phase = v1alpha1.ArrayNodePhaseSucceeding](https://github.com/flyteorg/flyte/blob/25c35961f7d39b75868c0cad312984d73aa245ae/flytepropeller/pkg/controller/nodes/array/handler.go#L407), and [arrayNodeState.Phase = v1alpha1.ArrayNodePhaseExecuting](https://github.com/flyteorg/flyte/blob/25c35961f7d39b75868c0cad312984d73aa245ae/flytepropeller/pkg/controller/nodes/array/handler.go#L407).

For arrayNodeState.Phase = v1alpha1.ArrayNodePhaseFailing and arrayNodeState.Phase = v1alpha1.ArrayNodePhaseSucceeding, there would have to be subnode phase change so we couldn't have a scenario where incrementPhase = False and currentArrayNodePhase != arrayNodeState.Phase.

For arrayNodeState.Phase = v1alpha1.ArrayNodePhaseExecuting, we shouldn't have a previous event emitted with the same TaskPhase and TaskPhaseVersion as we should only be in the v1alpha1.ArrayNodePhaseNone phase for the first pass through.

pvditt: The proceeding loops would have a new TaskPhase as well.

hamersaw: > The scenario of interest would be if we would lose some eventing data if we have incrementPhase = False and currentArrayNodePhase != arrayNodeState.Phase + have a previous event emitted with the same TaskPhase and TaskPhaseVersion.
Do we have a repro for this? I'm having difficulty understanding how this is possible. If `incrementPhase == False` then no subnode phases have been updated so `currentArrayNodePhase != arrayNodeState.Phase` cannot be true since they are determined from subNode phases - unless this is a system failure in the handler logic?

Then `taskPhase` is deterministic on `currentArrayNodePhase`, so if `currentArrayNodePhase != arrayNodeState.Phase` then we can not have emitted an event with the same `taskPhase` and `taskPhaseVersion`. FlyteAdmin does not take `taskPhaseVersion` into accord for new `taskPhase` values.

I'm probably missing something here, a repro would help.

pvditt: @hamersaw I noticed this when working on mapping over launch plans but don't think I actually ran into the bug. However this should repro an issue.

```
@task
def hello(num: int) -> int:
    if num > 9:
        time.sleep(10)
        raise Exception("This is a test exception")
    return num


@workflow
def map_workflow():
    map_task(hello, min_success_ratio=0.5)(num=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
```

A subnode task phase will be stuck in "Running" even though it has terminated/failed.

Emitting the event always fails due to the task phase not getting bumped but the task phase version getting reset.

pvditt: Wait, this doesn't occur when the staggered subtask succeeds. Looking back into this.

pvditt: Figured it out:

this doesn't bubble up for:

```
@task
def hello(num: int) -> int:
    if num > 9:
        time.sleep(10)
    return num


@workflow
def map_workflow():
    map_task(hello, min_success_ratio=0.5)(num=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
```

https://github.com/flyteorg/flyte/blob/fc1c92cce952f26242b60b59ef4e78d61f37e69c/flytepropeller/pkg/controller/nodes/array/handler.go#L374-L386 doesn't check for NodePhaseSucceeding so it counts as a running task so we don't bump the arrayNode phase here: https://github.com/flyteorg/flyte/blob/fc1c92cce952f26242b60b59ef4e78d61f37e69c/flytepropeller/pkg/controller/nodes/array/handler.go#L403-L409

Since we don't do that we bump task phase version instead of resetting to 0 to the event lands in admin. Meanwhile for NodePhaseFailing, we don't increment running so we update arrayNodeState.Phase  which then leads for the existing implementation to reset the task phase version.

hamersaw (approved):

PR #5454, opened 2024-06-07
[flyteagent] Remove redundant code in Agent

Future-Outlier:
## Tracking issue

Comment in the description below.

## How was this patch tested?
single binary.

### Screenshots
image

### DIFF
--- a/flyteplugins/go/tasks/plugins/webapi/agent/client.go
+++ b/flyteplugins/go/tasks/plugins/webapi/agent/client.go
@@ -64,13 +64,11 @@ func getGrpcConnection(ctx context.Context, agent *Deployment) (*grpc.ClientConn
 		return nil, err
 	}
 
-	defer func() {
-		go func() {
-			<-ctx.Done()
-			if cerr := conn.Close(); cerr != nil {
-				grpclog.Infof("Failed to close conn to %s: %v", agent, cerr)
-			}
-		}()
+	go func() {
+		<-ctx.Done()
+		if cerr := conn.Close(); cerr != nil {
+			grpclog.Infof("Failed to close conn to %s: %v", agent, cerr)
+		}
 	}()
 
 	return conn, nil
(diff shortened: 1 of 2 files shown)


flyteplugins/go/tasks/plugins/webapi/agent/client.go
```diff
@@ -63,13 +63,8 @@ func getGrpcConnection(ctx context.Context, agent *Deployment) (*grpc.ClientConn
 	if err != nil {
 		return nil, err
 	}
+
 	defer func() {
-		if err != nil {
-			if cerr := conn.Close(); cerr != nil {
```
Future-Outlier, line 68: Here will never be triggered

flyteplugins/go/tasks/plugins/webapi/agent/plugin.go
```diff
@@ -175,11 +176,7 @@ func (p Plugin) ExecuteTaskSync(
...
 
 	if err := stream.CloseSend(); err != nil {
-		return nil, nil, err
-	}
-
-	if err != nil {
-		logger.Errorf(ctx, "Failed to write output with err %s", err.Error())
+		logger.Errorf(ctx, "Failed to close stream with err %s", err.Error())
 		return nil, nil, err
 	}
```
Future-Outlier, line 181: Here will never be triggered, too.

flyteplugins/go/tasks/plugins/webapi/agent/client.go
```diff
@@ -63,13 +63,8 @@ func getGrpcConnection(ctx context.Context, agent *Deployment) (*grpc.ClientConn
 	if err != nil {
 		return nil, err
 	}
+
 	defer func() {
```
pingsutw, line 67: we can remove defer as well

Future-Outlier: Yes you are right, nice catch!

pingsutw (approved):

PR #5458, opened 2024-06-07
Make BaseURL insensitive to trailing slashes for metadata endpoint redirect.

### LINKED ISSUE #5456: [BUG] Handler for .well-known/openid-configuration constructs redirect path incorrectly
### Describe the bug

Let's assume the following configuration:
```
configmap:
  adminServer:
    auth:
      userAuth:
        openId:
          baseUrl: https://something.onelogin.com/oidc/2
```

Then request to Flyte's `/.well-known/openid-configuration` returns HTTP Code 303 (See Other) with the following header:
```
Location: https://something.onelogin.com/oidc/.well-known/openid-configuration
```

I think this happens because in [this](https://github.com/flyteorg/flyte/blob/25c35961f7d39b75868c0cad312984d73aa245ae/flyteadmin/auth/handlers.go#L487C60-L487C76) line [ResolveReference](https://pkg.go.dev/net/url#URL.ResolveReference) is used which has the following behaviour for merging paths:

>   return a string consisting of the reference's path component
>   appended to <ins>all but the last segment of the base URI's path</ins> (i.e.,
>   excluding any characters after the right-most "/" in the base URI
>   path, or excluding the entire base URI path if it does not contain
>   any "/" characters).

### Expected behavior

The correct URL would be `https://something.onelogin.com/oidc/2/.well-known/openid-configuration`

_No response_

### Screenshots

_No response_

- [X] Yes

- [X] Yes

Dlougach:
## Tracking issue

Example: Closes #5456

Fixes resolution of OpenID metadata URL redirect.

Without the change tests would fail:
```
expected: "https://login.microsoftonline.com/abc/v2.0/.well-known/openid-configuration"
actual  : "https://login.microsoftonline.com/abc/.well-known/openid-configuration"
```

Append a trailing slash to `BaseURL` before running `ResolveReference`

## How was this patch tested?

Unit tests only.

### DIFF
--- a/flyteadmin/auth/handlers.go
+++ b/flyteadmin/auth/handlers.go
@@ -484,7 +484,7 @@ func QueryUserInfoUsingAccessToken(ctx context.Context, originalRequest *http.Re
 // See https://tools.ietf.org/html/rfc8414 for more information.
 func GetOIdCMetadataEndpointRedirectHandler(ctx context.Context, authCtx interfaces.AuthenticationContext) http.HandlerFunc {
 	return func(writer http.ResponseWriter, request *http.Request) {
-		metadataURL := authCtx.Options().UserAuth.OpenID.BaseURL.ResolveReference(authCtx.GetOIdCMetadataURL())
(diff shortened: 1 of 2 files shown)


eapolinario:
Thank you!

PR #5463, opened 2024-06-10
Fix: Make 'flytectl compile' consider launchplans used within workflows

fellhorn:

```py
from flytekit import LaunchPlan, task, workflow

@task
def my_task(num: int) -> int:
    return num + 1


@workflow
def inner_workflow(num: int) -> int:
    return my_task(num=num)


@workflow
def outer_workflow() -> int:
    return LaunchPlan.get_or_create(inner_workflow, "name_override", default_inputs={"num": 42})()
```

This workflow can be registered with `pyflyte register` and executed in a cluster. However, `pyflyte --pkgs <module> package` + `flytectl compile` gives the following error:

```shell
...
Compiling workflow: wfs.wf.outer_workflow
:( Error Compiling workflow: wfs.wf.outer_workflow
Error: Collected Errors: 1
        Error 0: Code: WorkflowReferenceNotFound, Node Id: start-node, Description: Referenced Workflow [resource_type:LAUNCH_PLAN  name:"name_override"] not found.
 ```

The error occurs since flytectl compile does not consider all launchplans but only the default launchplan for the currently compiled workflow.

To do so, one needs to compile workflows in the correct order, which I do by recursively calling a new `handleWorkflow` function. Please point me to other places if there is already code handling this.

As far as I have checked there are no recursion checks necessary as they would be caught during packaging?

## How was this patch tested?

Added a unit test based on the example above. Additionally we tried the `flytectl compile` command on a number of our internal workflows.

Note to reviewers: We couldn't find whether the source files for the `testdata/*.tgz` should be checked in as well - should they?

## Related PRs

Replaces https://github.com/flyteorg/flyte/pull/5437

### CODE REVIEW

**Comment by fellhorn** [comment]
The docs generation in CI seems to be broken, I saw other Actions fail with the same error at the moment.

**Comment by davidmirror-ops** [comment]
@fellhorn the docs CI test failure should be fixed by https://github.com/flyteorg/flyte/pull/5468 so if you rebase and merge it should do the trick

**Comment by fellhorn** [comment]
> @fellhorn the docs CI test failure should be fixed by #5468 so if you rebase and merge it should do the trick

Rebased, thanks for the info :+1:

File: flytectl/cmd/compile/compile.go [hunk 1]
```diff
@@ -105,6 +100,65 @@ func compileFromPackage(packagePath string) error {
...
+	compiledWorkflows map[string]*core.CompiledWorkflowClosure,
+	compiledLaunchPlanProviders []common.InterfaceProvider,
+	plans map[string]*admin.LaunchPlan,
+	workflows map[string]*admin.WorkflowSpec) ([]common.InterfaceProvider, error) {
+	reqs, _ := compiler.GetRequirements(workflow.Template, workflow.SubWorkflows)
+	wfName := workflow.Template.Id.Name
+
+	// Check if all the subworkflows referenced by launchplan are compiled
+	for i := range reqs.GetRequiredLaunchPlanIds() {
+		lpID := &reqs.GetRequiredLaunchPlanIds()[i]
```
fg91, line 115: Nit: Can we do sth like this?

```go
for i, lpID := range reqs.GetRequiredLaunchPlanIds() {
```

flytectl/cmd/compile/compile.go
```diff
@@ -105,6 +100,65 @@ func compileFromPackage(packagePath string) error {
...
+		lpWfName := plans[lpID.Name].Spec.WorkflowId.Name
+		missingWorkflow := workflows[lpWfName]
```
fg91, line 117: In what sense is it missing? Would `currentWorkflow` also be a fitting name in your opinion?

flytectl/cmd/compile/compile.go
fellhorn: There is a mutex on the launchplans & a linter proposed changing to the current form.

fg91: Ok 👌

fg91 (approved):
Thank you for fixing this and adding a test that would have caught the problem 🙏

Issue #5476, opened 2024-06-14
[Housekeeping] Distributed Tracing Should Support OTLP Exporters


Sovietaced:
### Describe the issue

Current Flyte only supports exporting distributed traces to a file or using the jaeger protocol. The jaeger protocol has been deprecated in favor of OTLP. 

### What if we do not do this?

Newer versions of https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main have removed support for the jaeger protocol entirely, which would break Flyte as is. 

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #5477, opened 2024-06-14
Add support for otlp trace exporters

Sovietaced:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/5476

Distributed traces can only be exported using Jaeger with the current code. The[ jaeger protocol has been deprecated in favor of the open telemetry protocol (OTLP)](https://opentelemetry.io/blog/2022/jaeger-native-otlp/). Additionally, if we're using OTEL client libraries Flyte should also just support their native protocol as well. 

We should also deprecate use of the Jaeger exporter and slate it for removal. 

This pull request adds two new trace exporter strategies: `otlpGrpc` (OTLP over gRPC) and `otlpHttp` (OTLP over HTTP). 

## How was this patch tested?

This patch was tested locally in our non production environment.


flytestdlib/otelutils/config.go
```diff
@@ -33,7 +35,7 @@ var (
 )
 
 type Config struct {
-	ExporterType ExporterType `json:"type" pflag:",Sets the type of exporter to configure [noop/file/jaeger]."`
+	ExporterType ExporterType `json:"type" pflag:",Sets the type of exporter to configure [noop/file/jaeger/otlpGrpc/otlpHttp]."`
 	FileConfig   FileConfig   `json:"file" pflag:",Configuration for exporting telemetry traces to a file"`
 	JaegerConfig JaegerConfig `json:"jaeger" pflag:",Configuration for exporting telemetry traces to a jaeger"`
```
Sovietaced, line 40: All the OTEL exporter configuration can come from environment variables so I didn't really see any value in 1) deviating from OTEL and 2) trying to copy all of the settings that can be configured via the OTEL environment variables.

EngHabu:
@hamersaw @andrewwdye can you plz review this PR?

flytestdlib/otelutils/config.go
andrewwdye: Good callout @Sovietaced. My hesitation is that it deviates with other trace configuration and how flyte components are configured generally. It adds ambiguity to a component's behavior and how to configure it. I'm inclined to keep consistent.

If it's helpful I have a private version of this to use OTEL + expose some sampling config. I could pull that in instead and save you some cycles. Wdyt?

Sovietaced: I can see the user experience being a little nicer for some folks if more configuration came through the YAML dictionaries in the helm chart but I'm personally not interested in trying to duplicate the OTEL standardized environment variables so you're probably better of pulling in changes from your fork.

andrewwdye: Sounds good. I have a draft up at https://github.com/flyteorg/flyte/pull/5504. Let me pull in a few more of the changes you proposed here as well

Sovietaced:
Closing in favor of: https://github.com/flyteorg/flyte/pull/5504

PR #5480, opened 2024-06-14
Change Flyte CR naming scheme to better support namespace_mapping

ddl-ebrown:
@ddl-rliu did most of the work on this one - making this an upstream PR as it resolved a real issue for us.

 - Typically Flyte is configured so that each project / domain has its
   own Kubernetes namespace.

   Certain environments may change this behavior by using the Flyteadmin
   namespace_mapping setting to put all executions in fewer (or a singular)
   Kubernetes namespace. This is problematic because it can lead to
   collisions in the naming of the CR that flyteadmin generates.

 - This patch fixes 2 important things to make this work properly inside
   of Flyte:

   * it adds a random element to the CR name in Flyte so that the CR is
     named by the execution + some unique value when created by
     flyteadmin

     Without this change, an execution Foo in project A will prevent an
     execution Foo in project B from launching, because the name of the
     CR thats generated in Kubernetes *assumes* that the namespace the
     CRs are put into is different for project A and project B

     When namespace_mapping is set to a singular value, that assumption
     is wrong

   * it makes sure that when flytepropeller cleans up the CR resource
     that it uses Kubernetes labels to find the correct CR -- so instead
     of assuming that it can use the execution name, it instead uses the
     project, domain and execution labels

## How was this patch tested?

This is deployed in a live Flyte setup where we have automated tests. We observed that the CR names were correctly unique after this and the initial collision no longer occurred.

### DIFF
--- a/flytepropeller/pkg/compiler/transformers/k8s/workflow.go
+++ b/flytepropeller/pkg/compiler/transformers/k8s/workflow.go
@@ -2,17 +2,21 @@
 package k8s
 
 import (
+	"context"
 	"fmt"
 	"hash/fnv"
 	"strings"
 
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
 
 	"github.com/flyteorg/flyte/flyteidl/gen/pb-go/flyteidl/core"
 	"github.com/flyteorg/flyte/flytepropeller/pkg/apis/flyteworkflow/v1alpha1"
 	"github.com/flyteorg/flyte/flytepropeller/pkg/compiler/common"
 	"github.com/flyteorg/flyte/flytepropeller/pkg/compiler/errors"
+	"github.com/flyteorg/flyte/flytepropeller/pkg/controller/config"
 	"github.com/flyteorg/flyte/flytepropeller/pkg/utils"
+	"github.com/flyteorg/flyte/flytestdlib/logger"
 )
 
 const (
@@ -30,6 +34,11 @@ const (
 	ShardKeyLabel = "shard-key"
 	// The fully qualified FlyteWorkflow name
 	WorkflowNameLabel = "workflow-name"
+
+	// Length of hash to use as a suffix on the workflow CR name. Used when config.UseWorkflowCRNameSuffix is true.
+	// The workflow CR name should be at or under 63 characters long, here it is 52 + 1 + 10 = 63
+	workflowCRNameHashLength = 10
+	workflowCRNameSuffixFmt  = "%.52s-%s"
 )
 
 func requiresInputs(w *core.WorkflowTemplate) bool {
@@ -159,6 +168,20 @@ func generateName(wfID *core.Identifier, execID *core.WorkflowExecutionIdentifie
 	}
 }
 
+func hashIdentifier(identifier core.Identifier) uint64 {
+	h := fnv.New64()
+	_, err := h.Write([]byte(fmt.Sprintf("%s:%s:%s",
+		identifier.Project, identifier.Domain, identifier.Name)))
+	if err != nil {
+		// This shouldn't occur.
+		logger.Errorf(context.Background(),
+			"failed to hash execution identifier: %+v with err: %v", identifier, err)
+		return 0
+	}
+	logger.Debugf(context.Background(), "Returning hash for [%+v]: %d", identifier, h.Sum64())
+	return h.Sum64()
+}
+
 // BuildFlyteWorkflow builds v1alpha1.FlyteWorkflow resource. Returned error, if not nil, is of type errors.CompilerErrors.
 func BuildFlyteWorkflow(wfClosure *core.CompiledWorkflowClosure, inputs *core.LiteralMap,
 	executionID *core.WorkflowExecutionIdentifier, namespace string) (*v1alpha1.FlyteWorkflow, error) {
@@ -231,7 +254,19 @@ func BuildFlyteWorkflow(wfClosure *core.CompiledWorkflowClosure, inputs *core.Li
 		errs.Collect(errors.NewWorkflowBuildError(err))
 	}
 
-	obj.ObjectMeta.Name = name
+	if config.GetConfig().UseWorkflowCRNameSuffix {
+		// Seed the randomness before generating the name with random suffix
+		hashedIdentifier := hashIdentifier(core.Identifier{
+			Project: project,
+			Domain:  domain,
+			Name:    name,
+		})
+		rand.Seed(int64(hashedIdentifier))
+		obj.ObjectMeta.Name = fmt.Sprintf(workflowCRNameSuffixFmt, name, rand.String(workflowCRNameHashLength))
+	} else {
+		obj.ObjectMeta.Name = name
+	}
+
 	obj.ObjectMeta.GenerateName = generatedName
 	obj.ObjectMeta.Labels[ExecutionIDLabel] = label
 	obj.ObjectMeta.Labels[ProjectLabel] = project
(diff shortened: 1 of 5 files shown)


flytepropeller/pkg/compiler/transformers/k8s/workflow.go
```diff
@@ -159,6 +161,17 @@ func generateName(wfID *core.Identifier, execID *core.WorkflowExecutionIdentifie
 	}
 }
 
+const ExecutionIDSuffixLength = 21
+
+/* #nosec */
```
hamersaw, line 166: is there a reason not to use the crypto lib here rather than rand?

flytepropeller/pkg/compiler/transformers/k8s/workflow.go
```diff
@@ -159,6 +161,17 @@ func generateName(wfID *core.Identifier, execID *core.WorkflowExecutionIdentifie
 	}
 }
 
+const ExecutionIDSuffixLength = 21
```
hamersaw, line 164: Can we make this a configurable value and if set to `0` (default?) then have it disabled (so no random characters are appended). My concern is that if anything relies on the CR name, this will break it.

hamersaw (commented):
.

flytepropeller/pkg/compiler/transformers/k8s/workflow.go
hamersaw: Yeah, I understand the config bloat all to well :). As you suggest, my main concern is breaking backwards compatibility here. I know there are Flyte users that rely on the FlyteWorkflow CR to be named identical to the execution ID, which this would break. For me to be comfortable merging this, I feel it should be defaulted to the current behavior. cc @eapolinario thoughts?

kumare3:
I am not in favor of this, as randomness will lead to leaky workflows and duplicates. We should use the project id itself or generate a consistent hash to increase inter project execution entropy

EngHabu:
@ddl-ebrown I agree with not introducing randomization... specially that the name already starts with a random string :-)

Instead, I would update [this call](https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/manager/impl/util/shared.go#L28) to use something like `project-domain-rand(10)` and hash that and that becomes the execution name...

I would also make the length of the execution name configurable in flyteadmin. so in your deployment you can make it longer and give you better entropy...

flytepropeller/pkg/compiler/test/compiler_test.go
```diff
@@ -464,10 +460,6 @@ func runCompileTest(t *testing.T, dirName string) {
 				if !assert.NoError(t, err) {
 					t.FailNow()
 				}
-				// make sure real CR has randomized suffix
-				assert.Regexp(t, regexp.MustCompile("name-[bcdfghjklmnpqrstvwxz2456789]{20}"), flyteWf.Name)
-				// then reset for the sake of serialization comparison
-				flyteWf.Name = "name"
```
ddl-rliu, line 463: The cr-name-scheme-change is now non-default, and [tests are added here](flytepropeller/pkg/compiler/transformers/k8s/workflow_test.go), so removing these checks.

flytepropeller/pkg/compiler/transformers/k8s/workflow.go
```diff
@@ -244,7 +249,19 @@ func BuildFlyteWorkflow(wfClosure *core.CompiledWorkflowClosure, inputs *core.Li
 		errs.Collect(errors.NewWorkflowBuildError(err))
 	}
 
-	obj.ObjectMeta.Name = GetExecutionName(name, time.Now().UnixNano())
+	hashedIdentifier := hashIdentifier(core.Identifier{
+		Project: project,
+		Domain:  domain,
+		Name:    name,
+	})
+	rand.Seed(int64(hashedIdentifier))
```
ddl-rliu, line 257: By hashing exactly these fields `project, domain, name` we prevent duplicate CRs from being created for the same execution (they will hash to the same CR name and k8s will prevent those duplicate CRs).

flytepropeller/pkg/compiler/transformers/k8s/workflow.go
```diff
@@ -244,7 +249,19 @@ func BuildFlyteWorkflow(wfClosure *core.CompiledWorkflowClosure, inputs *core.Li
...
+
+	if workflowCRNameHashLength := config.GetConfig().WorkflowCRNameHashLength; workflowCRNameHashLength > 0 {
+		obj.ObjectMeta.Name = rand.String(workflowCRNameHashLength)
+	} else {
+		obj.ObjectMeta.Name = name
+	}
```
ddl-rliu, line 263: By default `WorkflowCRNameHashLength` is 0 so the hash CR naming scheme will only be used when the deployment is specially configured to do so.

flytepropeller/pkg/compiler/transformers/k8s/workflow.go
ddl-rliu: Thanks for the feedback! I've updated the PR to address this and explain a bit more here: https://github.com/flyteorg/flyte/pull/5480#discussion_r1683582220

flytepropeller/pkg/compiler/transformers/k8s/workflow.go
```diff
@@ -231,7 +252,25 @@ func BuildFlyteWorkflow(wfClosure *core.CompiledWorkflowClosure, inputs *core.Li
...
+			Name:    name,
+		})
+		rand.Seed(int64(hashedIdentifier))
+
+		base := name + "-"
+		maxNameLength := allowedExecutionNameLength - workflowCRNameHashLength
+		if len(base) > maxNameLength {
+			base = base[:maxNameLength]
+		}
+		obj.ObjectMeta.Name = fmt.Sprintf("%s%s", base, rand.String(workflowCRNameHashLength))
```
ddl-rliu, line 269: Based off of https://github.com/clarklee92/kubernetes/blob/519f1ba07379108c301c0474d5eb61b2012bd427/staging/src/k8s.io/apiserver/pkg/storage/names/generate.go#L57-L63

flytepropeller/pkg/compiler/transformers/k8s/workflow.go
```diff
@@ -231,7 +249,19 @@ func BuildFlyteWorkflow(wfClosure *core.CompiledWorkflowClosure, inputs *core.Li
...
-	obj.ObjectMeta.Name = name
+	hashedIdentifier := hashIdentifier(core.Identifier{
+		Project: project,
+		Domain:  domain,
+		Name:    name,
+	})
+	rand.Seed(int64(hashedIdentifier))
+
+	if workflowCRNameHashLength := config.GetConfig().WorkflowCRNameHashLength; workflowCRNameHashLength > 0 {
+		obj.ObjectMeta.Name = rand.String(workflowCRNameHashLength)
```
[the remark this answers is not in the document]
ddl-rliu: Updated https://github.com/flyteorg/flyte/pull/5480#discussion_r1688491088

Now, for a setting like `WorkflowCRNameHashLength: 32`, the CR name will look like `mywfname-f5ptmmd66pktp59nphfdjkqr9szp8fv5` or `mywfname-wnw9zlg22z5wmdjggdcrnzdrpw59m4bc`. Note that the suffix is deterministic, and uses a hash of the name+project+domain.

I opted for this approach rather than `mywfname-project`, this is because of the character limit of the CR name and how this affects executions that have a long execution name and a long project name, where improperly handled truncation will lead to undesired behavior such as duplicates.

flytepropeller/pkg/compiler/transformers/k8s/workflow.go
ddl-rliu: A fixed suffix length of 10 sounds good to me. Agreed about the sufficient amount of entropy on relatively high values for the suffix length. https://github.com/flyteorg/flyte/pull/5480/commits/15b57cb739f74cc00e133db79f19301b87d5f726

eapolinario:

PR #5490, opened 2024-06-20
Flyte core webhook pod settings should be separate

ddl-ebrown:

 - Previously, the webhook was sharing some pod level settings in the core chart with flytepropeller like:

   * podAnnotations
   * podEnv
   * podLabels
   * nodeSelector

   Since the webhook runs a separate pod, it should have separate
   settings. This allows for independently adding annotations, like for configuring Prometheus. propeller has a /metrics endpoint, while webhook does not.

NOTE: this isn't backwards compatible, but I also think it's a fairly minor issue that can be covered in docs. I suspect there aren't currently many users here and some of these were not introduced not long ago (as part of https://github.com/flyteorg/flyte/pull/4756)

Moves webhook level settings to their own section

### DIFF
--- a/charts/flyte-core/values.yaml
+++ b/charts/flyte-core/values.yaml
@@ -477,6 +477,14 @@ webhook:
     annotations:
       projectcontour.io/upstream-protocol.h2c: grpc
     type: ClusterIP
+  # -- Annotations for webhook pods
+  podAnnotations: {}
+  # -- Additional webhook container environment variables
+  podEnv: {}
+  # -- Labels for webhook pods
+  podLabels: {}
+  # -- nodeSelector for webhook deployment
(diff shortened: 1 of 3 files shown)


charts/flyte-core/values.yaml
```diff
@@ -477,6 +477,14 @@ webhook:
     annotations:
       projectcontour.io/upstream-protocol.h2c: grpc
     type: ClusterIP
+  # -- Annotations for webhook pods
+  podAnnotations: {}
+  # -- Additional webhook container environment variables
+  podEnv: {}
```
pingsutw, line 483: I think it's fine, not many people using these fields

pingsutw (approved):

PR #5497, opened 2024-06-21
Don't log auth tokens in debug mode

Sovietaced:
This pull request updates the `AuthenticationCustomMetadataInterceptor` so that it won't log authorization header values in debug mode. Access tokens fall under [OWASP's guidance](https://cheatsheetseries.owasp.org/cheatsheets/Logging_Cheat_Sheet.html#data-to-exclude) for what should never be logged unless it is legally sanctioned.


pingsutw (commented):
cc @pmahindrakar-oss mind taking a look at this as well

pmahindrakar-oss (approved):

PR #5504, opened 2024-06-22
Add OTLP and sampling to otelutils

andrewwdye:

This changes adds support for OLTP and sampling to the otelutils tracer provider abstraction.

Adds support for OLTP. This is the recommended replacement for the deprecated jaeger exporter.
> "go.opentelemetry.io/otel/exporters/jaeger" is deprecated: This module is no longer supported. OpenTelemetry dropped support for Jaeger exporter in July 2023. Jaeger officially accepts and recommends using OTLP

OLTP supports grpc and http, which are added as separate exporter types and configs.

Adds initial [sampling](https://opentelemetry.io/docs/languages/go/sampling/) support to the top level open telemetry config. Defaults to parent sampler `always`, but also adds a config for [TraceIdRatioBased](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#TraceIDRatioBased). See [these docs](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#ParentBased) for behavior of parent sampler.

## How was this patch tested?

Ran local sandbox with local jaeger all-in-one and verified
- [x] otlpgrpc with parent sampler always 
- [x] otlpgrpc with parent sampler traceid (along with other defaults)
- [x] otlphttp with parent sampler traceid (along with other defaults)

Flyte config
```
❯ docker run --rm -e COLLECTOR_OTLP_ENABLED=true -p 16686:16686 -p 4317:4317 -p 4318:4318 jaegertracing/all-in-one:1.52
```

Jaeger
```
otel:
  type: otlpgrpc
  sampler:
    parentSampler: traceid
```


flytestdlib/otelutils/factory.go
```diff
@@ -30,14 +33,15 @@ const (
 )
 
 var tracerProviders = make(map[string]*trace.TracerProvider)
-var noopTracerProvider = rawtrace.NewNoopTracerProvider()
+var noopTracerProvider = noop.NewTracerProvider()
 
 func RegisterTracerProvider(serviceName string, config *Config) error {
```
andrewwdye, line 38: @Sovietaced I ended up not taking a `context` param here to avoid a breaking change

Sovietaced: Maybe there should be another function that takes in a context. Otherwise you won't be able to gracefully shut down the OTLP exporters and the server may shut down with buffered data points.

andrewwdye: Added. Thanks for the suggestion.

Sovietaced (approved):

hamersaw (approved):

PR #5505, opened 2024-06-24
[BUG] array node sub node control flow inheritance

pvditt:

Array node increments workflow parallelism from the parent node. Right now, the same control flow that tracks parallelism for the workflow is passed down to subnodes allowing subnodes to bump the workflow parallelism which could lead to a double counting of nodes getting evaluated in a workflow.

If/when ArrayNode supports subworkflows or dynamics then we can pass back in the parent control flow.

Init a new control flow when creating the exec context for array node subnodes

## How was this patch tested?

Ran a few workflows locally


pvditt:
@hamersaw This would mess with the node + task count. I'll look back into this.

eapolinario:

PR #5508, opened 2024-06-24
Helm chart updates related to Prometheus, Webhook HPA, and Flyteconsole probes

mhotan:

Various Helm chart updates deemed appropriate to upstream from Union internal usage.

* Add flyteconsole livenessProbe and readinesProbe to mitigate downtime during releases or updates. Default to no probes to preserve backward-compatible behavior.
* Add flyte-console prometheus port and service monitor 
* Add flyte-pod-webhook HPA
* Add Prometheus annotations to propeller, webhook, and syncresources

## How was this patch tested? 

All three patches (commits) were tested individually within Union internal flyte deployment


pingsutw (approved):
LGTM, thank you!

PR #5512, opened 2024-06-26
add execution mode to ArrayNode proto

pvditt:

Support selecting toggling ArrayNode execution/storage and retain backwards compatibility.

Add execution_version to ArrayNode proto


flyteidl/protos/flyteidl/core/workflow.proto
```diff
@@ -134,6 +134,11 @@ message ArrayNode {
         // to determine when an ArrayNode can be marked successful.
         float min_success_ratio = 4;
     }
+
+    oneof execution_version_option {
```
hamersaw, line 138: Is there a reason to make this `oneof`? non-existence defaults to nil, but we could just as easily use the default of 0 for not being set I think?

pvditt: Complete brain fart on my end. Reverting the one_of change.

pvditt: I think the problem I noticed was if flyteidl isn't bumped on flytekit and then you pass in execution_version on the flytekit end. Answer would be to just ensure flyteidl version is bumped as part of that flyetkit change. Don't know why I did the one_of :face_palm:

flyteidl/protos/flyteidl/core/workflow.proto
```diff
@@ -134,6 +134,10 @@ message ArrayNode {
         // to determine when an ArrayNode can be marked successful.
         float min_success_ratio = 4;
     }
+
+    // execution_version determines the execution path for ArrayNode. The default, 0, will
+    // utilize the minimized storage implementation. 1 will utilize full subNode status storage.
+    uint32 execution_version = 5;
```
hamersaw, line 140: Unsure if `execution_mode` sounds better? Then it should probably be an enum.

hamersaw (dismissed):

hamersaw (approved):

PR #5521, opened 2024-06-26
Use jsonpb AllowUnknownFields everywhere

andrewwdye:

By default `jsonpb.Unmarshal()`, `jsonpb.UnmarshalString`, and `jsonpb.Unmarshaler{}` sets `AllowUnknownFields` to `false`, which means that unmarshaling a proto message with a newly added field will fail [here](https://github.com/golang/protobuf/blob/75de7c059e36b64f01d0dd234ff2fff404ec3374/jsonpb/decode.go#L381-L383).

This change updates all use of `jsonpb.Unmarshal*` to use the `flytestlib/utils.Unmarshal*` abstractions, which set `AllowUnknownFields` to `true`. 

There is some risk that this break existing assumptions about new field handling, but since protobuf silently drops fields, it seems reasonable to default to this behavior across the flyte codebase.

## How was this patch tested?

Unittests


squiishyy (dismissed):
lgtm, thank you for making the change!

pingsutw (approved):
lgtm, thank you!

eapolinario (approved):

Issue #5344, opened 2024-05-09
[Core feature] Rust Flytekit Remote Client


pingsutw:
### Motivation: Why do you think this is important?

The flytekit remote client relies on gRPC for communication with the control plane (FlyteAdmin). However, our experience has revealed numerous issues within Python gRPC, leading to test failures. Additionally, users frequently encounter challenges when working with specific versions of gRPC.

In order to eliminate the gRPC Python dependency from flytekit, our goal is to develop a Rust crate to serve as a replacement for the flytekit remote client. To achieve this, we plan to utilize Pyo3 to generate a Python binding, allowing flytekit to seamlessly integrate and utilize the new implementation.

### Goal: What should the final outcome look like, ideally?

https://github.com/flyteorg/flytekit/pull/2377 is a POC of a rust remote client that allows you to get/create tasks without Python grpc

Add a flag (enable_rust) to the FlyteRemote class. When this flag is true, FlyteRemote will use the rust implementation.

```python
remote_rs = FlyteRemote(Config.auto(), default_project=PROJECT, default_domain=DOMAIN, enable_rust=True)
flyte_task = remote_rs.register_task(
    entity=my_test_task,
    serialization_settings=SerializationSettings(
        image_config=ImageConfig.auto(img_name="flyte-cr.io/image-name:tag")
    ),
    version=VERSION_ID,
)
flyte_task = remote_rs.fetch_task(name=TASK_NAME, version=VERSION_ID)
```

Follow up Issues
- [ ] Add a test for the rust remote client
- [ ] Add a GA workflow to run rust remote client without installing Python grpc
- [ ] Add a GA workflow to make sure the output of the Rust remote and Python remote are the same
- [ ] Pass [credentials](https://github.com/flyteorg/flytekit/blob/86f566157833514e73440a2879356c59ee0705d1/flytekit/clients/auth/auth_client.py#L342-L345) from python to rust
- [ ] OAuth flow in the rust implementation

### Describe alternatives you've considered 

Use existing flytekit remote client

###  Propose: Link/Inline OR Additional context 

- https://github.com/flyteorg/flytekit/pull/2034
- https://github.com/flyteorg/flytekit/pull/913
- https://github.com/grpc/grpc/issues/36265

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

austin362667:
[...]

PR #5525, opened 2024-06-27
[WIP] Make Rust flyteidl PyO3-ready

austin362667:
## Tracking issue

https://github.com/flyteorg/flyte/issues/5344

This pull request focuses on modifying the Rust flyteidl library to pave the way for the introduction of `flyteidl-rust` as a Python binding. The changes address issues raised in [issues](https://github.com/flyteorg/flyte/issues/5344) and serve as a foundation for the [PR](https://github.com/flyteorg/flytekit/pull/2561).

### Key Modifications:

1. **Custom Code Generator:**
    
    Added a custom `tonic_builder` gRPC code generator. The code for this generator can be found in [src/gen.rs](https://github.com/flyteorg/flyte/blob/450ed8cd04ced7b075c75fa5ec1d6da830c87077/flyteidl/src/gen.rs). Will try to refactor into `buf` generator in follow-up PRs. 

2. **Compile Well-Known Types:**

    **Background:**

    1. Utilize the [`#[pyclass]` attribute macro](https://docs.rs/pyo3/latest/pyo3/attr.pyclass.html) on our Rust structure. This macro [implements the `PyClass` trait](https://docs.rs/pyo3/latest/pyo3/pyclass/trait.PyClass.html) for the underlying structure "under the hood".
    2. Implementing the `PyClass` trait for a structure from an external crate (like `prost_types`) can lead to orphan rule violations in our local crate. The Rust compiler will not permit implementation of traits for types outside of the crate where the type is defined. (This is known as the "orphan rule".)

    **Avoiding Orphan Rules:**
    
    In some protobuf builders, like `tonic_builder`, there's an [option to enable "compile well-known types"](https://github.com/flyteorg/flyte/pull/5525/files#diff-0a69de87d52b5e95a5adafe90be10a10053bf2c30d079e17e8b131b104d3ce84R35). This forces the builder to compile well-known types (WKTs) directly using Rust primitive types instead of relying on external structures compiled from the `prost_types` crate.

    **Alternatives Considered:**
    
    One alternative is to create local wrappers around each external structure, as [suggested by the PyO3 community](https://github.com/PyO3/pyo3/issues/287#issuecomment-800776622). However, this solution becomes impractical when dealing with nested member fields within these structures. Because we still have to add `PyClass` trait all the way down to the nesting hierarchy. We'll elaborate more on gRPC client cases.

3. **implement PyClass Traits for `Box<T>`:**

    `T` already has PyClass Traits. `Box<T>` needs but does not have PyClass Traits.
    For example, in `flyteidl.core.rs`, `Box<T>`: `LiteralType`, `Literal`, `Union`, `Scalar`, `ConjunctionExpression`, `BooleanExpression`, `IfBlock`, `BranchNode`, `ArrayNode`, a total of 9 types will be on the heap.

    **What situations will be converted to boxes?**

    - Recursive types, such as `ArrayNode`, can hold references to other Nodes and are converted to `ArrayNode {node: Option<Box<Node>>, ...}`.

    - This poses a challenge for PyO3, as it requires a PyClass to hold a reference to another PyClass. This is a Rust Non-lifetime-free issue.

    - While this [PR](https://github.com/PyO3/pyo3/pull/3014) supports `Cell<T>`, but didn't support `Box<T>`, it also highlights the difficulties of supporting `Box<T>`.
        
    - The PyO3 roadmap on addressing these issue seems like this [way](https://github.com/PyO3/pyo3/issues/1089).

    **Solution:**

    Introduce conversion implementations and lifetime markers for the 9 types that use Box<T> to prevent refs from outliving. For example, for Box<Node>, [implement these two conversion traits](https://pyo3.rs/v0.22.0/class#:~:text=To%20use%20these,implement%20FromPyObject) instead of directly adding PyClass traits to Box<Node>:
    ```rust
    impl pyo3::conversion::IntoPy<pyo3::PyObject> for Box<Node> {
        fn into_py(self, py: pyo3::marker::Python<'_>) -> pyo3::PyObject {
            self.as_ref().clone().into_py(py)
        }
    }
    impl<'py> pyo3::conversion::FromPyObject<'py> for Box<Node> {
        fn extract_bound(obj: & pyo3::Bound<'py, pyo3::types:: PyAny>) ->  pyo3::PyResult<Self> {
            Ok(Box::new(Node::extract_bound(obj).unwrap()))
        }
    }
    ```

    Prevents references from outliving their intended lifetime, ensuring memory safety and preventing potential memory leaks.
    

4. **IntoRequest Trait:**
    For cases like following example,
    ```rust
    let request = tonic::Request::new(HelloRequest {
        name: "Tonic".into(),
    });
    let response = client.say_hello(request).await?;
    ```
    Based on [Types implementing this trait can be used as arguments to client RPC](https://github.com/hyperium/tonic/blob/c7836521dd417434d625bd653fcf00fb7f7ae25e/tonic/src/request.rs#L28)
    We can directly pass them to gRPC services, like:
    ```rust
    client.get_feature(Request::new(Point {})); // can work.
    client.get_feature(Point {}); // can work, too.
    ```

5. **Tuple Variants in Complex Enum:** [PyO3 supports tuple enum variants](https://github.com/PyO3/pyo3/pull/4072/files#diff-761eec64821954515b6076c8ab0be3ec83b0738eecb9dbe335050697f0c1de19R88-R92) after version v0.22.0

6.  **Python submodules:** Use latest feature [Declarative modules](https://pyo3.rs/v0.22.0/module#declarative-modules) introduced in version v0.22.0.

---

1. Optional: **Concretize Generic Types**

    **Challenges:**

    1. Each instantiation, for example: `GenericType<i32>`, `GenericType<f64>`, would have to have a different Python type object. This is an infinite set, and code would need to be generated for each instantiation. [It isn't trivia for PyO3 to support.](https://pyo3.rs/v0.22.0/class#no-generic-parameters) 
    2. In our case, for any gRPC client service stubs may accepts arbitrary Channel, ChannelInterceptor, etc. We must specify the underlying concrete type.

    **The Macro Solution**

    1. Define Concrete Structures: We can create new structures (AdminStub in the example) with predefined concrete types for each gRPC stub. This allows specifying the exact types needed for PyO3 interaction.

    2. Using Macros for Automation: The following example leverages a Rust macro named `concrete_generic_structure!` to automate the creation of these concrete structures. It takes arguments for the structure name, the generic type it wraps, and the concrete type used. Additionally, it allows defining methods associated with the stub, specifying the request and response types.
    
    - The provided code demonstrates how the macro is used to create an `AdminStub` structure. This structure wraps the `AdminServiceClient` generic type, and specifies `tonic::transport::Channel` as the concrete type for the channel. Additionally, it defines two methods (`create_task` and `get_task`) with their specific request and response types. 
     
        ```rust
      macro_rules! concrete_generic_structure {
      ($name:ident, $generic:ident, $type:ty, $( ($method_name:ident, $request_type:ty, $response_type:ty) ),* ) => {
          #[pyo3::pyclass]
          #[derive(Debug)]
          pub struct $name {
              pub inner: $generic<$type>,
              pub rt:  Runtime,
          }
      }
  
      concrete_generic_structure!(
          AdminStub,
          AdminServiceClient,
          tonic::transport::Channel,
          (
              create_task,
              flyteidl::admin::TaskCreateRequest,
              flyteidl::admin::TaskCreateResponse
          ),
          (
              get_task,
              flyteidl::admin::ObjectGetRequest,
              flyteidl::admin::Task
          )
      )
        ```

    **Trade-offs:**

    - Predefined Stub Support: This approach requires defining the stubs you want to support beforehand (e.g., defining Interceptor for channels).

## Related PRs

https://github.com/flyteorg/flytekit/pull/2536


flyteidl/src/lib.rs
```diff
@@ -0,0 +1,802 @@
...
+    }
+
+    #[pymodule]
+    pub mod service {
+        #[pymodule_export]
+        use crate::flyteidl::service::{CreateUploadLocationRequest, CreateUploadLocationResponse};
+    }
+
+    #[pyclass(subclass, name = "RawSynchronousFlyteClient")]
+    pub struct RawSynchronousFlyteClient {
```
austin362667, line 425: The entry point that should wraps all the gRPC service stubs.

flyteidl/scripts/add_pyo3_subclass.sh
```diff
@@ -0,0 +1,12 @@
+sed '/pub struct /i\
+#[pyo3(subclass)]\
```
austin362667, line 2: Why this hacky workaround?
In prost builder configuration, protobuf compiler we used in `gen.rs` can't specify PyO3's `subclass` for **only** Rust `struct`. Because , `enum` is a subset of `type`, it's non-trivial to distinguish between these while adding attributes.

flyteidl/src/gen.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    tonic_build::configure()
+        .out_dir("gen/pb_rust")
+        .build_server(false)
+        .build_client(true)
+        // `type` already includes enum type
+        .type_attribute(".", "#[pyo3::pyclass(get_all, set_all)]")
+        // `enum` cannot be `subclass`
+        // .enum_attribute(".", "#[pyclass(get_all, set_all)]")
```
austin362667, line 36: Skip adding `subclass` for struct, please refer to `scripts/add_pyo3_subclass.sh`

flyteidl/src/lib.rs
```diff
@@ -0,0 +1,802 @@
...
+                ),
+            })
+            .into_inner();
+            Ok(res)
+        }
+    }
+}
+
+// #[macro_export]
+// macro_rules! concrete_generic_structure {
```
austin362667, line 683: At present, we don't need to specify the concrete type for generics parameterization, since we output the wrapped `FlyteRemote` as binding class instead of each gRPC stubs. So the macro `oncrete_generic_structure` is commented out.

flyteidl/scripts/add_pyo3_subclass.sh
austin362667: So we add this auxiliary scripts

flyteidl/pyo3_macro/src/lib.rs
```diff
@@ -87,19 +88,23 @@ pub fn with_new(input: TokenStream) -> TokenStream {
...
-                        // TODO:
-                        // pub fn SerializeToString(proto_obj: #name) -> Result<Vec<u8>, crate::flyteidl::MessageEncodeError> {
-                        //     let mut buf = vec![];
-                        //     proto_obj.encode(&mut buf)?;
-                        //     Ok(buf)
-                        // }
+                        pub fn SerializeToString(&self) -> Result<Vec<u8>, crate::_flyteidl_rust::MessageEncodeError>
+                        {
+                            // Bring `prost::Message` trait to scope here. Put it in outer impl block will leads to duplicated imports.
+                            use prost::Message;
```
austin362667, line 101: Check here: https://github.com/tokio-rs/prost/discussions/525

flyteidl/src/lib.rs
```diff
@@ -0,0 +1,1135 @@
...
+        types::{PyDict, PyTuple},
+    };
+    use std::fmt;
+    use tonic::Status;
+
+    convert_foreign_prost_error!();
+
+    #[pyclass]
+    #[pyo3(name = "FlyteUserException", subclass, extends = pyo3::exceptions::PyException)] // or PyBaseException
+    pub struct PyFlyteUserException {}
```
austin362667, line 269: XREF: https://github.com/PyO3/pyo3/discussions/4165

eapolinario:

PR #5535, opened 2024-07-03
Introduced SMTP notification

robert-ulbrich-mercedes-benz:

Currently E-Mail notifications can only be sent using either SendGrid or SES. In many enterprises none of the both services is used, but there is a custom SMTP server available in the company.

A new implementation of the `Emailer` intereface was added. It allows for notifications to be sent via a plain SMTP server. That SMTP server can be configured including authorization.

## How was this patch tested?

There are automated unit tests. Also the changes were tested against SMTP interface of SES running a one-binary dev environment.  The program was tested solely for our own use cases, which might differ from yours.

### Setup process

* Setup dev environment as documented
* Start local kubernetes cluster `flytectl demo start --dev `
* Modify `flyte-single-binary-local.yaml` as found below 
```
notifications:
  type: sandbox
  emailer:
    emailServerConfig:
      serviceName: smtp
      smtpServer: email-smtp.eu-central-1.amazonaws.com
      smtpPort: 25
      smtpUsername: <SES-Username>
      smtpPasswordSecretName: smtp_ses_password
    subject: "Notice: Execution \"{{ name }}\" has {{ phase }} in \"{{ domain }}\"."
    sender: "robert.ulbrich.89@gmail.com"
    body: >
      Execution \"{{ name }}\" has {{ phase }} in \"{{ domain }}\". View details at
      <a href=\http://example.com/projects/{{ project }}/domains/{{ domain }}/executions/{{ name }}>workflow execution</a>
```
* Create a file `smtp_ses_password` in `/etc/secrets` containing the SES password. 
* Compile the single binary `make compile`
* Start the single-binary `POD_NAMESPACE=flyte ./flyte start --config flyte-single-binary-local.yaml`
* Create a launchplan 
```
from flytekit import LaunchPlan, WorkflowExecutionPhase
from flytekit.core.notification import Email
from hello import hello_world_wf

hello_world_lp = LaunchPlan.create(
    name="hello_wf_new",
    workflow=hello_world_wf,
    notifications=[
        Email(
            phases=[
                WorkflowExecutionPhase.SUCCEEDED,
                WorkflowExecutionPhase.FAILED,
                WorkflowExecutionPhase.TIMED_OUT,
            ],
            recipients_email=["<your-mail"],
        )
    ],
)
```
for workflow:

```

@task
def say_hello() -> str:
    return "Hello, World!"

@workflow
def hello_world_wf() -> str:
    res = say_hello()
    return res
```
* An email should be sent about the successful execution of the workflow

## Docs link

Robert Ulbrich<robert.ulbrich@mercedes-benz.com>,  Mercedes-Benz Tech Innovation GmbH

[https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md](https://github.com/mercedes-benz/daimler-foss/blob/master/PROVIDER_INFORMATION.md)


robert-ulbrich-mercedes-benz:
I am sorry about the bad unit test coverage for the `smtp_emailer.go` file. Without implementing a simple SMTP server and actually performing an integration test - I have no idea how we can boost the test coverage since I cannot simply mock any of the external dependencies.

flyteadmin/pkg/runtime/interfaces/application_configuration.go
```diff
@@ -491,8 +491,13 @@ type NotificationsProcessorConfig struct {
 type EmailServerConfig struct {
 	ServiceName string `json:"serviceName"`
 	// Only one of these should be set.
-	APIKeyEnvVar   string `json:"apiKeyEnvVar"`
-	APIKeyFilePath string `json:"apiKeyFilePath"`
+	APIKeyEnvVar           string `json:"apiKeyEnvVar"`
+	APIKeyFilePath         string `json:"apiKeyFilePath"`
+	SMTPServer             string `json:"smtpServer"`
```
wild-endeavor, line 496: would this be cleaner in a separate object?

flyteadmin/pkg/async/notifications/implementations/smtp_emailer.go
```diff
@@ -0,0 +1,138 @@
...
+type SMTPEmailer struct {
+	config        *runtimeInterfaces.NotificationsEmailerConfig
+	systemMetrics emailMetrics
+	tlsConf       *tls.Config
+	auth          *smtp.Auth
+}
+
+func (s *SMTPEmailer) SendEmail(ctx context.Context, email admin.EmailMessage) error {
+
+	newClient, err := smtp.Dial(s.config.EmailerConfig.SMTPServer + ":" + s.config.EmailerConfig.SMTPPort)
```
wild-endeavor, line 29: Just to confirm, this is the intended behavior?  If admin needs to send a thousand emails, it has to make a new client each time?

wild-endeavor (commented):
Just a couple questions.  Would you mind please also making sure that the new settings are reachable using the flyte-core Helm chart?  An example would be lovely.  If not could you add to the chart in this PR, or cut another issue so we don't lose track of it?

Thank you!

flyteadmin/pkg/async/notifications/implementations/smtp_emailer.go
robert-ulbrich-mercedes-benz: Yes, this is intended behavior. When using the built-in email client of Go, most examples you find work that way. Also with the integration into the Notifications processor this is the simplest approach. Establishing a connection does not seem to be a very complex undertaking when looking at the source code of the email client. 

I agree that keeping the connection is more efficient. But it also makes the code more complex, because the connection state needs to be maintained over an instance - so whenever an email is meant to be sent, it first needs to be checked if the connection is still alive. 

If you want me to, I can refactor it to only open a connection once to the SMTP server.

flyteadmin/pkg/runtime/interfaces/application_configuration.go
robert-ulbrich-mercedes-benz: When moving it out to a separate object, I would rather move out the `APIKeyEnvVar` and `APIKeyFilePath` properties into their own type. Since the name `EmailServerConfig` better matches the newly introduced properties in my eyes. Doing that would be a breaking change. Maybe you can make a proposal how you want the objects structured / named?

robert-ulbrich-mercedes-benz:
> Just a couple questions. Would you mind please also making sure that the new settings are reachable using the flyte-core Helm chart? An example would be lovely. If not could you add to the chart in this PR, or cut another issue so we don't lose track of it?
> 
> Thank you!

The settings can already be set with the current Helm chart. Actually the notification settings with Sendgrid and AWS SES also do not have a good support in the Helm charts yet. Well I guess I did not made it better either, but also did not want to fully update the Helm chart here. 

The settings can be put into; `workflow_notifications.config.notifications.emailer` Helm key for AWS, Sendgrid and also the SMTP server notification settings. 

if you want me to update the whole Helm file here, I can do it, but I guess we will then soon lose the scope of the merge request.

wild-endeavor:
apologies for the delay @robert-ulbrich-mercedes-benz - mind taking a look at the errors in ci here?
much thanks.

robert-ulbrich-mercedes-benz:
Hi @wild-endeavor,

this PR is becoming a never ending story: After I thought that I finally fixed all linting issues - while I could not run the linter locally , since it always broke down - I tried to run it in different docker container with different Golang versions - it never works. 

When running the linter in the CI, it complains about issues in files that I did not touch. 

I have no idea how I will ever be able to get a successful build. It is simply annoying. That way new committers are actively kept away :(

flyteadmin/pkg/async/notifications/implementations/smtp_emailer.go
robert-ulbrich-mercedes-benz: I just checked again the internal implementation of Golang's smtp client: It appears that the client implementation seems to be not multi-thread-safe. Every client instance keeps an internal state between function calls - so when reusing the same client while sending an email for writing another mail can lead to weird behavior. 

My assumption is that there is only a single email processor per Flyte Admin instance. That means that there should be no concurrent email processing on any Flyte Admin instance, which makes it safe to reuse an existing Flyste client.

robert-ulbrich-mercedes-benz:
> The emailer interface changed a few days ago (in https://github.com/flyteorg/flyte/blame/master/flyteadmin/pkg/async/notifications/interfaces/emailer.go#L12C4-L12C4), after the last time you merged master.

Hi @eapolinario,

thanks for the hint! Actually Github was playing a trick on me: 
image

Yesterday it showed me that I am synced with the `master` branch of the original repository. Only this morning it told me, that I am 6 commits behind - those commits were already merged days ago - no idea why those changes were only recognized this morning and not already yesterday :(. So hopefully things will work out now.

neverett (commented):
Docs changes look good to me, but I'll let others approve the rest of the changes.

wild-endeavor:
helm test should be getting patched soon.

wild-endeavor (approved):

PR #5546, opened 2024-07-09
[flyteadmin] Refactor panic recovery into middleware

Sovietaced:

Previously all gRPC handlers would handle panics inside each RPC handler. This added a lot of repetitive boilerplate to all RPC handlers that was pretty fragile to maintain. This pull request introduces recovery middleware that will recover from panics for all RPCs mounted to the RPC server. 

This pull request also proposes a change to the panic recovery logic.

I made a change to the recovery logic such that it logs the panic at the error level instead of the fatal level. The previous fatal error level would call `os.Exit(1)` which immediately terminates the program ungracefully. My suspicion is that this made the existing prometheus panic metrics effectively useless given that prometheus metrics are polled on an interval and the server was likely killed when the metrics would normally be polled. (Arguably the panic metrics could be removed now).

IMO, it's better for high availability to have an RPC server that is alive and sending errors back (and reporting error metrics) than one that gets killed and is unresponsive until Kubernetes decides to boot up another healthy pod. As such, I have changed the behavior to return gRPC INTERNAL status codes instead of terminating the server. I'm open to debating this change so feel free to share your opinion. 

## How was this patch tested?

Unit tests


flyteadmin/pkg/rpc/adminservice/middleware/recovery_interceptor_test.go
```diff
@@ -0,0 +1,38 @@
+package middleware
```
Sovietaced, line 1: I'm open to changing where this lives but it feels like there should be a middleware package, and any interceptors should move here imo.

katrogan: Sounds good!

katrogan (approved):
looks good, thank you for refactoring and the detailed PR explanation!

katrogan:
cc @eapolinario who is looking into test failures

PR #5554, opened 2024-07-11
When there's a triggered LP SubWorkflow use the maxParallelism of LP

### LINKED ISSUE #5555: [Core feature] Sub-executions should respect LaunchPlan MaxParallelism if set

There could be many reasons why a remote LaunchPlan have the MaxParallelism set, I propose that any Sub-Execution should respect that limit instead of using the MaxParallelism of the main execution.

If Workflow A with maxParallelism of 100 triggers LaunchPlan B that have a maxParallelism of 1 then the Sub execution should have a MaxParallelism of 1.

Haven't considered any other idea

_No response_

- [X] Yes

- [X] Yes

RRap0so:
## Tracking issue
Closes #5555 

I'm proposing that Sub Executions should respect the limit set of a remote LaunchPlan. The authors of the LaunchPlan might have a reason to have this value set and could cause problems if the Main Execution has a higher limit

## How was this patch tested?
This was tested using a Workflow that triggers a remote launch plan in a test environment

### DIFF
--- a/flyteadmin/pkg/manager/impl/execution_manager.go
+++ b/flyteadmin/pkg/manager/impl/execution_manager.go
@@ -506,6 +506,14 @@ func (m *ExecutionManager) launchSingleTaskExecution(
 		return nil, nil, err
 	}
 
+	// If a parentNode launched this execution and it's a launchPlan, respect the maxParallelism set in the launchPlan.
+	if parentNodeExecutionID > 0 {
+		if launchPlan.Spec.MaxParallelism > 0 {
+			logger.Info(ctx, fmt.Sprintf("Max parallelism %d set in the LaunchPlan spec.", launchPlan.Spec.MaxParallelism))
+			requestSpec.MaxParallelism = launchPlan.Spec.MaxParallelism
+		}
+	}
+
 	// Dynamically assign task resource defaults.
 	platformTaskResources := util.GetTaskResources(ctx, workflow.Id, m.resourceManager, m.config.TaskResourceConfiguration())
 	for _, t := range workflow.Closure.CompiledWorkflow.Tasks {
@@ -907,6 +915,14 @@ func (m *ExecutionManager) launchExecutionAndPrepareModel(
 	parentNodeExecutionID, sourceExecutionID, err = m.getInheritedExecMetadata(ctx, requestSpec, &workflowExecutionID)
 	if err != nil {
 		return nil, nil, nil, err
+	}
+
+	// If a parentNode launched this execution and it's a launchPlan, respect the maxParallelism set in the launchPlan.
+	if parentNodeExecutionID > 0 {
+		if launchPlan.Spec.MaxParallelism > 0 {
+			logger.Info(ctx, fmt.Sprintf("Max parallelism %d set in the LaunchPlan spec.", launchPlan.Spec.MaxParallelism))
+			requestSpec.MaxParallelism = launchPlan.Spec.MaxParallelism
+		}
 	}
 
 	// Dynamically assign task resource defaults.


RRap0so:
Took a good look at the tests and couldn't really figure out to make sure this is covered. Let me know if you have any tips!

flyteadmin/pkg/manager/impl/execution_manager.go
```diff
@@ -909,6 +917,14 @@ func (m *ExecutionManager) launchExecutionAndPrepareModel(
 		return nil, nil, nil, err
 	}
 
+	// If a parentNode launched this execution and it's a launchPlan, respect the maxParallelism set in the launchPlan.
```
hamersaw, line 920: `and it's a launchPlan` can you elaborate here? Is this supposed to check the node type of the parent node?

IIUC this code is nullifying the overridden max parallelism configuration in the `requestSpec` if the execution has a `parentNodeExecutionID`.

hamersaw (commented):
I'm having a little difficulty following the use-case for sub-execution and parent execution max-parallelism relationship that this is addressing. Can you provide a concrete example maybe?

RRap0so:
@hamersaw sure thing.

We treat our LPs like libraries that are shared to whoever wants to use them. Depending on what systems they touch, the author of those LPs might want to be able to pre-define a max Parallelism. By adding this, we check if the launchPlan has a `parentNodeId` meaning that is not main execution, and if so, check if there's a parallelism defined by the author of said LP. Main execution keeps it's parallelism, sub-executions use defined LP first and if not defined then inherits whatever comes from the Main execution. 

There's also the missing piece from the workflow executor side which I'm planning to work on but just wanted to hear your thoughts first.
Screenshot 2024-07-16 at 10 03 53

hamersaw:
> @hamersaw sure thing.
> 
> We treat our LPs like libraries that are shared to whoever wants to use them. Depending on what systems they touch, the author of those LPs might want to be able to pre-define a max Parallelism. By adding this, we check if the launchPlan has a `parentNodeId` meaning that is not main execution, and if so, check if there's a parallelism defined by the author of said LP. Main execution keeps it's parallelism, sub-executions use defined LP first and if not defined then inherits whatever comes from the Main execution.
> 
> There's also the missing piece from the workflow executor side which I'm planning to work on but just wanted to hear your thoughts first.

What is the behavior right now? In FlytePropeller I see the `MaxParallelism` [being set on the ExecutionCreateRequest ](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/nodes/subworkflow/launchplan/admin.go#L143-L143)for executing a LP is set to the `MaxParallelism`[ for the current execution](https://github.com/flyteorg/flyte/blob/03a1c9e55014fdf5201ad4daccdc5201bf431abb/flytepropeller/pkg/controller/nodes/subworkflow/launchplan.go#L69-L69). This allows things like overriding `MaxParallelism` from the UI / flytekit instead of relying on the `LaunchPlanSpec` definition.

RRap0so:
> > @hamersaw sure thing.
> > We treat our LPs like libraries that are shared to whoever wants to use them. Depending on what systems they touch, the author of those LPs might want to be able to pre-define a max Parallelism. By adding this, we check if the launchPlan has a `parentNodeId` meaning that is not main execution, and if so, check if there's a parallelism defined by the author of said LP. Main execution keeps it's parallelism, sub-executions use defined LP first and if not defined then inherits whatever comes from the Main execution.
> > There's also the missing piece from the workflow executor side which I'm planning to work on but just wanted to hear your thoughts first.
> 
> What is the behavior right now? In FlytePropeller I see the `MaxParallelism` [being set on the ExecutionCreateRequest ](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/nodes/subworkflow/launchplan/admin.go#L143-L143)for executing a LP is set to the `MaxParallelism`[ for the current execution](https://github.com/flyteorg/flyte/blob/03a1c9e55014fdf5201ad4daccdc5201bf431abb/flytepropeller/pkg/controller/nodes/subworkflow/launchplan.go#L69-L69). This allows things like overriding `MaxParallelism` from the UI / flytekit instead of relying on the `LaunchPlanSpec` definition.

As it is right now, sub-executions will just inherit whatever is set from the main execution.

eapolinario:

RRap0so:
Hey @eapolinario @hamersaw we still have folks every now and then asking for this feature. Did you had a chance to take a look at it?

PR #5449, opened 2024-06-04
Add flyteconsole url to FlyteWorkflow CRD

eapolinario:
## Tracking issue
Related to https://github.com/flyteorg/flyte/issues/4773

## Why are the changes needed?
~This is a stacked PR on top of https://github.com/flyteorg/flyte/pull/5432. It adds a new field to the FlyteWorkflow CRD to help in the construction of flyteconsole urls pointing to node executions.~
edit: This PR started its life as a stacked PR, but can be completely separate from https://github.com/flyteorg/flyte/issues/5432. Let's separate those out in order to de-risk this change.

We need to provide a way for tasks to point to a flyteconsole address that corresponds to the execution of those particular tasks. This is useful, for example, if we want to provide third-party services with links to task executions. The first user of this feature is going to be the [wandb flytekit plugin](https://pypi.org/project/flytekitplugins-wandb/).

## What changes were proposed in this pull request?
In this PR we introduce a new field to the `FlyteWorkflow` CRD to represent a `flyteconsole` url. Most of the changes in the PR are about plumbing the necessary interfaces and implementations to use this new field.

The end result is that the pod will contain a new environment variable called `FLYTE_EXECUTION_URL` in case it requested it and the flyteconsole url is provided in the flyteadmin configuration. The way a task requests to have this new env var injected is to define `link_type` in its task template. This is the de factor protocol we've been using to enable dynamic log links, so this PR leans on it.

## How was this patch tested?
functional tests and local testing using the demo cluster.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

PR #5562, opened 2024-07-16
reverted mockery-v2 on ExecutionContext

hamersaw:
[This PR](https://github.com/flyteorg/flyte/pull/5449) updated the mockery command to only produce mocks for the `ExecutionContext` interface. This blocks updating any other construct.

Revert the mockery generate command.

## How was this patch tested?
Unit tests.

## Related PRs
_NA_

## Docs link
_NA_


flytepropeller/pkg/controller/executors/execution_context.go
```diff
@@ -4,7 +4,7 @@ import (
 	"github.com/flyteorg/flyte/flytepropeller/pkg/apis/flyteworkflow/v1alpha1"
 )
 
-//go:generate mockery-v2 --case=underscore --with-expecter --name ExecutionContext --output=mocks
+//go:generate mockery -all -case=underscore
```
pvditt, line 7: @hamersaw would we want to set this to --with-expecter -all ?

hamersaw: I reverted it to what it was previously. If we set `--with-expecter` then we need to refactor everywhere one of these mocks is used to have `EXPECT()`. For an idea of the breadth, this PR just removes the `EXPECT()` from the `ExecutionContext` mocks. So there would be 7x more updates. We can refactor this in another PR if it is important.

pvditt (approved):

PR #5579, opened 2024-07-23
Give flyte binary cluster role permission to create service accounts

shreyas44:
1. Flyte was failing to create service accounts on EKS since it didn't have permissions to do so
2. The service account template in the Single Cluster Simple Cloud Deployment [`eks-starter.yaml`](https://github.com/flyteorg/flyte/blob/master/charts/flyte-binary/eks-starter.yaml) file was failing for the above reason

Give the flyte binary cluster role permissions to create new service accounts.

## How was this patch tested?

Tested on an internal kubernetes cluster.

### Screenshots
Before:
Screenshot 2024-07-22 at 9 26 05 PM

After:
Screenshot 2024-07-22 at 9 49 23 PM

### DIFF
--- a/charts/flyte-binary/templates/clusterrole.yaml
+++ b/charts/flyte-binary/templates/clusterrole.yaml
@@ -25,6 +25,7 @@ rules:
     - namespaces
     - resourcequotas
     - secrets
+    - serviceaccounts
     verbs:
     - create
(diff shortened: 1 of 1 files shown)


eapolinario (approved):
Thank you!

Issue #5575, opened 2024-07-21
[Core feature] Updating annotations of pods belonging Ray cluster in order to adopting Yunikorn Gang scheduling


0yukali0:
### Motivation: Why do you think this is important?

Implementing mechanism eases user overhead about updating annotations and is ability to give different task group name for head pod and each work groups to monitor  each groups status with yunikorn gang scheduling .

### Goal: What should the final outcome look like, ideally?

Annotations of pod belonging head group
```
yunikorn.apache.org/schedulingPolicyParameters: “placeholderTimeoutSeconds=30”
yunikorn.apache.org/taskGroupName: “taskgroup-raycluster-head”
yunikorn.apache.org/taskGroup: “
    TaskGroups: [
      {
        name: taskgroup-raycluster-head
        minMember: 1
        minResource: {
          cpu: 1,
          memory: 500Mi,
        },
        nodeSelector: ...,
        tolerations: ...,
        affinity: ...,
      },
      {
        name: taskgroup-raycluster-worker-0
        minMember: 5
        minResource: {
          cpu: 2,
          memory: 1Gi,
        },
        {
        name: taskgroup-raycluster-worker-1
        minMember: 5
        minResource: {
          cpu: 2,
          memory: 1Gi,
        },
      },
    ]
    "
```
Pods of worker group 0
`name: taskgroup-raycluster-worker-0`
Pods of worker group 1
`name: taskgroup-raycluster-worker-1`

### Describe alternatives you've considered 

A mechanism updates annotations for pods according to the group they belongs. 
Available solution to activate yunikorn gang scheduling via defining pod template results in all pods sharing same task group name, resource and parameters annotations.
In order to distinguish task groups  between head and workers, ray plugin updating annotations including task group name, resource and parameters to a ray pod based on its role when creating ray cluster,
When user set "batch-scheduler" with "yunikorn", ray plugin is going to generate separate task group name for each pod, resource of placeholders and parameters.
Head pod annotations have task group names, placeholders' resource and parameters.
Meanwhile, pods belonging worker group has task group name generated from their worker group. 

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

0yukali0:
I am about to finish this issue in these days, if there are any suggestion and aspect, pls let me know.
We can make this idea better.
Following diagram is the progress i have, i will update this branch to my repo tonight.
[...]

PR #5594, opened 2024-07-26
Updating annotations of pods belonging Ray cluster in order to adopting Yunikorn Gang scheduling

0yukali0:
## Tracking issue
Related to [#5575](https://github.com/flyteorg/flyte/issues/5575)

Ability to classify pods to different gang scheduling group based on their role and worker group number. 
Currently, Pods under ray cluster share same gang scheduling task group via podTemplate.
With same pod spec, pods share same gang scheduling group name when adopting yunikorn.

For enable gang scheduling with yunikorn, updating annotations pods belonging to ray cluster based on role and number of worker group.

## How was this patch tested?
run make test_unit and make lint in flyteplugins.

### Setup process
1. adopting yaml including scheduler name and parameters
```
tasks:
  task-plugins:
    enabled-plugins:
      - container
      - sidecar
      - K8S-ARRAY
      - ray
    default-for-task-types:
      - container: container
      - container_array: K8S-ARRAY
      - ray: ray
    
plugins:
  ray:
    batchScheduler:
      scheduler: yunikorn
      parameters: "placeholderTimeoutInSeconds=10 gangSchedulingStyle=Soft"
```
2. Set [flyte dev env](https://docs.flyte.org/en/latest/community/contribute.html#how-to-setup-dev-environment-for-flyteidl-flyteadmin-flyteplugins-flytepropeller-datacatalog-and-flytestdlib)
3. Yunikorn installation
```yaml
embedAdmissionController: false
enableSchedulerPlugin: false
enableWebService: false
```
`helm install yunikorn yunikorn/yunikorn -f values.yaml -n yunikorn --create-namespace`
4. Set [flyte plugin ray](https://docs.flyte.org/en/latest/deployment/plugins/k8s/index.html)
5.Submit ray example
```
import typing
from flytekit import ImageSpec, Resources, task, workflow

custom_image = ImageSpec(
        registry="localhost:30000",
    packages=["flytekitplugins-ray"],
    apt_packages=["wget"],
)

import ray
from flytekitplugins.ray import HeadNodeConfig, RayJobConfig, WorkerNodeConfig

@ray.remote
def f(x):
    return x * x

ray_config = RayJobConfig(
    head_node_config=HeadNodeConfig(ray_start_params={"log-color": "True"}),
    worker_node_config=[WorkerNodeConfig(group_name="ray-group", replicas=1), WorkerNodeConfig(group_name="ray-group2", replicas=1)],
    runtime_env={"pip": ["numpy", "pandas"]},  # or runtime_env="./requirements.txt"
    enable_autoscaling=False,
    shutdown_after_job_finishes=True,
    ttl_seconds_after_finished=3600,
)

@task(
    task_config=ray_config,
    requests=Resources(mem="2Gi", cpu="2"),
    container_image=custom_image,
)
def ray_task(n: int) -> typing.List[int]:
    futures = [f.remote(i) for i in range(n)]
    return ray.get(futures)

@workflow
def ray_workflow(n: int) -> typing.List[int]:
    return ray_task(n=n)

if __name__ == "__main__":
    print(ray_workflow(n=10))
```
### Screenshots
craete
yunikornlog
results
chments/assets/d3d7d765-598f-4fcc-ae40-86bb8d684681)

## Docs link
[Yunikorn gang scheduling doc](https://yunikorn.apache.org/docs/user_guide/gang_scheduling/)


flyteplugins/go/tasks/plugins/k8s/ray/batchscheduler/config/config.go
```diff
@@ -0,0 +1,21 @@
+package config
+
+type Config struct {
+	Scheduler  string `json:"scheduler,omitempty"`
+	Parameters string `json:"parameters,omitempty"`
```
pingsutw, line 5: Could we add some comments or pflag to this config? here is an [example](https://github.com/flyteorg/flyte/blob/ba88e82a58145fe3062c7441cdd5c57c36f108f2/flytectl/cmd/config/config.go#L21-L24)

flyteplugins/go/tasks/plugins/k8s/ray/config.go
```diff
@@ -23,6 +24,7 @@ var (
 		DashboardHost:            "0.0.0.0",
 		EnableUsageStats:         false,
 		ServiceAccount:           "default",
+		BatchScheduler:           schedulerConfig.NewConfig(),
```
pingsutw, line 27:
```suggestion
		BatchScheduler:           schedulerConfig.Config{},
```

flyteplugins/go/tasks/plugins/k8s/ray/batchscheduler/config/config_test.go
```diff
@@ -0,0 +1,15 @@
+package config
```
pingsutw, line 1: should we move `batchscheduler` plugin to `flyteplugins/go/tasks/plugins/k8s`? spark/dask/kubeflow plugin can use it in the future.

flyteplugins/go/tasks/plugins/k8s/ray/batchscheduler/scheduler/yunikorn/yunikorn.go
```diff
@@ -0,0 +1,152 @@
...
+func (s *Plugin) ProcessWorker(metadata *metav1.ObjectMeta, worker *v1.PodSpec, index int) {
+	s.SetSchedulerName(worker)
+	s.AddGangSchedulingAnnotations(GenerateTaskGroupName(false, index), metadata)
+}
+
+func (s *Plugin) AfterProcess(metadata *metav1.ObjectMeta) {
+	if metadata == nil {
+		return
+	}
+	delete(metadata.Annotations, TaskGroupNameKey)
```
pingsutw, line 57: why do we need to delete annotation after process?

flyteplugins/go/tasks/plugins/k8s/ray/batchscheduler/plugins.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/flyteorg/flyte/flyteidl/gen/pb-go/flyteidl/plugins"
+	schedulerConfig "github.com/flyteorg/flyte/flyteplugins/go/tasks/plugins/k8s/ray/batchscheduler/config"
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/plugins/k8s/ray/batchscheduler/scheduler/kubernetes"
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/plugins/k8s/ray/batchscheduler/scheduler/yunikorn"
+)
+
+type SchedulerPlugin interface {
+	GetSchedulerName() string
+	ParseJob(config *schedulerConfig.Config, metadata *metav1.ObjectMeta, workerGroupsSpec []*plugins.WorkerGroupSpec, pod *v1.PodSpec, primaryContainerIdx int) error
```
pingsutw, line 15: Could we not use ray WorkerGroupSpec here? I think it's better to make it generic so we don't need to write another spark/dash scheduler plugin in the future.

flyteplugins/go/tasks/plugins/k8s/ray/batchscheduler/plugins.go
```diff
@@ -0,0 +1,28 @@
...
+	ProcessHead(metadata *metav1.ObjectMeta, head *v1.PodSpec, index int)
+	ProcessWorker(metadata *metav1.ObjectMeta, worker *v1.PodSpec, index int)
```
pingsutw, line 17: I think we could merge these two methods.
```
AnnotatePodSpec(metadata *metav1.ObjectMeta, podSpec *v1.PodSpec, index int)
```

flyteplugins/go/tasks/plugins/k8s/ray/batchscheduler/config/config_test.go
0yukali0: Agree.

0yukali0:
Hi @pingsutw, thanks for review.
I am about to improve this approach in these days.

davidmirror-ops:
@0yukali0 thank you, this is great. Let us know if you need help taking this to the finish line. This is important for several Flyte users

0yukali0:
Thanks @davidmirror-ops , i will.
Please feel free to mark any areas where i can improve this commit.
BTW, there's some exciting news regarding Spark.
The Spark operator has supported gang scheduling with Yunikorn since last week. Therefore, the future plugin for spark in Flyte can focus on generating Yunikorn application IDs and updating gang scheduling parameters when a new application is submitted.

[The commit in spark operator](https://github.com/kubeflow/spark-operator/pull/2107)

flyteplugins/go/tasks/plugins/k8s/batchscheduler/scheduler/yunikorn/rayhandler.go
```diff
@@ -0,0 +1,101 @@
+package yunikorn
+
+import (
+	"encoding/json"
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func MutateRayJob(parameters string, app *rayv1.RayJob) error {
```
kumare3, line 10: Lets add some comments explaining what this does?

flyteplugins/go/tasks/plugins/k8s/batchscheduler/scheduler/noop_scheduler.go
```diff
@@ -0,0 +1,16 @@
+package scheduler
```
kumare3, line 1: Lets add some module level comments that will help in generated docs

flyteplugins/go/tasks/plugins/k8s/batchscheduler/plugins.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"context"
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/plugins/k8s/batchscheduler/scheduler"
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/plugins/k8s/batchscheduler/scheduler/yunikorn"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type SchedulerManager interface {
+	// Mutate is responsible for mutating the object to be scheduled.
```
kumare3, line 11: Lets explain that the Mutate method will mutate the object in place.

flyteplugins/go/tasks/plugins/k8s/batchscheduler/plugins.go
```diff
@@ -0,0 +1,23 @@
+package batchscheduler
+
+import (
+	"context"
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/plugins/k8s/batchscheduler/scheduler"
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/plugins/k8s/batchscheduler/scheduler/yunikorn"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type SchedulerManager interface {
```
kumare3, line 10: How about docs that say,something to this effect)
```suggestion
// SchedulerManager is an interface that allows plugging in custom schedulers 
// such as Yunikorn, Kueue, or any other scheduler that manages pod scheduling 
// with advanced features like gang scheduling, preemption, and priority.
type SchedulerManager interface {
```

flyteplugins/go/tasks/plugins/k8s/batchscheduler/plugins.go
```diff
@@ -0,0 +1,23 @@
+package batchscheduler
```
kumare3, line 1:
```suggestion
// Flyte often deals with complex, distributed workloads that may require advanced scheduling features. Native Kubernetes scheduling may not suffice for scenarios that involve job-level scheduling policies like gang scheduling or job preemption. The SchedulerManager interface provides a flexible mechanism for integrating such schedulers into Flyte’s plugin system by allowing mutation of the scheduling-related properties of Kubernetes objects.
package batchscheduler
```

kumare3: also how about we call this scheduler?

flyteplugins/go/tasks/plugins/k8s/batchscheduler/scheduler/yunikorn/rayhandler.go
kumare3:
```suggestion
 // Mutate allows the custom scheduler to modify the given Kubernetes object (e.g., a Pod or
    // a custom resource representing a job) before it is submitted to the Kubernetes API.
    // This may include adding annotations, labels, or modifying fields related to scheduling.
    // The object is typically a Kubernetes Pod or a custom resource.
    // It returns an error if the mutation fails.
func MutateRayJob(parameters string, app *rayv1.RayJob) error {
```

flyteplugins/go/tasks/plugins/k8s/batchscheduler/plugins.go
```diff
@@ -0,0 +1,23 @@
...
+	"context"
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/plugins/k8s/batchscheduler/scheduler"
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/plugins/k8s/batchscheduler/scheduler/yunikorn"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type SchedulerManager interface {
+	// Mutate is responsible for mutating the object to be scheduled.
+	// It will add the necessary annotations, labels, etc. to the object.
+	Mutate(ctx context.Context, object client.Object) error
```
kumare3, line 13:
```suggestion
 // Mutate allows the custom scheduler to modify the given Kubernetes object (e.g., a Pod or
    // a custom resource representing a job) before it is submitted to the Kubernetes API.
    // This may include adding annotations, labels, or modifying fields related to scheduling.
    // The object is typically a Kubernetes Pod or a custom resource.
    // It returns an error if the mutation fails.
	Mutate(ctx context.Context, object client.Object) error
```

flyteplugins/go/tasks/plugins/k8s/batchscheduler/scheduler/yunikorn/yunikorn.go
```diff
@@ -0,0 +1,34 @@
...
+	TaskGroupParameters = "yunikorn.apache.org/schedulingPolicyParameters"
+)
+
+type YunikornSchedulerManager struct {
+	parameters string
+}
+
+func (y *YunikornSchedulerManager) Mutate(ctx context.Context, object client.Object) error {
+	switch object.(type) {
+	case *rayv1.RayJob:
```
kumare3, line 24: this is weird isnt it? 
WE have a common interface, but have a switch for each?
In this case, we should allow this overriding to exist next to ray job plugin and not with yunikorn. this will break when things update pretty quickly.

Why have a common interface?

flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go
```diff
@@ -203,6 +205,12 @@ func (e *PluginManager) launchResource(ctx context.Context, tCtx pluginsCore.Tas
 
 	key := backoff.ComposeResourceKey(o)
 
+	err = e.schedulerMgr.Mutate(ctx, o)
```
kumare3, line 208: I see this is why... So here is my suggestion, how about we handle the scheduler related changes directly to the plugin itself.

For example
MutateForGangScheduling

Also how about preemption etc if we add that in the future?

kumare3:
I love this PR, @0yukali0 but there is probably a better way to introduce this interface. I do not love the idea of adding cases in a separate place from the actual ray job plugin. My preference is to keep all that code under one module, also avoids imports

Also can you share an example of how preemption and priority works with yunikorn?

fg91:
Adding support for a gang scheduler is a great effort, thank you 🙏 

I personally am not a huge fan of the folder/module structure proposed in this PR because for each k8s plugin, we'd need to create another module under `scheduler/yunikorn`.

Is my understanding correct that we "only" have to add yunikorn annotations to the respective pod template specs?

If yes, would it be possible to have one central helper function defined that can be imported in the plugins and, if configured in the respective plugin's config, is called in the `BuildResource` function to append the annotations to the pod template spec?

PR #5598, opened 2024-07-27
Deterministic error propagation for distributed (training) tasks

fg91:

See introduction of RFC document.


rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,93 @@
...
+
+- @bgedik
+- @eapolinario
+- @fg91
+
+## 1 Executive Summary
+
+Flyte can schedule distributed training jobs leverging e.g. the [kubeflow training operator](https://github.com/kubeflow/training-operator/tree/f55a91d03f23498cdb465ac26c78566228077c51) and its `PyTorchJob`, `TFJob`, `MPIJob`, ...
+
+For these distributed jobs, multiple Kuberentes pods are launched. Any of these worker pods can crash, causing all other worker pods in the distributed job to fail subsequently because one worker disappeared.
```
fg91, line 13:
```suggestion
For these distributed jobs, multiple Kubernetes pods are launched. Any of these worker pods can crash, causing all other worker pods in the distributed job to fail subsequently because one worker disappeared.
```

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,93 @@
...
+
+
+### Flytepropeller (Backend)
+
+Currently, [here](https://github.com/flyteorg/flyte/blob/4514860cf56ba62717f6c207f269410a8c1a5461/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go#L290) in the plugin manager, upon completion of a node execution, a new [`RemoteFileOutputReader`](https://github.com/flyteorg/flyte/blob/d6da838627d57cd27d60beea004e974ce1fb3ca5/flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go#L14) is constructed which is responsible for reading the error file uploaded to blob storage. This `RemoteFileOutputReader` implements the [`OutputReader` interface](https://github.com/flyteorg/flyte/blob/1e54d21c4d4ee74245f799a57b4bb8a5534e8368/flyteplugins/go/tasks/pluginmachinery/io/iface.go#L32).
+
+We propose to implement a new `MultiErrorFileRemoteFileOutputReader` which (for future flexibility) can be configured with different policies the determine which of multiple errors to report downstream. Intially, the only available policy is "earliest".
```
fg91, line 60:
```suggestion
We propose to implement a new `MultiErrorFileRemoteFileOutputReader` which (for future flexibility) can be configured with different policies the determine which of multiple errors to report downstream. Initially, the only available policy is "earliest".
```

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,93 @@
...
+
+*What are other ways of achieving the same outcome?*
+
+## 7 Potential Impact and Dependencies
+
+The authors of this RFC have experience with pytorch (elastic and non-elastic) distributed training jobs. Are there any community members which have experience with mpi jobs or tenserflow jobs which we can include in the discussions?
+
+## 8 Unresolved questions
+
+Are there any problems regarding backwards compatability? What happens when the flytekit and distributed task plugin version do not upload multiple error files but the backend expects multiple ones (and vice versa)?
```
fg91, line 89:
```suggestion
Are there any problems regarding backwards compatibility? What happens when the flytekit and distributed task plugin version do not upload multiple error files but the backend expects multiple ones (and vice versa)?
```

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,93 @@
...
+
+## 1 Executive Summary
+
+Flyte can schedule distributed training jobs leverging e.g. the [kubeflow training operator](https://github.com/kubeflow/training-operator/tree/f55a91d03f23498cdb465ac26c78566228077c51) and its `PyTorchJob`, `TFJob`, `MPIJob`, ...
+
+For these distributed jobs, multiple Kubernetes pods are launched. Any of these worker pods can crash, causing all other worker pods in the distributed job to fail subsequently because one worker disappeared.
+
+Error propagation, in Flyte, happens by the pod entrypoint uploading a file called `error.pb` to blob storage which contains (among other things) the error message and the information whether the error is retriable.
+
+In a distributed training job, all worker pods currently try to create the same `error.pb` file in blob storage - leading to a race condition. It is not guaranteed that the root-cause error is the one being reported to the user and used to determine whether the task can be retried.
```
bgedik, line 17: As a side note, the current behavior typically results in the worst outcome, as the latter errors override the former ones, which is the exact opposite of the desired behavior of identifying the first error as the root cause.

Not that PyTorch elastic sets the earliest error as the root cause for the errors that happen within a pod, but across pods we typically end up having the opposite, the latest pod error taking over.

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,93 @@
...
+
+## 2 Motivation
+
+* As a Flyte user trying to understand why a distributed training task failed, I currently cannot rely on the error reported in the Flyte Console (UI) being the root cause error.
+    * Instead, I have to search the logs of each worker pod. For distributed training jobs with dozens or even hundreds of worker pods, this can be tedious.
+    * (Current remedies include combining all worker pods in stackdriver logs using a wildcard in the pod name and then filtering by severity.)
+* As a Flyte user marking specific errors that can occur in distributed training jobs as retriable (using a `FlyteRecoverableException`), I want Flyte to deterministically determine the root cause error so that the retry behaviour does not suffer from a race condition.
```
bgedik, line 24: Love this one. It is subtle, but in a typical PyTorch job, the first error is an instance of ChildError, and Flyte plugin for PyTorch handles that properly by checking if the root cause is a recoverable exception. The latter errors are often different, such as Rendezvous errors, and Flyte treats them as non-recoverable. With the current behavior of latter pod errors taking over, we end up treating these as non-recoverable.

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,93 @@
...
+
+Open questions:
+* How does the pod entrypoint determine that a specific task plugin requires separate error files for each worker pod?
+    * One of the task base classes, e.g. `PythonFunctionTask` (at which level should we do this?) could have an attribute `is_distributed` which is set to `False`. Distributed training tasks which inherit from this base class would overwrite this attribute to true. The entrypoint would check this attribute to determine the correct naming of the error file.
+* Should we not configure this in the task classes in flytekit but inject this information e.g. via an env var in `flyteplugins` (backend)?
```
bgedik, line 37: I assume Flyte backend knows whether a task is being implemented by multiple pods or not. If this information is available, it is a better design IMO not to burden the task plugin developers with remembering to override is_distributed.

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,93 @@
...
+
+
+We propose to make it possible to include timestamps in the error proto message reported by flytekit to the backend.
+
+Open questions:
+* Where exactly do we include the time stamp?
+    * [`message Error`](https://github.com/flyteorg/flyte/blob/d6da838627d57cd27d60beea004e974ce1fb3ca5/flyteidl/protos/flyteidl/core/types.proto#L202)
+    * [`message ErrorDocument`](https://github.com/flyteorg/flyte/blob/30d33149159c90d0de44f6351b8d5d7309242e59/flyteidl/protos/flyteidl/core/errors.proto#L32-L35)
+    * [`message ContainerError`](https://github.com/flyteorg/flyte/blob/30d33149159c90d0de44f6351b8d5d7309242e59/flyteidl/protos/flyteidl/core/errors.proto#L11) (I tend to include the timestamp here.)
+* At which level in the [flytekit exceptions](https://github.com/flyteorg/flytekit/tree/master/flytekit/exceptions) do we include the timestamp? Do we need system-scoped and user-scoped exceptions with timestamps? Do we need recoverable and non-recoverable exceptions with a timestamp?
```
bgedik, line 53: Would it make sense to add an optional timestamp to all exception kinds? 

If the timestamp is populated, then that is what we propagate as the timestamp to the pod error. If not we assign timestamp at the task plugin level (e.g. when it is coming from ChildError), and if that is also not available, at the entry point we can set the timestamp to the current timestamp (least desirable, but would work well in practice).

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,93 @@
...
+
+*What are other ways of achieving the same outcome?*
+
+## 7 Potential Impact and Dependencies
+
+The authors of this RFC have experience with pytorch (elastic and non-elastic) distributed training jobs. Are there any community members which have experience with mpi jobs or tenserflow jobs which we can include in the discussions?
+
+## 8 Unresolved questions
+
+Are there any problems regarding backwards compatibility? What happens when the flytekit and distributed task plugin version do not upload multiple error files but the backend expects multiple ones (and vice versa)?
```
bgedik, line 89: How about we generate a error.pb file for backward compatibility, but don't override it in entrypoint if it exists.

Here are the cases we have:
- If the task plugin is newer, then backend will work as expected, with the additional benefit that the latter failures will not overwrite the errors file
- If the backend is newer, then MultiErrorFileRemoteFileOutputReader will see that there are no error-<>.pb files, print a warning, and pick the error.pb file.

rfc/system/5598-deterministic-errors-distributed-training.md
fg91, line 17:
```suggestion
In a distributed training job, all worker pods currently try to create the same `error.pb` file in blob storage - leading to a race condition. It is not guaranteed that the root-cause error is the one being reported to the user and used to determine whether the task can be retried. In fact, the current behavior typically results in the worst outcome, as the latter errors override the former ones, which is the exact opposite of the desired behavior of identifying the first error as the root cause.
```

fg91: Added the first sentence of your remark https://github.com/flyteorg/flyte/pull/5598/commits/68a623a8010a475ecc5306158ade722c4365c2d5

Didn't add the note about elastic because I'd prefer the RFC to not be about one specific plugin and I feel we make the general problem more than apparent.

rfc/system/5598-deterministic-errors-distributed-training.md
fg91: Yes, the respective plugins in `flyteplugins` know and could inject an env var. We already set env vars such as `FLYTE_INTERNAL_EXECUTION_ID`, ... What do you think of adding `FLYTE_INTERNAL_ERROR_PROPAGATION=earliest`? (Not making it a binary flag leaves room for potential other future strategies.)

If we do it this way, backwards compatibility also becomes easier:
If the flytekit entrypoint reacts to the env var and if the to be added multi-error-file-output-reader checks for the default `error.pb` as a fallback, all cases (flytekit too old, backend too old, ...) should be covered.

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,93 @@
...
+
+*What are the main metrics we should be measuring? For example, when interacting with an external system, it might be the external system latency. When adding a new table, how fast would it fill up?*
+
+## 5 Drawbacks
+
+We don't see any drawbacks to making the error handling of distributed training tasks deterministic.
+
+## 6 Alternatives
+
+*What are other ways of achieving the same outcome?*
```
fg91, line 81:
```suggestion
A poor man's version would be to not override the error file, if it is already there. Not as accurate as the proposal in this RFC because there still is a race condition, but this would still better than the current behavior because at least we would *favor* earlier errors instead of later ones.
```

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,93 @@
...
```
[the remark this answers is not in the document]
fg91: https://github.com/flyteorg/flyte/pull/5598/commits/b3add894fd879d6b6c11e37ba292ae04bf60afeb

rfc/system/5598-deterministic-errors-distributed-training.md
fg91: > Would it make sense to add an optional timestamp to all exception kinds?

I like this idea.


If we go with an env var as proposed [here](https://github.com/flyteorg/flyte/pull/5598#discussion_r1695754162), I feel we could just assert that the exception combing from the task has a timestamp. It would then be the plugin's responsibility to catch all errors and re-raise them with an appropriate timestamp. What do you think of this?

rfc/system/5598-deterministic-errors-distributed-training.md
fg91: > but don't override it in entrypoint if it exists

I fear with this approach we enter the realm of consistency and guarantees of blob storage (which might be different for different cloud providers?).

With the following proposal, the pod entrypoints don't have to check whether the `error.pb` exists:
 https://github.com/flyteorg/flyte/pull/5598#discussion_r1695754162

What do you think about this?

rfc/system/5598-deterministic-errors-distributed-training.md
fg91: If we use the pod name in the error file, we also have to set `FLYTE_INTERNAL_POD_NAME` via the downward api.

bgedik: Agreed, this makes backward compatibility easier.

rfc/system/5598-deterministic-errors-distributed-training.md
bgedik: I like it. It handles the case where the backend is old and the task plugin is new. 

For the case where the backend is new and the task plugin is old, we still need the MultiErrorFileRemoteFileOutputReader to tolerate a single error.pb, isn't it?

bgedik (approved):

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,105 @@
...
+
+Open questions:
+
+* How do we configure for distributed plugins to use this new `MultiErrorFileRemoteFileOutputReader` reader instead of the default one?
+    * We could add a `MultipleErrorFiles` property to `PluginProperties` (see https://github.com/flyteorg/flyte/blob/4514860cf56ba62717f6c207f269410a8c1a5461/flyteplugins/go/tasks/pluginmachinery/k8s/plugin.go#L34). The PyTorch plugin, for instance, would then pass `true` for `MultipleErrorFiles` [here](https://github.com/flyteorg/flyte/blob/4514860cf56ba62717f6c207f269410a8c1a5461/flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go#L31).
+    
+    Currently, [here](https://github.com/flyteorg/flyte/blob/4514860cf56ba62717f6c207f269410a8c1a5461/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go#L290) in the plugin manager, where we call `NewRemoteFileOutputReader`, we do have access to `e.plugin`, and thus to `PluginProperties` and could make use of that information to instantiate another output reader.
```
eapolinario, line 65: +1 to this.

If we feel that `MultiErrorFileRemoteFileOutputReader` ends up being used more widely in plugins we can "bump" it up in the configuration of plugins.

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,105 @@
...
+* If flytekit uses an older version that does not support multiple error files while the backend does, a single error file will be uploaded despite `FLYTE_INTERNAL_ERROR_PROPAGATION` being set. The output reader will, however, fall back to reading the single `error.pb`.
+
+### Flyteconsole (UI)
+
+We propose that in the UI, in addition to the error message, we display the name of the pod which experienced the root-cause error. As a user trying to debug the failure, this allows me to quickly identify the logs of the relevant pod in a distributed job with potentially hundreds of workers.
+
+To transport this information from `pyflyte-execute` to flytepropeller, we propose to add an additional field `pod_name` (or `container_name`) to `message ContainerError`.
+
+Open question:
+* Where does flytepropeller add this info for it to be displayed as part of the error in the UI?
```
eapolinario, line 81: @pvditt , let's set some time to clarify how this is done currently.

wild-endeavor:
Flytekit related
* each pod uploads a unique error file name.
  * this is good yes, but i'm not sure it's necessary to pin it to the pod name.  when it makes sense, flyte will open up its abstraction layer and expose the underlying k8s construct (like pod template), but i think in this case maybe it doesn't make sense to.  the file name that propeller passes to the plugin can still be the pod name, esp if that's helpful, but maybe we can use a different env var?
* the `FLYTE_INTERNAL_ERROR_PROPAGATION` - other than signaling to flytekit to use a special error file name, will this do anything else?
* I think adding timestamps to errors is fine, and if that's the only signal a plugin has to go off on, then okay, but in general i don't like relying on timestamps.  wall clock time can be off.
* Should we put the new files in an `errors/` folder as a sibling to the `error.pb` today?

So broadly speaking there's a map/reduce problem with error docs.  With flytekit changes, we can take care of the map part pretty easily. On the reduce side...
* Are there cases where there can be one special Python worker that can iterate through all the child errors and merge them into one error document?  (I'm assuming the answer is even if there are, this is rare and more likely it'll need to be handled by the backend plugin code)
* On the back-end side, there's two potential kinds of plugins - K8s plugins and non-K8s plugins.
  * For non-K8s plugins, we'd probably just ask plugin writers to do this as a special case of `Handle()` right?  Scan output path for multiple error pb files, and somehow collate or merge them into one.
  * For K8s plugins, we're saying that it feels too weird to do it as part of `GetTaskPhase()` right?  I kinda agree.

fg91:
FYI I made the PR a draft again and moved the RFC back to "new" in the Kanban. I don't have a full picture yet of how we communicate from propeller to admin/the UI in which of the workers the error occurred. Once I understood the details, I'll update the RFC and mark it as ready for review again.

fg91:
Thank you for your review @wild-endeavor ! 🙇 

> * each pod uploads a unique error file name.
>   
>   * this is good yes, but i'm not sure it's necessary to pin it to the pod name.  when it makes sense, flyte will open up its abstraction layer and expose the underlying k8s construct (like pod template), but i think in this case maybe it doesn't make sense to.  the file name that propeller passes to the plugin can still be the pod name, esp if that's helpful, but maybe we can use a different env var?

Ultimately we want to show in the UI which worker pod failed so that one immediately knows which pod's logs to check. I was first thinking that propeller takes this information from the error file name. But now, thinking about it, maybe it's better if we properly persist the pod name in `message ContainerError`? The file name could then just contain a random UUID to avoid collisions. That leaves the question how the pod entrypoint knows the pod name. I see two options: 1) either from the `HOSTNAME` env var or 2) from a potential `FLYTE_INTERNAL_POD_NAME` env var that would be set via the k8s downward api. Option 1 would be simpler but I'm not 100% sure whether there are edge cases where the `HOSTNAME` is not equal to the pod name.

> * the `FLYTE_INTERNAL_ERROR_PROPAGATION` - other than signaling to flytekit to use a special error file name, will this do anything else?

In the RFC we discuss that in the future there might be potential other strategies to determine the root cause error other than earliest timestamp. I could imagine that the pod entrypoint *might* have to do different things based on this strategy. For now, `FLYTE_INTERNAL_ERROR_PROPAGATION=earliest` would tell the pod entrypoint that is has to include the timestamp in `ContainerError`.
In theory we could not distinguish between strategies (with a single case for now) but just set `FLYTE_INTERNAL_MULTI_ERROR`.


> * I think adding timestamps to errors is fine, and if that's the only signal a plugin has to go off on, then okay, but in general i don't like relying on timestamps.  wall clock time can be off.

I agree with the criticism that time is not a reliable indicator but it is the strategy that torch distributed elastic launch uses within a local worker group (processes running in a single pod) to determine which of the processes in the pod died first. In this sense, we are just extending the mechanism to multiple nodes.
I would argue that we treat error propagation in a best-effort manner here which will improve the situation drastically given that currently there is a race condition which favors the non-root cause errors. If sometimes we still don't identify the right root cause errors, users can still combine logs of multiple pods and search all of them.

> * Should we put the new files in an `errors/` folder as a sibling to the `error.pb` today?

I'm open to both, if you have a preference, I'll put that into the rfc.


> So broadly speaking there's a map/reduce problem with error docs. With flytekit changes, we can take care of the map part pretty easily. On the reduce side...
> 
> * Are there cases where there can be one special Python worker that can iterate through all the child errors and merge them into one error document?  (I'm assuming the answer is even if there are, this is rare and more likely it'll need to be handled by the backend plugin code)

I can't come up with a case where we would want to handle this in python/tasks. For torch distributed, communication between workers to aggregate errors would require them to rendezvous but this is very likely not possible anymore after the worker group crashed.

> * On the back-end side, there's two potential kinds of plugins - K8s plugins and non-K8s plugins.
>   
>   * For non-K8s plugins, we'd probably just ask plugin writers to do this as a special case of `Handle()` right?  Scan output path for multiple error pb files, and somehow collate or merge them into one.

For this RFC, I'm only thinking about K8s plugins to be honest. I would expect authors of non-k8s plugins to do this in `Handle()`, yes.

>   * For K8s plugins, we're saying that it feels too weird to do it as part of `GetTaskPhase()` right?  I kinda agree.

Yes, definitely, plugins don't read error files in `GetTaskPhase()` currently afaik and I'd like to not blur that boundary.

rfc/system/5598-deterministic-errors-distributed-training.md
pvditt: @eapolinario @fg91 apologies for just now seeing this.

[DemystifyFailure](https://github.com/flyteorg/flyte/blob/5c147eff5905e162244bd84995a25186208d8c5a/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L830) and [DemystifyPending](https://github.com/flyteorg/flyte/blob/5c147eff5905e162244bd84995a25186208d8c5a/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L602) get the messages from the pods.

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,132 @@
...
+}
+```
+
+Currently, [here](https://github.com/flyteorg/flyte/blob/4514860cf56ba62717f6c207f269410a8c1a5461/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go#L290) in the plugin manager, upon completion of a node execution, a new [`RemoteFileOutputReader`](https://github.com/flyteorg/flyte/blob/d6da838627d57cd27d60beea004e974ce1fb3ca5/flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go#L14) is constructed which is responsible for reading the error file uploaded to blob storage. This `RemoteFileOutputReader` implements the [`OutputReader` interface](https://github.com/flyteorg/flyte/blob/1e54d21c4d4ee74245f799a57b4bb8a5534e8368/flyteplugins/go/tasks/pluginmachinery/io/iface.go#L32).
+
+We propose to implement a new `MultiErrorFileRemoteFileOutputReader` which (for future flexibility) can be configured with the different strategies we define. Initially, the only available strategy will be `"earliest"` which the RFC authors aim to use for the kubeflow pytorch plugin. This output reader will search for all error files in the `/errors` folder under the raw output prefix and aggregate the error as specified by the strategy.
```
bgedik, line 87: Instead of a `MultiErrorFileRemoteFileOutputReader` class, why not have `RemoteFileOutputReader` take an `ErrorAggregationStrategy`?

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,132 @@
...
+
+If in [the plugin manager](https://github.com/flyteorg/flyte/blob/4514860cf56ba62717f6c207f269410a8c1a5461/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go#L290) the respective plugin is found to configure an error aggregation strategy other than `Default`, we instantiate such a `MultiErrorFileRemoteFileOutputReader` reader (instead of the existing `RemoteFileOutputReader`) and configure it with the respective strategy.
```
bgedik, line 89: How does the plugin manager get access to the plugin properties?

rfc/system/5598-deterministic-errors-distributed-training.md
```diff
@@ -0,0 +1,132 @@
...
+We propose that the new `MultiErrorFileRemoteFileOutputReader` falls back to reading the `error.pb` (behaviour of the default `RemoteFileOutputReader`) if no `error-<pod-name>.pb` files are found in order to solve the problem of backwards compatibility:
+
+* If flytekit uses a version that supports multiple error files but the backend does not yet, `pyflyte-execute` will not upload multiple error files for distributed tasks since the `FLYTE_INTERNAL_DIST_ERROR_STRATEGY` environment variable will not be set.
+* If flytekit uses an older version that does not support multiple error files while the backend does, a single error file will be uploaded despite `FLYTE_INTERNAL_DIST_ERROR_STRATEGY` being set. The output reader will, however, fall back to reading the single `error.pb`.
+
+
+### Displaying the name of the worker which experienced the root cause error in the UI
+
+We propose that in the UI, in addition to the root-cause error message, for distributed tasks we display the name of the worker pod which experienced the root-cause error. As a user trying to debug a failure, this allows to quickly identify the logs of the relevant pod out of potentially hundreds of pods.
```
bgedik, line 102: name of the worker pod 
->
name of the worker 

based on the latest updates

rfc/system/5598-deterministic-errors-distributed-training.md
bgedik: The `RemoteFileOutputReader`'s main goal is to read outputs. We can have multiple classes to extract the errors, which would be an internal detail to `RemoteFileOutputReader`.

bgedik: The `RemoteFileOutputReader`'s main goal is to read outputs. We can have multiple classes to extract the errors, which would be an internal detail to `RemoteFileOutputReader`.

rfc/system/5598-deterministic-errors-distributed-training.md
fg91: [Here](https://github.com/flyteorg/flyte/blob/4514860cf56ba62717f6c207f269410a8c1a5461/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go#L260), `e.plugin` gives you the plugin.

davidmirror-ops:
09/26/2024 Contributors sync notes: Fabio introduced the idea, @pvditt to please review.

bgedik (approved):

fg91:
@wild-endeavor and @EngHabu had these questions:

> had a quick chat with @EngHabu about this... few questions
> * Is there nothing in the pytorch operator that will help with the aggregation across workers?
> * Is there a possible race condition in the logs? Like worker 1 experiences an error, but then gets delayed writing to s3, so worker 2 writes its error log first, and then worker 1's log appears later in the series.
> * is it possible to just head for the error.pb first, and only do a list if that is not there? then it won't waste the list.
anything else @EngHabu?

@bgedik responded:

> 1/ I’ll research this and get back.
2/ The timestamp is not the timestamp of the write time but the timestamp of the exception happening. However, I don’t know when Flyte retrieves the error. So it is possible that we try to aggregate before everything is ready. I can do some code reading to understand this, but if anyone already has an answer, let me know.
3/ If we are going with the no strategy design and unify, we can certainly do that.

bgedik:
> This means that for distributed jobs we do an "unnecessary" check if error.pb exists (unnecessary because we could know it won't) but this is far better in my eyes than doing an "unnecessary" ls for all non distributed tasks. So I think we could do this if you prefer this behavior over the currently proposed different remote file output readers/strategies in propeller.

@fg91  I thought about point 3 a bit more and I think this is a bit more than just having an extra  HEAD call on `error.pb`. 

Say we looked at `error.pb` and it is not there. Is it not there because we are using multiple error files or is it not there because there is no error? We would then need to do a listing to answer for sure. Then we are back to doing a listing. Specifically:

For non distributed cases:  If `error.pb` file is missing, we need to have additional information on this being non distributed inside the remote output reader to avoid the listing, as lack of the file is not sufficient proof that we have no error.

For distributed cases:  If `error.pb` file is missing, we need to have additional information on whether distributed error reporting is enabled or not inside  the remote output reader to avoid an unnecessary listing.

Overall, having a strategy seems cleaner to me. May be we should discuss a bit what the downside is with the strategy approach.

davidmirror-ops:
10/24/2024 Contributor's sync notes: pending sync with Haytham to keep refining implementation approach.

PR #5599, opened 2024-07-27
Feat: Declarative management of matcheable resources

fg91:

### DIFF
--- a/flyteadmin/pkg/repositories/config/config.go
+++ b/flyteadmin/pkg/repositories/config/config.go
@@ -1,0 +1,125 @@
+package config
+
+import (
+	"github.com/flyteorg/flyte/flytestdlib/config"
+)
+
+const SectionKey = "matcheableResources"
+
+type MatcheableResourcesConfig struct {
+	ExecutionClusterLabels    ExecutionClusterLabelConfig     `json:"executionClusterLabels" pflag:", Defines the configuration for execution cluster labels."`
+	TaskResourceAttributes    TaskResourceAttributesConfig    `json:"taskResourceAttributes" pflag:", Defines the configuration for task resource attributes."`
+	ClusterResourceAttributes ClusterResourceAttributesConfig `json:"clusterResourceAttributes" pflag:", Defines the configuration for cluster resource attributes."`
+	ExecutionQueueAttributes  ExecutionQueueAttributesConfig  `json:"executionQueueAttributes" pflag:", Defines the configuration for execution queue attributes."`
+	PluginOverrides           PluginOverridesConfig           `json:"pluginOverrides" pflag:", Defines the configuration for plugin overrides."`
+	WorkflowExecutionConfigs  WorkflowExecutionConfigConfig   `json:"workflowExecutionConfigs" pflag:", Defines the config for workflow execution configs."`
+}
+
+type TaskResourceAttributesConfig struct {
+	Declarative bool                     `json:"declarative" pflag:", Whether the task resource attributes are managed declaratively, i.e. whether non-listed attributes are deleted."`
+	Values      []TaskResourceAttributes `json:"values" pflag:", The list of task resource attributes to be managed."`
+}
+
+// Config matching admin.TaskResourceSpec
+type TaskResourceSpec struct {
+	Cpu              string `json:"cpu" pflag:", Number of CPUs."`
+	Gpu              string `json:"gpu" pflag:", Number of GPUs."`
+	Memory           string `json:"memory" pflag:", Amount of memory."`
+	Storage          string `json:"storage" pflag:", Amount of storage."`
+	EphemeralStorage string `json:"ephemeralStorage" pflag:", Amount of ephemeral storage."`
+}
+
+type TaskResourceAttributes struct {
+	Project  string           `json:"project" pflag:", The project to which the task resource attribute is applied."`
+	Domain   string           `json:"domain" pflag:", The domain to which the task resource attribute is applied."`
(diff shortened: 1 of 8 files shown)


fg91:
@katrogan my initial plan was to invoke this logic together with the `seed-projects` logic in the init container. That won't work though because the init container is not re-executed when the config map changes. Do you have a suggestion where to best move this? Treat it the same way we treat the "cluster resources sync"?

flyteadmin/pkg/repositories/config/config.go
```diff
@@ -0,0 +1,125 @@
...
+	PluginId              []string `json:"plugin_id" pflag:", Plugin id(s) to be used in place of the default for the task type."`
+	MissingPluginBehavior int      `json:"missing_plugin_behavior" pflag:", Behavior when no specified plugin_id has an associated handler. 0 : FAIL , 1: DEFAULT."`
+}
+
+type WorkflowExecutionConfigConfig struct {
+	Declarative bool                      `json:"declarative" pflag:", Whether the workflow execution configs are managed declaratively, i.e. whether non-listed configs are deleted."`
+	Values      []WorkflowExecutionConfig `json:"values" pflag:", The list of workflow execution configs to be managed."`
+}
+
+type WorkflowExecutionConfig struct {
```
fg91, line 100: This config override is not fully implemented yet because it has a lot of nested sub configs. Might do so in another PR to not overload this one.

flyteadmin/pkg/repositories/config/config.go
```diff
@@ -0,0 +1,125 @@
+package config
+
+import (
+	"github.com/flyteorg/flyte/flytestdlib/config"
+)
+
+const SectionKey = "matcheableResources"
```
fg91, line 7: TODO:

Matcheable resources is "internal naming" according to the [docs](https://docs.flyte.org/en/latest/deployment/configuration/customizable_resources.html#customizing-project-domain-and-workflow-resources-with-flytectl).

How should we call this? `configCustomizations` or `configOverrides`?

RRap0so:
> @katrogan my initial plan was to invoke this logic together with the `seed-projects` logic in the init container. That won't work though because the init container is not re-executed when the config map changes. Do you have a suggestion where to best move this? Treat it the same way we treat the "cluster resources sync"?

I'm not entirely sure this is the correct approach, how would it work if a user of a flyte platform wants to declare their resources? Or are we just considering this feature to be used by the owners/maintainers of the platform?

fg91:
> I'm not entirely sure this is the correct approach, how would it work if a user of a flyte platform wants to declare their resources? Or are we just considering this feature to be used by the owners/maintainers of the platform?

`flytectl update` continues to work of course in case organizations want to allow users to manage this themselves.

> Or are we just considering this feature to be used by the owners/maintainers of the platform?

Yes, exactly.

> We're very much pro-self-serve and take those responsibilities out of the platform maintainers, we've ended up building a controller that has it's own Object that creates/updates/deletes flyte projects and some of their resources. Happy to demo it :)

Would be curious to see it!

flyteadmin/pkg/repositories/config/config.go
katrogan: +1 to anything that's not matchable resources :) I like configOverrides although both sound great!

flyteadmin/pkg/repositories/config/seed_data.go
```diff
@@ -27,3 +30,349 @@ func SeedProjects(db *gorm.DB, projects []string) error {
...
+
+func CreateAttributeModel(project string, domain string, workflow string, attribute *admin.MatchingAttributes, resourceType admin.MatchableResource) (models.Resource, error) {
+	err := ValidateMatcheableResourceLevel(project, domain, workflow)
+	if err != nil {
+		return models.Resource{}, err
+	}
+
+	var model models.Resource
+
+	if workflow != "" {
```
katrogan, line 52: nit: maybe use a switch statement here?

flyteadmin/pkg/repositories/config/seed_data.go
```diff
@@ -27,3 +30,349 @@ func SeedProjects(db *gorm.DB, projects []string) error {
...
+			},
+			resourceType,
+		)
+
+	}
+	return model, err
+}
+
+func CreateKey(project string, domain string, workflow string, resourceType string) string {
+	return fmt.Sprintf("%s-%s-%s-%s", project, domain, workflow, resourceType)
```
katrogan, line 86: nit: should we version this string perhaps? (if we want to one day modify the format to support task or launch plan level overrides, for example) e.g. `v1-%s-%s...`

katrogan (commented):
this is awesome, thank you for putting up this PR!

> @katrogan my initial plan was to invoke this logic together with the seed-projects logic in the init container. That won't work though because the init container is not re-executed when the config map changes. Do you have a suggestion where to best move this? Treat it the same way we treat the "cluster resources sync"?

I actually think this is okay, we never re-read config map values in flyteadmin as it is, and depend on the service deployment being rolled for those to take effect

One optimization we could add is a lightweight marker database table that stores a checksum of the last applied configmap so we don't always have to re-run the migration script on every deploy rollout

katrogan:
re: flytectl update continues to work of course in case organizations want to allow users to manage this themselves.

how does this work in a configmap managed environment? if a user decides to use the cli to update a matchable attribute in the configmap, then the service restart will wipe our their changes, right? should we disable api access in deployments where we use the configmap style approach?

PR #5602, opened 2024-07-29
Introduce more generous resource requests and remove duplicate file

davidmirror-ops:
In line with some of the observations documented [in the docs](https://docs.flyte.org/en/latest/deployment/configuration/performance.html#automatic-scale-out) and as a response to multiple users hitting memory limits quickly on the single binary, this PR introduces more generous resource requests for this Pods. 
Limits are not specified to as they could affect scheduling and influence the behavior on `empyDir` volumes used by Flyte.  ([ref](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#how-pods-with-resource-limits-are-run))


charts/flyte-binary/values.yaml
```diff
@@ -234,12 +234,10 @@ deployment:
...
-  # Uncomment and update to specify resources for deployment
-  # resources:
-  #   limits:
-  #     memory: 1Gi
-  #   requests:
-  #     cpu: 1
+  resources:
+    requests:
+      cpu: "4"
+      memory: "4G"
```
eapolinario, line 240: nit: historically we've always used Binary units instead of Decimal. I know it's a matter of preference, but let's stick to one in our docs/examples/values files.

neverett (approved):

Issue #5318, opened 2024-05-03
[Core feature] Support for Native JSON types in flyte


kumare3:
### Motivation: Why do you think this is important?

Currently flyte supports JSON through protobuf Struct. This causes a lot of problems for the users, especially when using ints as Structs only store numbers - which are double values.

### Goal: What should the final outcome look like, ideally?

Support a new FlyteIDL.LiteralType called LiteralType.JSON. So users can simply use any object that dumps to a JSON using binary / bytearray as the transport layer.

It should be possible to also have customer json serializers and deserializers for the data like JsonPickle etc

### Describe alternatives you've considered 

New literalType called LiteralType.JSON
This maps to a Literal of type binary (wire format)
Optionally the metadata / schema can be extracted and stored

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

kumare3:
Related to https://github.com/flyteorg/flyte/issues/4505

MortalHappiness:
#take

thomasjpfan:
From https://github.com/flyteorg/flyte/issues/4505#issuecomment-2060451042, I see the follow Python types that can use `Literal.JSON`:

1. `dicts` -> `JsonPickle`
2. `dataclasses` -> Use `Mashumuro's` encoders/decorders or the object's `to_json` + `from_json` (These methods are defined with `DataClassJSONMixin`)

@kumare3 Are you considering other Python types that would use `Literal.JSON`?

kumare3:
I am saying wholesale migration to json for these unsupported types from struct

wild-endeavor:
https://github.com/flyteorg/flyte/pull/5337

Future-Outlier:
follow up:
for flytekit, we can suport `Dict[int, dataclass]` as input, but this need to also change `click_types.py`, which will be big enough to open a new PR.

Future-Outlier:
follow up 2:
we should consider rewrite `get_literal_type` for `Dict[int, dataclasss]` attribute access.

wild-endeavor:
> we should consider rewrite get_literal_type for Dict[int, dataclasss] attribute access
what is this now and what should it be?

Future-Outlier:
> > we should consider rewrite get_literal_type for Dict[int, dataclasss] attribute access
> 
> what is this now and what should it be?

now: Dict[int, dataclasss] not supported, and Dict[int, dataclasss] not supported with attribute access
[...]

Issue #5489, opened 2024-06-19
[BUG] Union types fail for e.g. two different dataclasses


fg91:
### Describe the bug

```py
from typing import Union
from dataclasses import dataclass
from flytekit import task, workflow

@dataclass
class A:
    a: int

@dataclass
class B:
    b: int

@task
def foo(inp: Union[A, B]):
    ...

@workflow
def wf():
    foo(inp=B(b=1))

if __name__ == "__main__":
    wf()
```

The workflow works when executing locally with `python wf.py` but fails to register with flyteadmin:

```console
flytekit.exceptions.user.FlyteInvalidInputException: USER:BadInputToAPI: error=None, cause=<_InactiveRpcError of RPC that terminated with:
        status = StatusCode.INVALID_ARGUMENT
        details = "failed to compile workflow for [resource_type:WORKFLOW project:”…” domain:"development" name:”…f” version:”…”] with err failed to compile workflow with err Collected Errors: 1
        Error 0: Code: MismatchingTypes, Node Id: n0, Description: Variable [inp] (type [simple:STRUCT]) doesn't match expected type [union_type:{variants:{simple:STRUCT metadata:{fields:{key:"additionalProperties" value:{bool_value:false}} fields:{key:"properties" value:{struct_value:{fields:{key:"a" value:{struct_value:{fields:{key:"type" value:{string_value:"integer"}}}}}}}} fields:{key:"required" value:{list_value:{values:{string_value:"a"}}}} fields:{key:"title" value:{string_value:"A"}} fields:{key:"type" value:{string_value:"object"}}} structure:{tag:"Object-Dataclass-Transformer"}} variants:{simple:STRUCT metadata:{fields:{key:"additionalProperties" value:{bool_value:false}} fields:{key:"properties" value:{struct_value:{fields:{key:"b" value:{struct_value:{fields:{key:"type" value:{string_value:"integer"}}}}}}}} fields:{key:"required" value:{list_value:{values:{string_value:"b"}}}} fields:{key:"title" value:{string_value:"B"}} fields:{key:"type" value:{string_value:"object"}}} structure:{tag:"Object-Dataclass-Transformer"}}}].
```

### Expected behavior

As a python developer using Flyte, I would expect this workflow to work. 

### Additional context to reproduce

The root cause for this error is the following:

[...]

PR #5607, opened 2024-07-30
[RFC] JSON IDL

Future-Outlier:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5318

### DIFF
--- a/rfc/system/5606-json-idl.md
+++ b/rfc/system/5606-json-idl.md
@@ -61,7 +61,7 @@ def t2(a: dict):
 - We will use the same type interface and ensure the backward compatibility.
 
 ### How to turn a value to bytes?
-#### Use MsgPack to convert value to a byte string
+#### Use MsgPack to convert a value into bytes
 ##### Python
 ```python
 import msgpack
@@ -217,7 +217,7 @@ func literalTypeForScalar(scalar *core.Scalar) *core.LiteralType {
   return literalType 
 }
 ```
-3. Support input and default input
+3. Support input and default input.
 ```go
 // Literal Input
 func ExtractFromLiteral(literal *core.Literal) (interface{}, error) {
@@ -229,7 +229,7 @@ func ExtractFromLiteral(literal *core.Literal) (interface{}, error) {
     }
 }
 // Default Input
-func MakeDefaultLiteralForType(type *core.LiteralType) (*core.Literal, error) {
+func MakeDefaultLiteralForType(typ *core.LiteralType) (*core.Literal, error) {
     switch t := typ.GetType().(type) {
 	case *core.LiteralType_Simple:
         ...
@@ -265,8 +265,8 @@ We will pass the value to our class, which inherits from `click.ParamType`, and 
 | **Stage** | **Conversion** | **Description**                                                                                                                                                                             |
 | --- | --- |---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | **Before** | Python Value to Literal | 1. `Dict[type, type]` uses type hints to construct a LiteralMap. <br> 2. `dict` uses `JSON.dumps` to turn a `dict` value to a JSON string, and store it to Protobuf Struct.                 |
-| | Literal to Python Value | 1. `Dict[type, type]` uses type hints to convert LiteralMap to Python Value. <br> 2. `dict` uses `JSON.loads` to turn a JSON string to a dict value and store it to Protobuf Struct.     |
-| **After** | Python Value to Literal | 1. `Dict[type, type]` stays the same. <br> 2. `dict` uses `msgpack.dumps` to turn a dict to msgpack bytes, and store is to Protobuf JSON.                                            |
+| | Literal to Python Value | 1. `Dict[type, type]` uses type hints to convert LiteralMap to Python Value. <br> 2. `dict` uses `JSON.loads` to turn a JSON string into a dict value and store it to Protobuf Struct.     |
+| **After** | Python Value to Literal | 1. `Dict[type, type]` stays the same. <br> 2. `dict` uses `msgpack.dumps` to turn a dict into msgpack bytes, and store it to Protobuf JSON.                                            |
 | | Literal to Python Value | 1. `Dict[type, type]` uses type hints to convert LiteralMap to Python Value. <br> 2.  `dict` conversion: msgpack bytes -> dict value, method: `msgpack.loads`. |
 
 ### Dataclass Transformer
@@ -287,6 +287,8 @@ We will pass the value to our class, which inherits from `click.ParamType`, and 
 | **After** | Python Value to Literal | Converts the Pydantic `BaseModel` to a dictionary, then serializes it into msgpack bytes using `msgpack.dumps`. |
 | | Literal to Python Value | Deserializes `msgpack` bytes into a dictionary, then converts it back into a Pydantic `BaseModel`. |
 
+Note: Pydantic BaseModel can't be serialized directly by `msgpack`, but this implementation will still ensure 100% correct.
+
 ### FlyteCtl
 In FlyteCtl, we can construct input for the execution, so we have to make sure the values we passed to FlyteAdmin 
 can all be constructed to Literal.
@@ -295,12 +297,11 @@ reference: https://github.com/flyteorg/flytectl/blob/131d6a20c7db601ca9156b8d43d
 
 ### FlyteConsole
 #### Show input/output on FlyteConsole
-We will get node’s input output literal value by FlyteAdmin’s API, and get the JSON byte string in the literal value.
+We will get the node's input and output literal values by FlyteAdmin’s API, and obtain the JSON IDL bytes from the literal value.
 
-We can use MsgPack dumps the MsgPack to a dictionary, and shows it to the flyteconsole.
+We can use MsgPack dumps the MsgPack into a dictionary, and shows it to the flyteconsole.
 #### Construct Input
 We should use `msgpack.encode` to encode input value and store it to the literal’s JSON field.
-
 
 ## 4 Metrics & Dashboards
 
@@ -311,6 +312,7 @@ None
 None
 
 ## 6 Alternatives
+
 None, it's doable.
 
 
--- a/rfc/system/5606-json-idl.md
+++ b/rfc/system/5606-json-idl.md
@@ -8,13 +8,25 @@
 - [@Yee Hing Tong](https://github.com/wild-endeavor)
 
 ## 1 Executive Summary
+- To Literal
 
-Use byte string in protobuf instead of json string to fix int is not supported in protobuf struct.
+| Before                            | Now                                          |
+|-----------------------------------|----------------------------------------------|
+| JSON String -> Protobuf Struct    | JSON String -> Byte String -> Protobuf Json  |
 
+- To Python Value
+
+| Before                            | Now                                          |
+|-----------------------------------|----------------------------------------------|
+| Protobuf Struct -> JSON String | Protobuf Json -> Byte String -> JSON String  |
+
+Use byte string in protobuf instead of json string to fix case that int is not supported in protobuf struct.
 
 ## 2 Motivation
 
-In Flytekit, when handling dataclasses, Pydantic base models, and dictionaries, we store data using JSON strings within struct protobuf. This approach causes issues with integers, as protobuf does not support int types, leading to their conversion to floats. This results in performance issues since we need to recursively iterate through all attributes/keys in dataclasses and dictionaries to ensure floats types are converted to int. In addition to performance issues, the required code is complicated and error prone.
+In Flytekit, when handling dataclasses, Pydantic base models, and dictionaries, we store data using JSON strings within struct protobuf.
+This approach causes issues with integers, as protobuf does not support int types, leading to their conversion to floats.
+This results in performance issues since we need to recursively iterate through all attributes/keys in dataclasses and dictionaries to ensure floats types are converted to int. In addition to performance issues, the required code is complicated and error prone.
 
 Note: We have more than 10 issues about dict, dataclass and Pydantic.
 
@@ -53,7 +65,7 @@ def t2(a: dict):
 - ~~It makes it easier for the frontend to support JSON IDL after these features are merged.~~
 - ~~If users only upgrade Flytekit, we can ensure they won’t face error when using dict only.
 (Since we have to upgrade both flytepropeller, flyteidl and flytekit to support JSON IDL.)~~
-- We will use the same type interface, but ensure the backward compatibility.
+- We will use the same type interface and ensure the backward compatibility.
 
 ### How to create a byte string?
 #### Use MsgPack to convert value to a byte string
@@ -230,71 +242,37 @@ func MakeDefaultLiteralForType(typ *core.LiteralType) (*core.Literal, error) {
 The behavior will remain unchanged. 
 We will pass the value to our class, which inherits from `click.ParamType`, and use the corresponding type transformer to convert the input to the correct type.
 
-#### Dict Transformer
-##### Before
-###### Convert Python Value to Literal
-- Dict[type, type] uses type hints to construct a LiteralMap.
-- `dict` uses `json.dumps` to turn a `dict` value to a JSON string, and store it to protobuf Struct .
-###### Convert Literal to Python Value
-- `Dict[type, type]` uses type hints to convert LiteralMap to Python Value.
-- `dict` uses `json.loads` to turn a JSON string to a dict value and store it to protobuf Struct .
-##### After
-###### Convert Python Value to Literal
-- `Dict[type, type]` stays the same.
-- ~~`dict` uses `msgpack.dumps` to turn a dict value to a JSON byte string, and store is to protobuf Json .~~
-- `dict` uses `msgpack.dumps` to turn a JSON string to a byte string, and store is to protobuf Json .
+### Dict Transformer
 
-###### Convert Literal to Python Value
-- `Dict[type, type]` uses type hints to convert LiteralMap to Python Value.
-- ~~`dict` uses `msgpack.loads` to turn a byte string to a JSON string and `dict` value and store it to protobuf 
-  `Struct` .~~
-- `dict` conversion: byte string -> JSON string -> dict value, method: `msgpack.loads` -> `json.loads`
+| **Stage** | **Conversion** | **Description**                                                                                                                                                                                                                                                                                                               |
+| --- | --- |-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **Before** | Python Value to Literal | 1. `Dict[type, type]` uses type hints to construct a LiteralMap. <br> 2. `dict` uses `json.dumps` to turn a `dict` value to a JSON string, and store it to protobuf Struct.                                                                                                                                                   |
+| | Literal to Python Value | 1. `Dict[type, type]` uses type hints to convert LiteralMap to Python Value. <br> 2. `dict` uses `json.loads` to turn a JSON string to a dict value and store it to protobuf Struct.                                                                                                                                          |
+| **After** | Python Value to Literal | 1. `Dict[type, type]` stays the same. <br> 2. ~~`dict` uses `msgpack.dumps` to turn a dict value to a JSON byte string, and store is to protobuf Json.~~ <br> 3. `dict` uses `msgpack.dumps` to turn a JSON string to a byte string, and store is to protobuf Json.                                                           |
+| | Literal to Python Value | 1. `Dict[type, type]` uses type hints to convert LiteralMap to Python Value. <br> 2. ~~`dict` uses `msgpack.loads` to turn a byte string to a JSON string and `dict` value and store it to protobuf `Struct`.~~ <br> 3. `dict` conversion: byte string -> JSON string -> dict value, method: `msgpack.loads` -> `json.loads`. |
 
-#### Dataclass Transformer
-##### Before
-###### Convert Python Value to Literal
-Uses `mashumaro JSON Encoder` to turn a dataclass value to a JSON string, and store it to protobuf `Struct` .
+### Dataclass Transformer
 
-Note: For `FlyteTypes`, we will inherit mashumaro `SerializableType` to define our own serialization behavior, which includes upload file to remote storage.
+| **Stage** | **Conversion** | **Description**                                                                                                                                                                                                                                                                                                                                                                                                                                                                   |
+| --- | --- |-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **Before** | Python Value to Literal | Uses `mashumaro JSON Encoder` to turn a dataclass value to a JSON string, and store it to protobuf `Struct`. <br><br> ~~Note: For `FlyteTypes`, we will inherit mashumaro `SerializableType` to define our own serialization behavior, which includes uploading files to remote storage.~~                                                                                                                                                           |
+| | Literal to Python Value | Uses `mashumaro JSON Decoder` to turn a JSON string to a python value, and recursively fixed int attributes to int (it will be float because we stored it in to `Struct`). <br><br> ~~Note: For `FlyteTypes`, we will inherit the mashumaro `SerializableType` to define our own deserialization behavior, which includes download file from remote storage.~~                                                                                                                        |
+| **After** | Python Value to Literal | ~~Uses `msgpack.dumps()` to turn a Python value into a byte string.~~ <br> Uses `mashumaro JSON Encoder` to turn a dataclass value to a JSON string, and uses `msgpack.dumps()` to turn the JSON string into a byte string, and store it to protobuf `Json`. <br><br> ~~Note: For `FlyteTypes`, we will need to customize serialization behavior by msgpack reference [here](https://github.com/msgpack/msgpack-python?tab=readme-ov-file#packingunpacking-of-custom-data-type).~~ |
+| | Literal to Python Value | ~~Uses `msgpack.loads()` to turn a byte string into a Python value.~~ <br> Uses `msgpack.loads()` to turn a byte string into a JSON string, and uses `mashumaro JSON Decoder` to turn the JSON string into a Python value. <br><br> ~~Note: For `FlyteTypes`, we will need to customize deserialization behavior by `msgpack` reference [here](https://github.com/msgpack/msgpack-python?tab=readme-ov-file#packingunpacking-of-custom-data-type).~~                              |
 
-###### Convert Literal to Python Value
-Uses `mashumaro JSON Decoder` to turn a JSON string to a python value, and recursively fixed int attributes to int (it will be float because we stored it in to `Struct`).
+### Pydantic Transformer
 
-Note: For `FlyteTypes`, we will inherit the mashumaro `SerializableType` to define our own serialization behavior, which includes uploading files to remote storage.
-##### After
-###### Convert Python Value to Literal
-~~Uses `msgpack.dumps()` to turn a Python value into a byte string.~~
-Uses `mashumaro JSON Encoder` to turn a dataclass value to a JSON string, and uses `msgpack.dumps()` to turn the JSON string into a byte string, and store it to protobuf `Json`.
+| **Stage** | **Conversion** | **Description** |
+| --- | --- | --- |
+| **Before** | Python Value to Literal | Convert `BaseModel` to a JSON string, and then convert it to a Protobuf `Struct`. |
+| | Literal to Python Value | Convert Protobuf `Struct` to a JSON string and then convert it to a `BaseModel`. |
+| **After** | Python Value to Literal | ~~Convert the Pydantic `BaseModel` to a JSON string, then convert the JSON string to a `dictionary`, and finally, convert it to a `byte string` using msgpack.~~ <br> Convert the Pydantic `BaseModel` to a JSON string, then convert the JSON string to a `byte string` using msgpack. |
+| | Literal to Python Value | ~~Convert `byte string` to a `dictionary` using `msgpack`, then convert dictionary to a JSON string, and finally, convert it to Pydantic `BaseModel`.~~ <br> Convert `byte string` to a JSON string using `msgpack`, then convert it to Pydantic `BaseModel`. |
 
-Note: For `FlyteTypes`, we will need to customize serialization behavior by msgpack reference here.
 
-https://github.com/msgpack/msgpack-python?tab=readme-ov-file#packingunpacking-of-custom-data-type 
-
-###### Convert Literal to Python Value
-
-~~Uses `msgpack.loads()` to turn a byte string into a Python value.~~
-Uses `msgpack.loads()` to turn a byte string into a JSON string, and uses `mashumaro JSON Decoder` to turn the JSON string into a Python value.
-
-Note: For `FlyteTypes`, we will need to customize deserialization behavior by `msgpack` reference here.
-
-https://github.com/msgpack/msgpack-python?tab=readme-ov-file#packingunpacking-of-custom-data-type 
-
-#### Pydantic Transformer
-##### Before
-###### Convert Python Value to Literal
-Convert `BaseModel` to a JSON string, and then convert it to a Protobuf `Struct`.
-###### Convert Literal to Python Value
-Convert Protobuf `Struct` to a JSON string and then convert it to a `BaseModel`.
-##### After
-###### Convert Python Value to Literal
-~~Convert the Pydantic `BaseModel` to a JSON string, then convert the JSON string to a `dictionary`, and finally, convert it to a `byte string` using msgpack.~~
-
-Convert the Pydantic `BaseModel` to a JSON string, then convert the JSON string to a `byte string` using msgpack.
-###### Convert Literal to Python Value
-~~Convert `byte string` to a `dictionary` using `msgpack`, then convert dictionary to a JSON string, and finally, convert it to Pydantic `BaseModel`.~~
-Convert `byte string` to a JSON string using `msgpack`, then convert it to Pydantic `BaseModel`.
 ### FlyteCtl
-In Flytectl, we can construct input for the execution, so we have to make sure the values we passed to FlyteAdmin can all be constructed to Literal.
+In FlyteCtl, we can construct input for the execution, so we have to make sure the values we passed to FlyteAdmin 
+can all be constructed to Literal.
 
 https://github.com/flyteorg/flytectl/blob/131d6a20c7db601ca9156b8d43d243bc88669829/cmd/create/serialization_utils.go#L48 
 
@@ -312,10 +290,13 @@ We should use `msgpack.encode` to encode input value and store it to the literal
 None
 
 ## 5 Drawbacks
-There's no breaking changes if we use `Annotated[dict, Json]`, but we need to be very careful about will there be any breaking changes.
+~~There's no breaking changes if we use `Annotated[dict, Json]`, but we need to be very careful about will there be any breaking changes.~~
+
+None
 
 ## 6 Alternatives
-Use UTF-8 format to encode and decode, this will be more easier for implementation, but maybe will cause performance issue when using Pydantic Transformer.
+~~Use UTF-8 format to encode and decode, this will be more easier for implementation, but maybe will cause performance issue when using Pydantic Transformer.~~
+None, it's doable.
 
 ## 7 Potential Impact and Dependencies
 
@@ -325,7 +306,9 @@ Use UTF-8 format to encode and decode, this will be more easier for implementati
 - *How could this be exploited by malicious attackers?*
 
 ## 8 Unresolved questions
-I am not sure use UTF-8 format or msgpack to encode and decode is a better option.
+~~I am not sure use UTF-8 format or msgpack to encode and decode is a better option.~~
+MsgPack is better because it's more smaller and faster.
 
 ## 9 Conclusion
-Whether use UTF-8 format or msgpack to encode and decode, we will definitely do it.
+~~Whether use UTF-8 format or msgpack to encode and decode, we will definitely do it.~~
+We will use msgpack to do it.
--- a/rfc/system/5606-json-idl.md
+++ b/rfc/system/5606-json-idl.md
@@ -24,8 +24,8 @@ Use byte string in protobuf instead of json string to fix case that int is not s
 
 ## 2 Motivation
 
-In Flytekit, when handling dataclasses, Pydantic base models, and dictionaries, we store data using JSON strings within struct protobuf.
-This approach causes issues with integers, as protobuf does not support int types, leading to their conversion to floats.
+In Flytekit, when handling dataclasses, Pydantic base models, and dictionaries, we store data using JSON strings within protobuf struct datatype.
+This approach causes issues with integers, as protobuf struct does not support int types, leading to their conversion to floats.
 This results in performance issues since we need to recursively iterate through all attributes/keys in dataclasses and dictionaries to ensure floats types are converted to int. In addition to performance issues, the required code is complicated and error prone.
 
 Note: We have more than 10 issues about dict, dataclass and Pydantic.
--- a/rfc/system/5606-json-idl.md
+++ b/rfc/system/5606-json-idl.md
@@ -6,6 +6,7 @@
 - [@Ping-Su](https://github.com/pingsutw)
 - [@Fabio M. Graetz](https://github.com/fg91)
 - [@Yee Hing Tong](https://github.com/wild-endeavor)
+- [@Eduardo Apolinario](https://github.com/eapolinario)
 
 ## 1 Executive Summary
 - To Literal
@@ -33,8 +34,7 @@ Note: We have more than 10 issues about dict, dataclass and Pydantic.
 This feature can solve them all.
 
 ## 3 Proposed Implementation
-### Flytekit Example
-#### Before
+### Before
 ```python
 @task
 def t1() -> dict:
@@ -45,7 +45,7 @@ def t1() -> dict:
 def t2(a: dict):
   print(a["integer"]) # wrong, will be a float
 ```
-#### After
+### After
 ```python
 @task
 def t1() -> dict: # Json Byte Strings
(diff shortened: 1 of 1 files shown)


rfc/system/5606-json-idl.md
```diff
@@ -0,0 +1,343 @@
+# JSON IDL
+
+**Authors:**
+
+- [@Han-Ru](https://github.com/future-outlier)
+- [@Ping-Su](https://github.com/pingsutw)
+
+## 1 Executive Summary
+
+Use json byte string in protobuf instead of json string to fix int is not supported in protobuf struct.
```
fg91, line 10:
```suggestion
Use json byte string in protobuf instead of json string to avoid the issue that integers are not supported in protobuf structs.
```

rfc/system/5606-json-idl.md
```diff
@@ -0,0 +1,343 @@
...
+
+
+## 2 Motivation
+
+In Flytekit, when handling dataclasses, Pydantic base models, and dictionaries, we store data using JSON strings within struct protobuf. This approach causes issues with integers, as protobuf does not support int types, leading to their conversion to floats. This results in performance issues since we need to recursively iterate through all attributes/keys in dataclasses and dictionaries to ensure floats types are converted to int.
```
fg91, line 15:
```suggestion
In Flytekit, when handling dataclasses, Pydantic base models, and dictionaries, we store data using JSON strings within struct protobuf. This approach causes issues with integers, as protobuf does not support int types, leading to their conversion to floats. This results in performance issues since we need to recursively iterate through all attributes/keys in dataclasses and dictionaries to ensure floats types are converted to int. In addition to performance issues, the required code is complicated and error prone.
```

rfc/system/5606-json-idl.md
```diff
@@ -0,0 +1,343 @@
...
+This feature can solve them all.
+
+## 3 Proposed Implementation
+### Flytekit Example
+#### Before
+```python
+@task
+def t1() -> dict:
+  ...
+  return {"a": 1} -> protobuf Struct {"a": 1.0}
```
fg91, line 28:
```suggestion
  return {"a": 1}  # -> protobuf Struct {"a": 1.0}
```

rfc/system/5606-json-idl.md
```diff
@@ -0,0 +1,343 @@
...
+
+@task
+def t2(a: <dict>):
+  print(a["integer"]) # wrong, will be float point
```
fg91, line 32:
```suggestion
def t2(d: dict):
  print(d["a"]) # will be float instead of int
```

rfc/system/5606-json-idl.md
```diff
@@ -0,0 +1,343 @@
...
+```
+#### After
+```python
+Json = "json"
+
+@task
+def t1() -> Annotated[dict, Json]: // Json Byte Strings
+  ...
+  return {"a": 1} -> protobuf Json b'{"a": 1}'
```
fg91, line 41:
```suggestion
  return {"a": 1}  # -> protobuf Json b'{"a": 1}'
```

rfc/system/5606-json-idl.md
```diff
@@ -0,0 +1,343 @@
...
+@task
+def t2(a: <dict>):
+  print(a["integer"]) # wrong, will be float point
+```
+#### After
+```python
+Json = "json"
+
+@task
+def t1() -> Annotated[dict, Json]: // Json Byte Strings
```
fg91, line 39: Will the user have to use this function signature? As a user I would want to be able to just use ` ->  dict` and not worry about this.

rfc/system/5606-json-idl.md
```diff
@@ -0,0 +1,343 @@
...
+### Flytekit Example
+#### Before
+```python
+@task
+def t1() -> dict:
+  ...
+  return {"a": 1} -> protobuf Struct {"a": 1.0}
+
+@task
+def t2(a: <dict>):
```
fg91, line 31:
```suggestion
def t2(a: dict):
```

rfc/system/5606-json-idl.md
```diff
@@ -0,0 +1,343 @@
...
+Json = "json"
+
+@task
+def t1() -> Annotated[dict, Json]: // Json Byte Strings
+  ...
+  return {"a": 1} -> protobuf Json b'{"a": 1}'
+
+@task
+def t2(a: <dict>):
+  print(a["integer"]) # wrong
```
fg91, line 45:
```suggestion
def t2(a: dict):
  print(a["integer"]) # Now actually an integer
```

Is this what you mean? Or is it still wrong?

rfc/system/5606-json-idl.md
```diff
@@ -0,0 +1,343 @@
...
+```
+
+#### Note
+- We use Annotated[dict, Json]  instead of dict to ensure backward compatibility.
+    - This helps us avoid breaking changes.
```
fg91, line 50: As a user, I wouldn't want to use `Annotated[dict, Json]` to be honest :/

rfc/system/5606-json-idl.md
```diff
@@ -0,0 +1,343 @@
...
+                                            Value: []byte("{}"),
+                                        },
+                                    },
+                                },
+                            },
+                        }, nil
+    }
+}
+```
+### FlyteKit
```
fg91, line 222: Nit:

I feel this detail is not needed in the RFC, the addition of the Json message to flyteidl contains all the information.

rfc/system/5606-json-idl.md
```diff
@@ -0,0 +1,343 @@
...
+### FlyteCtl
+In Flytectl, we can construct input for the execution, so we have to make sure the values we passed to FlyteAdmin can all be constructed to Literal.
+
+https://github.com/flyteorg/flytectl/blob/131d6a20c7db601ca9156b8d43d243bc88669829/cmd/create/serialization_utils.go#L48 
+
+### FlyteConsole
+#### Show input/output on FlyteConsole
+We will get node’s input output literal value by FlyteAdmin’s API, and get the json byte string in the literal value.
+
+We can use MsgPack dumps the json byte string to a dictionary, and shows it to the flyteconsole.
```
fg91, line 316: To make sure I understand correctly, the user will still be able to see the content of the dict, ... because flyte console too will deserialize?
If yes, this is quite neat.

rfc/system/5606-json-idl.md
fg91: Do you think it would be possible that we "put some kind of metadata into the Literal" that "tells the type engine to use the new mechanism"?

fg91:
A while ago we discussed [this](https://github.com/flyteorg/flyte/issues/5489) issue which is also related to how dataclasses etc. are serialized. @eapolinario remarked:

> We're in the process of revisiting the dataclass transformer and will include this in the process.
A good point raised during today's contributor meeting is that simply adding the tags to the struct message is not enough as that would break inheritance.

Is this RFC the mentioned process to revisit the dataclass transformer?

rfc/system/5606-json-idl.md
wild-endeavor: @Future-Outlier is there any way to just make the reading backwards compatible?  if this happens to read a cached value from an upstream execution that written with the old code, fall back to reading the old code (and just log a warning).

I think most people update admin & propeller at the same time.  If they just upgrade flytekit with this new type, it'll fail at registration time right?  if they just upgrade flytekit & admin, but not propeller, it'll fail at launch time?

eapolinario:
> A while ago we discussed [this](https://github.com/flyteorg/flyte/issues/5489) issue which is also related to how dataclasses etc. are serialized. @eapolinario remarked:
> 
> > We're in the process of revisiting the dataclass transformer and will include this in the process.
> > A good point raised during today's contributor meeting is that simply adding the tags to the struct message is not enough as that would break inheritance.
> 
> Is this RFC the mentioned process to revisit the dataclass transformer?

yes, exactly. We've been [experimenting](https://github.com/flyteorg/flytekit/pull/2600) with this idea and decided to make it official by following the [RFC process](https://github.com/flyteorg/flyte/tree/master/rfc).

rfc/system/5606-json-idl.md
wild-endeavor: I think this is not needed yeah, we don't want to draw attention to model files in general.  But could you add an example showing how pyflyte run would work?

rfc/system/5606-json-idl.md
Future-Outlier: Yes, you're right, thank you

rfc/system/5606-json-idl.md
Future-Outlier: @fg91 @wild-endeavor 
For flytekit, there's defintely a way to make it backward compatible by adding metadata.

Future-Outlier: > it'll fail at registration time right?
If we specify with default input, and propeller didn't upgrade (take sandbox as an example)
propeller will crash. (This is definitely a bug in propeller)
```python

flytekit_hash = "4795f8ef3291e5f90730c0a030025ec6bc189190"
flyteidl_hash = "c1182565d62b4885156454da9d0a100b2facf313"

json_idl = f"git+https://github.com/flyteorg/flyte.git@{flyteidl_hash}#subdirectory=flyteidl"
flytekit = f"git+https://github.com/flyteorg/flytekit.git@{flytekit_hash}"

image = ImageSpec(
    packages=[flytekit, json_idl],
    apt_packages=["git"],
    registry="localhost:30000",
)

@task(container_image=image)
def t2(input: dict) -> dict:
    print(type(input["a"]))
    print(type(input["a"]))
    return input

@workflow
def wf(input: dict={'a': 1, }) -> dict:
    return t2(input=input)
```

image


image

Future-Outlier: > if they just upgrade flytekit & admin, but not propeller, it'll fail at launch time?
It'll fail before launch time, but I also think it'll fail at launch time too.
The worst part is that the propeller will crash.
Do you think it's ok?
We didn't have good error handling about handling unknown literal types registered to the propeller's compiler should return a good error message response to us, instead out propeller will just crash.

Future-Outlier:
> A while ago we discussed [this](https://github.com/flyteorg/flyte/issues/5489) issue which is also related to how dataclasses etc. are serialized. @eapolinario remarked:
> 
> > We're in the process of revisiting the dataclass transformer and will include this in the process.
> > A good point raised during today's contributor meeting is that simply adding the tags to the struct message is not enough as that would break inheritance.
> 
> Is this RFC the mentioned process to revisit the dataclass transformer?

I can tell what might be a potential solution for it.
We should put `identifier` to metadata for different dataclasses.

rfc/system/5606-json-idl.md
Future-Outlier: For the `wrong` one, it means that we will get a float value but not in value due to the `Struct` protobuf doesn't have `int` field.

rfc/system/5606-json-idl.md
```diff
@@ -0,0 +1,343 @@
...
+
+
+## 4 Metrics & Dashboards
+
+None
+
+## 5 Drawbacks
+There's no breaking changes if we use `Annotated[dict, Json]`, but we need to be very careful about will there be any breaking changes.
+
+## 6 Alternatives
```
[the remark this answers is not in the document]
Future-Outlier: Let me clarify the serialization process (deserialization is the reverse):

- Using msgpack:
basemodel -> json string -> dictionary -> msgpack bytes

- Using utf-8:
basemodel -> json string -> json byte string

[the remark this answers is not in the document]
Future-Outlier: But the benefit is that if we use `msgpack`, we can have smaller data bytes when uploading to remote storage and downloading from it.
This is really a HUGE benefit and lots of people wants it.

pingsutw (dismissed):

Future-Outlier:
@fg91 @eapolinario @wild-endeavor
Does it look good to you?
Can we merge it?

rfc/system/5606-json-idl.md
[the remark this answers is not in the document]
eapolinario: UTF-8 and msgpack are not the only possible formats, right? Can we add more color to this? What if of instead dictating msgpack as the serialization format everywhere we added that as an explicit field as part of the new literal type? This way we could have more specialized serialization formats used by plugins in case they want/need to.

eapolinario (commented):
Can you remove all uses of strikethrough (i.e. `~~`)?

Also, can we clarify why we need to encode objects to json and then msgpack before storing those values? In other words, why can't we go straight to msgpack-encoded objects whenever applicable?

Future-Outlier:
Hi, all I just came out with performance tests.
However, I found that `mashumaro` can't support `pydantic BaseModel` , so it means if we have to have different behaviors between `dataclass and dict` and `pydantic BaseModel`, and we have to handle those differernt behavior in the flyte backend and frontend.

EX:
dataclass and dict: python val -> msgpack bytes
pydantic BaseModel: python val -> json str -> msgpack bytes

## performance
### encode

### decode

## example
### encode
```python
import msgpack
from dataclasses import dataclass, fields
from typing import List, Dict
from mashumaro.mixins.msgpack import DataClassMessagePackMixin
from mashumaro.codecs.json import JSONEncoder, JSONDecoder
from mashumaro.codecs.msgpack import MessagePackEncoder
import timeit


@dataclass
class InnerDataClass:
    inner_attr1: int
    inner_attr2: str

@dataclass
class MyDataClassWithMsgPackMixin(DataClassMessagePackMixin):
    attr1: int
    attr2: str
    attr3: float
    attr4: bool
    attr5: List[int]
    attr6: InnerDataClass
    attr7: str
    attr8: float
    attr9: Dict[str, int]
    attr10: bool

@dataclass
class MyDataClass:
    attr1: int
    attr2: str
    attr3: float
    attr4: bool
    attr5: List[int]
    attr6: InnerDataClass
    attr7: str
    attr8: float
    attr9: Dict[str, int]
    attr10: bool


# Initialize objects
dc_with_mixin = MyDataClassWithMsgPackMixin(
    attr1=1, attr2='2', attr3=3.0, attr4=True, attr5=[1, 2, 3],
    attr6=InnerDataClass(1, '2'), attr7='7', attr8=8.0,
    attr9={'9': 9}, attr10=True
)

dc = MyDataClass(
    attr1=1, attr2='2', attr3=3.0, attr4=True, attr5=[1, 2, 3],
    attr6=InnerDataClass(1, '2'), attr7='7', attr8=8.0,
    attr9={'9': 9}, attr10=True
)

new_dc = type(MyDataClass.__name__ + "WithMessagePackMixin", (MyDataClass, DataClassMessagePackMixin), {})
encoder = JSONEncoder(MyDataClass)

# Test function 1: Directly use msgpack mixin
def test_msgpack_mixin():
    return dc_with_mixin.to_msgpack()

# Test function 2: Dynamically create a class with msgpack mixin
def test_dynamic_msgpack_mixin():
    new_instance = new_dc(**{field.name: getattr(dc_with_mixin, field.name) for field in fields(dc_with_mixin)})
    return new_instance.to_msgpack()

def test_create_new_dc():
    return type(MyDataClass.__name__ + "WithMessagePackMixin", (MyDataClass, DataClassMessagePackMixin), {})

# Test function 3: Convert dataclass to JSON string and then encode to msgpack
def test_json_str_to_msgpack():
    return msgpack.dumps(encoder.encode(dc))

def test_create_encoder():
    return JSONEncoder(MyDataClass)

# Test function 4: Use MessagePackEncoder from mashumaro (without initialization)
def test_mashumaro_msgpack_encoder():
    return mashumaro_encoder.encode(dc)

# Test function for initializing the mashumaro MessagePackEncoder
def test_create_mashumaro_encoder():
    return MessagePackEncoder(MyDataClass)

# Initialize mashumaro encoder outside the test loop
mashumaro_encoder = test_create_mashumaro_encoder()

# Use timeit to test each method's performance, running 10000 times
iterations = 10000
time_mixin = timeit.timeit(test_msgpack_mixin, number=iterations)
time_dynamic = timeit.timeit(test_dynamic_msgpack_mixin, number=iterations)
time_dynamic_create_base_class = timeit.timeit(test_create_new_dc, number=1)
time_json = timeit.timeit(test_json_str_to_msgpack, number=iterations)
time_create_json_encoder = timeit.timeit(test_create_encoder, number=1)
time_mashumaro_msgpack = timeit.timeit(test_mashumaro_msgpack_encoder, number=iterations)
time_create_mashumaro_encoder = timeit.timeit(test_create_mashumaro_encoder, number=1)

# Output the results
print(f'{iterations} times dataclass with msgpack mixin elapsed time:', time_mixin)
print(f'{iterations} times dynamically create msgpack mixin elapsed time:', time_dynamic + time_dynamic_create_base_class)
print(f'{iterations} times dataclass -> json str -> msgpack bytes elapsed time:', time_json + time_create_json_encoder)
print(f'{iterations} times mashumaro msgpack encoder elapsed time:', time_mashumaro_msgpack + time_create_mashumaro_encoder)
# print(f'1 time creating mashumaro msgpack encoder elapsed time:', time_create_mashumaro_encoder)

import matplotlib.pyplot as plt

# Labels and timings for the four methods
methods = ['Msgpack Mixin', 'Dynamic Msgpack Mixin', 'JSON String to Msgpack', 'Mashumaro Msgpack Encoder']
times = [time_mixin, time_dynamic + time_dynamic_create_base_class, time_json + time_create_json_encoder, time_mashumaro_msgpack + time_create_mashumaro_encoder]

# Plotting the comparison
plt.figure(figsize=(30, 5))
plt.bar(methods, times, color=['blue', 'green', 'red', 'orange'])
plt.xlabel('Methods')
plt.ylabel('Elapsed Time (seconds)')
plt.title('Comparison of Serialization Methods')
plt.savefig('/Users/future-outlier/code/dev/flytekit/build/PR/JSON/eduardo/performance_comparison/encode.png')
```

### decode
```python
import msgpack
from dataclasses import dataclass, fields
from typing import List, Dict
from mashumaro.mixins.msgpack import DataClassMessagePackMixin
from mashumaro.codecs.json import JSONEncoder, JSONDecoder
from mashumaro.codecs.msgpack import MessagePackEncoder, MessagePackDecoder
import timeit


@dataclass
class InnerDataClass(DataClassMessagePackMixin):
    inner_attr1: int
    inner_attr2: str

@dataclass
class MyDataClassWithMsgPackMixin(DataClassMessagePackMixin):
    attr1: int
    attr2: str
    attr3: float
    attr4: bool
    attr5: List[int]
    attr6: InnerDataClass
    attr7: str
    attr8: float
    attr9: Dict[str, int]
    attr10: bool

@dataclass
class MyDataClass:
    attr1: int
    attr2: str
    attr3: float
    attr4: bool
    attr5: List[int]
    attr6: InnerDataClass
    attr7: str
    attr8: float
    attr9: Dict[str, int]
    attr10: bool


# Initialize objects
dc_with_mixin = MyDataClassWithMsgPackMixin(
    attr1=1, attr2='2', attr3=3.0, attr4=True, attr5=[1, 2, 3],
    attr6=InnerDataClass(1, '2'), attr7='7', attr8=8.0,
    attr9={'9': 9}, attr10=True
)

dc = MyDataClass(
    attr1=1, attr2='2', attr3=3.0, attr4=True, attr5=[1, 2, 3],
    attr6=InnerDataClass(1, '2'), attr7='7', attr8=8.0,
    attr9={'9': 9}, attr10=True
)

new_dc = type(MyDataClass.__name__ + "WithMessagePackMixin", (MyDataClass, DataClassMessagePackMixin), {})
encoder = JSONEncoder(MyDataClass)
decoder = JSONDecoder(MyDataClass)

# Serialize the object first to test decoding
msgpack_bytes_mixin = dc_with_mixin.to_msgpack()
msgpack_bytes_dynamic = new_dc(**{field.name: getattr(dc_with_mixin, field.name) for field in fields(dc_with_mixin)}).to_msgpack()
msgpack_bytes_json = msgpack.dumps(encoder.encode(dc))

# Test function 1: Decode using msgpack mixin
def test_msgpack_mixin_decode():
    return MyDataClassWithMsgPackMixin.from_msgpack(msgpack_bytes_mixin)

# Test function 2: Dynamically create a class and decode using msgpack mixin
def test_dynamic_msgpack_mixin_decode():
    new_instance = new_dc(**{field.name: getattr(dc_with_mixin, field.name) for field in fields(dc_with_mixin)})
    return new_instance.from_msgpack(msgpack_bytes_dynamic)

def test_create_new_dc():
    return type(MyDataClass.__name__ + "WithMessagePackMixin", (MyDataClass, DataClassMessagePackMixin), {})

# Test function 3: Convert msgpack bytes to JSON string and decode back to dataclass
def test_json_str_to_msgpack_decode():
    return decoder.decode(msgpack.loads(msgpack_bytes_json))
def test_create_decoder():
    return JSONDecoder(MyDataClass)
# Test function 4: Use MessagePackDecoder from mashumaro
def test_create_mashumaro_decoder():
    return MessagePackDecoder(MyDataClass)
# Initialize mashumaro decoder outside the test loop
mashumaro_decoder = test_create_mashumaro_decoder()
def test_mashumaro_msgpack_decoder():
    return mashumaro_decoder.decode(msgpack_bytes_mixin)

# Test function for initializing the mashumaro MessagePackDecoder


# Use timeit to test each method's performance, running 10000 times
iterations = 10000
time_mixin_decode = timeit.timeit(test_msgpack_mixin_decode, number=iterations)
time_dynamic_decode = timeit.timeit(test_dynamic_msgpack_mixin_decode, number=iterations)
time_dynamic_create_base_class = timeit.timeit(test_create_new_dc, number=1)
time_json_decode = timeit.timeit(test_json_str_to_msgpack_decode, number=iterations)
time_create_json_decoder = timeit.timeit(test_create_decoder, number=1)
time_mashumaro_decode = timeit.timeit(test_mashumaro_msgpack_decoder, number=iterations)
time_create_mashumaro_decoder = timeit.timeit(test_create_mashumaro_decoder, number=1)

# Output the results
print(f'{iterations} times dataclass with msgpack mixin decode elapsed time:', time_mixin_decode)
print(f'{iterations} times dynamically create msgpack mixin decode elapsed time:', time_dynamic_decode + time_dynamic_create_base_class)
print(f'{iterations} times dataclass -> json str -> msgpack bytes decode elapsed time:', time_json_decode + time_create_json_decoder)
print(f'{iterations} times mashumaro msgpack decode elapsed time:', time_mashumaro_decode + time_create_mashumaro_decoder)

import matplotlib.pyplot as plt

# Labels and timings for the four methods
methods = ['Msgpack Mixin Decode', 'Dynamic Msgpack Mixin Decode', 'JSON String to Msgpack Decode', 'Mashumaro Msgpack Decode']
times = [time_mixin_decode, time_dynamic_decode + time_dynamic_create_base_class, time_json_decode + time_create_json_decoder, time_mashumaro_decode + time_create_mashumaro_decoder]

# Plotting the comparison
plt.figure(figsize=(30, 5))
plt.bar(methods, times, color=['blue', 'green', 'red', 'orange'])
plt.xlabel('Methods')
plt.ylabel('Elapsed Time (seconds)')
plt.title('Comparison of Decoding Methods')
plt.savefig('/Users/future-outlier/code/dev/flytekit/build/PR/JSON/eduardo/performance_comparison/decode.png')

```

Future-Outlier:
Building a `JSON schema` is not an issue for the `Dataclass Transformer`.

Whether you have a pure dataclass or one that incorporates DataClassMessagePackMixin, DataClassJSONMixin, or both, it will successfully generate a JSON schema.

You can verify this by running the following example:

```python
import os
import tempfile
from flytekit.core.type_engine import TypeEngine
import pandas as pd
from flytekit.types.structured import StructuredDataset
from mashumaro.mixins.json import DataClassJSONMixin
from mashumaro.mixins.msgpack import DataClassMessagePackMixin

@dataclass
class Datum:
    x: int
    y: str
    z: dict[int, str]

@dataclass
class FlyteTypes:
    dataframe: StructuredDataset
    file: FlyteFile
    directory: FlyteDirectory

transformer = TypeEngine.get_transformer(FlyteTypes)
lt1_1 = transformer.get_literal_type(Datum)
lt1_2 = transformer.get_literal_type(FlyteTypes)

@dataclass
class Datum(DataClassMessagePackMixin, DataClassJSONMixin):
    x: int
    y: str
    z: dict[int, str]

@dataclass
class FlyteTypes(DataClassMessagePackMixin, DataClassJSONMixin):
    dataframe: StructuredDataset
    file: FlyteFile
    directory: FlyteDirectory

lt2_1 = transformer.get_literal_type(Datum)
lt2_2 = transformer.get_literal_type(FlyteTypes)
@dataclass
class Datum(DataClassJSONMixin):
    x: int
    y: str
    z: dict[int, str]

@dataclass
class FlyteTypes(DataClassJSONMixin):
    dataframe: StructuredDataset
    file: FlyteFile
    directory: FlyteDirectory

lt3_1 = transformer.get_literal_type(Datum)
lt3_2 = transformer.get_literal_type(FlyteTypes)

@dataclass
class Datum(DataClassMessagePackMixin):
    x: int
    y: str
    z: dict[int, str]

@dataclass
class FlyteTypes(DataClassMessagePackMixin):
    dataframe: StructuredDataset
    file: FlyteFile
    directory: FlyteDirectory

lt4_1 = transformer.get_literal_type(Datum)
lt4_2 = transformer.get_literal_type(FlyteTypes)

assert lt1_1 == lt2_1 and lt1_1 == lt3_1 and lt1_1 == lt4_1
assert lt1_2 == lt2_2 and lt1_2 == lt3_2 and lt1_2 == lt4_2
```

Future-Outlier:
I also found that directly encode by `msgpack.dumps` is the best way to do it.
## terminal output
```python
MsgPack serialized data size: 4 bytes
MsgPack serialized JSON string size: 9 bytes
UTF-8 encoded JSON string size: 8 bytes
```
## code example
```
import msgpack
import json

# Step 1: Serialize dictionary using msgpack
d = {'a': 1}
msgpack_data = msgpack.dumps(d)
msgpack_data_size = len(msgpack_data)

# Step 2: Convert dictionary to JSON string and serialize using msgpack
json_str = json.dumps(d)
msgpack_json_data = msgpack.dumps(json_str)
msgpack_json_data_size = len(msgpack_json_data)

# Step 3: Encode the JSON string using UTF-8 and compare size
encoded_json_str = json_str.encode('utf-8')
encoded_json_str_size = len(encoded_json_str)

# Print size comparison
print(f"MsgPack serialized data size: {msgpack_data_size} bytes")
print(f"MsgPack serialized JSON string size: {msgpack_json_data_size} bytes")
print(f"UTF-8 encoded JSON string size: {encoded_json_str_size} bytes")
```

Future-Outlier:
> I also found that directly encoding by `msgpack.dumps` is the best way to do it.
> 
> ## terminal output
> output 1
> 
> ```python
> MsgPack serialized data size: 15757449 bytes
> MsgPack serialized JSON string size: 21777797 bytes
> UTF-8 encoded JSON string size: 21777792 bytes
> ```
> 
> output 2 image
> 
> ## code example
> output 1
> 
> ```python
> import msgpack
> import json
> 
> # Step 1: Create a dictionary with 100000 key-value pairs using a for loop
> d = {f'key_{i}': i for i in range(1, 1000001)}
> 
> # Step 2: Serialize dictionary using msgpack
> msgpack_data = msgpack.dumps(d)
> msgpack_data_size = len(msgpack_data)
> 
> # Step 3: Convert dictionary to JSON string and serialize using msgpack
> json_str = json.dumps(d)
> msgpack_json_data = msgpack.dumps(json_str)
> msgpack_json_data_size = len(msgpack_json_data)
> 
> # Step 4: Encode the JSON string using UTF-8 and compare size
> encoded_json_str = json_str.encode('utf-8')
> encoded_json_str_size = len(encoded_json_str)
> 
> # Print size comparison
> print(f"MsgPack serialized data size: {msgpack_data_size} bytes")
> print(f"MsgPack serialized JSON string size: {msgpack_json_data_size} bytes")
> print(f"UTF-8 encoded JSON string size: {encoded_json_str_size} bytes")
> ```
> 
> output 2
> 
> ```python
> import msgpack
> import json
> import matplotlib.pyplot as plt
> 
> # Function to measure sizes for given number of attributes
> def measure_sizes(n):
>     # Create a dictionary with n key-value pairs
>     d = {f'key_{i}': i for i in range(1, n+1)}
> 
>     # Serialize dictionary using msgpack
>     msgpack_data = msgpack.dumps(d)
>     msgpack_data_size = len(msgpack_data)
> 
>     # Convert dictionary to JSON string and serialize using msgpack
>     json_str = json.dumps(d)
>     msgpack_json_data = msgpack.dumps(json_str)
>     msgpack_json_data_size = len(msgpack_json_data)
> 
>     # Encode the JSON string using UTF-8
>     encoded_json_str = json_str.encode('utf-8')
>     encoded_json_str_size = len(encoded_json_str)
> 
>     return msgpack_data_size, msgpack_json_data_size, encoded_json_str_size
> 
> # Define test points for 100000, 200000, ..., 1000000 attributes
> num_attributes = [100000 * i for i in range(1, 11)]
> msgpack_sizes = []
> msgpack_json_sizes = []
> utf8_json_sizes = []
> 
> # Measure sizes for each point
> for n in num_attributes:
>     msgpack_size, msgpack_json_size, utf8_json_size = measure_sizes(n)
>     msgpack_sizes.append(msgpack_size)
>     msgpack_json_sizes.append(msgpack_json_size)
>     utf8_json_sizes.append(utf8_json_size)
> 
> # Plot the results
> plt.plot(num_attributes, msgpack_sizes, label='MsgPack serialized data size')
> plt.plot(num_attributes, msgpack_json_sizes, label='MsgPack serialized JSON string size')
> plt.plot(num_attributes, utf8_json_sizes, label='UTF-8 encoded JSON string size')
> 
> plt.xlabel('Number of attributes')
> plt.ylabel('Size (bytes)')
> plt.title('Comparison of Serialized Data Sizes')
> plt.legend()
> plt.grid(True)
> plt.show()
> ```

The official msgpack documentation says: "MessagePack is an efficient binary serialization format. It lets you exchange data among multiple languages like JSON. But it's faster and smaller. Small integers are encoded into a single byte, and typical short strings require only one extra byte in addition to the strings themselves."

I also found that if we use `msgpack` to serialize strings, we can't get any benefit from it.

Future-Outlier:
# JSON IDL Discussion

## Prerequisite:
The idea is to store `bytes` in literal, so that we can serialize and deserialize it successfully.

There's 2 dimension we can think of, the JSON types and the way to do  serialization and deserialization.

### JSON types
Now we only support python's default JSON type, see [here](https://docs.python.org/3/library/json.html).
But there's [other JSON format](https://stackoverflow.com/questions/18692060/different-types-of-json) and each has some difference, maybe there will be user wants it in the future.

### serialization and deserialization APIs

We decided to use `msgpack` now, but there's also other API to do serialization and deserialization.

for example, `Zlib library` can compress the JSON data to more lighter, but also requires more time.

#### size
encode

#### speed
encode




## Discussion
We have 2 ways to implement it, either use a new IDL type or reuse the Binary IDL.
### Add a new IDL called JSON
```proto
message Json {
    bytes value = 1;
}
```
#### propeller compiler
check if literal scalar simple type is `JSON` or not.

#### propeller attribute access
we can use `TypeStructure` in the literal to provide metadata.
```proto
message TypeStructure {
    // Must exactly match for types to be castable
    string tag = 1;
    // dataclass_type only exists for dataclasses.
    // This is used to resolve the type of the fields of dataclass
    // The key is the field name, and the value is the literal type of the field
    // e.g. For dataclass Foo, with fields a, and a is a string
    // Foo.a will be resolved as a literal type of string from dataclass_type
    map<string, LiteralType> dataclass_type = 2;
}
```
if `TypeStructure.tag == "json-msgpack", then we deserialize it using msgpack api and to json type.

#### flytectl
check if the type is `scalar.JSON`

#### flyteconsole
check if the type is (`scalar.JSON` and `TypeStructure.tag has postfix "msgpack"`)


### Reuse BINARY IDL
```proto
message Binary {
    bytes value = 1;
    string tag = 2;
}
```


#### propeller compiler
check if literal scalar simple type is (`Binary` and `TypeStructure.tag == "json-msgpack"`) or not.

#### propeller attribute access
The same as `new IDL called JSON`.

#### flytectl
check if the type is (`scalar.BINARY` and `TypeStructure.tag has prefix "json"`)

#### flyteconsole
check if the type is (`scalar.BINARY` and `TypeStructure.tag has prefix "json" and postfix "msgpack"`)

fg91:
I have a preference for the "type alias" JSOn (why not Messagepack?) over reusing Binary as we'd avoid comparing string tags.

Future-Outlier:
> I have a preference for the "type alias" JSON (why not call it Messagepack?) over reusing Binary as we'd avoid comparing string tags.

I have 1 idea, but implementing, will show all of you today, and thank for your extra work in the weekend, really appreciate that.

Future-Outlier:
> For these reasons it was thought (again, correct me if i'm wrong) that this would be a powerful abstraction. A couple things that might be enabled by a stronger json notion.

I agree that having a good abstraction layer in Flytekit would be perfect.
The current implementation might be simple and useful,
but it doesn't exactly align with what the original issue describes.

Future-Outlier:
> We can somehow instruct flytekit that the incoming data is of JSON form, rather than flyteidl proto binary form. And sure we can pick some encoding (utf-8, msgpack, doesn't matter).

Not entirely understood this case, if possible, someone please explain it, thank you.

Future-Outlier:
> Writing this here so others can comment. @Future-Outlier and I had a chat today about the most recent draft. I think we've strayed a bit far from the intent of the original issue.
> 
> The current implementation now doesn't really consider the JSON component very much at all. I'm basing this off of the conversation we had but it's also summarized by the first to literal block in the summary...
> 
> > `Python Val -> Bytes -> Protobuf JSON`
> 
> In the original issue, the JSON component is the primary motivator, and msgpack was only picked because it is a more efficient transport mechanism for bytes across the wire than say, utf-8, or some other normal encoding format. (it's also understandable by browsers, unlike a generic compression format.) Now the json component seems to be an afterthought with msgpack being the primary motivator for this work. As far as I know (please correct me if I'm wrong), this is not the intent. If this was switched and intentionally so, could someone please add the context here?
> 
> If I understand correctly, the genesis of this project was the realization that json
> 
> * is a format that gets around a lot of the struct issues
> * is something natively supported by popular libraries like dataclass and pydantic
> * is a general enough format that we can write a general transformer for it in the future, thus enabling anything that has a `to_json`/`from_json` function to seamlessly integrate with flyte.
> * is something that both back and front-end can understand
> 
> For these reasons it was thought (again, correct me if i'm wrong) that this would be a powerful abstraction. A couple things that might be enabled by a stronger json notion.
> 
> ### Compatibility
> This was covered back in the [issue with unions](https://github.com/flyteorg/flyte/issues/5489) and dataclasses, but the existence of a json schema allows for more useful type/compatibility checking I think.
> 
> ### Attribute access of dataclass-like structures
> The second part that I think is worth thinking about and improving upon we uncovered in our chat today. The data accessor featured added last year works well for native flyte idl container types (like lists and maps) but doesn't work with dataclasses. This example fails both subsequent tasks
> 
> ```python
> from dataclasses import dataclass
> from mashumaro.mixins.json import DataClassJSONMixin
> 
> from flytekit import task, workflow, WorkflowFailurePolicy
> from flytekit.types.file import FlyteFile
> 
> 
> @dataclass
> class MyDC(DataClassJSONMixin):
>     data: FlyteFile
>     number: int = 10
> 
> 
> @task
> def produce_data() -> MyDC:
>     return MyDC(data=FlyteFile(path=__file__))
> 
> 
> @task
> def print_file(input_file: FlyteFile):
>     with open(input_file, "r") as f:
>         print(f.read())
> 
> 
> @task
> def print_number(nn: int):
>     print(f"Number is {nn}")
> 
> 
> @workflow(failure_policy=WorkflowFailurePolicy.FAIL_AFTER_EXECUTABLE_NODES_COMPLETE)
> def wf():
>     my_dc = produce_data()
>     print_file(my_dc.data)
>     print_number(my_dc.number)
> ```
> 
> The errors look pretty similar for both tasks
> 
> _Error for `t1`_
> 
> ```
> Cannot convert from Flyte Serialized object (Literal):
>   scalar:
>     generic:
>       fields: {'path': string_value: "s3://my-s3-bucket/data/o5/ak52wd7h59fhhtbckjgf-n0- [...] to <class 'flytekit.types.file.file.FlyteFile'>
> ```
> 
> _Error for `t2`_
> 
> ```
>   Cannot convert literal Flyte Serialized object (Literal):
>   scalar:
>     primitive:
>       float_value: 10.0 to <class 'int'>
> ```
> 
> The second issue (the mismatch in float/int) will be addressed by the PR as it stands, but I don't think the second can be. The reason is that propeller is just grabbing the struct object that's pointed to by the `data` attribute, which is not a FlyteFile object... it's a generic struct. I am very much against teaching flytekit type transformers how to handle proto generic structs.
> 
> However I think teaching type transformers how to handle `to/from_json` is entirely valid. We have the Python type when `t1` kicks off, we know it's a FlyteFile, and so we can just parse the JSON that's pointed to by the attribute.
> 
> We can somehow instruct flytekit that the incoming data is of JSON form, rather than flyteidl proto binary form. And sure we can pick some encoding (utf-8, msgpack, doesn't matter).

@wild-endeavor @eapolinario @pingsutw 

Serialization: Python value -> `json_str` -> bytes -> JSON IDL

I think if we require everyone using `JSON IDL` to include a layer called `json_str`, it will be closer to the original idea. 
Additionally, I can create a base class called `JsonTransformer` and use it as the parent class for `DataclassTransformer`, `DictTransformer`, and `PydanticTransformer`.
This way, if we want to add other transformers in the future, it will be easier. What do you folks think?

Thank you.
If I didn't understand what Yee said, please correct me, thank you all so much.

Future-Outlier:
> > Writing this here so others can comment. @Future-Outlier and I had a chat today about the most recent draft. I think we've strayed a bit far from the intent of the original issue.
> > The current implementation now doesn't really consider the JSON component very much at all. I'm basing this off of the conversation we had but it's also summarized by the first to literal block in the summary...
> > > `Python Val -> Bytes -> Protobuf JSON`
> > 
> > 
> > In the original issue, the JSON component is the primary motivator, and msgpack was only picked because it is a more efficient transport mechanism for bytes across the wire than say, utf-8, or some other normal encoding format. (it's also understandable by browsers, unlike a generic compression format.) Now the json component seems to be an afterthought with msgpack being the primary motivator for this work. As far as I know (please correct me if I'm wrong), this is not the intent. If this was switched and intentionally so, could someone please add the context here?
> > If I understand correctly, the genesis of this project was the realization that json
> > 
> > * is a format that gets around a lot of the struct issues
> > * is something natively supported by popular libraries like dataclass and pydantic
> > * is a general enough format that we can write a general transformer for it in the future, thus enabling anything that has a `to_json`/`from_json` function to seamlessly integrate with flyte.
> > * is something that both back and front-end can understand
> > 
> > For these reasons it was thought (again, correct me if i'm wrong) that this would be a powerful abstraction. A couple things that might be enabled by a stronger json notion.
> > ### Compatibility
> > This was covered back in the [issue with unions](https://github.com/flyteorg/flyte/issues/5489) and dataclasses, but the existence of a json schema allows for more useful type/compatibility checking I think.
> > ### Attribute access of dataclass-like structures
> > The second part that I think is worth thinking about and improving upon we uncovered in our chat today. The data accessor featured added last year works well for native flyte idl container types (like lists and maps) but doesn't work with dataclasses. This example fails both subsequent tasks
> > ```python
> > from dataclasses import dataclass
> > from mashumaro.mixins.json import DataClassJSONMixin
> > 
> > from flytekit import task, workflow, WorkflowFailurePolicy
> > from flytekit.types.file import FlyteFile
> > 
> > 
> > @dataclass
> > class MyDC(DataClassJSONMixin):
> >     data: FlyteFile
> >     number: int = 10
> > 
> > 
> > @task
> > def produce_data() -> MyDC:
> >     return MyDC(data=FlyteFile(path=__file__))
> > 
> > 
> > @task
> > def print_file(input_file: FlyteFile):
> >     with open(input_file, "r") as f:
> >         print(f.read())
> > 
> > 
> > @task
> > def print_number(nn: int):
> >     print(f"Number is {nn}")
> > 
> > 
> > @workflow(failure_policy=WorkflowFailurePolicy.FAIL_AFTER_EXECUTABLE_NODES_COMPLETE)
> > def wf():
> >     my_dc = produce_data()
> >     print_file(my_dc.data)
> >     print_number(my_dc.number)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The errors look pretty similar for both tasks
> > _Error for `t1`_
> > ```
> > Cannot convert from Flyte Serialized object (Literal):
> >   scalar:
> >     generic:
> >       fields: {'path': string_value: "s3://my-s3-bucket/data/o5/ak52wd7h59fhhtbckjgf-n0- [...] to <class 'flytekit.types.file.file.FlyteFile'>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > _Error for `t2`_
> > ```
> >   Cannot convert literal Flyte Serialized object (Literal):
> >   scalar:
> >     primitive:
> >       float_value: 10.0 to <class 'int'>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The second issue (the mismatch in float/int) will be addressed by the PR as it stands, but I don't think the second can be. The reason is that propeller is just grabbing the struct object that's pointed to by the `data` attribute, which is not a FlyteFile object... it's a generic struct. I am very much against teaching flytekit type transformers how to handle proto generic structs.
> > However I think teaching type transformers how to handle `to/from_json` is entirely valid. We have the Python type when `t1` kicks off, we know it's a FlyteFile, and so we can just parse the JSON that's pointed to by the attribute.
> > We can somehow instruct flytekit that the incoming data is of JSON form, rather than flyteidl proto binary form. And sure we can pick some encoding (utf-8, msgpack, doesn't matter).
> 
> @wild-endeavor @eapolinario @pingsutw
> 
> Serialization: Python value -> `json_str` -> bytes -> JSON IDL
> 
> I think if we require everyone using `JSON IDL` to include a layer called `json_str`, it will be closer to the original idea. Additionally, I can create a base class called `JsonTransformer` and use it as the parent class for `DataclassTransformer`, `DictTransformer`, and `PydanticTransformer`. (`class JsonTransformer(TypeTransformer)` is required to implement 2 methods, `to_json` and `from_json`.) This way, if we want to add other transformers in the future, it will be easier. What do you folks think?
> 
> Thank you. If I didn't understand what Yee said, please correct me, thank you all so much.

To make the JSON IDL more useful for users, I think that if we require users to have a `json_str` layer, it will ensure that the object is 100% a JSON object. For example, if we use another serialization API to serialize our object, it can't guarantee that it's a JSON object. (Since MsgPack is a superset of JSON, it can also support `bytes`, `tuple`, and `set`.)

A feasible approach is as follows:
- Serialization: language value -> JSON string -> bytes -> JSON IDL
- Deserialization: JSON IDL -> bytes -> JSON string -> language value

EngHabu:
@Future-Outlier, @wild-endeavor and myself met to discuss the direction of this RFC. Here is what we landed on:
1. We will use msgpack to serialize python types directly.
2. The `LiteralType` will be just `SimpleType.STRUCT` like it's for DataClasses today. 
3. We will fill in `TypeStructure` in `LiteralType` to describe the type using Json Schema.
4. The `Literal` (serialization medium) will be `Literal.Scalar.Binary` with a `tag=msgpack` to allow runtime code in various languages to decode it.
5. The UI will need to be updated to be able to display the inputs/outputs in Binary msgpack format.
6. The launch form will behave correctly in that it will render the fields based on the TypeStructure. When it creates an execution it will, however, use a `Literal.scalar.generic` instead of `Literal.scalar.binary` which should be ok (from a system design perspective as it's just a different medium of passing the same information) and propeller/flytekit should be able to access it normally. It can, however, also be updated at some point in the future to serialize into msgpack to avoid loss of fidelity and give compaction benefits.

The rationale:
1. msgpack is more efficient to store data than JSON is. We have seen people generate really large dataclasses as outputs and storing json as a string is probably not the right thing.
2. msgpack library in Golang is lossless when you unmarshal and marshal data. Json marshaller unfortunately is lossy (`1.0 float` converts to `1 int`)
3. We do not need a new `LiteralType` since the "type" the user is defining isn't new. We are just introducing a more efficient (and more correct) way of maintaining the values-fidelity.
4. We do not need a new `Literal` since the `value` the user is storing is already supported as `Binary`

@wild-endeavor and @Future-Outlier will add more details about the implementation and flytekit changes needed...

cc @eapolinario

Future-Outlier:
Move to https://github.com/flyteorg/flyte/pull/5742

PR #5542, opened 2024-07-05
JSON IDL

Future-Outlier:
## Tracking issue

https://github.com/flyteorg/flyte/issues/5318

## Why are the changes needed?
1. solve more than 10+ issues due to the protobuf struct doesn't support int.
2. using msgpack bytes, which are smaller than JSON bytes, we can reduce storage size and improve I/O performance when interacting with database.

### flyteidl
- [x] add JSON IDL type

### flytepropeller
- [x] attribute access for remote execution
- [x] backward compatibility (compiler)
    upstreamType (Struct) -> downstreamType (Json)

### flytectl
- [x] flytectl create execution with inputs
- [x] flytectl get execution

## How was this patch tested?
unit tests and remote execution.
Example are from [here](https://github.com/flyteorg/flytekit/pull/2600)

####  [flytepropeller] attribute access
- dataclass example
image

#### [flytepropeller] backward compatible
- reference task 
upstream (struct) -> downstream (json)
image

- cache
upstream (struct) -> downstream (json)
image

#### [flytepropeller] others
dict example
image

#### [flytectl] create execution with inputs
```bash
(dev) future@outlier ~ % flytectl create execution --execFile build/PR/JSON/demo/flytectl/create_json_dataclass.yaml -p flytesnacks -d development

execution identifier project:"flytesnacks" domain:"development" name:"f6dee626da95445a8894"
```
image

```yaml
iamRoleARN: ""
inputs:
    input:
        a: 1
        b: 3.14
        c: example_string
        d:
            "1": 100
            "2": 200
        e:
            a: 1
            b: 3.14
envs: {}
kubeServiceAcct: ""
targetDomain: ""
targetProject: ""
task: dataclass_example.dataclass_task
version: x1QUSQo64Vd9FuIo7xBkvg
```

#### [flytectl] get task
```
(dev) future@outlier ~ % flytectl get task -d development -p flytesnacks dataclass_example.dataclass_task --execFile build/PR/JSON/demo/flytectl/get_json_dataclass.yaml --version x1QUSQo64Vd9FuIo7xBkvg
 ------------------------ ---------------------------------- ------------- -------- --------- -------------- ------------------- ----------------------------- 
| VERSION                | NAME                             | TYPE        | INPUTS | OUTPUTS | DISCOVERABLE | DISCOVERY VERSION | CREATED AT                  |
 ------------------------ ---------------------------------- ------------- -------- --------- -------------- ------------------- ----------------------------- 
[...]

PR #5612, opened 2024-08-01
[flytepropeller][compiler] Error Handling when Type is not found

Future-Outlier:
## Tracking issue
https://linear.app/unionai/issue/DX-948/flytepropeller-compiler-error-handling-when-type-is-not-found

When developing `JSON IDL`, I found that if I registered task to flytepropeller but flytepropeller didn't  support the corresponding `flyteidl`, flyteadmin will crash and the error message from flyteadmin doesn't have a clear error message.

#### flytekit
image

#### flytepropeller
image

### After
We should give a good error message and avoid crashing the propeller.

image

Improve error message when `inputType` is `nil`, which means flytepropeller didn't support that type.

## How was this patch tested?
single binary and unit test.

### Setup process
flytekit: https://github.com/flyteorg/flytekit/pull/2600
flyteidl: https://github.com/flyteorg/flyte/pull/5542
flyte: this branch
```python

@task()
def t5(input: dict) -> dict:
    print(input)
    print("different")
    return input.update({"b": 2}).update({"c": 3})

@workflow
def wf2(input: dict={'d': 1, }) -> dict:
    return t5(input=input)
```
```python
pyflyte run --remote build/PR/JSON/example.py wf2
```

### Screenshots
image

## Related PRs
https://github.com/flyteorg/flyte/pull/5542


flyteadmin/pkg/manager/impl/validation/validation.go
```diff
@@ -282,11 +283,26 @@ func validateParameterMap(inputMap *core.ParameterMap, fieldName string) error {
 			defaultValue := defaultInput.GetDefault()
 			if defaultValue != nil {
 				inputType := validators.LiteralTypeForLiteral(defaultValue)
+
+				if inputType == nil {
+					return errors.NewFlyteAdminErrorf(codes.InvalidArgument,
+						fmt.Sprintf(
+							"Flyte Propeller encountered an issue while determining\n"+
+								"the type of the default value for Parameter '%s' in '%s'.\n"+
+								"Registered type from FlyteKit: [%s].\n"+
```
pingsutw, line 292: The user may use flytekit-java instead of flytekit.

eapolinario:
@Future-Outlier , can you fix the [lint error](https://github.com/flyteorg/flyte/actions/runs/10193513817/job/28198271351?pr=5612)?

wild-endeavor:
@Future-Outlier just wondering does this fail in the same way for pyflyte register?  or is it just run?

Future-Outlier:
> @Future-Outlier just wondering does this fail in the same way for pyflyte register?  or is it just run?

Both failed
They will go through the same code piece

pingsutw (approved):

PR #5622, opened 2024-08-02
Fix nil pointer when task plugin load returns error

Sovietaced:
I accidentally misconfigured one of my plugins and the result was a nil pointer because `pluginCore.LoadPlugin` errored and returned a nil `cp`. 

This pull request updates the logic such that it checks the error before operating on `cp`.


Future-Outlier (approved):
LGTM, thank you so much

PR #5623, opened 2024-08-02
Log stack trace when refresh cache sync recovers from panic

Sovietaced:
This pull request improves the auto refresh cache's `sync` panic recovery logic such that it logs the stack trace for the panic that occurred. This makes it much easier to understand what exactly failed in the cache sync logic (ie. instead of  just something like "invalid memory address or nil pointer dereference")


Future-Outlier (approved):
thank you

PR #5637, opened 2024-08-06
Improve execution name readability

wayner0628:
- Improve execution name readability

- Add `EnableHumanHash` config in [flyteadmin/pkg/runtime/interfaces/application_configuration.go](https://github.com/flyteorg/flyte/pull/5637/commits/5f37b11138bdd32231151d40d794d30a0986b41f#diff-e4e11bcd1e36c8870c82a37fcab52de08d6de4911b1fcf83195cea4156f9a0bb)
- Add human hash feature in [flyteadmin/pkg/common/executions.go](https://github.com/flyteorg/flyte/pull/5637/commits/815f597d4f66aa56d25c8ce89fd482171e7c570c#diff-331f07e355b610d0581bd6dcc08b8f31bddcea598ecafeaab61836aa7b7c77bc)

## How was this patch tested?
unit testing and integration testing

### Setup process
To enable human-readable execution names, add the following configuration:

```yaml
flyteadmin:
  featureGates:
    enableHumanHash: true
```

### Screenshots
Screenshot 2024-08-11 at 9 55 04 PM

## Related PRs
https://github.com/flyteorg/flytekit/pull/2678


flytectl/cmd/create/execution_util.go
```diff
@@ -150,7 +153,12 @@ func recoverExecution(ctx context.Context, executionName string, project string,
 func createExecutionRequest(ID *core.Identifier, inputs *core.LiteralMap, envs *admin.Envs, securityContext *core.SecurityContext, authRole *admin.AuthRole, targetExecName string, targetExecutionCluster string) *admin.ExecutionCreateRequest {
 
 	if len(targetExecName) == 0 {
-		targetExecName = "f" + strings.ReplaceAll(uuid.New().String(), "-", "")[:19]
+		config := runtime.NewApplicationConfigurationProvider()
```
pingsutw, line 156: Let's update flyteadmin first. Ideally, the ID should only be generated from the backend instead of the client.

flyteadmin/pkg/common/executions.go
```diff
@@ -3,17 +3,24 @@ package common
...
+const ExecutionIDLengthLimit = 63
 const ExecutionStringFormat = "a%s"
 
 /* #nosec */
-func GetExecutionName(seed int64) string {
+func GetExecutionName(seed int64, enableHumanHash bool) string {
 	rand.Seed(seed)
+	if enableHumanHash {
+		hashKey := []byte(rand.String(ExecutionIDLength))
+		result, _ := humanhash.Humanize(hashKey, 3)
```
pingsutw, line 21: should we handle the error here?
```suggestion
		result, err := humanhash.Humanize(hashKey, 3)
```

flyteadmin/pkg/async/schedule/aws/workflow_executor.go
```diff
@@ -206,8 +215,10 @@ func (e *workflowExecutor) run() error {
 			}
 			continue
 		}
-		executionRequest := e.formulateExecutionCreateRequest(launchPlan, scheduledWorkflowExecutionRequest.KickoffTime)
-
+		executionRequest, err := e.formulateExecutionCreateRequest(launchPlan, scheduledWorkflowExecutionRequest.KickoffTime)
+		if err != nil {
+			continue
```
pingsutw, line 220: We should log an error here

flyteadmin/pkg/manager/impl/util/shared.go
```diff
@@ -15,17 +15,23 @@ import (
...
-		return request.Name
+		return request.Name, nil
 	}
-	return common.GetExecutionName(time.Now().UnixNano())
+	config := runtime.NewApplicationConfigurationProvider()
+	executionName, err := common.GetExecutionName(time.Now().UnixNano(), config.GetTopLevelConfig().FeatureGates.EnableHumanHash)
+	if err != nil {
+		return "", err
+	}
+	return executionName, nil
```
eapolinario, line 34: we shouldn't need this once we go back to the original signature for `GetExecutionName`.

flyteadmin/pkg/runtime/interfaces/application_configuration.go
```diff
@@ -50,6 +50,7 @@ type PostgresConfig struct {
 
 type FeatureGates struct {
 	EnableArtifacts bool `json:"enableArtifacts" pflag:",Enable artifacts feature."`
+	EnableHumanHash bool `json:"enableHumanHash" pflag:",Enable humanhash feature."`
```
eapolinario, line 53: Can we rename this to `EnableFriendlyNames`?

```suggestion
	EnableFriendlyNames bool `json:"enableFriendlyNames" pflag:",Enable generation of friendly execution names feature."`
```

flyteadmin/scheduler/executor/executor_impl.go
```diff
@@ -66,10 +68,22 @@ func (w *executor) Execute(ctx context.Context, scheduledTime time.Time, s model
...
+	config := runtime.NewApplicationConfigurationProvider()
+	if config.GetTopLevelConfig().FeatureGates.EnableHumanHash {
+		executionName, err = common.GetExecutionName(time.Now().UnixNano(), config.GetTopLevelConfig().FeatureGates.EnableHumanHash)
+		if err != nil {
+			logger.Errorf(ctx, "failed to generate execution name for schedule %+v due to %v", s, err)
+			return err
+		}
+	} else {
+		executionName = "f" + strings.ReplaceAll(executionIdentifier.String(), "-", "")[:19]
+	}
```
eapolinario, line 81: Are friendly names important for scheduled executions?

flyteadmin/pkg/common/executions.go
```diff
@@ -1,20 +1,33 @@
...
 const ExecutionStringFormat = "a%s"
 
 /* #nosec */
-func GetExecutionName(seed int64) string {
+func GetExecutionName(seed int64, enableHumanHash bool) (string, error) {
 	rand.Seed(seed)
-	return fmt.Sprintf(ExecutionStringFormat, rand.String(ExecutionIDLength-1))
+	if enableHumanHash {
+		hashKey := []byte(rand.String(ExecutionIDLength))
+		result, err := humanhash.Humanize(hashKey, 3)
```
eapolinario, line 23: IIUC, by setting this to 3 we get about 16M unique ids to choose from. The reason is that humanhash maintains a list of [unique 256 words](https://github.com/wolfeidau/humanhash/blob/master/human_hash.go#L28-L64), so 256^3 = 16.8M. 

We should have a conversation about whether this is enough for most big deployments. My gut feeling is that it's not sufficient. How about we extend this to 4? If we do this and end up choosing the largest word in that set 4 times we're still under the 63, which is the max char limit for an execution id. Out of curiosity, the largest word in that dataset is "massachusetts" at 13 characters, so that would be 55 characters (13*4 + 3 to account for the separators).

wayner0628:
Hello @pingsutw @eapolinario,

I've made the following changes based on your feedback:

1. Removed all unnecessary error handlers.
2. Added a comment explaining why errors in humanhash generation can be safely ignored.
3. Updated the humanhash package version to `v1.1.0`.
4. Renamed the configuration setting to `EnableFriendlyNames`.
5. Extend the name to `4` subwords.
6. Now reading the config directly within the `getExecutionName` function, and adjusted the file path accordingly.
7. Updated the unit tests to use mocking objects, as the config is now read within the function.

Please review these changes and let me know if you have any suggestions or concerns.

Thank you!

wayner0628:
Hi @pingsutw @eapolinario,

I've thoroughly investigated the issue but couldn't find any connection between my changes and the CI unit test failure. Additionally, there are no error messages logged that could provide further insight. Could you please help retrigger the CI build? Thank you!

pingsutw (approved):
I just reverted the flytectl changes. It seems like we can't change that in the same PR. You could create a follow-up PR to update flytectl.

wayner0628:
> I just reverted the flytectl changes. It seems like we can't change that in the same PR. You could create a follow-up PR to update flytectl.

I noticed some issues still persist, so I'll continue checking them.

pingsutw (approved):

pingsutw:
@ddl-ebrown, we could make `suffix` opt-in. Need to add one more field to the config map. something like
```
flyteadmin
  featureGates:
    enableFriendlyNames: true
  execution_name_with_random_suffix: true
```

Feel free to open a PR!

kumare3:
This is pretty cool - cc @katrogan

PR #5103, opened 2024-03-25
[RFC] Offloaded Raw Literals

wild-endeavor:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/<number>_

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

PR #5653, opened 2024-08-12
RFC into RFC

katrogan:

Updates draft RFC here: https://github.com/flyteorg/flyte/pull/5103


rfc/system/5103-offloaded-literal.md
```diff
@@ -4,32 +4,77 @@
...
+        LiteralCollection collection = 2;
+        // A map of strings to literals.
+        LiteralMap map = 3;
+    }
+    ...
+    // ** new below this line **
+    // If this literal is offloaded, this field will contain metadata including the offload location.
+    string uri = 6;
+    // Includes information about the size of the literal.
+    uint64 size_bytes = 7;
```
wild-endeavor, line 51: why is size important again?  I mean there's other metadata as well (etag information). The assumption here is that size is super important so we want to be able to show that without making a head call?

rfc/system/5103-offloaded-literal.md
```diff
@@ -4,32 +4,77 @@
...
+Once offloading is enabled in the deployment config, flytepropeller can read from the [RuntimeMetadata](https://github.com/flyteorg/flyte/blob/f448a0358d8706a09b65b96543134f629327d755/flyteidl/protos/flyteidl/core/tasks.proto#L71-L87) in the task config to determine the SDK version.
+
+When writing outputs in the [remote_file_output_writer](https://github.com/flyteorg/flyte/blob/2ca31119d6b9258661a71f38e450f93b6692402c/flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_writer.go#L56-L84) the source code should detect whether the literal size exceeds the configured minimum and
+- if the task is using a newer SDK version that supports reading offloaded literals, offload the literal to the configured storage backend and update the literal with the offload URI and size.
+- if the task is using an older SDK version that doesn't support offloaded literals, fail the task with an error message indicating that the task output is too large and the user should update their SDK version.
```
wild-endeavor, line 61: let's add a line to say downstream tasks also have to upgraded?  that is, if you have a reference/remote task downstream that consumes the map task output, but it hasn't been updated, then it'll fail.

rfc/system/5103-offloaded-literal.md
```diff
@@ -4,32 +4,77 @@
...
 
 ### 3.3 Flytekit & Copilot
 Flytekit and Copilot will both need to detect that a Literal has been offloaded and know to download it.
+- in Flytekit this can be done by checking the `uri` field in the Literal message when converting a literal [to_python_value](https://github.com/flyteorg/flytekit/blob/e394af0be9f904fbf8be675eaa8b8cdc24311ced/flytekit/core/type_engine.py#L1134)
+- in Copilot, the data downloader [literal handling](https://github.com/flyteorg/flyte/blob/5f4199899922ca63f7690c82dfca42a783db64c3/flytecopilot/data/download.go#L219-L248) should fetch the value
 
-For large outputs (like large maps of large dataclasses), Flytekit should also know how to offload the data. This should be done transparently to the user. How will propeller know to fail though if propeller hasn't been updated?
+As a follow-up, we can also implement literal offloading in the SDK for conventional python tasks. Flytekit should also know how to offload the data. This should be done transparently to the user.
```
wild-endeavor, line 68: should we cover that here?  or is this only for map tasks for now? for the general case, we were going to go with the solution of propeller setting an environment variable that turns on offloading on the flytekit side?

for erroring if things get too big, i don't know that there's a solution.  We should just add a size limit asap in flytekit right @eapolinario?  Some env var based setting with a 10MB default.  If a literal is more than 10MBs then error.  considering we don't know when we'll get to the general case, by the time we do, most users might've already upgraded.

wild-endeavor (approved):
thanks for the updates!

rfc/system/5103-offloaded-literal.md
katrogan: general consensus seemed to be this is useful, I guess it's nice for clients who want to decide whether to pull massive datasets?

rfc/system/5103-offloaded-literal.md
katrogan: I think we're leaning towards not tackling the implementation bits for this proposal but I think it's okay to cover future work here?

katrogan:
thanks for the review @wild-endeavor, updated - mind taking another look?

wild-endeavor (approved):

Issue #5125, opened 2024-03-27
[Core feature] Workflow level throttling


a4501150:
### Motivation: Why do you think this is important?

Currently flyte does not support workflow level throttling that limits the `same` concurrent workflows that can be running under a flyte project / domain. (same workflow means the same binary identified by workflow name).

In our use case, we use flyte both for workflow scheduling and orchestration. One of our workflow will hit our backend API heavily and we want to limit the resource pressure on our backend API. Although we can implement throttling on backend API but backend API will also use in synchronized manner and we don't want to affect the current synchronized API performance and behavior, thus implement throttling on workflow side is the best choice.

Compare to other workflow engines, this is the major feature that lacks in flyte.

### Goal: What should the final outcome look like, ideally?

For instance, when we trigger a workflow by creating a `CreateExecution` with a launch plan (preferred) OR when we register a workflow, we should be able to configure an extra parameter, eg. `max_concurrent_execution` as type long which can limit the concurrent workflows that running at the same time.

For example, let's say we have a `workflow A`, with a `Task A`. We created a launch plan for this workflow with `max_concurrent_execution` as 5.

- We trigger the concurrent workflows by creating 100 `CreateExecution` requests with same launch plan to flyte grpc API
- The first 5 workflow executions should be running
- The rest of 95 executions should be put in a queue at flyte backend, in status of `QUEUED`
- The ultimate goal is there should be only `max_concurrent_execution` workflows in status of `RUNNING` at the same time
- The execution id should be always returned

### Describe alternatives you've considered 

We are now using a hacky way by creating a `resource lock` task to achieve the functionalities:

- A new task with `for loop` and `Thread.sleep()` inside the for loop along with querying the flyte endpoint to get the top `max_concurrent_execution` running executions sorted by execution created time.
- If the returned top X running executions count >= `max_concurrent_execution`, and the execution id obtained by the the task via ENV `FLYTE_INTERNAL_EXECUTION_ID` is not one of the returned top X executions, then the task will just wait there in the for loop with `thread.sleep()`

This effectively limit the concurrent executions of a given launchPlan (workflow) by have the throttling task as the first task in the given workflow DAG.

However, this will create many semi-dangling `resource lock` tasks if the amount of concurrent execution requests is large and `max_concurrent_execution` is relatively small.

### Another alternatives solution

Task level throttling can further enhance the fine control level, however this maybe harder to implement than workflow level throttling.

### Are you sure this issue hasn't been raised already?

- [X] Yes

[...]

Issue #420, opened 2020-07-20
[Feature] Support serializing Scheduled Executions


EngHabu:
**Motivation: Why do you think this is important?**
With a scheduled executions, users might expect no more than one execution will be running at the same time (assuming an older run takes a long time to finish).

**Goal: What should the final outcome look like, ideally?**
Users can signify exactly once flag when creating a schedule for a launch plan.

**Flyte component**
- [x] Overall
- [ ] Flyte Setup and Installation scripts
- [ ] Flyte Documentation
- [ ] Flyte communication (slack/email etc)
- [ ] FlytePropeller
- [ ] FlyteIDL (Flyte specification language)
- [ ] Flytekit (Python SDK)
- [ ] FlyteAdmin (Control Plane service)
- [ ] FlytePlugins
- [ ] DataCatalog 
- [ ] FlyteStdlib (common libraries)
- [ ] FlyteConsole (UI)
- [ ] Other

kumare3:
Exactly once means that something will be done exactly once, not twice and not zero times. @EngHabu  Do you rather mean that only one scheduled execution runs at a time effectively serial iz Ing the schedules?

EngHabu:
Yes exactly that

kumare3:
Cc @EngHabu this should not be closed

kdhingra307:
@kumare3 any possibility of making this happen?

kumare3:
This is on the roadmap, cannot be prioritized right now as short staffed. Please contribute!

kumare3:
Cc @EngHabu @eapolinario fyi

kumare3:
@kdhingra307 please share a motivating example / usecase

kdhingra307:
@kumare3 One use-case could involve processing a batch of data from a stream. For example, if we are at index 10 and want to process the next 500 samples, we increase the counter once the workflow is complete. However, if the workflow takes longer to finish than the frequency at which new workflows are launched, another workflow might start and process the same data.


Also, i can try working on this

kdhingra307:
@kumare3 i feel the logic to implement skip feature would be relatively easier. We can add a check in [PrepareFlyteWorkflow](https://github.com/flyteorg/flyte/blob/00d0a816c66141b9d9066ec4c779caa16c688467/flyteadmin/pkg/workflowengine/impl/prepare_execution.go#L114) function and check if a copy of same workflow is running or not. I did a small poc locally and it seems to be working fine.

```go

[...]

Issue #267, opened 2020-04-14
[Feature] Prevent concurrent execution 


art4ul:
**Motivation: Why do you think this is important?**
In my project I use scheduled launch plan to process a dataset each 1 hour. But sometimes the workflow may process the data more the 1 hour. In this case the scheduler creates the new concurrent execution of the launch plan. But the concurrent execution of the job is unacceptable for my use case (I get dublicated output). 

**Goal: What should the final outcome look like, ideally?**
I think it would be nice to have an option for launch plan to allow concurrent execution  or not.

**Describe alternatives you've considered**
The only one option fo me to start implementing some distributed locks using some external systems like Zookeeper 

**Flyte component**
- [ ] Overall
- [ ] Flyte Setup and Installation scripts
- [ ] Flyte Documentation
- [ ] Flyte communication (slack/email etc)
- [ ] FlytePropeller
- [ ] FlyteIDL (Flyte specification language)
- [x] Flytekit (Python SDK)
- [ ] FlyteAdmin (Control Plane service)
- [ ] FlytePlugins
- [ ] DataCatalog 
- [ ] FlyteStdlib (common libraries)
- [x] FlyteConsole (UI)
- [ ] Other

**Is this a blocker for you to adopt Flyte**
Yes

kumare3:
Let me preface by saying that, the feature you are asking does make sense, its just that we need to think about how long should we keep waiting, is there a timeout?
@art4ul  on the other hand, for your problem Ideally you want to prevent side-effects in Flyte Workflows. So why do you not lock the dataset that will be loaded by the execution every 1 hour, and thus the subsequent workflows will always work on independent units. This also gives you complete replayability.

Lets take an example, you have a dataset that materializes at some frequency. 
datasets, d1( time x), d2( time x+delta)....
Your workflow runs every hour, at time t.

You could structure your workflow like this
```
Task1: lock-dataset(kickoff-time, execution_window="1 hour") -> []dataset_references
    collect_datasets for times (t - execution_window)
    return collected_datasets

Task2: process-dataset(refs []dataset_references)
     process-dataset(refs)

outputs...

```

Thus you can see, you are always locking the datasets so no matter when you run, but if you run with the right kickoff-time, you will end up with the right datasets to be processed

art4ul:
Thank you @kumare3 for reaching out!
 
> Let me preface by saying that, the feature you are asking does make sense, its just that we need to think about how long should we keep waiting, is there a timeout?

[...]

PR #5659, opened 2024-08-13
RFC: Add workflow execution concurrency

katrogan:
## Related Issues
https://github.com/flyteorg/flyte/issues/5125
https://github.com/flyteorg/flyte/issues/420
https://github.com/flyteorg/flyte/issues/267


rfc/system/RFC-0000-execution-concurrency.md
```diff
@@ -0,0 +1,165 @@
...
+   1. open question: also update its phase in the database to `QUEUED`?
+   1. let execution proceed
+
+We should consider adding an index to the executions table to include
+- launch_plan_id
+- phase
+- created_at
+
+#### Open Questions
+- Should we always attempt to schedule pending executions in ascending order of creation time?
```
pingsutw, line 101: Maybe make it configurable? FIFO, FILO

katrogan: yeah I wasn't sure!  any suggestions here? we could introduce an enum and choose fifo to begin with and expand support

granthamtaylor: I have mixed thoughts on making the queue's order of execution configurable.

If we support a limited number of parallel executions (more than 1), the order of these executions would naturally start as FIFO up until that limit is reached.

To me, providing an option to begin executing FILO after that limit is reached feels a little strange.

However, that brings a different question to mind: If multiple workflows are queued up, should we provide an option to enable loud notifications?

In other words, if backlogged executions have the possibility of impacting downstream operations, can we enable users to receive loud notifications, including the number of queued executions?

I can imagine a use case where: holiday shopping -> increased purchase volume -> increased data size -> multiple, consecutive execution delays -> cascading backlog of executions. In this scenario, the owners of the workflow may be out on leave and not be aware of the growing backlog.

katrogan: interesting, we have workflow notifications enabled for terminal state but we've talked more about richer, customizable notifications and I think this slates neatly into that

I think for a v1 having the default behavior be fifo with an extended description/explanation for the pending state may provide some visibility here to start off with

rfc/system/RFC-0000-execution-concurrency.md
```diff
@@ -0,0 +1,165 @@
...
+  string description = X;
+  
+  
+}
+
+// Can also add to ExecutionSpec to specify execution time overrides
+
+```
+
+### FlyteAdmin
```
eapolinario, line 79: During last week's contributors meeting someone asked a question about having this concurrency control work across versions. Can we either have a discussion in this PR about it or list that use case as not being supported explicitly in the RFC?

rfc/system/RFC-0000-execution-concurrency.md
```diff
@@ -0,0 +1,165 @@
...
+At a broad level
+1. At CreateExecution time, if the launch plan in the ExecutionSpec has a concurrency policy
+   1. Create the execution in the database with a new `PENDING` execution phase and reason populated in `ExecutionStateChangeDetails`.
+      1. or fail the request when the concurrency policy is set to `ABORT`
+   1. Do not create the workflow CRD
+
+Introduce an async reconciliation loop in FlyteAdmin to poll for all pending executions:
```
eapolinario, line 86: Do we have prior art for this kind of reconciliation loop in flyteadmin?

rfc/system/RFC-0000-execution-concurrency.md
```diff
@@ -0,0 +1,165 @@
...
+1. Query all pending executions by timestamp ascending (open question, should we prefer more recent executions instead? should we make this configurable?)
+   1. as an optimization, could even parallelize this into goroutines, one per distinct launch plan ID that has any `PENDING` execution
+2. Check the database to see if there are fewer than `MAX_CONCURRENCY` non-terminal executions with an identical launch plan ID
+3. If there are fewer than `MAX_CONCURRENCY` executions running, select the oldest pending execution for that launch plan
+   1. create the workflow CRD
+   1. open question: also update its phase in the database to `QUEUED`?
+   1. let execution proceed
+
+We should consider adding an index to the executions table to include
```
eapolinario, line 95: What do you think of adding a section describing the proposed solution in more details?

rfc/system/RFC-0000-execution-concurrency.md
eapolinario: Can we add this suggestion of having an enum listing the policies to the Implementation details section?

rfc/system/RFC-0000-execution-concurrency.md
```diff
@@ -0,0 +1,165 @@
...
+- phase
+- created_at
+
+#### Open Questions
+- Should we always attempt to schedule pending executions in ascending order of creation time?
+- Should we propagate concurrency policies to child executions?
+
+## 4 Metrics & Dashboards
+
+*What are the main metrics we should be measuring? For example, when interacting with an external system, it might be the external system latency. When adding a new table, how fast would it fill up?*
```
eapolinario, line 106: How is this feature going to be rolled out? Should we have an explicit list of metrics used to help the health of the feature? (e.g. total number of attempts of a given launchplan )

rfc/system/RFC-0000-execution-concurrency.md
```diff
@@ -0,0 +1,165 @@
...
+
+## 5 Drawbacks
+
+*Are there any reasons why we should not do this? Here we aim to evaluate risk and check ourselves.*
```
eapolinario, line 110: Do we have any reservations about more load on the DB (even with indexes, etc)?

eapolinario (commented):
This is looking pretty good. I'd feel more comfortable if we fleshed out the implementation a bit more, but otherwise, I feel like we're on the same page.

katrogan:
> I'd feel more comfortable if we fleshed out the implementation a bit more, but otherwise, I feel like we're on the same page.

Sounds good, just wanted overall alignment before diving into the implementation. Will do that next and thank you already for all the feedback

rfc/system/RFC-0000-execution-concurrency.md
katrogan: Interesting question. I think scheduling attempts here is based on the polling interval right? But could be useful to understand time spent in PENDING

rfc/system/RFC-0000-execution-concurrency.md
katrogan: good point, we already have a ton of indices on executions - there is definitely a tradeoff to adding a new one

rfc/system/RFC-0000-execution-concurrency.md
katrogan: yes, the scheduler!

katrogan:
added some more implementation details, mind taking another look @eapolinario

rfc/system/RFC-0000-execution-concurrency.md
corleyma: I can say that something that works across versions would be really useful for us.

rfc/system/RFC-0000-execution-concurrency.md
```diff
@@ -0,0 +1,232 @@
...
+
+
+type ConcurrencyScheduling enum {
+  FIFO = 0;
+  FILO = 1;
+  ...
+}
+```
+
+Furthermore, we may want to introduce a max pending period to fail executions that have been in `PENDING` for too long
```
fg91, line 194: :+1: I agree that this would be good.

rfc/system/RFC-0000-execution-concurrency.md
```diff
@@ -0,0 +1,232 @@
...
+
+*Here, we aim to be mindful of our environment and generate empathy towards others who may be impacted by our decisions.*
+
+- *What other systems or teams are affected by this proposal?*
+- *How could this be exploited by malicious attackers?*
+
+## 8 Unresolved questions
+
+- Should we always attempt to schedule pending executions in ascending order of creation time?
+    - Decision: We'll use FIFO scheduling by default but can extend scheduling behavior with an enum going forward.
```
fg91, line 207: If this has been decided (I'm ok with it), could you please reformulate in the text above where this is still discussed as an open question? 🙏

rfc/system/RFC-0000-execution-concurrency.md
fg91, line 1: Can you please change the filename to include the PR number?

rfc/system/RFC-0000-execution-concurrency.md
fg91: For us too because we very often `pyflyte run` which means we often don't have two executions of the same version.

This could be made configurable here:

```py
 concurrency=Concurrency(
        max=1,  # defines how many executions with this launch plan can run in parallel
        policy=ConcurrencyPolicy.WAIT  # defines the policy to apply when the max concurrency is reached,
        level=ConcurrencyLevel.Version, # or ConcurrencyLevel.LaunchPlan
    )
```

corleyma:
I'd add one thing, possibly out of scope for this RFC: it would be really nice to be able to define a "max execution concurrency" on the backend, either propeller-wide or per project/domain.  Flyte would benefit from more controls that allow operators to protect quality of service and aren't dependent on workflow authors to set reasonable limits.

katrogan:
hi @corleyma thanks for reviewing! re your comment on platform-max execution concurrency, that's really intriguing - would you want to start a separate discussion on that here: https://github.com/flyteorg/flyte/discussions so we don't lose track of the suggestion?

execution namespace quota is meant to help address quality of service and fairness in a multitenant system but it would be cool to flesh out other mechanisms for managing overall executions

rfc/system/RFC-0000-execution-concurrency.md
fiedlerNr9: Can add a customers feedback here, where the desired behaviour is to actually replace (terminate) the current executions by subsequent executions. Sounds like too much for the initial scope but still interested if this would be possible to add later with the current approach?

rfc/system/RFC-0000-execution-concurrency.md
katrogan: done, thanks

rfc/system/RFC-0000-execution-concurrency.md
katrogan: updated the discussion above!

rfc/system/RFC-0000-execution-concurrency.md
katrogan: thanks @eapolinario @corleyma @fg91 for the feedback, I don't think this will be too much of a lift but added a proposal for different levels of precision here too

rfc/system/RFC-0000-execution-concurrency.md
katrogan: @fiedlerNr9 added a section under Alternatives. I don't think this is precluded by this implementation but not in scope atm

corleyma:
> execution namespace quota is meant to help address quality of service and fairness in a multitenant system but it would be cool to flesh out other mechanisms for managing overall executions

execution namespace quota can help protect against workloads that would otherwise utilize too many cluster resources, but it doesn't really help protect e.g. flyte propeller from too many concurrent executions.

I am happy to start a separate conversation though!

rfc/system/RFC-5659-execution-concurrency.md
```diff
@@ -0,0 +1,313 @@
...
+```protobuf
+message Concurrency {
+  // Defines how many executions with this launch plan can run in parallel
+  uint32 max = 1;
+  
+  // Defines how to handle the execution when the max concurrency is reached.
+  ConcurrencyPolicy policy = 2;
+}
+
+enum ConcurrencyPolicy {
```
nikp1172, line 49: we should have a replace option also?
To stop the previous execution and replace it with the current one. This is what k8s job does.

https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#concurrency-policy

Would be great if the behaviour can be made as close to this.

davidmirror-ops:
09/26/2024 Contributors sync notes: no updates

rfc/system/RFC-5659-execution-concurrency.md
katrogan: great point, updated

rfc/system/RFC-5659-execution-concurrency.md
```diff
@@ -195,6 +198,7 @@ WHERE  ( launch_plan_named_entity_id, created_at ) IN (SELECT launch_plan_named_
                                           GROUP  BY launch_plan_named_entity_id); 
 ```
 
+Note, in this proposal, registering a new version of the launch plan and setting it to active will determine the concurrency policy across all launch plan versions.
```
eapolinario, line 201: ❤️

kumare3:
I would not implement this with too many db calls. I would make the decision simple and completely in memory using Singletons like the scheduler. This is because 
1. We can easily update the algorithm of how to order, probably support backfilling etc
2. Create priorities in the future
3. Alleviate DB pressure and make it really scalable
image

rfc/system/RFC-5659-execution-concurrency.md
```diff
@@ -0,0 +1,317 @@
...
+```
+
+### FlyteAdmin
+At a broad level
+1. At CreateExecution time, if the launch plan in the ExecutionSpec has a concurrency policy
+   1. Create the execution in the database with a new `PENDING` execution phase and reason populated in `ExecutionStateChangeDetails`.
+      1. or fail the request when the concurrency policy is set to `ABORT`
+   1. Do not create the workflow CRD
+
+Introduce an async reconciliation loop in FlyteAdmin to poll for all pending executions:
```
kumare3, line 89: this has to be done very carefully. Lets use the controller pattern. Just simply run everything in memory on one machine. We have seen with scheduler and propeller this is stable and highly scalable

rfc/system/RFC-5659-execution-concurrency.md
```diff
@@ -0,0 +1,317 @@
...
+```sql
+SELECT e.*
+FROM   executions AS e
+WHERE  ( launch_plan_id, created_at ) IN (SELECT launch_plan_id,
+                                                 Min(created_at)
+                                          FROM   executions
+                                          WHERE  phase = 'PENDING'
+                                          GROUP  BY launch_plan_id); 
+```
+2. For each execution returned by the above query, `Add()` the pending execution to a [rate limiting workqueue](https://github.com/kubernetes/client-go/blob/master/util/workqueue/rate_limiting_queue.go#L27-L40) (as a suggestion)
```
kumare3, line 100: i would not actually do this, this is because it avoids making across execution generalizations like priority etc. This is infact a scheduler. 
Important thing to understand is when scheduler is and when is it not.

rfc/system/RFC-5659-execution-concurrency.md
```diff
@@ -0,0 +1,317 @@
...
+3. In a separate goroutine, fetch items from the workqueue and individually process each execution entry
+   1. Check the database to see if there are fewer than `MAX_CONCURRENCY` non-terminal executions matching the launch plan ID in the pending execution model
```
kumare3, line 102: i would do this once in memory and keep a copy of running executions in memory. Then as the plan changes we can simply update the inmemory state, reducing the db load greatly

rfc/system/RFC-5659-execution-concurrency.md
```diff
@@ -0,0 +1,317 @@
...
+                                                 Min(created_at)
+                                          FROM   executions
+                                          WHERE  phase = 'PENDING' AND concurrency_level = 2;
+                                          GROUP  BY launch_plan_named_entity_id); 
+```
+
+Note, in this proposal, registering a new version of the launch plan and setting it to active will determine the concurrency policy across all launch plan versions.
+
+#### Prior Art
+The flyteadmin native scheduler (https://github.com/flyteorg/flyte/tree/master/flyteadmin/scheduler) already implements a reconciliation loop to catch up on any missed schedules.
```
kumare3, line 204: exactly look at the prior art. The number of db queries is minimal. Actually one per cycle.

I would keep all running executions with concurrency_level set in memory and all lps with concurrency_level set in memory (only the concurrency policies)
We should periodically update these and its ok to be eventually consistent

rfc/system/RFC-5659-execution-concurrency.md
katrogan: how do we reconcile the terminated executions to unblock concurrency gates if we don't read from the db?

SebSIK:
It was suggested I add a use case to this PR -- [slack thread](https://flyte-org.slack.com/archives/CP2HDHKE1/p1732738537754739?thread_ts=1732718388.864849&cid=CP2HDHKE1)

I have a cron scheduled workflow (wf) that should not execute if the wf is already executing (from a previous schedule).

Why?  The wf executes a legacy app that checks if new files are available on S3.  If so, then the app begins processing, which involves other resources that allow only single use.  Processing time can vary from minutes to hours, depending on the data received.  If another process is started, both will eventually fail due to resource contention.

Considerations:
* It is impractical to change the legacy app.
* It is desired to set cron to a rather small period.  So several wf's can start while the original is running.
* It is not desirable to have tasks stack up waiting for the first to finish (i.e. cache serialize)

LemontreeN:
Hello,
I would like to know if this feature is under development? We really need this feature!

rfc/system/RFC-5659-execution-concurrency.md
```diff
@@ -0,0 +1,284 @@
...
+  
+  // Defines how to handle the execution when the max concurrency is reached.
+  ConcurrencyPolicy policy = 2;
+}
+
+enum ConcurrencyPolicy {
+  UNSPECIFIED = 0;
+  
+  // wait for previous executions to terminate before starting a new one
+  WAIT = 1;
```
ZhouGongZiBBS, line 53: For my understanding, this policy is on the scheduler side, and wait essentially means the scheduler does not fire request to admin before the previous one is completed ?

rfc/system/RFC-5659-execution-concurrency.md
```diff
@@ -0,0 +1,284 @@
...
+  
+  // fail the CreateExecution request and do not permit the execution to start
+  ABORT = 2;
```
ZhouGongZiBBS, line 56: For abort, since this is part of the policy in the scheduler, how does it address the issue with other workflows that does not go through shceduler ?
such as https://github.com/flyteorg/flyte/issues/5125
where users can create workflows without schedule and still create lot of workflows

katrogan: this doesn't go through the native scheduler but rather is managed by a separate async routine that inspects _all_ pending executions

thomasjhuang:
> Hello, I would like to know if this feature is under development? We really need this feature!

We are in progress! Still cleaning up final parts of the RFC and I think implementation can be underway soon.

wild-endeavor:
Met with met up with @ ZhouGongZi_yizluo and @thomasjhuang yesterday to discuss this RFC.  We just went over the doc in detail to make sure we were all on the same page.  Have a series of clarifications we wanted more information/questions that we weren’t able to reach a conclusion on, detailed below.
* Reiterate we’re not going to do concurrency per version.  Need a compelling user story here to justify this work.
* We asked/discussed at the end of whether there was talk ever of having an Abort threshold - like the behavior should be to wait, until there’s a backup of N executions all in Pending state.  At that point, the N + 1 execution should fail immediately.  Definitely also not doing this unless there is a compelling user requirement… it adds a lot of complexity to the project unless we can really justify it.
* In “Creating an execution” item 2 - two questions here: a) what’s the scenario in which the process tries to write QUEUED and the database already has QUEUED ? and b) by “conditionally mark” you just mean a sql statement that checks the current value first right?  (compare-and-swap semantics).
* In “Creating an execution” item 3 - from “and swallow any errors…” onwards we didn’t really understand.  Could you clarify this point please?
  * one scenario we wanted clarity on: let’s say step 1 completes (“create the workflow CRD”).  Before step 2 can run, the process crashes, the node running the concurrency manager goes away completely.  When it comes up again, it’ll again query the Admin database and see that the execution is still Pending (let’s say propeller hasn’t picked up the crd yet).  So we enter the execution create cycle again, only this time let’s say there’s a network partition, and it fails, the kube api is down.  But in the background propeller now starts working on the CRD and updates admin to Running.  At this point we should move to step 3 right, attempt to mark the execution as Failed?
    * related question due to my lack of understanding of the K8s workqueue library - can the same execution be in the workqueue twice?  That is, in between steps 1 and 2, is there a chance that another goroutine comes up and tries to run step 1 (‘create the workflow CRD’) again?
* In “Launch Plan informer” section, the last paragraph discusses the launch plan informer object running queries in the background to get updated launch plan concurrency information.  The last sentence though is: “If an execution has terminated since the last time the query ran, it won’t be in the result set and we’ll want to update the in memory map to remove the execution.”  What does executions terminating have anything to do with launch plan table update times?
* In “Introduce the Concurrency Controller” - item 2. ii. - “update the map…“.  By ‘update’ we mean replace right?  Otherwise the Sets will just grow forever.
* In “Introduce the Concurrency Controller” - item 4. i. - “Check a separate in-memory map…“.   Shouldn’t this just be the launch plan informer?   Or are there two of these informers?

Other points:
* Was there some consideration given to the ‘active’ launch plan marker?  One core use-case for this concurrency feature is to make sure only one scheduled workflow is running at a given time.  Given that schedules key off of the active launch plan, would it make sense that the concurrency setting on the active launch plan be the one that’s used rather than the latest version by time?
* We should have some sort of a CLI that can show the current status of pending executions.  What this API looks like will is to be determined by Thomas, Yizhou et. al. in the course of implementing this feature - they’ll learn what is helpful to see/debug.

rfc/system/RFC-5659-execution-concurrency.md
ZhouGongZiBBS: how does this helps with the native scheduler ? (in the case where the cron schedule is frequent, before the next run starts, the previous run has not finished, and we want to ensure only 1 execution there.)

from my understanding, we can specify ABORT or WAIT:
- If we specify ABORT for a scheduled, the desired effect we wanted to achieve is **period = min(schedule, run_time)**, but in this case, there will be some executions being aborted leaving "space" between executions: imagine a 5 mins periodic schedule, while each run take ~10 mins. When the second run starts, it sees the previous run is still running, so it is ABORTED, then flyte would wait for the third run to start, which takes a bit more time than desired. I.e. not exactly by the time the previous execution finishes.
- If we specify WAIT for the schedule, the cron job will all be PENDING state in the DB. In this case, we can ensure the 1 execution at a time exactly, however, the DB will grow larger as time goes.

Should we have sth in the native/original scheduler to it has the knowledge when the previous run finished before starting the next execution (so it waits a bit even when it is the time for next execution, since max=1)

hamersaw (commented):
Has there been previous conversation about the term "concurrency"? My concern is this will be difficult to disambiguate from the `parallelism` / `max-parallelism` we can define at the LP level to define the number of tasks within a workflow that can run at the same time. In the `cache_serialize` proposal we had some similar conversation and decided to use a completely separate term, ie. `serialize`. IMO something like `SerializationStrategy` aligns with the `cache_serialize` concept and disambiguates from any `parallelism` controls we have elsewhere.

rfc/system/RFC-5659-execution-concurrency.md
```diff
@@ -0,0 +1,284 @@
...
+    ...
+    concurrency=Concurrency(
+        max=1,  # defines how many executions with this launch plan can run in parallel
+        policy=ConcurrencyPolicy.WAIT  # defines the policy to apply when the max concurrency is reached
+    )
+)
+```
+
+### FlyteIDL
+We propose adding a new IDL message to capture concurrency behavior at CreateExecutionTime and embedding it in the existing [Schedule](https://github.com/flyteorg/flyte/blob/master/flyteidl/protos/flyteidl/admin/schedule.proto) message
```
hamersaw, line 38: Is there a specific reason to embed this in the Schedule proto? Does this mean it only applies to executions which are scheduled? So concurrency policy does not apply if a user manually starts a workflow?

katrogan: it was originally a suggestion from Ketan since it applies to scheduling policies

rfc/system/RFC-5659-execution-concurrency.md
katrogan: I think separation of concerns may be helpful here to simplify the implementation, wdyt? The schedule is only responsible for managing create execution requests at a periodic interval, the concurrency controller will handle when to actually create execution objects for all pending executions.

Abort scenario: I would argue that if a workflow execution takes 6 minutes to execute, and the schedule interval is every 5 minutes then this represents an incompatibility in the schedule, it's basically destined to keep terminating executions before they complete. Maybe a future investment would be a workflow triggering itself as a final step before completing, if you wanted to daisy chain executions.

Wait scenario: I would also argue this is misconfigured and the schedule interval should be increased to allow the average duration job to complete. In any case the execution load remains the same, we're just delaying the processing with the queue build-up like you mentioned.

katrogan:
@wild-endeavor, @ZhouGongZiBBS  and @thomasjhuang thank you for the very detailed feedback, I've updated the RFC in response but wanted to address individual points in a reply as well

| Reiterate we’re not going to do concurrency per version. Need a compelling user story here to justify this work.
Done, called this out explicitly as a non-goal although included a potential design approach in case we decide to pursue this in the future to ensure this is option isn't closed off to us

| We asked/discussed at the end of whether there was talk ever of having an Abort threshold - like the behavior should be to wait, until there’s a backup of N executions all in Pending state. At that point, the N + 1 execution should fail immediately. Definitely also not doing this unless there is a compelling user requirement… it adds a lot of complexity to the project unless we can really justify it.
Really interesting point, we could always extend the `SchedulerPolicy` with the abort threshold and have the concurrency controller manage the auto-aborts.

| In “Creating an execution” item 2 - two questions here: a) what’s the scenario in which the process tries to write QUEUED and the database already has QUEUED ? and b) by “conditionally mark” you just mean a sql statement that checks the current value first right? (compare-and-swap semantics).
a) For the state machine I envisioned something that could recover at any failure point. So if for whatever reason the process stops after we write a QUEUED event and we reboot the concurrency controller, we shouldn't undo the execution progress we've already recorded.
b) yes exactly! clarified in the doc


| In “Creating an execution” item 3 - from “and swallow any errors…” onwards we didn’t really understand. Could you clarify this point please?
- one scenario we wanted clarity on: let’s say step 1 completes (“create the workflow CRD”). Before step 2 can run, the process crashes, the node running the concurrency manager goes away completely. When it comes up again, it’ll again query the Admin database and see that the execution is still Pending (let’s say propeller hasn’t picked up the crd yet). So we enter the execution create cycle again, only this time let’s say there’s a network partition, and it fails, the kube api is down. But in the background propeller now starts working on the CRD and updates admin to Running. At this point we should move to step 3 right, attempt to mark the execution as Failed?
- related question due to my lack of understanding of the K8s workqueue library - can the same execution be in the workqueue twice? That is, in between steps 1 and 2, is there a chance that another goroutine comes up and tries to run step 1 (‘create the workflow CRD’) again?
Great points, updated this so we only overwrite for 'PENDING' or 'QUEUED' statuses if the execution progresses due to network partition.
Re: same execution, this will have an identical execution identifier, correct? We upstreamed https://github.com/flyteorg/flyte/pull/6025 which gracefully handles re-enqueing the same item.


| In “Launch Plan informer” section, the last paragraph discusses the launch plan informer object running queries in the background to get updated launch plan concurrency information. The last sentence though is: “If an execution has terminated since the last time the query ran, it won’t be in the result set and we’ll want to update the in memory map to remove the execution.” What does executions terminating have anything to do with launch plan table update times?
I have no idea, I'm sorry. I updated this section.

| In “Introduce the Concurrency Controller” - item 2. ii. - “update the map…“. By ‘update’ we mean replace right? Otherwise the Sets will just grow forever.
Yes, just overwriting the map entry. Clarified in the doc.

| In “Introduce the Concurrency Controller” - item 4. i. - “Check a separate in-memory map…“. Shouldn’t this just be the launch plan informer? Or are there two of these informers?
Yes clarified, thank you!

| Was there some consideration given to the ‘active’ launch plan marker? One core use-case for this concurrency feature is to make sure only one scheduled workflow is running at a given time. Given that schedules key off of the active launch plan, would it make sense that the concurrency setting on the active launch plan be the one that’s used rather than the latest version by time?
Yes this is exactly what we should use to get the active Scheduler Policy! I clarified the proposal but let me know if this is still ambiguous.

| We should have some sort of a CLI that can show the current status of pending executions. What this API looks like will is to be determined by Thomas, Yizhou et. al. in the course of implementing this feature - they’ll learn what is helpful to see/debug.
This sounds like an excellent plan!

erikvanoosten:
We have a rather complex scheduling requirement that could be solved by this issue, but only when we can limit concurrency on an arbitrary label (a resource id?), with potentially having thousands of labels in use at the same time.

A bit more details:

- We need to run a hundred to a few thousands tasks, each task updates around 10 partitions of our database.
- The list of partitions (they have a clear id) a task updates is known when the task is added to the task graph (which is not at the start but later).
- A partition may only be updated by 1 task at a time.
- Tasks are not short lived, they take 30 to 90 minutes.
- We want to run as many tasks in parallel as possible.
- The concurrency limit needs to work across task graphs.

Mapping our use case to the ideas in this RFC:

- a task declares which resources it needs (where resource = partition id).
- for every used resource (partition id used in the graph) we declare the maximum concurrency to be 1.

**Bonus**

With smart scheduling the total run time can be greatly reduced. We do this by unblocking as many tasks as possible by first running tasks that have the most resources in common with other tasks.

**Discussion**

This all sounds pretty complex, but when there is a singleton singleton, the algorithm is not that difficult.

erikvanoosten:
This RFC points to https://github.com/flyteorg/flyte/issues/267. However, this RFC does not fully cover the use cases of that issue. This RFC does not mention something like a resource/lock/mutex-id. The result is that this RFC only works at the task level and it does not cover use cases where we need to restrict access to the same resource from different tasks (even across launch plans).

As mentioned in my previous comment, locking multiple resources brings even more use cases in scope.

thomasjhuang:
Work is currently in progress - #6309 6309 . This isn't fully flushed out yet, but not new adds to the RFC, it is mostly following the defined outline in the RFC so far. Will add more details + docs in the PR as we make progress.

katrogan:
@eapolinario since it sounds like we've settled on a design here, should we merge the RFC?

eapolinario (approved):

eapolinario:
@erikvanoosten , let's take this idea of providing a lock per-resource into a separate discussion. 

It's not too hard to extrapolate this idea to also cover the proposal in this RFC (which only covers workflow execution concurrency) with a more general resource concurrency, but it's worth separating these two out as there are implications to both the SDK and how this is implemented in the backend.

erikvanoosten:
> the basic algorithm could be implemented in 20 lines of e.g. python

Because someone asked for it, the '20 lines' are:

```python
import enum


class TaskState(enum.Enum):
    NEW = "new"
    RUNNING = "started"
    SUCCESS = "success"
    FAILED = "failed"


@dataclass
class Task:
    task_id: str
    lock_ids: list[str]
    state: SchedulerTaskState


def tasks_to_start(tasks: list[Task], capacity: int) -> list[Task]:
	# Collect locks held by a running task
	active_lock_ids: set[str] = set()
	for task in tasks:
		if task.state == TaskState.RUNNING:
			active_lock_ids |= set(task.lock_ids)

	startable_tasks: list[Task] = []
	for task in tasks:
		if (
			task.state == TaskState.NEW and
			all(lock_id not in active_lock_ids for lock_id in task.lock_ids)
		):
			startable_tasks.append(task)

	# Build a dict from lock_id to number of tasks that require the lock
	task_count_by_lock_id = defaultdict(int)
	for task in startable_tasks:
		for lock_id in task.lock_ids:
			task_count_by_lock_id[lock_id] += 1
	task_count_by_lock_id: dict[str, int] = dict(task_count_by_lock_id)

	# For each task calculate the number of tasks that have a lock in common
	common_lock_count_per_task: dict[str, int] = defaultdict(int)
	for task in startable_tasks:
		for lock_id in task.lock_ids:
			common_lock_count_per_task[task.task_id] += task_count_by_lock_id[lock_id]

	# Order the tasks by:
	#   - the number of locks it has in common with other tasks (most first)
	#   - task id
	ordered_tasks = sorted(
		startable_tasks,
		key=lambda task: (-common_lock_count_per_task[task.task_id], task.task_id)
	)

	# Select the highest ranked tasks without locking conflicts
	selected_tasks: list[Task] = []
	selected_task_counts: int = 0
	selected_lock_ids: set[str] = set()
	for task in ordered_tasks:
		task_lock_ids = set(task.lock_ids)
		if (
			selected_task_counts < capacity and
			len(selected_lock_ids.intersection(task_lock_ids)) == 0
		):
			selected_task_counts += 1
			selected_lock_ids = selected_lock_ids.union(task_lock_ids)
			selected_tasks.append(task)

	return selected_tasks
```

Issue #5135, opened 2024-03-29
[BUG] Fix ArrayNode not reporting terminal state in TaskExecutionEvent


pvditt:
### Describe the bug

ArrayNode currently reports events identically to maptask plugin where we encapsulate subNode executions within the ExternalResourceInfo field in TaskExecutionEvent proto message. The ExternalResourceInfo proto message has a phase field, which corresponds to the phase of the subNode. The top-level node phase is reported using (and displayed in the UI from) the NodeExecutionEvent - the task phase on the TaskExecutionEvent that is not currently being used by console.

However, this task phase is incorrect. We attempt to set it on terminal states [here](https://github.com/flyteorg/flyte/blob/513c3e1b8dba948cb2443609b27aae04d5609af4/flytepropeller/pkg/controller/nodes/array/handler.go#L531-L534) , but this code is never reached because [terminal states return prematurely](https://github.com/flyteorg/flyte/blob/513c3e1b8dba948cb2443609b27aae04d5609af4/flytepropeller/pkg/controller/nodes/array/handler.go#L510-L516). We need to ensure that the phase on the TaskExecutionEvent is correctly set on terminal phases.

### Expected behavior

task execution phases match terminal state for array nodes

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

cw75:
@runllm is there a pull request that fixes this issue?

pvditt:
fixes:
https://github.com/flyteorg/flyte/pull/5680
https://github.com/flyteorg/flyte/pull/5681
https://github.com/flyteorg/flyte/pull/5451/files

pvditt:
@cw75 we just upstreamed a few of the fixes. Will close this after https://github.com/flyteorg/flyte/pull/5681 is merged

PR #5681, opened 2024-08-21
[BUG] use deep copy of bit arrays when getting array node state

pvditt:
Upstreaming https://github.com/unionai/flyte/pull/428
https://github.com/flyteorg/flyte/issues/5135

ArrayNodeState is not supposed to get persisted unless the handle loop completes and successfully calls [PutArrayNodeState](https://github.com/unionai/flyte/blob/1fd4a0084a75797a97acf284c50dc73dea57b67f/flytepropeller/pkg/controller/nodes/array/handler.go#L661)

However, any [SetItem](https://github.com/unionai/flyte/blob/1fd4a0084a75797a97acf284c50dc73dea57b67f/flytestdlib/bitarray/compact_array.go#L39) call on a bitarray ([ex](https://github.com/unionai/flyte/blob/1fd4a0084a75797a97acf284c50dc73dea57b67f/flytepropeller/pkg/controller/nodes/array/state_store.go#L167)) that is used to store subnode phases + subtask phases will persist even if the Handle loop does not successfully call PutArrayNodeState at the end of the handle loop.

This is because the [BitSet](https://github.com/unionai/flyte/blob/1fd4a0084a75797a97acf284c50dc73dea57b67f/flytestdlib/bitarray/compact_array.go#L16) field in a bitarray is a pointer -> SetItem modifies the underlying data with an explicit call to PutArrayNodeState.

This can cause issues as we don't finish handling a subnode at a given phase if there is an error causing a return UnknownTransition after we [persist subnode state](https://github.com/unionai/flyte/blob/1fd4a0084a75797a97acf284c50dc73dea57b67f/flytepropeller/pkg/controller/nodes/array/handler.go#L366).

Example issue:

subnode "1" gets set to a terminal node phase
[eventing fails](https://github.com/unionai/flyte/blob/1fd4a0084a75797a97acf284c50dc73dea57b67f/flytepropeller/pkg/controller/nodes/array/handler.go#L644)
next handle loops has the accidentally updated state so doesn't get re-evaluated at the previous, non-terminal phase so we don't propagate the necessary eventing to set the subnode's correct phases

## How was this patch tested?
- deployed in Union 
- ran a myriad of workflows, mapping over tasks and launch plans, that had varying levels of successes, failures, and manual aborts.
- manually set eventing to fail to ensure that the next loop properly evaluated with the correct phases

## Related PRs

branched off https://github.com/flyteorg/flyte/pull/5680 instead of master to reduce conflicts while upstreaming. Will merge that in first to clear up diff.


.run/single-binary.run.xml
```diff
@@ -1,13 +0,0 @@
-<component name="ProjectRunConfigurationManager">
```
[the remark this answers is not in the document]
pvditt: oh thanks for catching this. Unsure how I did that. Let me add it back.

flytepropeller/pkg/apis/flyteworkflow/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -548,6 +548,10 @@ func (in *NodeStatus) DeepCopyInto(out *NodeStatus) {
 		*out = new(DynamicNodeStatus)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.ArrayNodeStatus != nil {
+		in, out := &in.ArrayNodeStatus, &out.ArrayNodeStatus
+		*out = (*in).DeepCopy()
+	}
```
eapolinario, line 554: can you document how you generated the change in this file?

flytepropeller/pkg/controller/config/config.go
```diff
@@ -259,6 +259,7 @@ const (
 type EventConfig struct {
 	RawOutputPolicy           RawOutputPolicy `json:"raw-output-policy" pflag:",How output data should be passed along in execution events."`
 	FallbackToOutputReference bool            `json:"fallback-to-output-reference" pflag:",Whether output data should be sent by reference when it is too large to be sent inline in execution events."`
+	ErrorOnAlreadyExists      bool            `json:"error-on-already-exists" pflag:",Whether to return an error when an event already exists."`
```
eapolinario, line 262: worth enabling this [in single-binary](https://github.com/flyteorg/flyte/blob/ff12f36d385cdcc0ad2acc86803940cd3812ea23/charts/flyte-binary/templates/configmap.yaml#L42-L43) by default?

flytepropeller/pkg/apis/flyteworkflow/v1alpha1/zz_generated.deepcopy.go
pvditt: I did this manually on our fork when working on an array node feature a couple months ago. I couldn't get the script to run.

flytepropeller/pkg/controller/config/config.go
pvditt: This should never really be enabled except for array node. I didn't want to update the record task event function, but I guess that might be better if users were to unknowingly change this and have executions fail.

pvditt: actually I think I can do this a little better and not expose a new config

flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -316,6 +316,27 @@ func (in *ArrayNodeStatus) SetTaskPhaseVersion(taskPhaseVersion uint32) {
...
+		*out = *in
+	}
+}
+
+func (in *ArrayNodeStatus) DeepCopy() *ArrayNodeStatus {
+	if in == nil {
+		return nil
+	}
+
+	out := &ArrayNodeStatus{}
```
eapolinario, line 335: Why are we copying into an empty `ArrayNodeStatus`?

flytepropeller/pkg/controller/config/config.go
```diff
@@ -259,7 +259,8 @@ const (
 type EventConfig struct {
 	RawOutputPolicy           RawOutputPolicy `json:"raw-output-policy" pflag:",How output data should be passed along in execution events."`
 	FallbackToOutputReference bool            `json:"fallback-to-output-reference" pflag:",Whether output data should be sent by reference when it is too large to be sent inline in execution events."`
-	ErrorOnAlreadyExists      bool            `json:"error-on-already-exists" pflag:",Whether to return an error when an event already exists."`
+	// only meant to be overridden for certain node types that have different eventing behavior such as ArrayNode
+	ErrorOnAlreadyExists bool `json:"-"`
```
eapolinario, line 263: I understand the purpose of this now, thanks for the comment. We're adding a field to a central config object and assuming that such field will only be used in a particular situation. This feels hacky to me, but maybe there's a precedent and good reason to follow this pattern.

pvditt: yup - it definitely is hacky. Looking forward when/if we eventually re-work eventing in propeller

flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
pvditt: This method returns a deep copy of the ArrayNodeStatus.

This is similar to [func (in *TaskNodeStatus) DeepCopy() *TaskNodeStatus](https://github.com/flyteorg/flyte/blob/38cc76dd78e7f46283f160189cb6b38cf8fb1c8d/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go#L999)

eapolinario: sorry, got the direction wrong. Thank for clarifying!

eapolinario (approved):

Issue #5606, opened 2024-07-30
[BUG] FlytePropeller 1.13.0 is crashing Flyteadmin and making the overall service in a bad state


RRap0so:
### Describe the bug

We've tried several times upgrading to flyte v1.13.0 and we've found some serious issues blocking the upgrade, we don't know the cause yet but overall it seems that Flyte Propeller calls to FlyteAdmin makes FlyteAdmin make many calls too two GRPC Endpoints: 

- `GetOauth2Metadata`
- `GetPublicClientConfig`

What we see in Flyteadmin is a lot of pressure in the GRPC side and an increase in memory until it gets killed by the k8s with OOM. This in turn causes the FlytePropeller calls to fail with grpc errors and eventually also cause a crash. What we also noticed is that this versions added a retry mechanism which in our case is too aggressive but we've tuned that down.

Hard to say looking at the code but I suspect this PR introduced some kind of bug or missing config unaware to us which makes FlyteAdmin constantly trying to fetch a token (and fails).

https://github.com/flyteorg/flyte/pull/5407

When we rollback to FlytePropeller v1.11.0 things are fine and FlyteAdmin behaves nicely (still running in v1.13.0).

### Expected behavior

Flyteadmin doesn't get massively impacted by grpc calls.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

wild-endeavor:
I think what is happening here is that the auth intercepter, provided by the `NewAuthInterceptor` [function](https://github.com/flyteorg/flyte/blob/0a441a97c990024c0f6e07a39939fb3aa1f4b7f0/flyteidl/clients/go/admin/auth_interceptor.go#L140), [calls](https://github.com/flyteorg/flyte/blob/0a441a97c990024c0f6e07a39939fb3aa1f4b7f0/flyteidl/clients/go/admin/auth_interceptor.go#L147C11-L147C33) the `MaterializeCredentials` function at every invocation, even if there is a token.

Inside [MaterializeCredentials](https://github.com/flyteorg/flyte/blob/0a441a97c990024c0f6e07a39939fb3aa1f4b7f0/flyteidl/clients/go/admin/auth_interceptor.go#L23), the code does a couple things that maybe can be optimized.
* it always creates a new `authMetadataClient`
* and it always calls `NewTokenSourceProvider`
[...]

PR #5686, opened 2024-08-23
Auth/prevent lookup per call

wild-endeavor:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5606

Clients configured with client credentials secret were calling Admin's auth metadata endpoints for every call.  This moves the initialization of all that above the interceptor.

* Move initialization logic into a new object, synchronized with a Once.
* Make the flytectl `config` command disable the flyte client.

## How was this patch tested?
Tested by running locally and hitting our development deployment with client secret.


eapolinario (approved):

flyteidl/clients/go/admin/auth_interceptor.go
```diff
@@ -137,14 +114,40 @@ func setHTTPClientContext(ctx context.Context, cfg *Config, proxyCredentialsFutu
...
+	ctx := context.Background()
+
+	authMetadataClient, err := InitializeAuthMetadataClient(ctx, cfg, proxyCredentialsFuture)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialized Auth Metadata Client. Error: %w", err)
+	}
+
+	tokenSourceProvider, err := NewTokenSourceProvider(ctx, cfg, tokenCache, authMetadataClient)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialized token source provider. Err: %w", err)
```
eapolinario, line 127: nit: can we fix this typo? initialized -> initialize

eapolinario (approved):

Sovietaced:
Ideally it would be nice if the metadata was not cached permanently considering it contains trusted public key material. If there is interest I can follow up with a strategy that periodically refreshes the metadata so it will handle key rotations gracefully.

PR #5697, opened 2024-08-27
Added literal offloading for array node map tasks

pmahindrakar-oss:
## Tracking issue
RFC https://github.com/flyteorg/flyte/pull/5103
This PR tackles the mapped o/p results exceeeding the offloaded min size 

Without these changes we hit grpc limit when propeller tries to create an execution by passing inputs inline which it got from output of large map task 

Here you can see the failure where we are hitting the default 4MB grpc message size limit
This is the line where it  fails
https://github.com/flyteorg/flyte/blob/7136919f271ec1a5b0fa0ccd29175191d5ceffee/flytepropeller/pkg/controller/nodes/subworkflow/launchplan/admin.go#L151

```
[UserError] failed to launch workflow, caused by: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (26624663 vs. 4194304)
```

See the testing details for the workflow used. In particular this section 

```
def ref_wf(mbs: int):
    processed_results = my_wf(mbs=mbs)  # Promise of List[str]
    big_inputs_wf_lp(input=processed_results) 
```
Where we have large task output being returned in processed results and is being passed to launchplan and propeller sends these inputs inline and hence if the value is too large then we exceed the grpc limit

* Adds literal offloading config which provides a way to define min and max offloading size for the literal and also minimum supported SDK version
* After the outputs are gathered from the map tasks and before writing the outputs of the combined literal, we write the data to offloaded location and clear out the literal value and store the offloaded location in the literal and in that way when we create execution from propeller the inputs are small and only contain references to the data and not the actual data.

## How was this patch tested?

Using the following workflow which generates ~ 20 MB map output when launched with input =20 for the following workflow `ref_wf`

### Workflow used for testing

```
import logging
from typing import List
from flytekit import map_task, task, workflow, LaunchPlan

# Set the logging level to DEBUG
logging.basicConfig(level=logging.DEBUG)
logger = logging.getLogger("flytekit")
logger.setLevel(logging.DEBUG)

# Task to generate 1MB of string
@task
def my_1mb_task(i: str) -> str:
    return f"Hello world {i}" * 100 * 1024

# Task to generate a list of strings
@task
def generate_strs(count: int) -> List[str]:
    return ["a"] * count

# Workflow to process the list of strings and return processed results
@workflow
def my_wf(mbs: int) -> List[str]:
    strs = generate_strs(count=mbs)
    return map_task(my_1mb_task)(i=strs)

# Task to handle inputs (with metadata printing)
@task
def noop_with_inputs(input: List[str]):
    # Print metadata about the input
    print("Input Metadata:")
    print(f"Number of items: {len(input)}")
    print(f"Total size: {sum(len(s) for s in input)} bytes")

# Workflow to handle big inputs
@workflow
def big_inputs_wf(input: List[str]):
    noop_with_inputs(input=input)

# LaunchPlan for big inputs workflow
big_inputs_wf_lp = LaunchPlan.get_or_create(name="big_inputs_wf_lp", workflow=big_inputs_wf)

# Main workflow that orchestrates other workflows
@workflow
def ref_wf(mbs: int):
    processed_results = my_wf(mbs=mbs)  # Promise of List[str]
    big_inputs_wf_lp(input=processed_results) 
```

### Before the change receive the following error
```
[UserError] failed to launch workflow, caused by: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (26624663 vs. 4194304)
```

### After the Change

The workflow succeeds

Following is the subworkflow crd which shows the offloaded uri being used as input when propeller launches and execution

```
inputs:
  literals:
    input:
      offloadedMetadata:
        inferredType:
          collectionType:
            simple: STRING
        sizeBytes: "26624325"
        uri: s3://my-s3-bucket/metadata/propeller/flytesnacks-development-argv96zc87ljhdz2fmgt/n0/data/0/n1/0/o0_offloaded_metadata.pb
```

Logs from flytekit reading the offloaded literal and passing this in subworkflow. This uses the flytekit changes from here 

https://github.com/flyteorg/flytekit/pull/2685
which reads the offloaded literal if present

```
Download data to local from s3://my-s3-bucket/metadata/propeller/flytesnacks-development-argv96zc87ljhdz2fmgt/n0/data/0/n1/0/o0_offloaded_metadata.pb. [Time: 0.081294s]", "taskName": null}

```

Pending : Unit tests

### DIFF
--- a/flyteadmin/pkg/manager/impl/validation/execution_validator.go
+++ b/flyteadmin/pkg/manager/impl/validation/execution_validator.go
@@ -100,7 +100,13 @@ func CheckAndFetchInputsForExecution(
 			}
 			executionInputMap[name] = expectedInput.GetDefault()
 		} else {
-			inputType := validators.LiteralTypeForLiteral(executionInputMap[name])
+			var inputType *core.LiteralType
+			switch executionInputMap[name].GetValue().(type) {
+			case *core.Literal_OffloadedMetadata:
+				inputType = executionInputMap[name].GetOffloadedMetadata().GetInferredType()
+			default:
+				inputType = validators.LiteralTypeForLiteral(executionInputMap[name])
+			}
 			if !validators.AreTypesCastable(inputType, expectedInput.GetVar().GetType()) {
 				return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument, "invalid %s input wrong type. Expected %s, but got %s", name, expectedInput.GetVar().GetType(), inputType)
 			}
--- a/flyteadmin/pkg/manager/impl/validation/validation.go
+++ b/flyteadmin/pkg/manager/impl/validation/validation.go
@@ -234,7 +234,7 @@ func validateLiteralMap(inputMap *core.LiteralMap, fieldName string) error {
 			if name == "" {
 				return errors.NewFlyteAdminErrorf(codes.InvalidArgument, "missing key in %s", fieldName)
 			}
-			if fixedInput == nil || fixedInput.GetValue() == nil {
+			if fixedInput.GetValue() == nil && fixedInput.GetOffloadedMetadata() == nil {
 				return errors.NewFlyteAdminErrorf(codes.InvalidArgument, "missing valid literal in %s %s", fieldName, name)
 			}
 			if isDateTime(fixedInput) {
--- a/flytepropeller/pkg/controller/config/config.go
+++ b/flytepropeller/pkg/controller/config/config.go
@@ -125,7 +125,7 @@ var (
 			DefaultParallelismBehavior: ParallelismBehaviorUnlimited,
 		},
 		LiteralOffloadingConfig: LiteralOffloadingConfig{
-			Enabled: true, // Default keep this enabled and in case of issues allow users to disable using config map.
+			Enabled: false, // Default keep this disabled and we will followup when flytekit is released with the offloaded changes.
 			SupportedSDKVersions: map[string]string{ // The key is the SDK name (matches the supported SDK in core.RuntimeMetadata_RuntimeType)  and the value is the minimum supported version
 				"FLYTE_SDK": "1.13.5", // Expected release number with flytekit support from this PR https://github.com/flyteorg/flytekit/pull/2685
 			},
--- a/flytepropeller/pkg/controller/config/config.go
+++ b/flytepropeller/pkg/controller/config/config.go
@@ -34,12 +34,16 @@
 package config
 
 import (
+	"context"
+	"fmt"
 	"time"
 
+	"github.com/Masterminds/semver"
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/flyteorg/flyte/flytestdlib/config"
 	"github.com/flyteorg/flyte/flytestdlib/contextutils"
+	"github.com/flyteorg/flyte/flytestdlib/logger"
 )
 
 //go:generate pflags Config --default-var=defaultConfig
@@ -120,6 +124,14 @@ var (
 			EventVersion:               0,
 			DefaultParallelismBehavior: ParallelismBehaviorUnlimited,
 		},
+		LiteralOffloadingConfig: LiteralOffloadingConfig{
+			Enabled: true, // Default keep this enabled and in case of issues allow users to disable using config map.
+			SupportedSDKVersions: map[string]string{ // The key is the SDK name (matches the supported SDK in core.RuntimeMetadata_RuntimeType)  and the value is the minimum supported version
+				"FLYTE_SDK": "1.13.5", // Expected release number with flytekit support from this PR https://github.com/flyteorg/flytekit/pull/2685
+			},
+			MinSizeInMBForOffloading: 10,   // 10 MB is the default size for offloading
+			MaxSizeInMBForOffloading: 1000, // 1 GB is the default size before failing fast.
+		},
 	}
 )
 
@@ -127,40 +139,79 @@ var (
 // the base configuration to start propeller
 // NOTE: when adding new fields, do not mark them as "omitempty" if it's desirable to read the value from env variables.
 type Config struct {
-	KubeConfigPath           string               `json:"kube-config" pflag:",Path to kubernetes client config file."`
-	MasterURL                string               `json:"master"`
-	Workers                  int                  `json:"workers" pflag:",Number of threads to process workflows"`
-	WorkflowReEval           config.Duration      `json:"workflow-reeval-duration" pflag:",Frequency of re-evaluating workflows"`
-	DownstreamEval           config.Duration      `json:"downstream-eval-duration" pflag:",Frequency of re-evaluating downstream tasks"`
-	LimitNamespace           string               `json:"limit-namespace" pflag:",Namespaces to watch for this propeller"`
-	ProfilerPort             config.Port          `json:"prof-port" pflag:",Profiler port"`
-	MetadataPrefix           string               `json:"metadata-prefix,omitempty" pflag:",MetadataPrefix should be used if all the metadata for Flyte executions should be stored under a specific prefix in CloudStorage. If not specified, the data will be stored in the base container directly."`
-	DefaultRawOutputPrefix   string               `json:"rawoutput-prefix" pflag:",a fully qualified storage path of the form s3://flyte/abc/..., where all data sandboxes should be stored."`
-	Queue                    CompositeQueueConfig `json:"queue,omitempty" pflag:",Workflow workqueue configuration, affects the way the work is consumed from the queue."`
-	MetricsPrefix            string               `json:"metrics-prefix" pflag:",An optional prefix for all published metrics."`
-	MetricKeys               []string             `json:"metrics-keys" pflag:",Metrics labels applied to prometheus metrics emitted by the service."`
-	EnableAdminLauncher      bool                 `json:"enable-admin-launcher" pflag:"Enable remote Workflow launcher to Admin"`
-	MaxWorkflowRetries       int                  `json:"max-workflow-retries" pflag:"Maximum number of retries per workflow"`
-	MaxTTLInHours            int                  `json:"max-ttl-hours" pflag:"Maximum number of hours a completed workflow should be retained. Number between 1-23 hours"`
-	GCInterval               config.Duration      `json:"gc-interval" pflag:"Run periodic GC every 30 minutes"`
-	LeaderElection           LeaderElectionConfig `json:"leader-election,omitempty" pflag:",Config for leader election."`
-	PublishK8sEvents         bool                 `json:"publish-k8s-events" pflag:",Enable events publishing to K8s events API."`
-	MaxDatasetSizeBytes      int64                `json:"max-output-size-bytes" pflag:",Deprecated! Use storage.limits.maxDownloadMBs instead"`
-	EnableGrpcLatencyMetrics bool                 `json:"enable-grpc-latency-metrics" pflag:",Enable grpc latency metrics. Note Histograms metrics can be expensive on Prometheus servers."`
-	KubeConfig               KubeClientConfig     `json:"kube-client-config" pflag:",Configuration to control the Kubernetes client"`
-	NodeConfig               NodeConfig           `json:"node-config,omitempty" pflag:",config for a workflow node"`
-	MaxStreakLength          int                  `json:"max-streak-length" pflag:",Maximum number of consecutive rounds that one propeller worker can use for one workflow - >1 => turbo-mode is enabled."`
-	EventConfig              EventConfig          `json:"event-config,omitempty" pflag:",Configures execution event behavior."`
-	IncludeShardKeyLabel     []string             `json:"include-shard-key-label" pflag:",Include the specified shard key label in the k8s FlyteWorkflow CRD label selector"`
-	ExcludeShardKeyLabel     []string             `json:"exclude-shard-key-label" pflag:",Exclude the specified shard key label from the k8s FlyteWorkflow CRD label selector"`
-	IncludeProjectLabel      []string             `json:"include-project-label" pflag:",Include the specified project label in the k8s FlyteWorkflow CRD label selector"`
-	ExcludeProjectLabel      []string             `json:"exclude-project-label" pflag:",Exclude the specified project label from the k8s FlyteWorkflow CRD label selector"`
-	IncludeDomainLabel       []string             `json:"include-domain-label" pflag:",Include the specified domain label in the k8s FlyteWorkflow CRD label selector"`
-	ExcludeDomainLabel       []string             `json:"exclude-domain-label" pflag:",Exclude the specified domain label from the k8s FlyteWorkflow CRD label selector"`
-	ClusterID                string               `json:"cluster-id" pflag:",Unique cluster id running this flytepropeller instance with which to annotate execution events"`
-	CreateFlyteWorkflowCRD   bool                 `json:"create-flyteworkflow-crd" pflag:",Enable creation of the FlyteWorkflow CRD on startup"`
-	NodeExecutionWorkerCount int                  `json:"node-execution-worker-count" pflag:",Number of workers to evaluate node executions, currently only used for array nodes"`
-	ArrayNode                ArrayNodeConfig      `json:"array-node-config,omitempty" pflag:",Configuration for array nodes"`
+	KubeConfigPath           string                  `json:"kube-config" pflag:",Path to kubernetes client config file."`
+	MasterURL                string                  `json:"master"`
+	Workers                  int                     `json:"workers" pflag:",Number of threads to process workflows"`
+	WorkflowReEval           config.Duration         `json:"workflow-reeval-duration" pflag:",Frequency of re-evaluating workflows"`
+	DownstreamEval           config.Duration         `json:"downstream-eval-duration" pflag:",Frequency of re-evaluating downstream tasks"`
+	LimitNamespace           string                  `json:"limit-namespace" pflag:",Namespaces to watch for this propeller"`
+	ProfilerPort             config.Port             `json:"prof-port" pflag:",Profiler port"`
+	MetadataPrefix           string                  `json:"metadata-prefix,omitempty" pflag:",MetadataPrefix should be used if all the metadata for Flyte executions should be stored under a specific prefix in CloudStorage. If not specified, the data will be stored in the base container directly."`
+	DefaultRawOutputPrefix   string                  `json:"rawoutput-prefix" pflag:",a fully qualified storage path of the form s3://flyte/abc/..., where all data sandboxes should be stored."`
+	Queue                    CompositeQueueConfig    `json:"queue,omitempty" pflag:",Workflow workqueue configuration, affects the way the work is consumed from the queue."`
+	MetricsPrefix            string                  `json:"metrics-prefix" pflag:",An optional prefix for all published metrics."`
+	MetricKeys               []string                `json:"metrics-keys" pflag:",Metrics labels applied to prometheus metrics emitted by the service."`
+	EnableAdminLauncher      bool                    `json:"enable-admin-launcher" pflag:"Enable remote Workflow launcher to Admin"`
+	MaxWorkflowRetries       int                     `json:"max-workflow-retries" pflag:"Maximum number of retries per workflow"`
+	MaxTTLInHours            int                     `json:"max-ttl-hours" pflag:"Maximum number of hours a completed workflow should be retained. Number between 1-23 hours"`
+	GCInterval               config.Duration         `json:"gc-interval" pflag:"Run periodic GC every 30 minutes"`
+	LeaderElection           LeaderElectionConfig    `json:"leader-election,omitempty" pflag:",Config for leader election."`
+	PublishK8sEvents         bool                    `json:"publish-k8s-events" pflag:",Enable events publishing to K8s events API."`
+	MaxDatasetSizeBytes      int64                   `json:"max-output-size-bytes" pflag:",Deprecated! Use storage.limits.maxDownloadMBs instead"`
+	EnableGrpcLatencyMetrics bool                    `json:"enable-grpc-latency-metrics" pflag:",Enable grpc latency metrics. Note Histograms metrics can be expensive on Prometheus servers."`
+	KubeConfig               KubeClientConfig        `json:"kube-client-config" pflag:",Configuration to control the Kubernetes client"`
+	NodeConfig               NodeConfig              `json:"node-config,omitempty" pflag:",config for a workflow node"`
+	MaxStreakLength          int                     `json:"max-streak-length" pflag:",Maximum number of consecutive rounds that one propeller worker can use for one workflow - >1 => turbo-mode is enabled."`
+	EventConfig              EventConfig             `json:"event-config,omitempty" pflag:",Configures execution event behavior."`
+	IncludeShardKeyLabel     []string                `json:"include-shard-key-label" pflag:",Include the specified shard key label in the k8s FlyteWorkflow CRD label selector"`
+	ExcludeShardKeyLabel     []string                `json:"exclude-shard-key-label" pflag:",Exclude the specified shard key label from the k8s FlyteWorkflow CRD label selector"`
+	IncludeProjectLabel      []string                `json:"include-project-label" pflag:",Include the specified project label in the k8s FlyteWorkflow CRD label selector"`
+	ExcludeProjectLabel      []string                `json:"exclude-project-label" pflag:",Exclude the specified project label from the k8s FlyteWorkflow CRD label selector"`
+	IncludeDomainLabel       []string                `json:"include-domain-label" pflag:",Include the specified domain label in the k8s FlyteWorkflow CRD label selector"`
+	ExcludeDomainLabel       []string                `json:"exclude-domain-label" pflag:",Exclude the specified domain label from the k8s FlyteWorkflow CRD label selector"`
+	ClusterID                string                  `json:"cluster-id" pflag:",Unique cluster id running this flytepropeller instance with which to annotate execution events"`
+	CreateFlyteWorkflowCRD   bool                    `json:"create-flyteworkflow-crd" pflag:",Enable creation of the FlyteWorkflow CRD on startup"`
+	NodeExecutionWorkerCount int                     `json:"node-execution-worker-count" pflag:",Number of workers to evaluate node executions, currently only used for array nodes"`
+	ArrayNode                ArrayNodeConfig         `json:"array-node-config,omitempty" pflag:",Configuration for array nodes"`
+	LiteralOffloadingConfig  LiteralOffloadingConfig `json:"literalOffloadingConfig" pflag:",config used for literal offloading."`
+}
+
+type LiteralOffloadingConfig struct {
+	Enabled bool
+	// Maps flytekit and union SDK names to minimum supported version that can handle reading offloaded literals.
+	SupportedSDKVersions map[string]string
+	// Default, 10Mbs. Determines the size of a literal at which to trigger offloading
+	MinSizeInMBForOffloading int64
+	// Fail fast threshold
+	MaxSizeInMBForOffloading int64
+}
+
+// IsSupportedSDKVersion returns true if the provided SDK and version are supported by the literal offloading config.
+func (l LiteralOffloadingConfig) IsSupportedSDKVersion(sdk string, versionString string) bool {
+	if leastSupportedVersion, ok := l.SupportedSDKVersions[sdk]; ok {
+		c, err := semver.NewConstraint(fmt.Sprintf(">= %s", leastSupportedVersion))
+		if err != nil {
+			// This should never happen
+			logger.Warnf(context.TODO(), "Failed to parse version constraint %s", leastSupportedVersion)
+			return false
+		}
+		version, err := semver.NewVersion(versionString)
+		if err != nil {
+			// This should never happen
+			logger.Warnf(context.TODO(), "Failed to parse version %s", versionString)
+			return false
+		}
+		return c.Check(version)
+	}
+	return false
+}
+
+// GetSupportedSDKVersion returns the least supported version for the provided SDK.
+func (l LiteralOffloadingConfig) GetSupportedSDKVersion(sdk string) string {
+	if leastSupportedVersion, ok := l.SupportedSDKVersions[sdk]; ok {
+		return leastSupportedVersion
+	}
+	return ""
 }
 
 // KubeClientConfig contains the configuration used by flytepropeller to configure its internal Kubernetes Client.
--- a/flytepropeller/pkg/controller/nodes/array/handler.go
+++ b/flytepropeller/pkg/controller/nodes/array/handler.go
@@ -499,7 +499,6 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 		// attempt best effort at initializing outputLiterals with output variable names. currently
 		// only TaskNode and WorkflowNode contain node interfaces.
 		outputLiterals := make(map[string]*idlcore.Literal)
-		canOffloadLiteral := false
 		switch arrayNode.GetSubNodeSpec().GetKind() {
 		case v1alpha1.NodeKindTask:
 			taskID := *arrayNode.GetSubNodeSpec().TaskRef
@@ -508,9 +507,6 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 				return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoFailure(idlcore.ExecutionError_SYSTEM,
 					errors.BadSpecificationError, fmt.Sprintf("failed to find ArrayNode subNode task with id: '%s'", taskID), nil)), nil
 			}
-			runtimeData := taskNode.CoreTask().GetMetadata().GetRuntime()
-			// check if the literal can be offloaded based on the runtime data and the offloading config
-			canOffloadLiteral = a.literalOffloadingConfig.Enabled && a.literalOffloadingConfig.IsSupportedSDKVersion(runtimeData.GetType().String(), runtimeData.GetVersion())
 
 			if outputs := taskNode.CoreTask().GetInterface().GetOutputs(); outputs != nil {
 				for name := range outputs.Variables {
@@ -552,7 +548,7 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 		}
 
 		// only offload literal if canOffloadLiteral is true which currently checks if the sdk is compatible to use with this backend change.
-		if canOffloadLiteral {
+		if a.literalOffloadingConfig.Enabled {
 			for outputLiteralKey, outputLiteral := range outputLiterals {
 				// if the size of the output Literal is > threshold then we write the literal to the offloaded store and populate the literal with its zero value and update the offloaded url
 				// use the OffloadLargeLiteralKey to create  {OffloadLargeLiteralKey}_offloaded_metadata.pb file in the datastore.
--- a/flytepropeller/pkg/controller/nodes/array/handler.go
+++ b/flytepropeller/pkg/controller/nodes/array/handler.go
@@ -547,7 +547,7 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 			return handler.UnknownTransition, fmt.Errorf("worker error(s) encountered: %s", workerErrorCollector.Summary(events.MaxErrorMessageLength))
 		}
 
-		// only offload literal if canOffloadLiteral is true which currently checks if the sdk is compatible to use with this backend change.
+		// only offload literal if config is enabled for this feature.
 		if a.literalOffloadingConfig.Enabled {
 			for outputLiteralKey, outputLiteral := range outputLiterals {
 				// if the size of the output Literal is > threshold then we write the literal to the offloaded store and populate the literal with its zero value and update the offloaded url
--- a/flytepropeller/pkg/controller/nodes/array/handler.go
+++ b/flytepropeller/pkg/controller/nodes/array/handler.go
@@ -45,6 +45,7 @@ var (
 // arrayNodeHandler is a handle implementation for processing array nodes
 type arrayNodeHandler struct {
 	eventConfig                 *config.EventConfig
+	literalOffloadingConfig     config.LiteralOffloadingConfig
 	gatherOutputsRequestChannel chan *gatherOutputsRequest
 	metrics                     metrics
 	nodeExecutionRequestChannel chan *nodeExecutionRequest
@@ -498,7 +499,7 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 		// attempt best effort at initializing outputLiterals with output variable names. currently
 		// only TaskNode and WorkflowNode contain node interfaces.
 		outputLiterals := make(map[string]*idlcore.Literal)
-
+		canOffloadLiteral := false
 		switch arrayNode.GetSubNodeSpec().GetKind() {
 		case v1alpha1.NodeKindTask:
 			taskID := *arrayNode.GetSubNodeSpec().TaskRef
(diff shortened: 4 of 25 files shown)


flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -631,6 +651,99 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
...
+		return err
+	}
+
+	toBeOffloaded.Uri = dataReference.String()
+	toBeOffloaded.SizeBytes = literalSizeBytes
+	return nil
+}
+
+// createZeroValueRecursively creates a zero value for the literal recursively
+func createZeroValueRecursively(toBeOffloaded *idlcore.Literal) error {
```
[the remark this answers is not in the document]
pmahindrakar-oss: As discussed yesterday, we have flyteadmin which does type checks which fail if we dont do this.
Thanks for starting the Slack discussion https://unionai.slack.com/archives/C07FC1FF0TC/p1724751606931679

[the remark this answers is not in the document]
katrogan: ah super cool, so this gets around that?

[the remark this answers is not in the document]
pmahindrakar-oss: yes exactly , we dont need to store the type info in the literal message, but just infer it from the zero valued literal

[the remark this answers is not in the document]
pmahindrakar-oss: @EngHabu let me know what do you think about this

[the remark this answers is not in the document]
EngHabu: 1. Can you verify this actually does what you think it does? 
    * Generate a zero value for literal scalar String for example,
    * Marshal it into bytes,
    * Unmarshal from bytes
    * Check that .GetScalar().Value.(type) == *idlcore.Primitive_StringValue

    I do think proto3 keeps that information for zero values but I could be wrong..
2. Why is this better than storing the LiteralType? Are we losing fidelity if we use LiteralTypes? If we are saying we do need type checks, I would rather store the LiteralType instead

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -631,6 +651,99 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
...
+			toBeOffloaded.GetScalar().Value = &idlcore.Scalar_StructuredDataset{StructuredDataset: &idlcore.StructuredDataset{}}
+		case *idlcore.Scalar_Schema:
+			toBeOffloaded.GetScalar().Value = &idlcore.Scalar_Schema{Schema: &idlcore.Schema{}}
+		case *idlcore.Scalar_Union:
+			toBeOffloaded.GetScalar().Value = &idlcore.Scalar_Union{Union: &idlcore.Union{}}
+		default:
+			return fmt.Errorf("unsupported scalar type [%T]", toBeOffloaded.GetScalar().GetValue())
+		}
+	} else if toBeOffloaded.GetCollection() != nil {
+		for _, l := range toBeOffloaded.GetCollection().Literals {
```
EngHabu, line 730: If we proceed with this approach, we should do this once (for the first element of the literals then bail... While Literals support heterogeneous collections, LiteralType doesn't and therefore we don't throughout the system.

flytepropeller/pkg/controller/nodes/array/handler.go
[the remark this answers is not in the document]
pmahindrakar-oss: 1. works well and the types match as even with the zero value
2. I wanted to see if we infer this information from the smaller size literal itself and the motivation was just reduce the size of the literal to bare minimum and offload the bulky data. I will go with the explicit LiteralType though which works well too.  If we go with 2. I am proposing if we introduce new message 

```

    // Offloaded literal metadata
    LiteralOffloadedMetadata offloaded_metadata = 6;
}

// A message that contains the metadata of the offloaded data.
message LiteralOffloadedMetadata {
    // The location of the offloaded data.
    string uri = 1;

    // The size of the offloaded data.
    uint64 size_bytes = 2;

    // The literal type of the offloaded data.
    LiteralType type = 3;
}
```

flytepropeller/pkg/controller/nodes/array/handler.go
pmahindrakar-oss: I would prefer to go with explicit LiteralType which i find it better too. Discard this change

flytepropeller/pkg/controller/nodes/array/handler.go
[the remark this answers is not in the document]
katrogan: as an alternative, can we add the LiteralType message to the top-level `core.Literal` inline?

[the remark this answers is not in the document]
pmahindrakar-oss: We have a way to infer this info https://github.com/flyteorg/flyte/blob/1cd775a417e4280fd78c9334f7e9329e53b017d3/flytepropeller/pkg/compiler/validators/utils.go#L247 which is why i was wondering if we needed to store in the regular case.

We need this only in special case where we make the value nil and offload the literal

flyteadmin/pkg/manager/impl/validation/execution_validator.go
```diff
@@ -100,7 +100,13 @@ func CheckAndFetchInputsForExecution(
 			}
 			executionInputMap[name] = expectedInput.GetDefault()
 		} else {
-			inputType := validators.LiteralTypeForLiteral(executionInputMap[name])
+			var inputType *core.LiteralType
```
katrogan, line 103: nit: mind adding a unit test?

flyteadmin/pkg/manager/impl/validation/validation.go
```diff
@@ -234,7 +234,8 @@ func validateLiteralMap(inputMap *core.LiteralMap, fieldName string) error {
 			if name == "" {
 				return errors.NewFlyteAdminErrorf(codes.InvalidArgument, "missing key in %s", fieldName)
 			}
-			if fixedInput == nil || fixedInput.GetValue() == nil {
+			fixedInput.GetOffloadedMetadata()
```
katrogan, line 237: what's the purpose of this call?

flyteadmin/pkg/manager/impl/validation/validation.go
```diff
@@ -234,7 +234,8 @@ func validateLiteralMap(inputMap *core.LiteralMap, fieldName string) error {
...
+			if fixedInput == nil || (fixedInput.GetValue() == nil && fixedInput.GetOffloadedMetadata() == nil) {
```
katrogan, line 238: nit: I don't think we need the `fixedInput == nil` check anymore since we're using generated getters :)

flyteidl/protos/flyteidl/core/literals.proto
```diff
@@ -93,7 +93,8 @@ message Scalar {
 
 // A simple value. This supports any level of nesting (e.g. array of array of array of Blobs) as well as simple primitives.
 message Literal {
-    reserved 6, 7;
+
```
katrogan, line 96: are the changes in this file intended?

flytepropeller/pkg/controller/config/config.go
```diff
@@ -120,47 +124,86 @@ var (
...
+	if leastSupportedVersion, ok := l.SupportedSDKVersions[sdk]; ok {
+		c, err := semver.NewConstraint(fmt.Sprintf(">= %s", leastSupportedVersion))
+		if err != nil {
+			// This should never happen
+			logger.Errorf(context.TODO(), "Failed to parse version constraint %s", leastSupportedVersion)
+			return false
+		}
+		version, err := semver.NewVersion(versionString)
+		if err != nil {
+			// This should never happen
```
katrogan, line 200: this could fail for user provided input, right (someone working with a prerelease or private version of the SDK for example). could we downgrade to Info?

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -508,6 +515,10 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 					errors.BadSpecificationError, fmt.Sprintf("failed to find ArrayNode subNode task with id: '%s'", taskID), nil)), nil
 			}
 
+			runtimeData := taskNode.CoreTask().GetMetadata().GetRuntime()
+			// check if the literal can be offloaded based on the runtime data and the offloading config
+			canOffloadLiteral = a.literalOffloadingConfig.Enabled && a.literalOffloadingConfig.IsSupportedSDKVersion(runtimeData.GetType().String(), runtimeData.GetVersion())
```
katrogan, line 520: nit, maybe define this closer to where it's used?

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -631,6 +655,47 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 	return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoRunning(&handler.ExecutionInfo{})), nil
 }
 
+// OffloadLargeLiteral offloads the large literal if meets the threshold conditions
+func (a *arrayNodeHandler) OffloadLargeLiteral(ctx context.Context, datastore *storage.DataStore, dataReference storage.DataReference, toBeOffloaded *idlcore.Literal) error {
+	// TODO : also add sdk version checks to ensure that the literal is not offloaded if the sdk version does not support it
```
katrogan, line 660: this is now done in this change, right? :)

katrogan (commented):
looks great to me!

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -631,6 +655,47 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 	return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoRunning(&handler.ExecutionInfo{})), nil
 }
 
+// OffloadLargeLiteral offloads the large literal if meets the threshold conditions
+func (a *arrayNodeHandler) OffloadLargeLiteral(ctx context.Context, datastore *storage.DataStore, dataReference storage.DataReference, toBeOffloaded *idlcore.Literal) error {
```
hamersaw, line 659: Does it make sense to make this a fucntion of the `DataStore`? Or put this in flytestdlib under the storage package somehwere? It seems like we will certainly use it elsewhere as this work progresses.

flytepropeller/pkg/controller/nodes/array/handler.go
hamersaw: +1

hamersaw (commented):
High-level question - it looks like we're using semver comparison for the flytekit version for the specific task that is executing. Is this right? For example we have a workflow with and ArrayNode (`n0`) that writes sends data to a TaskNode (`n1`). We're checking if the flytekit version of the ArrayNode task for `n0` supports offloading. IIUC that task will execute normally, we will transparently offload, and we are really concerned whether the task for `n1` supports offloading.

flyteadmin/pkg/manager/impl/validation/validation.go
pmahindrakar-oss: oops. thanks test code

flyteidl/protos/flyteidl/core/literals.proto
pmahindrakar-oss: yup just formatting changes

flytepropeller/pkg/controller/config/config.go
pmahindrakar-oss: Sure. Let me make it warning instead of info and we are not failing which is what i intended to do

flytepropeller/pkg/controller/nodes/array/handler.go
pmahindrakar-oss: The reason being,this info is only available under taskMetadata and hence i moved it inside the case statement instead of introducing another conditional to check the same a few lines below
```
switch arrayNode.GetSubNodeSpec().GetKind() {
		case v1alpha1.NodeKindTask:
```
Let me know what you guyz think.

flytepropeller/pkg/controller/nodes/array/handler.go
pmahindrakar-oss: thanks. stale comment

flytepropeller/pkg/controller/nodes/array/handler.go
pmahindrakar-oss: Datastore is working on generic proto message and would want to avoid putting literal functionality over there  as this function does handle infering the literal type and knowing about the offloaded metadata message.

I think if we encounter more use cases we can define a utility in propeller to do this instead of moving to stdlib which  doesn't need to know about the type system IMO. 

Let me know what you think

pmahindrakar-oss:
>High-level question - it looks like we're using semver comparison for the flytekit version for the specific task that is executing. Is this right? For example we have a workflow with and ArrayNode (n0) that writes sends data to a TaskNode (n1). We're checking if the flytekit version of the ArrayNode task for n0 supports offloading. IIUC that task will execute normally, we will transparently offload, and we are really concerned whether the task for n1 supports offloading.

@hamersaw  I agree with this concern and i think we had raised a similar concern during the spec review @katrogan and we had decided to allow things to fail for these heterogeneous scenarios . 

Still trying to get to understand the type system and propeller code but If we need to support this what do you think would be needed here. 
We could fail fast at execution time if we know this is bound to happen and prevent it from happening ?

flyteadmin/pkg/manager/impl/validation/execution_validator.go
pmahindrakar-oss: yup planning to add more coverage. just needed reviews on the main flow

flytepropeller/pkg/controller/nodes/array/handler.go
pmahindrakar-oss: Let me refactor this

pmahindrakar-oss: How does this sound, 


```
// objects
type Options struct {
	// This metadata is passed to stow store implementations.
	Metadata map[string]interface{}
	// Only currently used as an option for protobuf store
	ValidationConfig *ValidationConfig
}

// OffloadingConfig contains configurations for offloading.
// Currently, it is used to configure offloading of large literals.
type ValidationConfig struct {
	MinSizeInMBForOffloading uint64
	MaxSizeInMBForOffloading uint64
}
```

We define a validation Config in store and its used by protbuf store

and the default protobuf implementation does this if the opts contains a validation config


```
func (s DefaultProtobufStore) WriteProtobuf(ctx context.Context, reference DataReference, opts Options, msg proto.Message) error {
	ctx, span := otelutils.NewSpan(ctx, otelutils.BlobstoreClientTracer, "flytestdlib.storage.DefaultProtobufStore/WriteProtobuf")
	defer span.End()

	t := s.metrics.MarshalTime.Start()
	if opts.ValidationConfig != nil {
		literalSizeBytes := uint64(proto.Size(msg))
		literalSizeMB := literalSizeBytes / MB
		// check if the literal is large
		if literalSizeMB > opts.ValidationConfig.MaxSizeInMBForOffloading {
			errWithDetails := fmt.Errorf("message size [%d] MB exceeds the allowed max size [%d] MB", literalSizeMB, opts.ValidationConfig.MaxSizeInMBForOffloading)
			logger.Errorf(ctx, "%v", errWithDetails)
			return errWithDetails
		}
		if literalSizeMB < opts.ValidationConfig.MinSizeInMBForOffloading {
			logger.Debugf(ctx, "message size [%d] MB is smaller than the min size [%d] MB for offloading", literalSizeMB, opts.ValidationConfig.MinSizeInMBForOffloading)
			return nil
		}
	}

```


The caller would just be doing this  and that way we can control how this config is used 

```
opts := storage.Options{
		ValidationConfig: &storage.ValidationConfig{
			MinSizeInMBForOffloading: a.literalOffloadingConfig.MinSizeInMBForOffloading,
			MaxSizeInMBForOffloading: a.literalOffloadingConfig.MaxSizeInMBForOffloading,
		},
	}
	if err := datastore.WriteProtobuf(ctx, dataReference, opts, toBeOffloaded); err != nil {
		logger.Errorf(ctx, "Failed to offload literal at location [%s] with error [%s]", dataReference, err)
		return err
	}
```


A wrapper utlility will take of getting the url and setting it in the literal

```
OffloadLargeLiteral(ctx context.Context, datastore *storage.DataStore, dataReference storage.DataReference, toBeOffloaded *idlcore.Literal) error {
	opts := storage.Options{
		ValidationConfig: &storage.ValidationConfig{
			MinSizeInMBForOffloading: a.literalOffloadingConfig.MinSizeInMBForOffloading,
			MaxSizeInMBForOffloading: a.literalOffloadingConfig.MaxSizeInMBForOffloading,
		},
	}
	if err := datastore.WriteProtobuf(ctx, dataReference, opts, toBeOffloaded); err != nil {
		logger.Errorf(ctx, "Failed to offload literal at location [%s] with error [%s]", dataReference, err)
		return err
	}
	inferredType := validators.LiteralTypeForLiteral(toBeOffloaded)
	if inferredType == nil {
		errString := "Failed to determine literal type for offloaded literal"
		logger.Errorf(ctx, errString)
		return fmt.Errorf(errString)
	}
	// update the literal with the offloaded URI, size and inferred type
	toBeOffloaded.Value = &idlcore.Literal_OffloadedMetadata{
		OffloadedMetadata: &idlcore.LiteralOffloadedMetadata{
			Uri:          dataReference.String(),
			SizeBytes:    uint64(proto.Size(toBeOffloaded)),
			InferredType: inferredType,
		},
	}

	return nil
}
```

Let me know what you think

flytepropeller/pkg/controller/nodes/array/handler.go
hamersaw: I'm OK with whatever you want. If this stops adding a different switch just to get this to work that's fine.

hamersaw:
> Still trying to get to understand the type system and propeller code but If we need to support this what do you think would be needed here.
This is really difficult, because it's unclear if we want to fail just because propeller supports offloading and the task (or downstream tasks) do not. If the outputs aren't large enough to offload, we would want it to run successfully rather than fail even though it could succeed right?

If we absolutely need to validate the flytekit version of tasks to support offloading the only place that makes sense to me is when propeller begins evaluation ([in the handleReadyWorkflow function](https://github.com/unionai/flyte/blob/master/flytepropeller/pkg/controller/workflow/executor.go#L90-L90)). We can't do it during registration or compilation because flyteadmin doesn't know if propeller has offloading enabled. In this function we could iterate through all tasks in the `WorkflowSpec` and validate the flytekit versions the same way you already have. This would have to parse through SubWorkflows and LaunchPlans as well. Of course, this still wouldn't catch issues with dynamics since we do not know what the underlying workflow looks like.

> We could fail fast at execution time if we know this is bound to happen and prevent it from happening ?
This would be much easier. In the node executor we copy inputs from previous nodes when transitioning the node from `NotYetStarted` to `Queued` [here](https://github.com/unionai/flyte/blob/master/flytepropeller/pkg/controller/nodes/executor.go#L751-L787) if the inputs are offloaded literals and the node is a TaskNode that does not have a compatible flytekit version we can fail. It might make more sense to do this in the [TaskHandler](https://github.com/unionai/flyte/blob/master/flytepropeller/pkg/controller/nodes/task/handler.go), or [PodPlugin](https://github.com/unionai/flyte/blob/master/flyteplugins/go/tasks/plugins/k8s/pod/plugin.go)? Offloaded literals won't work for things like Spark, Dask, Agents, etc right? I have to admit I wasn't able to read the doc in depth.

pmahindrakar-oss:
Thanks @hamersaw . Would want to avoid complicated logic if possible and rely on a rollout process to solve it.
Since we are enabling offloading behind a config, there is controlled rollout that is possible .
By default i will keep this disabled and have notes to enable this only when all the workflows/tasks/launchplans have been upgraded to use the new SDK version. This statement though sounds very heavy handed and not sure how many of our users  would be willing to do this, but given that this solves a very peculiar problem when the data outputed by the task is large, i am assuming the user has already worked around this by breaking his workflows or increasing the grpc message size and when he is trying to enable this would be aware of the release note actions and hence my point of allowing things to fail for this heterogeneous case . 

If we cant rely on this process and we want automated checks then we should think about how we can make this fully work (even dynamic case) for old workflows without failing them

cc : @katrogan

pmahindrakar-oss:
@katrogan @hamersaw tested with new changes which now does version checks in heterogeneous tasks scenario and heres the result
Screenshot 2024-09-05 at 5 09 01 PM

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -631,6 +655,46 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
...
+	}
+
+	// offload the literal
+	if err := datastore.WriteProtobuf(ctx, dataReference, storage.Options{}, toBeOffloaded); err != nil {
+		logger.Errorf(ctx, "Failed to offload literal at location [%s] with error [%s]", dataReference, err)
+		return err
+	}
+
+	// update the literal with the offloaded URI, size and inferred type
+	toBeOffloaded.Value = &idlcore.Literal_OffloadedMetadata{
```
katrogan, line 687: nit: might be nice to debug log here that we offloaded and dump the uri?

flytepropeller/pkg/controller/nodes/array/handler.go
katrogan: I actually disagree this should live in the datastore since it's more of a logical service layer implementation (we call the datastore only when we choose to offload, right?) and the storage layer is currently agnostic to the message type of the input protobuf

but to Dan's point if we want to use this in copilot down the road or possibly other use cases I quite like having a general purpose helper utility, and not making this function a member of the `arrayNodeHandler` struct

flytepropeller/pkg/controller/nodes/executor.go
```diff
@@ -790,6 +795,34 @@ func (c *nodeExecutor) preExecute(ctx context.Context, dag executors.DAGStructur
 	return handler.PhaseInfoNotReady("predecessor node not yet complete"), nil
 }
 
+// checkOffloadingCompat checks if the upstream and downstream nodes are compatible with the literal offloading feature
+func (c *nodeExecutor) checkOffloadingCompat(ctx context.Context, nCtx interfaces.NodeExecutionContext, inputLiterals map[string]*core.Literal, node v1alpha1.ExecutableNode) handler.PhaseInfo {
+	isOffloadLiteral := false
```
katrogan, line 800:
```suggestion
	consumesOffloadLiteral := false
```

or something else you'd prefer, only because we can have multiple literals here and this var name sounds like there is just one

flytepropeller/pkg/controller/nodes/executor.go
```diff
@@ -790,6 +795,34 @@ func (c *nodeExecutor) preExecute(ctx context.Context, dag executors.DAGStructur
...
+	}
+	switch node.GetKind() {
+	case v1alpha1.NodeKindTask:
+		taskID := *node.GetTaskID()
+		taskNode, err := nCtx.ExecutionContext().GetTask(taskID)
+		if err != nil {
+			return handler.PhaseInfoFailure(core.ExecutionError_SYSTEM, "GetTaskIDFailure", err.Error(), nil)
+		}
+		runtimeData := taskNode.CoreTask().GetMetadata().GetRuntime()
+		if isOffloadLiteral && (c.literalOffloadingConfig.Enabled && !c.literalOffloadingConfig.IsSupportedSDKVersion(runtimeData.GetType().String(), runtimeData.GetVersion())) {
```
katrogan, line 815: when we have an offloaded literal but the config isn't set does that mean we silently give up attempting to check compatibility? worth at least debug logging in this scenario?

katrogan (commented):
looks great!

flytepropeller/pkg/controller/nodes/executor.go
```diff
@@ -790,6 +795,34 @@ func (c *nodeExecutor) preExecute(ctx context.Context, dag executors.DAGStructur
...
+			logger.Debugf(ctx, "literal offloading : sdk version check failed for task [%s]", taskID)
+			return handler.PhaseInfoFailure(core.ExecutionError_USER, "LiteralOffloadingNotSupported", "Literal offloading is not supported for this task", nil)
```
katrogan, line 817: since this is user facing, want to include the min supported SDK version here so this error is hopefully more self-diagnosable?

flytepropeller/pkg/controller/nodes/array/handler.go
pmahindrakar-oss: Done moved it to utils

flytepropeller/pkg/controller/nodes/array/handler.go
pmahindrakar-oss: Added

flytepropeller/pkg/controller/nodes/executor.go
pmahindrakar-oss: done

flytepropeller/pkg/controller/nodes/executor.go
pmahindrakar-oss: Adding error when the config is off and the consuming a offloaded output so that its clear in the UI

flytepropeller/pkg/controller/nodes/executor.go
pmahindrakar-oss: done

flyteadmin/pkg/manager/impl/validation/execution_validator.go
pmahindrakar-oss: done

flytepropeller/pkg/controller/nodes/common/utils.go
```diff
@@ -67,3 +78,82 @@ func GetTargetEntity(ctx context.Context, nCtx interfaces.NodeExecutionContext)
...
+
+// CheckOffloadingCompat checks if the upstream and downstream nodes are compatible with the literal offloading feature and returns an error if not contained in phase info object
+func CheckOffloadingCompat(ctx context.Context, nCtx interfaces.NodeExecutionContext, inputLiterals map[string]*core.Literal, node v1alpha1.ExecutableNode, literalOffloadingConfig config.LiteralOffloadingConfig) *handler.PhaseInfo {
+	consumesOffloadLiteral := false
+	for _, val := range inputLiterals {
+		if val != nil && val.GetOffloadedMetadata() != nil {
+			consumesOffloadLiteral = true
+			break
+		}
+	}
```
katrogan, line 131: nit: can return early if `consumesOffloadLiteral` is false

flytepropeller/pkg/controller/nodes/common/utils.go
```diff
@@ -67,3 +78,82 @@ func GetTargetEntity(ctx context.Context, nCtx interfaces.NodeExecutionContext)
...
+	var phaseInfo handler.PhaseInfo
+	switch node.GetKind() {
+	case v1alpha1.NodeKindTask:
```
katrogan, line 134: should we move the check for the node kind earlier and return if it's not `NodeKindTask`?

flytepropeller/pkg/controller/nodes/common/utils.go
```diff
@@ -67,3 +78,82 @@ func GetTargetEntity(ctx context.Context, nCtx interfaces.NodeExecutionContext)
...
+				return &phaseInfo
+			}
+			leastSupportedVersion := literalOffloadingConfig.GetSupportedSDKVersion(runtimeData.GetType().String())
+			errMsg := fmt.Sprintf("Literal offloading is not supported for this task as its registered with SDK version [%s] which is less than the least supported version [%s] for this feature", runtimeData.GetVersion(), leastSupportedVersion)
+			logger.Errorf(ctx, errMsg)
+			phaseInfo = handler.PhaseInfoFailure(core.ExecutionError_USER, "LiteralOffloadingNotSupported", errMsg, nil)
+			return &phaseInfo
+		}
+	default:
+		logger.Warnf(ctx, "literal offloading : skipping sdk version check for node kind '%s'", node.GetKind())
```
katrogan, line 156: should we log the node id here too?

katrogan: cc @hamersaw is there any way to validate the SDK version for downstream array nodes here?

katrogan (approved):
LGTM but please wait for @hamersaw's review too

flytepropeller/pkg/controller/nodes/common/utils.go
pmahindrakar-oss: Done

flytepropeller/pkg/controller/nodes/common/utils.go
pmahindrakar-oss: done

flytepropeller/pkg/controller/nodes/common/utils.go
```diff
@@ -67,3 +78,89 @@ func GetTargetEntity(ctx context.Context, nCtx interfaces.NodeExecutionContext)
...
+		}
+	}
+	if !consumesOffloadLiteral {
+		return nil
+	}
+	var phaseInfo handler.PhaseInfo
+
+	// Return early if the node is not of type NodeKindTask
+	if node.GetKind() != v1alpha1.NodeKindTask {
+		logger.Warnf(ctx, "literal offloading : skipping sdk version check for node kind [%s'] with id [%v] ", node.GetKind(), node.GetID())
```
hamersaw, line 139: IIUC this will result in an insane amount of `warn` level logging if we offload a at a high rate.

flytepropeller/pkg/controller/nodes/common/utils.go
```diff
@@ -67,3 +78,89 @@ func GetTargetEntity(ctx context.Context, nCtx interfaces.NodeExecutionContext)
...
+			break
+		}
+	}
+	if !consumesOffloadLiteral {
+		return nil
+	}
+	var phaseInfo handler.PhaseInfo
+
+	// Return early if the node is not of type NodeKindTask
+	if node.GetKind() != v1alpha1.NodeKindTask {
```
hamersaw, line 138: If we are only validating this for task nodes should this logic be in the `TaskHandler` instead? [maybe here](https://github.com/unionai/flyte/blob/master/flytepropeller/pkg/controller/nodes/task/handler.go#L576-L576)?

flytepropeller/pkg/controller/nodes/common/utils.go
```diff
@@ -67,3 +78,89 @@ func GetTargetEntity(ctx context.Context, nCtx interfaces.NodeExecutionContext)
...
+	}
+
+	// Process NodeKindTask
+	taskID := *node.GetTaskID()
+	taskNode, err := nCtx.ExecutionContext().GetTask(taskID)
+	if err != nil {
+		phaseInfo = handler.PhaseInfoFailure(core.ExecutionError_SYSTEM, "GetTaskIDFailure", err.Error(), nil)
+		return &phaseInfo
+	}
+	runtimeData := taskNode.CoreTask().GetMetadata().GetRuntime()
```
hamersaw, line 150: I don't have good context here, but TaskNodes encompass all task executions in Flyte (actors, agents, spark, dask, ray, etc) how is this `runtimeData` set for each? For example, if we offload into a agent task is there a specific check we need to do there?

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -547,6 +551,18 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
...
 		}
 
+		// only offload literal if canOffloadLiteral is true which currently checks if the sdk is compatible to use with this backend change.
+		if canOffloadLiteral {
+			for outputLiteralKey, outputLiteral := range outputLiterals {
+				// if the size of the output Literal is > threshold then we write the literal to the offloaded store and populate the literal with its zero value and update the offloaded url
+				// use the OffloadLargeLiteralKey to create  {OffloadLargeLiteralKey}_offloaded_metadata.pb file in the datastore.
+				// Update the url in the outputLiteral with the offloaded url and also update the size of the literal.
+				offloadedOutputFile := v1alpha1.GetOutputsLiteralMetadataFile(outputLiteralKey, nCtx.NodeStatus().GetOutputDir())
+				if err := common.OffloadLargeLiteral(ctx, nCtx.DataStore(), offloadedOutputFile, outputLiteral, a.literalOffloadingConfig); err != nil {
```
hamersaw, line 561: This is offloading data from individual subNode executions right? Below this is where we actually write the aggregated data (~570). IIUC that is where we actually want to offload. For example, each subNode outputs 50M, which is under the offloading configuration threshold, but if there are 1000 subnodes that exceeds the threshold.

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -507,6 +508,9 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 				return handler.DoTransition(handler.TransitionTypeEphemeral, handler.PhaseInfoFailure(idlcore.ExecutionError_SYSTEM,
 					errors.BadSpecificationError, fmt.Sprintf("failed to find ArrayNode subNode task with id: '%s'", taskID), nil)), nil
 			}
+			runtimeData := taskNode.CoreTask().GetMetadata().GetRuntime()
+			// check if the literal can be offloaded based on the runtime data and the offloading config
+			canOffloadLiteral = a.literalOffloadingConfig.Enabled && a.literalOffloadingConfig.IsSupportedSDKVersion(runtimeData.GetType().String(), runtimeData.GetVersion())
```
hamersaw, line 513: I thought we decided this was unnecessary, please correct me if I am misunderstanding. This portion of the code is retrieving outputs from an already executed task (or launchplan). If the task that already executed doesn't have a valid SDK version (that supports literal offloading) that is OK. The only check we need is if the actual task that is executing over an offloaded literal supports it.

flytepropeller/pkg/controller/nodes/common/utils.go
hamersaw: Maybe something we should discuss. @pvditt do you have strong feelings?

flytepropeller/pkg/controller/nodes/common/utils.go
katrogan: if my understanding is correct, anything we serialize from flytekit should have the runtime set from the SDK, right? 

e.g. here for spark: https://github.com/flyteorg/flytekit/blob/ae9c6f8de21eb716dbda8d1af555c2057655db1d/plugins/flytekit-spark/flytekitplugins/spark/task.py#L153-L159 we call the super method for `PythonFunctionTask`

this https://github.com/flyteorg/flytekit/blob/ae9c6f8de21eb716dbda8d1af555c2057655db1d/flytekit/core/python_function_task.py#L81 inherits from `PythonAutoContainerTask`

which itself inherits from `PythonTask` (https://github.com/flyteorg/flytekit/blob/ae9c6f8de21eb716dbda8d1af555c2057655db1d/flytekit/core/python_auto_container.py#L29)

which sets the metadata https://github.com/flyteorg/flytekit/blob/ae9c6f8de21eb716dbda8d1af555c2057655db1d/flytekit/core/base_task.py#L167-L169

katrogan: for agents in particular, we use the pyflyte entrypoint for all task types so I believe we should get typetransformer changes for reading offloaded inputs for free

e.g here https://github.com/flyteorg/flytekit/blob/master/flytekit/extend/backend/agent_service.py#L176 and elsewhere we call `LiteralMap.from_flyte_idl` in the agent service

we call agent.do https://github.com/flyteorg/flytekit/blob/master/flytekit/extend/backend/agent_service.py#L178

or agent.create https://github.com/flyteorg/flytekit/blob/master/flytekit/extend/backend/agent_service.py#L124

and then call `input_python_value = TypeEngine.literal_map_to_kwargs` https://github.com/flyteorg/flytekit/blob/ae9c6f8de21eb716dbda8d1af555c2057655db1d/plugins/flytekit-openai/flytekitplugins/openai/chatgpt/agent.py#L33

which itself calls `to_python_value`: https://github.com/flyteorg/flytekit/blob/ae9c6f8de21eb716dbda8d1af555c2057655db1d/flytekit/core/type_engine.py#L1181

which is what Eduardo's PR modifies https://github.com/flyteorg/flytekit/pull/2685/files#diff-68624e201f644896b015de1a5716eabef9b8b5a106caba1d9f21c51c0708684bR1101

katrogan: cc @eapolinario let me know if this sounds correct to you, would love someone from OSS to verify 🙂

flytepropeller/pkg/controller/nodes/common/utils.go
pmahindrakar-oss: thanks removing this.

flytepropeller/pkg/controller/nodes/common/utils.go
pmahindrakar-oss: Discussed offline that we should be ok right now to keep this in utils and use it in array handler. Once we move to individual task level literal checks we can reuse or move it accordingly

flytepropeller/pkg/controller/nodes/array/handler.go
pmahindrakar-oss: thanks  for the catch . removing the SDK version check here

flytepropeller/pkg/controller/nodes/common/utils.go
pmahindrakar-oss: cc : @wild-endeavor @pingsutw

pingsutw: > anything we serialize from flytekit should have the runtime set from the SDK, right?

yes, you should be able to see these metadata on flyteconsole for every kind of task.

However, this is the Flytekit version you use locally, and it doesn't match the one you use in the image.

In other words, users may use flytekit 1.13.5 to serialize the workflow, but use flytekit 1.12 in the container.

Screenshot 2024-09-12 at 2 45 29 PM

pingsutw: > for agents in particular, we use the pyflyte entrypoint for all task types so I believe we should get typetransformer changes for reading offloaded inputs for free

yes, type transformer is used to ser/de literal, so we should only change the transformer.

https://github.com/flyteorg/flytekit/blob/c06ef30518dec2057e554fbed375dfa43b985c60/plugins/flytekit-bigquery/flytekitplugins/bigquery/agent.py#L52

katrogan: thanks @pingsutw I think following the implementation of `literal_map_to_kwargs` which calls

to_python_value: https://github.com/flyteorg/flytekit/blob/ae9c6f8de21eb716dbda8d1af555c2057655db1d/flytekit/core/type_engine.py#L1181

this does trigger Eduardo's modified code path https://github.com/flyteorg/flytekit/pull/2685/files#diff-68624e201f644896b015de1a5716eabef9b8b5a106caba1d9f21c51c0708684bR1101

 
the more concerning bit is that the runtime metadata doesn't correspond to the container image. is the only way to verify the SDK version using the code to introspect the image then?

pmahindrakar-oss: >users may use flytekit 1.13.5 to serialize the workflow, but use flytekit 1.12 in the container.

Which should be fine , the version checks make sure of the compatibility 


>yes, type transformer is used to ser/de literal, so we should only change the transformer.
Given the above path @katrogan pointed out, i dont see we need to change anything here as the base changes already take care of it.

pingsutw: > the more concerning bit is that the runtime metadata doesn't correspond to the container image. is the only way to verify the SDK version using the code to introspect the image then?

I think so. We could probably release flytekit first and ensure most users have upgraded it.

pmahindrakar-oss: We handle the flytekit version check, so IMO it shouldn't block this PR and if folks are ok with the current version of the code then lets check this in unless there are any other concerns. @hamersaw @katrogan @pingsutw

katrogan: sounds good to me, I think this is the best check we have for compatibility

flytepropeller/pkg/controller/config/config.go
```diff
@@ -120,47 +124,94 @@ var (
 			EventVersion:               0,
 			DefaultParallelismBehavior: ParallelismBehaviorUnlimited,
 		},
+		LiteralOffloadingConfig: LiteralOffloadingConfig{
+			Enabled: true, // Default keep this enabled and in case of issues allow users to disable using config map.
```
katrogan, line 128: until we have the SDK version checked in, I think we should keep this false just in case it ends up taking multiple releases for the expected SDK change

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -547,6 +547,18 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 			return handler.UnknownTransition, fmt.Errorf("worker error(s) encountered: %s", workerErrorCollector.Summary(events.MaxErrorMessageLength))
 		}
 
+		// only offload literal if canOffloadLiteral is true which currently checks if the sdk is compatible to use with this backend change.
```
hamersaw, line 550: Can we update the comment? This isn't checking sdk version now IIUC.

hamersaw (dismissed):
Small nit.

hamersaw (approved):

Issue #4358, opened 2023-11-03
[BUG] Unable to compile task with (NamedTuple, str) as the output


ByronHsu:
### Describe the bug

```
from typing import NamedTuple
from flytekit import task, workflow

class TaskOutputs(NamedTuple):
    train_data: str
    val_data: str


# namedtuple + str
@task
def t() -> (TaskOutputs, str):
    return TaskOutputs("train", "val"), "foo"

@workflow
def wf():
    t()
```

```
$ pyflyte run --remote test_workflow.py wf                  
Failed with Unknown Exception <class 'flytekit.core.type_engine.RestrictedTypeError'> Reason: Transformer for type <class 'tuple'> is restricted currently
Transformer for type <class 'tuple'> is restricted currently
```

### Expected behavior

Able to compile

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

Mecoli1219:
#take

Issue #3158, opened 2022-12-15
[BUG] dict type converts int values to float


cosmicBboy:
### Describe the bug

This is an issue in unionml, where hyperparameters fall back to using `dict` or `typing.Dict` automatically converts ints to floats:

```python
from flytekit import task, workflow
from typing import Dict


@task
def t1(x: Dict) -> Dict:
    return x


@workflow
def wf() -> Dict:
    return t1(x={"a": 100})


if __name__ == "__main__":
    print(wf())
    # output: {'a': 100.0}

```

This is a problem because some frameworks like sklearn will do hyperparameter argument validation:
```
InvalidParameterError: The 'max_iter' parameter of LogisticRegression must be an int in the range [0, inf). Got 1000.0 instead.
```

### Expected behavior

The types inside `dict` should be preserved

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

cosmicBboy:
ideas moving forward:

- support for tuples: if this was supported, we could encode generic dictionaries as tuples when encoding `LiteralType`
- logical types: use python value when deriving literal type

cosmicBboy:
Wanna say this is still an issue... types ought to be preserved even with a dict type.

Issue #1337, opened 2021-08-14
flytekit - support NamedTuple as input type


tekumara:
**Motivation: Why do you think this is important?**

[NamedTuples](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/named_outputs.html) can be used to name outputs of tasks and workflows. But when used as inputs to a task, eg:

```
from typing import NamedTuple
from flytekit import task

nt = NamedTuple("nt", its_a_kind_of=str)

# this works
@task
def generate_reference_data() -> nt:
    return nt(its_a_kind_of="MAGIC")

# this will error
@task
def process_reference_data(data: nt) -> str:
    return data.its_a_kind_of
```

Flytekit will error with:

```
Transformer for type <class 'tuple'> is restricted currently
```

**Goal: What should the final outcome look like, ideally?**

Ideally a NamedTuple could be passed directly as an input to a task.

**Describe alternatives you've considered**

At the moment, NamedTuples have to be manually unpacked and their individual fields passed as arguments.

kumare3:
@tekumara we have reserved namedtuples to created named outputs from tasks and workflows.
You can instead use- data classes would that not suffice?

tekumara:
Dataclasses would work but would require converting the output into a dataclass before passing to the next task I think?

A NamedTuple input would be more fluent when composing workflows, eg: composing the tasks above could just be

```
@workflow
def main():
    reference_data = generate_reference_data()
    result = process_reference_data(reference_data)
```

PR #5699, opened 2024-08-27
[RFC] Tuple IDL

Mecoli1219:
## Tracking issue
https://github.com/flyteorg/flyte/issues/1337
https://github.com/flyteorg/flyte/issues/3158
https://github.com/flyteorg/flyte/issues/4358

- [] All new and existing tests passed.

### DIFF
--- a/rfc/system/4358-tuple-idl.md
+++ b/rfc/system/4358-tuple-idl.md
@@ -8,116 +8,117 @@
 
 Goals:
 
-- Implement support for tuple types (list of different types) in FlyteIDL
-  - Example: `a : Tuple[int, str, float]` must be assigned with integer for first element, string for second element, and floating value for third element.
-- Implement support for new Python types in Flytekit
-  - `typing.NamedTuple`
-    - `typing.NamedTuple("NAME", ("K1": T1), ("K2": T2))`
-  - `typing.Tuple`:
-    - `typing.Tuple[T1, T2, T3]` (Simplify case for NamedTuple)
-    - `typing.Tuple[T1, ...]` (Not the goal of this tuple IDL, since it is quite similar to `typing.List[T1]`)
+- Implement support for tuple types (lists of different types) in FlyteIDL.
+	- Example: a : Tuple[int, str, float] must be assigned an integer for the first element, a string for the second element, and a floating value for the third element.
+- Implement support for new Python types in Flytekit:
+	- `typing.NamedTuple`
+	    - `typing.NamedTuple("NAME", ("K1", T1), ("K2", T2))`
+	- `typing.Tuple`
+	    - `typing.Tuple[T1, T2, T3]` (Simplified case for NamedTuple)
+	    - `typing.Tuple[T1, ...]` (It is quite similar to `typing.List[T1]`, we can address it in this RFC or use `LiteralType_CollectionType` to support it for now.)
 
 Implementation:
 
-- Each element in the tuple should be considered independently, while the order of elements should be preserved. In order to support NamedTuple, we need to store the name of each element.
-  ```proto
-  message LiteralField {
-      string name = 1;
-      Literal literal = 2;
-  }
-  ```
+-  A new `LiteralType` will be introduced to represent tuple types in FlyteIDL. Each element in the tuple will be considered independently, preserving the order of elements. For NamedTuple, the name of each element will be stored along with its type, while for regular tuples, the names will be left empty.
+    ```proto
+    message TupleFieldType {
+        string name = 1;
+        LiteralType type = 2;
+    }
+
+    message TupleType {
+        string tuple_name = 1;
+        repeated TupleFieldType fields = 2;
+    }
+
+    message LiteralType {
+        oneof type {
+            // ...
+            TupleType tuple_type = 11;
+        }
+        // ...
+    }
+    ```
 
 ## 2 Motivation
 
-Flytekit restricts the support of tuple types currently, which are commonly used in Python. This RFC aims to add support for tuple types in FlyteIDL and Flytekit.
+Flytekit currently restricts the support of tuple types, which are commonly used in Python. This RFC aims to add support for tuple types in FlyteIDL and Flytekit.
 
-Note: We have several issues about NamedTuple and tuple support in Flytekit.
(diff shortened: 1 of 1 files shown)


rfc/system/4358-tuple-idl.md
```diff
@@ -0,0 +1,284 @@
...
+    ```
+
+- The above code should be added to [flyteidl/protos/flyteidl/core/types.proto](https://github.com/flyteorg/flyte/blob/master/flyteidl/protos/flyteidl/core/types.proto).
+
+#### Literal of NamedTuple & Tuple
+
+- Since the type, including the names of the fields and the tuple, is stored in `LiteralType`, we only need to store the value of each field in the tuple. We can use the existing `LiteralCollection` message from [flyteidl/protos/flyteidl/core/types.proto](https://github.com/flyteorg/flyte/blob/master/flyteidl/protos/flyteidl/core/literals.proto) to store the value of each field.
+
+    ```proto
+    message LiteralCollection {
```
pingsutw, line 100: I think we still need a new literal for tuple. Otherwise, we can't properly get the literal type from the literal [here](https://github.com/flyteorg/flyte/blob/1cd775a417e4280fd78c9334f7e9329e53b017d3/flytepropeller/pkg/compiler/validators/utils.go#L247).

rfc/system/4358-tuple-idl.md
```diff
@@ -0,0 +1,284 @@
...
+    }
+    ```
+
+#### Other Changes
+
+- Update `MakeDefaultLiteralForType()` and `MakeLiteralForType()` functions in [flyteidl/clients/go/coreutils/literals.go](https://github.com/flyteorg/flyte/blob/master/flyteidl/clients/go/coreutils/literals.go)
+
+### FlytePropeller
+
+1. Implement a new type checker in [flytepropeller/pkg/compiler/validators/typing.go](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/compiler/validators/typing.go). We need to discuss whether to check the name of each field and the tuple itself, as this will affect the casting between NamedTuple and Tuple. (#TO_DISCUSS)
```
pingsutw, line 121: > We need to discuss whether to check the name of each field and the tuple itself

yup. I think we need it

rfc/system/4358-tuple-idl.md
```diff
@@ -0,0 +1,284 @@
...
+
+None
+
+## 6 Alternatives
+
+None
+
+## 7 Potential Impact and Dependencies
+
+- This feature will affect FlyteIDL, FlytePropeller, and FlyteKit.
```
pingsutw, line 273: also flytectl and flyteconsole

pingsutw:
cc @fg91 I know you want this too

fg91:
> cc @fg91 I know you want this too

Yes, I started an incubator discussion about this a while back: https://github.com/flyteorg/flyte/discussions/4544

rfc/system/4358-tuple-idl.md
```diff
@@ -0,0 +1,562 @@
...
+- Implement support for tuple types (lists of different types) in FlyteIDL.
+    - Example: a : Tuple[int, str, float] must be assigned an integer for the first element, a string for the second element, and a floating value for the third element.
+- Implement support for new Python types in Flytekit:
+    - `typing.NamedTuple`
+        - `typing.NamedTuple("NAME", ("K1", T1), ("K2", T2))`
+    - `typing.Tuple`
+        - `typing.Tuple[T1, T2, T3]` (Simplified case for NamedTuple)
+        - `typing.Tuple[T1, ...]` (It is quite similar to `typing.List[T1]`, we can address it in this RFC or use `LiteralType_CollectionType` to support it for now.)
+
+Implementation:
```
fg91, line 20: Nit: could you please move this further below? I feel this is too much detail for the executive summary.

fg91: Let's move this to `## 3 Proposed Implementation`.

Mecoli1219 (commented):
@task
def my_task() -> Tuple[Tuple[int, str], str]:
    return (1, "foo"), "bar"
    # protobuf: {
    #     o1: {
    #         tuple_name: ""
    #         fields: [
    #             {name: "", value: 1},
    #             {name: "", value: "foo"}
    #         ]
    #     }
    #     o2: "bar"
    # }

cosmicBboy:
One comment here re: `FlyteRemote` API. When accessing outputs in the `{Workflow, Task}Execution` objects, one thought is to introduce another attribute `execution.output` (as opposed to the current `outputs` attribute) since tasks will only ever output single output (since multiple outputs should be wrapped in `Tuple[...]`)

Mecoli1219:
I thought Flyte implements multiple outputs for task/workflow (aka tuple-wrapped output) with `LiteralMap`, which is quite different from the new IDL proposed by this RFC. In my implementation [#5720](https://github.com/flyteorg/flyte/pull/5720), I used other types that are more similar to the original implementation, including a `map[string]Literal` type and also two more extra fields (`string tuple_name` and `[]string order`) to enable the reconstruction of the Python `tuple` and `NamedTuple` types. 

---

@fg91 After reading the original incubator discussion, there are some points I would like to address. 

1. I don't think `def my_task() -> (int, str)` is the legal output annotation in Python, but Flyte supports it somehow.
2. As mentioned in [official doc](https://docs.python.org/3/library/typing.html#typing.Tuple), `typing.Tuple` is a deprecated alias for `tuple` after Python version 3.9, so I think we can view them as the same type in flytekit. In my current implementation, I will treat both of them the same.

--- 
> one thought is to introduce another attribute execution.output (as opposed to the current outputs attribute) since tasks will only ever output a single output (since multiple outputs should be wrapped in Tuple[...])

I believe that this can be enabled in the future. I have tried to make my implementation close to the multiple outputs structure.
However, I am not sure whether we should add a new attribute `output`, or replace `outputs` with `output`. Also, I am not really sure how this will affect the flyte system, could anyone provide more comments?

Future-Outlier (commented):
Why int and str tuple key attribute access?
NamedTuple: order(int) and key(str) both supported
Tuple: order(int) only

davidmirror-ops:
10/24/2024 Contributor's sync notes: looking to have a less controversial approach by not creating a new Literal for Tuple. Michale running experiments

Issue #5701, opened 2024-08-28
[Core feature] Allow flyteadmin to start even if OIDC is unavailable (Improve flyteadmin startup resiliency)


ddl-rliu:
### Motivation: Why do you think this is important?

Today, the flyteadmin pod is blocked from starting up until the OIDC provider is healthy and available (the pod gets stuck in Error state). In some Kubernetes configurations, this erroring-pod could cause deployment-wide issues. The current behavior could be made more resilient.

(Note that this applies to configurations using `useAuth=true`)

### Goal: What should the final outcome look like, ideally?

A better approach in these configurations is to allow flyte to start up, even if the OIDC provider is unavailable. Then, try to re-initialize the OIDC provider later in the deployment lifespan. This is a more resilient approach, and it can be made configurable.

### Describe alternatives you've considered 

A workaround is to disable Flyte until the OIDC provider is available.

###  Propose: Link/Inline OR Additional context 

Proposed fix here: https://github.com/flyteorg/flyte/pull/5702

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

Sovietaced:
> A better approach in these configurations is to allow flyte to start up, even if the OIDC provider is unavailable. Then, try to re-initialize the OIDC provider later in the deployment lifespan. This is a more resilient approach, and it can be made configurable.

This is a double edged sword. If I recall correctly there was some work done recently to indefinitely cache something related to OIDC on boot. If the OIDC provider is down and you do a normal rolling deployment you could end up in a worse state since you previously had working pods but now you have broken ones.

eapolinario:
> If I recall correctly there was some work done recently to indefinitely cache something related to OIDC on boot.

@Sovietaced , can you point to this change? Are you thinking of https://github.com/flyteorg/flyte/pull/5621/?

Sovietaced:
> @Sovietaced , can you point to this change? Are you thinking of #5621?

Yeah I think so

PR #5702, opened 2024-08-28
[Core feature] Allow flyteadmin to start even if OIDC is unavailable (Improve flyteadmin startup resiliency)

ddl-rliu:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5701

Today, the flyteadmin pod is blocked from starting up until the OIDC provider is healthy and available (the pod gets stuck in Error state). In some Kubernetes configurations, this erroring-pod could cause deployment-wide issues. The current behavior could be made more resilient.

(Note that this applies to configurations using `useAuth=true`)

A better approach in these configurations is to allow flyte to start up, even if the OIDC provider is unavailable. Then, try to re-initialize the OIDC provider later in the deployment lifespan. This is a more resilient approach, and it can be made configurable.

Adds an `onlyStartIfOIDCIsAvailable` config which controls this behavior.

## How was this patch tested?

A writeup is here which shows the "good" flow when `onlyStartIfOIDCIsAvailable` is enabled and OIDC is unhealthy for a period: https://gist.github.com/ddl-rliu/4c09862404f46a5adbc451025160e0eb

### DIFF
--- a/flyteadmin/auth/config/config.go
+++ b/flyteadmin/auth/config/config.go
@@ -72,6 +72,7 @@ var (
 					"openid",
 					"profile",
 				},
+				OnlyStartIfOIDCIsAvailable: true,
 			},
 			CookieSetting: CookieSettings{
 				Domain:         "",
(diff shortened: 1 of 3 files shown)


flyteadmin/auth/config/config.go
```diff
@@ -72,6 +72,7 @@ var (
 					"openid",
 					"profile",
 				},
+				OnlyStartIfOIDCIsAvailable: true,
```
ddl-rliu, line 75: Configurable `onlyStartIfOIDCIsAvailable` is `true` by default. This matches the behavior today, where if OIDC is unavailable due to e.g. DNS issues, flyteadmin will not start.

For reference, another project which takes a similar approach to OIDC issues/service startup resiliency is https://github.com/juanfont/headscale/blob/main/hscontrol/app.go#L157-L166

eapolinario:
@Sovietaced [brings up a good point](https://github.com/flyteorg/flyte/issues/5701#issuecomment-2445893317) regarding this change.

eapolinario:

Issue #3991, opened 2023-08-26
[BUG] Investigate memory leak in single-binary deployment


eapolinario:
### Describe the bug

A user in the OSS slack [reported](https://flyte-org.slack.com/archives/CP2HDHKE1/p1692894721174089) that single-binary pods keep getting killed due to OOM errors. Here's a graph of memory consumption in this case (different colors refer to different pods):

![image](https://github.com/flyteorg/flyte/assets/653394/650b4d94-5355-4037-ace2-ab0de1debc1f)

They are running stock Flyte v1.8.1 release.

### Expected behavior

Memory consumption should remain relatively constant.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

wild-endeavor:
look at prometheus metrics, educated guess from dan - every time a task runs, prometheus metrics in propeller might be contributing to memory usage.

cpaulik:
I'm pretty sure this is what I'm seeing as well. See https://flyte-org.slack.com/archives/CP2HDHKE1/p1723819039330009 for more discussion.

eapolinario:
After some investigation, it turns out that accumulating metrics data is the expected behavior of the prometheus golang client, as per https://github.com/prometheus/client_golang/discussions/920. We emit high-cardinality metrics in flytepropeller, as an example you can see [here](https://gist.github.com/eapolinario/4626e39c4d40086454a79764c8d0df6c) the emitted metrics after running a few hundred workflow executions. Note how we maintain a lot of metrics per execution id. This is the [symptom](https://github.com/prometheus/client_golang/discussions/920#discussioncomment-1448783) described in the prometheus client gh issue.

One could argue that emitting metrics per-execution id is the wrong granularity, but that's exactly what we do in the case of [single-binary](https://github.com/flyteorg/flyte/blob/master/cmd/single/start.go#L252).

> would deploying via flyte-core help here?

Knowing what we know now, the set of default metrics emitted by flytepropeller in the [default cause](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/config/config.go#L111-L112) does not cause the issue. Similar argument can be made to [flyteadmin](https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/runtime/application_config_provider.go#L26-L29). In other words, this was a single-binary issue-only.

[...]

PR #5704, opened 2024-08-29
Do not emit execution id label by default in single binary

eapolinario:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/3991

Single binary emits execution-id labels to a bunch of flytepropeller and flyteadmin metrics, but since the golang prometheus golang client is setup in such a way that no metric is ever dropped (as described in https://github.com/prometheus/client_golang/discussions/920#discussioncomment-1448783), this ends up causing an apparent memory leak due to how these high cardinality metrics are stored in-memory.

This PR removes the execution-id label from the single binary metrics and adds a comment to hint at the potential problems caused by that setting.

I confirmed that the metric is not enabled by default in any component, nor is set in any of the shipped helm charts, while being left as a last resort debugging tool.

One could argue that we should never have had that label in the first place and instead reach out to other debugging means, like open telemetry, etc, but that's a separate discussion.

## How was this patch tested?
I had a local test where I'd have a workflow composed of a single cached task. Running that in a tight loop and confirming that the increase in memory of the single binary executable is now orders of magnitude slower than before the PR.

### DIFF
--- a/flytestdlib/contextutils/context.go
+++ b/flytestdlib/contextutils/context.go
@@ -12,14 +12,15 @@ import (
 type Key string
 
 const (
-	AppNameKey         Key = "app_name"
-	NamespaceKey       Key = "ns"
-	TaskTypeKey        Key = "tasktype"
-	ProjectKey         Key = "project"
-	DomainKey          Key = "domain"
-	WorkflowIDKey      Key = "wf"
-	NodeIDKey          Key = "node"
-	TaskIDKey          Key = "task"
+	AppNameKey    Key = "app_name"
+	NamespaceKey  Key = "ns"
+	TaskTypeKey   Key = "tasktype"
+	ProjectKey    Key = "project"
+	DomainKey     Key = "domain"
+	WorkflowIDKey Key = "wf"
+	NodeIDKey     Key = "node"
+	TaskIDKey     Key = "task"
+	// Adding this label to a metric will cause higher cardinality. Use with caution.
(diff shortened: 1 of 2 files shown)


wild-endeavor (dismissed):

flytestdlib/contextutils/context.go
```diff
@@ -12,14 +12,15 @@ import (
...
-	TaskIDKey          Key = "task"
+	AppNameKey    Key = "app_name"
+	NamespaceKey  Key = "ns"
+	TaskTypeKey   Key = "tasktype"
+	ProjectKey    Key = "project"
+	DomainKey     Key = "domain"
+	WorkflowIDKey Key = "wf"
+	NodeIDKey     Key = "node"
+	TaskIDKey     Key = "task"
+	// Adding this label to a metric will cause higher cardinality. Use with caution.
```
thomasjpfan, line 23: Nit: "this" is unclear to me:

```suggestion
	// Adding the ExecIDKey label to a metric will cause higher cardinality. Use with caution.
```

wild-endeavor (approved):

davidmirror-ops:
Thank you! This is great as it simplifies a bit some of the queries in the Grafana user dashboard ([example](https://github.com/flyteorg/flyte/blob/5c147eff5905e162244bd84995a25186208d8c5a/stats/flyteuser.dashboard.py#L182))

pingsutw (approved):

PR #5712, opened 2024-08-31
OIDC config to allow mismatched discovery / issuer

ddl-ebrown:

 - There are a number of cases where the OIDC discovery url returns one issuer, but its desirable to use a separately configured / named issuer for validation instead.

   There are cases in Azure where this is necessary due to their non-standard OIDC configuration -- which is why this was originally added: https://github.com/coreos/go-oidc/pull/315

   There are also cases where it's necessary to use an in-cluster service address, but browser clients are using the external ingress address. Due to cluster DNS configuration, it's possible that flyteadmin may be unable to resolve or use the public ingress address for an Idp, but the internal service address is available. This configuration change allows for that.

 - Regenerated flyteadmin code through mise with:
 > mise x go@1.22 -- make generate

### DIFF
--- a/flyteadmin/auth/auth_context.go
+++ b/flyteadmin/auth/auth_context.go
@@ -141,6 +141,12 @@ func NewAuthenticationContext(ctx context.Context, sm core.SecretManager, oauth2
 	// Construct an oidc Provider, which needs its own http Client.
 	oidcCtx := oidc.ClientContext(ctx, httpClient)
 	baseURL := options.UserAuth.OpenID.BaseURL.String()
+	// use a different issuer for token validation if configured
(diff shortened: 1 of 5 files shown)


wild-endeavor (approved):
Had a chat with @EngHabu.  This pr is good to go thank you!  There was some discussion about whether this config option needs to be piped in to the resource_server so that the issuer can be verified but it looks like we don't do that.  In the self auth server case, the issuer is already read from config or matched against authorized urls so that's okay.

wild-endeavor:
mind taking care of the checks though @ddl-ebrown

Sovietaced (approved):

PR #5717, opened 2024-09-02
Refactor flyteadmin to pass proto structs as pointers

Sovietaced:

Go structs generated by `protoc-gen-go` contain a [DoNotCopy](https://github.com/protocolbuffers/protobuf-go/blob/v1.34.2/internal/pragma/pragma.go#L29) mutex so that `go vet` will complain and discourage people from copying these go structs.

The `flyteadmin` code doesn't respect these intentions and has many of these warnings. 1,135 to be exact. The orange warnings are a bit noisy in the IDE and I noticed the `golangci` config ignores them.

As someone who comes from a Java background I'm guessing that the intent of passing around copies of these go structs is to prevent nil pointer errors and employ defensive programming. I think in most circumstances this makes sense, but in this case gRPC and `protoc-gen-go` are very intentional about using pointers. The structs are never nil in the gRPC API handlers, they are just passed with pointers so that the structs are used safely. 

Having said that, this pull request attempts to refactor `flyteadmin` to use pointers for protobuf structs throughout the codebase to eliminate these warnings. It updates the `golangci` config to ignore the singular unavoidable and safe copy that is made. 

see: https://stackoverflow.com/questions/64183794/why-do-the-go-generated-protobuf-files-contain-mutex-locks
 
## How was this patch tested?
Unit tests and integration test have been updated. 

### DIFF
--- a/flyteadmin/.golangci.yml
+++ b/flyteadmin/.golangci.yml
@@ -36,5 +36,6 @@ linters-settings:
       - prefix(github.com/flyteorg)
     skip-generated: true
 issues:
-  exclude:
-    - copylocks
+  exclude-rules:
+    - path: pkg/workflowengine/impl/prepare_execution.go
+      text: "copies lock"
--- a/flyteadmin/pkg/manager/impl/testutils/mock_requests.go
+++ b/flyteadmin/pkg/manager/impl/testutils/mock_requests.go
@@ -47,8 +47,8 @@ func GetValidTaskRequest() *admin.TaskCreateRequest {
 	}
 }
 
-func GetValidTaskRequestWithOverrides(project string, domain string, name string, version string) admin.TaskCreateRequest {
-	return admin.TaskCreateRequest{
+func GetValidTaskRequestWithOverrides(project string, domain string, name string, version string) *admin.TaskCreateRequest {
+	return &admin.TaskCreateRequest{
 		Id: &core.Identifier{
 			ResourceType: core.ResourceType_TASK,
 			Project:      project,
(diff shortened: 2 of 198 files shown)


flyteadmin/.golangci.yml
```diff
@@ -36,5 +36,6 @@ linters-settings:
       - prefix(github.com/flyteorg)
     skip-generated: true
 issues:
-  exclude:
-    - copylocks
+  exclude-rules:
```
Sovietaced, line 39: Fine grained ignore

flyteadmin/pkg/manager/impl/testutils/mock_requests.go
```diff
@@ -78,20 +78,15 @@ func GetValidTaskRequestWithOverrides(project string, domain string, name string
 	}
 }
 
-func GetValidTaskSpecBytes() []byte {
```
Sovietaced, line 81: This was unused so I just removed it

katrogan: sounds good and thank you!

flyteadmin/.golangci.yml
katrogan: why does this still need to be excluded?

katrogan (approved):
this is awesome!

flyteadmin/.golangci.yml
Sovietaced: https://github.com/flyteorg/flyte/pull/5717/files/0f75532e95adab4fb43c5ce380a9b11808eced30#diff-68d7b4389844bb5dec320a0854bdb20122c9fcb4356c32dce5dff80a55ee486eR33-R34

The code copies the a security context to the flyte workflow custom resource, which stores it as a non-pointer so I think its inevitable unless I write some conversion method to traverse the security context and reconstruct a non-pointer version. I felt it was probably less error prone and more future proof to do a deep copy and ignore the warning.

katrogan: thanks for explaining, leaving this as an exception makes sense!

PR #5720, opened 2024-09-03
[1st version] Add Tuple idl

Mecoli1219:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5908
https://github.com/flyteorg/flyte/issues/1337
https://github.com/flyteorg/flyte/issues/3158
https://github.com/flyteorg/flyte/issues/4358

## Related PRs
RFC: https://github.com/flyteorg/flyte/pull/5699


flyteidl/clients/go/coreutils/literals.go
```diff
@@ -637,6 +667,47 @@ func MakeLiteralForType(t *core.LiteralType, v interface{}) (*core.Literal, erro
...
+		//
+		// [Example usage]
+		// inputs:
+		//     tuple_data:
+		//         key1: "foo"
+		//         key2: 123
+
+		vMap, ok := v.(map[string]interface{})
+		if !ok {
+			return nil, errors.Errorf("Expected a map[string]interface{} for tuple type, got [%v]", v)
```
Future-Outlier, line 681: Can you explain more?

Mecoli1219: In this part, the LiteralType is already given, so the name of the tuple and the order of each field should be already defined. The only thing that is missing is the value of each field.
Therefore, we only need to provide the value of each tuple field (via the key-value pair of each field) in `flytectl` for the registered task or workflow for further execution needs.

flyteidl/protos/flyteidl/core/literals.proto
```diff
@@ -140,6 +143,18 @@ message LiteralMap {
...
+// A collection of fields for tuple literal.
+message LiteralTupleMap {
+    // The name of the NamedTuple. If it is original tuple, it would be empty string.
+    string tuple_name = 1;
+
+    // The order of each fields stored in the tuple.
+    repeated string order = 2;
+    
+    // A map of literals.
+    map<string, Literal> literals = 3;
```
pingsutw, line 155: Can we use `Literal.collection`?

flyteidl/protos/flyteidl/core/types.proto
```diff
@@ -108,6 +108,14 @@ message UnionType {
     repeated LiteralType variants = 1;
 }
 
+// Defines a named tuple type
+// TODO: Add more docs
+message TupleType {
+    string tuple_name = 1;
+    repeated string order = 2;
+    map<string, LiteralType> fields = 3;
```
pingsutw, line 116: Can we use repeated here, like

```proto
message TupleEntry {
  string name = 1;
  LiteralType literal_type = 2;
}
message TupleType {
    string tuple_name = 1;
    repeated TupleEntry = 2;
}
```

flyteidl/protos/flyteidl/core/literals.proto
Mecoli1219: `Literal.collection` is not enough to store the name of the fields for `NamedTuple`.

flyteidl/protos/flyteidl/core/types.proto
Mecoli1219: > Other Considerations
An alternative approach to storing the literal values for NamedTuple and Tuple would be combining the order and the map into a list of messages, as shown below:

// !!! This is not the proposed implementation !!!
message FakeLiteralTupleField {
    string name = 1;
    Literal value = 2;
}

message FakeLiteralTupleMap {
    string tuple_name = 1;
    repeated FakeLiteralTupleField fields = 2;
}
While this structure may seem more straightforward, it has a drawback: looking up a field’s value by name requires iterating through the list, which results in O(n) complexity. In contrast, using a map allows for more efficient O(1) lookup times when retrieving the value of a field by its name. Although in practice the tuple size is usually small, the map-based approach is more scalable and efficient.

PR #5723, opened 2024-09-04
Remove explict go toolchain versions

ddl-ebrown:

 - https://github.com/flyteorg/flyte/pull/5032 updated builds to use Go 1.22 -- but in practice it looks like flytectl is being built with an outdated toolchain based on Go 1.22.0 instead of Go 1.22.5 or 1.22.6 (i.e. latest)

   For instance, flytectl v0.9.0 reports being built with Go 1.22.0

>   ❯ go version -m -v 
>   flytectl flytectl: go1.22.0

   Ensure the proper Go environment for security reasons

### DIFF
--- a/datacatalog/go.mod
+++ b/datacatalog/go.mod
@@ -2,8 +2,6 @@ module github.com/flyteorg/flyte/datacatalog
 
 go 1.22
 
-toolchain go1.22.1
-
 require (
 	github.com/Selvatico/go-mocket v1.0.7
 	github.com/flyteorg/flyte/flyteidl v0.0.0-00010101000000-000000000000
--- a/flytectl/go.mod
+++ b/flytectl/go.mod
@@ -2,8 +2,6 @@ module github.com/flyteorg/flyte/flytectl
 
 go 1.22
 
-toolchain go1.22.0
-
 require (
 	github.com/apoorvam/goterminal v0.0.0-20180523175556-614d345c47e5
 	github.com/avast/retry-go v3.0.0+incompatible


eapolinario:
> If I build source locally, I get my local 1.23.0 toolchain already

https://go.dev/doc/toolchain#select has the answer for this. It's kind of convoluted, but essentially, if `GOTOOLCHAIN` is set to `auto` and you have a lower version than the one set in the `go.mod` file, then the version set in the `go.mod` is installed and used to run the go command.

I verified this by running:
```
❯ docker run --rm -it --workdir /code --volume $(pwd):/code golang:1.21.12-bookworm bash
root@fff6de1b2bf3:/code# go version
go version go1.21.12 linux/arm64
root@fff6de1b2bf3:/code# GOTOOLCHAIN=auto make -C flytectl compile
make: Entering directory '/code/flytectl'
go build -o bin/flytectl -ldflags="-s -w -X github.com/flyteorg/flyte/flytestdlib/version.Version=v0.9.1-39-gd4db50cb2 -X github.com/flyteorg/flyte/flytestdlib/version.Build=d4db50cb2 -X github.com/flyteorg/flyte/flytestdlib/version.BuildTime=2024-09-04" main.go
go: downloading go1.22.0 (linux/arm64)
...
root@fff6de1b2bf3:/code# go version -m ./flytectl/bin/flytectl | head -n1
./flytectl/bin/flytectl: go1.22.0
```

That env var is set in the case of that gh action: https://github.com/flyteorg/flyte/actions/runs/10269861937/job/28416285221#step:3:42.

If you run the same code but using the `golang:1.23.0-bookworm` image you'll see that the version is set to go1.23.0.


> how flytectl is being built / shipped

we use this gh action to release flytectl: https://github.com/flyteorg/flyte/blob/master/.github/workflows/flytectl-release.yml.

The last time we released flytectl: https://github.com/flyteorg/flyte/actions/runs/10269861937. Notice how we specify go 1.21 but somehow we end up with go 1.22.0 installed. Reading [the code](https://github.com/actions/setup-go/blob/b26d40294f8ad76fcc90b915dac85892322fe62d/src/main.ts#L139) for that specific version of `setup-go@v4` I don't see how we end up installing 1.22.0. 

That said, I feel like removing toolchain is the right call. Also, we should probably switch to setting up the [go version from the go.mod](https://github.com/actions/setup-go/tree/v4?tab=readme-ov-file#getting-go-version-from-the-gomod-file) file (so that it's easier to maintain the right version across go.mod and gh workflows). wdyt, @ddl-ebrown ?

eapolinario (approved):

eapolinario:
> we should probably switch to setting up the [go version from the go.mod](https://github.com/actions/setup-go/tree/v4?tab=readme-ov-file#getting-go-version-from-the-gomod-file) file (so that it's easier to maintain the right version across go.mod and gh workflows).

Trying out this idea in https://github.com/flyteorg/flyte/pull/5736.

PR #5724, opened 2024-09-04
fix: Use deterministic execution names in scheduler

pingsutw:
## Tracking issue
NA

https://github.com/flyteorg/flyte/pull/5637/files changes the execution name to a random string in the scheduler, but it should use a deterministic name. Otherwise, the scheduler will create multiple executions for the same time slot.

generate a deterministic hash and than use it to generate a human id

## How was this patch tested?
unit test

## Related PRs
NA

## Docs link
NA


eapolinario (approved):
good catch!

pmahindrakar-oss (approved):

PR #5739, opened 2024-09-11
Enable caching in flyte-core helmchart

cpaulik:

Without this caching was actually not enabled when deploying with the flyte-core helm chart

Adding `storage` to the `cache.yaml` file so that the [configuration values](https://docs.flyte.org/en/latest/deployment/configuration/generated/flytepropeller_config.html#cache-storage-cachingconfig) for `storage.cache.max_size_mbs` and `storage.cache.target_gc_percet` are picked up correctly

## How was this patch tested?

I've deployed flyte core without this change and caching was not enabled. After I added 

```
configmap:
  core:
    storage:
      cache:
        max_size_mbs: 5000
        target_gc_percent: 70
```

caching started working so I assume that adding `storage` in the `cache.yaml` would also work.


cpaulik:
For DCO I'm waiting for our legal department so this might take a bit for the first contribution.

cpaulik:
DCO should be fixed

pingsutw (approved):
LGTM, thank you for fixing it

davidmirror-ops:
@cpaulik Thank so much! Also to fix the problem detected by generate_helm, you should be able to run `make helm` in your fork and push the changes

cpaulik:
I ran `make helm` and updated the commit

eapolinario (approved):
Thank you.

PR #5556, opened 2024-07-11
SecurtyContext: applying PodTemplate for InitContainers and removing SYS_PTRACE 

vlibov:
## Tracking issue

Related to #5462

## Why are the changes needed?

The Issue https://github.com/flyteorg/flyte/issues/5462 describes the problem in detail.

## What changes were proposed in this pull request?

1. Remove the SYS_PTRACE capability
2. Apply the PodTemplate settings to the InitContainers. In particular, this allows the SecurityContext settings to be applied also to the InitContainer.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

Issue #5462, opened 2024-06-10
More flexible configuration of SecurityContext for Pods/Containers started by flyte


vlibov:
### Motivation: Why do you think this is important?

To run workloads in a security-hardenend kubernetes cluster, the pods and containers must fullfill certain requirements, in particular regarding the SecurityContext fields (pod level, container level).

It is possible to set those in flyte e.g. by utilizing the PodTemplate native k8s resource. However it seems that these settings won't be applied to the initContainers (flyte-copilot-downloader). Furthermore, there are some settings which are applied to the main container (addéd capability SYS_PTRACE) without a possibility to disable it.

This prevents running flyte workflows in a hardened namespace, which is somewhat unfortunate.

It would be nice to be able to configure those, i.e. apply the PodTemplate also for the initContainers and make the SYS_PTRACE capability optional. 

### Goal: What should the final outcome look like, ideally?

Ultimately we need to

1. Be able to disable the SYS_PTRACE capability, i.e. make it optional in the config
2. Apply PodTemplates also to the initContainers. 

### Describe alternatives you've considered 

One alternative is to run flyte-workloads in a separate namespace which has more elevated rights (i.e. less secure) and restrict the users from creating workloads in the namespace direclty. However this still poses a security risk, because the user still can create workloads indirectly (e.g. create a malicious python code or image and start it as a flyte worklfow).

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

vlibov:
Including a configuration for initContainer in the podtemplate as suggested above leads to a duplication error:

`Workflow[kodi:development:workflows.kodi-training.kodi_training_wf] failed. RuntimeExecutionError: max number of system retry attempts [11/10] exhausted. Last known status message: failed at Node[n0]. RuntimeExecutionError: failed during plugin execution, caused by: failed to execute handle for plugin [container]: [Invalid] failed to create resource, caused by: Pod "aqmll4kts4gghgvgqljv-n0-0-n0-n1-0-n0-0" is invalid: spec.initContainers[1].name: **Duplicate value: "flyte-copilot-downloader"**`

vlibov:
https://github.com/vlibov/flyte/tree/podsecurity

davidmirror-ops:
@vlibov do you plan to contribute the proposed changes?

vlibov:
Hi @davidmirror-ops, yes!

davidmirror-ops:
[...]

PR #5746, opened 2024-09-13
Add SYS_PTRACE setting to Co-pilot container config.

dylanspag-lmco:
## Tracking issue

Related to #5462

The issue https://github.com/flyteorg/flyte/issues/5462 describes the problem in detail.

I made `SYS_PTRACE` an optional setting for co-pilot containers.

## How was this patch tested?

I added the default setting to the "happy" and "happy stow backend" Co-pilot Container tests. I also added a test to ensure the SYS_PTRACE capability is present when the setting is enabled.

## Related PRs

See #5556.

### DIFF
--- a/docs/deployment/configuration/generated/flyteadmin_config.rst
+++ b/docs/deployment/configuration/generated/flyteadmin_config.rst
@@ -2720,7 +2720,7 @@ k8s (`config.K8sPluginConfig`_)
     output-vol-name: flyte-outputs
     start-timeout: 1m40s
     storage: ""
-    add-sys-ptrace-capability: false
+    add-sys-ptrace-capability: true
   create-container-config-error-grace-period: 0s
   create-container-error-grace-period: 3m0s
   default-annotations:
@@ -3041,7 +3041,7 @@ Co-Pilot Configuration
   output-vol-name: flyte-outputs
   start-timeout: 1m40s
   storage: ""
-  add-sys-ptrace-capability: false
+  add-sys-ptrace-capability: true
   
 
 delete-resource-on-finalize (bool)
@@ -3386,7 +3386,7 @@ Used to enable SYS_PTRACE for co-pilot containers
 
 .. code-block:: yaml
 
-  "false"
+  "true"
   
 
 resource.Quantity
--- a/docs/deployment/configuration/generated/flyteadmin_config.rst
+++ b/docs/deployment/configuration/generated/flyteadmin_config.rst
@@ -2720,6 +2720,7 @@ k8s (`config.K8sPluginConfig`_)
     output-vol-name: flyte-outputs
     start-timeout: 1m40s
     storage: ""
+    add-sys-ptrace-capability: false
   create-container-config-error-grace-period: 0s
   create-container-error-grace-period: 3m0s
   default-annotations:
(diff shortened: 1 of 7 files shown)


neverett:
Pinging @davidmirror-ops for deployment docs changes.

davidmirror-ops:
@dylanspag-lmco Thank you for contributing! I was wondering if this wouldn't be a breaking change once released? Shouldn't we default to opt-out?

dylanspag-lmco:
@davidmirror-ops I'm not sure of all of the downstream implications. The conversation in #5462 seemed to indicate removing it might be relatively low impact, which is why I had set the default as opt-in. Which would you prefer?

davidmirror-ops:
@dylanspag-lmco I think the goal is to make it configurable as opposed to hardcoded. Disabling it by default may break functionality for current users.

dylanspag-lmco:
@davidmirror-ops makes sense. I just swapped the default behavior and updated the tests accordingly.

davidmirror-ops:
@dylanspag-lmco thanks so much for your effort so far. I've been talking to other maintainers about this change and we think the best path forward is to no accept it as it would break `RawContainer` functionality, the only instance where the copilot container is used in Flyte.
Flyte needs a way to determine that the user code finished execution and to capture the outcome (Success/Failure) and three mechanisms were envisioned to achieve that goal as described in the [original flytecopilot design](https://github.com/flyteorg/flyte/tree/master/flytecopilot#raw-notes)

1. Poll the kube-api: it would require less process on the user code but a deeper investigation/benchmarking to determine the sane defaults to configure the kube-client and avoid overloading the kube-api server. Recent testing showed that this won't work, and it would require to spend time updating [the code](https://github.com/flyteorg/flyte/blob/master/flytecopilot/cmd/containerwatcher/kubeapi_watcher.go).
2. [Implement a file-based protocol](https://github.com/flyteorg/flyte/blob/master/flytecopilot/cmd/containerwatcher/file_watcher.go). With this option, the user code had to include logic to drop the outputs to a certain directory and make the User code use a FileWatcher to determine the result of its own execution. Depending on the outcome (SUCCESS/FAILURE) it would need to write a different file, all of that before returning. Recent testing shows that it works butthere are potential performance implications going this route.Also, there is now wat to inspect that user code ran, just waiting for a file to drop.

Notice that for the two options above there is no way to activate them or select them, no flag.

4. Share process namespace between copilot and the user code container (current default). With this mechanism, the copilot container downloads the inputs, watches the user code container for a result (using the shared process namespace and, hence, requiring the `SYS_PTRACE` capability), and copies the outputs, or updates flyteadmin,  depending on whether the task succeeded or failed, respectively.

Considering this context, we cannot make `SYS_PTRACE` configurable because then we'd also have to disable `RawContainer`, and also there is no flag today to do so.
The closest acceptable change would be about creating that flag and exposing it on the Helm chart. 

I hope the above explanation help in some way, and please let us know how we can help you with your current use case.

dylanspag-lmco:
@davidmirror-ops Thank you for explanation. I'll consult with my organization's flyte admins to see if there's anything we can do internally to get ContainerTasks working without disabling this.

vlibov:
> Considering this context, we cannot make `SYS_PTRACE` configurable because then we'd also have to disable `RawContainer`, and also there is no flag today to do so. The closest acceptable change would be about creating that flag and exposing it on the Helm chart.
> 

What is the principal difference between the proposed changes in this PR and "exposing on the Helm chart"?
Thanks!

PR #5752, opened 2024-09-17
Fixes monodocs by updating intersphinx mapping

thomasjpfan:

The monodocs are failing because PyTorch has moved to using `main` as their development branch.

This PR fixes the `intersphinx_mapping`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -294,6 +294,7 @@ intersphinx_mapping = {
 }
 
 myst_enable_extensions = ["colon_fence"]
+myst_heading_anchors = 6
 
 # Sphinx-mermaid config
 mermaid_output_format = "raw"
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -287,14 +287,13 @@
...
     "pandas": ("https://pandas.pydata.org/pandas-docs/stable/", None),
-    "torch": ("https://pytorch.org/docs/master/", None),
+    "torch": ("https://pytorch.org/docs/main/", None),
     "scipy": ("https://docs.scipy.org/doc/scipy/reference", None),
     "matplotlib": ("https://matplotlib.org", None),
     "pandera": ("https://pandera.readthedocs.io/en/stable/", None),
 }
 
 myst_enable_extensions = ["colon_fence"]
-myst_heading_anchors = 6
```
neverett, line 297: This should not be deleted -- this is the heading level depth to assign HTML anchors. (The default is zero, so if removed, I think we lose HTML anchors.)

neverett (changes requested):
One change requested, otherwise LGTM

neverett:
Looks like this was fixed in https://github.com/flyteorg/flyte/pull/5755

docs/conf.py
thomasjpfan: Ah sorry, this was a mistake. Reverted.

thomasjpfan:
Looks like this was already fix on master. Closing.

PR #5767, opened 2024-09-22
Use pluggable clock for auto refresh cache and make unit tests fast

### LINKED ISSUE #5711: [BUG] flytestdlib/cache unit tests are flaky
### Describe the bug

During a recent build in CI I encountered a unit test failure. I noticed that pluggable / test time sources aren't being used in thees unit tests which may lead to flakes depending on how slowly the unit tests are executed. It would be best to use a test time source instead, it will also speed up the tests because they won't have to sleep. 

```
--- FAIL: TestCacheFour (8.11s)
    --- FAIL: TestCacheFour/Test_panic_on_sync_and_shutdown (5.50s)
        auto_refresh_test.go:200: 
            	Error Trace:	/home/runner/work/flyte/flyte/flytestdlib/cache/auto_refresh_test.go:200
            	Error:      	Condition never satisfied
            	Test:       	TestCacheFour/Test_panic_on_sync_and_shutdown
        auto_refresh_test.go:208: 
            	Error Trace:	/home/runner/work/flyte/flyte/flytestdlib/cache/auto_refresh_test.go:208
            	Error:      	Not equal: 
            	            	expected: 10
            	            	actual  : 0
            	Test:       	TestCacheFour/Test_panic_on_sync_and_shutdown
```

### Expected behavior

The unit tests pass and are deterministic.

_No response_

### Screenshots

_No response_

- [X] Yes

- [X] Yes

Sovietaced:
## Tracking issue
Closes #5711 

Previously this unit test was relying on a real clock which makes unit tests slow and flaky due to nondeterministic execution duration. If code takes a dependency on time it should use a pluggable clock. 

Before 
```
ok      github.com/flyteorg/flyte/flytestdlib/cache     8.828s
```

After
```
ok      github.com/flyteorg/flyte/flytestdlib/cache     0.277s
```

`autoRefresh` and the work queue it relies were updated to use a pluggable clock. A fake clock is used for unit tests while a real clock is used under all other circumstances.

### DIFF
--- a/flytestdlib/.golangci.yml
+++ b/flytestdlib/.golangci.yml
@@ -26,7 +26,6 @@ linters:
     - structcheck
     - typecheck
     - unconvert
-    - unparam
     - unused
     - varcheck
 
--- a/flytestdlib/cache/auto_refresh.go
+++ b/flytestdlib/cache/auto_refresh.go
@@ -3,13 +3,13 @@ package cache
 import (
 	"context"
 	"fmt"
+	"k8s.io/utils/clock"
 	"runtime/debug"
 	"sync"
 	"time"
 
 	lru "github.com/hashicorp/golang-lru"
 	"github.com/prometheus/client_golang/prometheus"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/workqueue"
 
 	"github.com/flyteorg/flyte/flytestdlib/contextutils"
@@ -125,6 +125,7 @@ type autoRefresh struct {
 	workqueue   workqueue.RateLimitingInterface
 	parallelizm int
 	lock        sync.RWMutex
+	clock       clock.Clock
 }
 
 func getEvictionFunction(counter prometheus.Counter) func(key interface{}, value interface{}) {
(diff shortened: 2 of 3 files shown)


flytestdlib/cache/auto_refresh_test.go
```diff
@@ -167,24 +174,27 @@ func TestCacheFour(t *testing.T) {
 		})
 		assert.NoError(t, err)
 
-		// Wait half a second for all resync periods to complete
 		// If the cache tries to enqueue the item, a panic will be thrown.
-		time.Sleep(500 * time.Millisecond)
+		fakeClock.Step(testResyncPeriod)
```
Sovietaced, line 178: These tests aren't perfect with how things are asynchronous... It's hard to verify that a sync actually happened. I wanted to add a unexpected sync counter to the auto refresh to use in unit tests but the constructor imo incorrectly returns the interface instead of a pointer to the struct (which should adhere to the interface). As such, the unit tests can't access anything package level.

Sovietaced: I wanted to fix that but I'm trying to reduce the blast radius of this change.

flytestdlib/cache/auto_refresh.go
```diff
@@ -406,3 +428,8 @@ func NewAutoRefreshCache(name string, syncCb SyncFunc, syncRateLimiter workqueue
 
 	return NewAutoRefreshBatchedCache(name, SingleItemBatches, syncCb, syncRateLimiter, resyncPeriod, parallelizm, size, scope)
 }
+
+func newAutoRefreshCacheWithClock(name string, syncCb SyncFunc, syncRateLimiter workqueue.RateLimiter, resyncPeriod time.Duration,
```
Sovietaced, line 432: I've added these package lever constructors for testing but honestly with all the arguments that can be configured on the auto refresh cache it would be nicer to have a constructor with `Options`. I can follow up with that in another pull request if you'd like.

flytestdlib/cache/auto_refresh.go
```diff
@@ -165,17 +166,29 @@ func (w *autoRefresh) Start(ctx context.Context) error {
 	}
 
 	enqueueCtx := contextutils.WithGoroutineLabel(ctx, fmt.Sprintf("%v-enqueue", w.name))
-
-	go wait.Until(func() {
```
Sovietaced, line 169: Moved away from using `wait.Until` so I could use a pluggable clock.

flytestdlib/.golangci.yml
```diff
@@ -26,7 +26,6 @@ linters:
     - structcheck
     - typecheck
     - unconvert
-    - unparam
```
Sovietaced, line 29: Disabling since I was getting 
```
cache/auto_refresh.go:433:2: `newAutoRefreshCacheWithClock` - `parallelizm` always receives `10` (unparam)
        parallelizm, size int, scope promutils.Scope, clock clock.WithTicker) (AutoRefresh, error) {
        ^
```

I could omit this from my specific line but people generally seem to think this linter is pretty annoying.

flytestdlib/cache/auto_refresh.go
eapolinario: We can do that in a separate PR.

eapolinario (approved):
Thank you for working on this.

PR #5705, opened 2024-08-29
Using new offloaded metadata literal message for literal offloading

pmahindrakar-oss:
## Tracking issue
RFC https://github.com/flyteorg/flyte/pull/5103

## Why are the changes needed?

Without these changes we hit grpc limit when propeller tries to create an execution by passing inputs inline which it got from output of large map task

## What changes were proposed in this pull request?

Adding a new sub message for offloaded literal which would be populated initially by propeller for map task user when the collected values are greater than configured threshold and later flytekit which write the o/p where we will be adding similar checks

## How was this patch tested?

Will be tested as part of this PR https://github.com/flyteorg/flyte/pull/5697
These are just IDL changes

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

PR #5771, opened 2024-09-23
Fix propeller crash when inferring literal type for an offloaded literal

pmahindrakar-oss:

Followup to this ticket https://github.com/flyteorg/flyte/pull/5705
Fixes propeller crash when the type inferred is nil.

The LiteralTypeFromLiteral function didn't handle Offloaded type

Along with this adds the following changes.
* Reading of offloaded literal in array node map task if the input itself is an offloaded literal. This is to support passing the inputs to all the sub nodes
* Promise input supports for offloaded literal
* Removes logging of the inputs outputs which might be large from datacatlog
* Adds hashing logic for offloaded literal which is used for caching
* adds flytectl support to o/p offloaded metadata uri
* bumps the maxDownloadMbs to 1000 MB to be inline with the max offloaded literal size

This does increase the memory profile for propeller whenever it needs to download large literal

## How was this patch tested?
Tested using sandbox and the changes working fine for the following examples . Tested for variation of caching aswell

```
import typing
from flytekit import task, workflow, map_task
from mashumaro.mixins.json import DataClassJSONMixin

@dataclass
# class InnerDC(DataClassJSONMixin):
class InnerDC:
    a: bool
    b: int
    c: float
    d: list[int]
    e: dict[str, int]
    f: FlyteFile

@dataclass
# class DC(DataClassJSONMixin):
class DC:
    name: str
    other: str
    f: FlyteFile
    inner: InnerDC


@task(cache=True, cache_version="1.1")
def f(n: int) -> DC:
    with open("/tmp/abc", "w") as f:
        f.write("wow")
    name = "a" * 500_000
    return DC(
        name=name,
        other=name,
        f=FlyteFile("/tmp/abc"),
        inner=InnerDC(
            a=True,
            b=42,
            c=3.1415,
            d=[1,2,3],
            e={"a": 1, "b": 2},
            f=FlyteFile("/tmp/abc"),
        )
    )

@task(cache=True, cache_version="1.1")
def get_list(n: int) -> list[int]:
    return list(range(n))

@task(cache=True, cache_version="1.1")
def id_dataclass(dc: DC) -> DC:
    with open(dc.f, "r") as f:
        print(f.read())
    return dc


@task(cache=True, cache_version="1.1")
def consume_list_of_objects(xs: list[DC]) -> list[DC]:
    with open(xs[0].f, "r") as f:
        print(f.read())
    return xs[0:2]


@workflow
def working_wf(n: int = 15) -> list[DC]:
    xs=map_task(f)(n=get_list(n=n))
    return consume_list_of_objects(xs=xs)


@workflow
def passing_map_output_wf(n: int = 15) -> list[DC]:
    xs=map_task(f)(n=get_list(n=n))
    # Notice how we pass the offloaded literal to map_task
    return map_task(id_dataclass)(dc=xs) 

@task(cache=True, cache_version="1.1")
def consume_single_element(y: DC) -> DC:
    with open(y.f, "r") as f:
        print(f.read())
    return y

@workflow
def indexing_map_output_wf(n: int = 15):
    a=passing_map_output_wf(n=n)
    consume_single_element(y=a[0])
```

Notice `consume_list_objects` returns a slice since the outputting of offloaded data is yet to be supported in flytekit without which if the data is too large the datacatlog call errors out with grpc error due to max size breached


wild-endeavor (approved):
i think this is fine, but cc @eapolinario

wild-endeavor:
this doesn't break any flyte releases right?

pmahindrakar-oss:
@wild-endeavor yes right now it doesn't since the feature is disabled, but this happens in peculiar case of map task accepting an offloaded literal. I am working on more elaborate fix for this as the current fix doesn't completely solve it.

pmahindrakar-oss:
@wild-endeavor added more fixes for the crash along with array node support to read offloaded literal ,promise support etc,

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -192,11 +192,20 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 
 		size := -1
 		for _, variable := range literalMap.Literals {
+			if variable.GetOffloadedMetadata() != nil {
+				// variable will be overwritten with the contents of the offloaded data which contains the actual large literal.
+				// We need this for the map task to be able to create the subNodeSpec
```
katrogan, line 197: just so I understand, this is also needed so the map task can index into the inputs as well too?

flytepropeller/pkg/controller/nodes/catalog/datacatalog/datacatalog.go
```diff
@@ -122,7 +122,7 @@ func (m *CatalogClient) Get(ctx context.Context, key catalog.Key) (catalog.Entry
 		logger.Debugf(ctx, "DataCatalog failed to get artifact by tag %+v, err: %+v", tag, err)
 		return catalog.Entry{}, err
 	}
-	logger.Debugf(ctx, "Artifact found %v from tag %v", artifact, tag)
+	logger.Debugf(ctx, "Artifact found %v from tag %v", artifact.GetId(), tag)
```
katrogan, line 125: curious why these changes are necessary? do we ever send too big inputs/outputs to cache service? shouldn't we be using the offloaded literal?

flytepropeller/pkg/controller/nodes/common/utils.go
```diff
@@ -79,6 +81,24 @@ func GetTargetEntity(ctx context.Context, nCtx interfaces.NodeExecutionContext)
...
 }
 
+// ReadLargeLiteral reads the offloaded large literal needed by array node task
+func ReadLargeLiteral(ctx context.Context, datastore *storage.DataStore,
+	tobeRead *idlcore.Literal) error {
+	if tobeRead.GetOffloadedMetadata() == nil {
+		return fmt.Errorf("unsupported type for reading offloaded literal")
+	}
+	// read the offloaded literal
+	dataReference := tobeRead.GetOffloadedMetadata().GetUri()
```
katrogan, line 91: nit: worth checking that `tobeRead.GetOffloadedMetadata().GetUri()` is non-empty?

flytepropeller/pkg/controller/nodes/common/utils.go
```diff
@@ -107,6 +127,12 @@ func OffloadLargeLiteral(ctx context.Context, datastore *storage.DataStore, data
 		logger.Errorf(ctx, "Failed to offload literal at location [%s] with error [%s]", dataReference, err)
 		return err
 	}
+	// compute the hash of the literal
+	literalDigest, err := pbhash.ComputeHash(ctx, toBeOffloaded)
```
katrogan, line 131: nit: can we move this to the if block under

`if toBeOffloaded.GetHash() == "" {`

so we only compute it when necessary?

katrogan (approved):
LGTM but hopefully @hamersaw or @pvditt can review here too!

flytepropeller/pkg/controller/nodes/catalog/datacatalog/datacatalog.go
pmahindrakar-oss: Even they are not too big breaching the 10 MB limit, but anything less than that threshold will get logged which is also huge and hence removed it. Let me know if you disagree

flytepropeller/pkg/controller/nodes/common/utils.go
pmahindrakar-oss: I wanted to do something before overrwriting the value, moved the value assignment at the end along with this being computed in conditional

flytepropeller/pkg/controller/nodes/array/handler.go
pmahindrakar-oss: yeah subNodeSpec covers that as it contains the inputs for each subnode

flytepropeller/pkg/controller/nodes/catalog/datacatalog/datacatalog.go
katrogan: sounds good, I could still see this being useful, maybe we only log a deterministic prefix?

pmahindrakar-oss: It was logging an entire object and hence i think logging an identifier should good for long term too and we have api's to fetch the aritifact using the ID if we want to debug further. Logging a prefix of object converted to string format doesn't seem useful IMO

hamersaw (approved):

katrogan (approved):

PR #5785, opened 2024-09-30
Write abort event for aborted k8s workflows

### LINKED ISSUE #5547: [BUG] Workflow status is stuck at Aborting
### Describe the bug

https://flyte-org.slack.com/archives/CP2HDHKE1/p1720478153981989

Workflow status is stuck at Aborting and task status is stuck at Running Forever when I abort the workflow

Screenshot 2024-07-09 at 10 50 21 AM

Minimal repro:

```
flytectl demo start
pyflyte run --remote wf.py wf
flytectl demo start # restart the demo cluster; it will delete the CRD, and the propeller won't propagate the event to admin.
```

```python
@task(enable_deck=True)
def t1(a: int) -> int:
    sleep(3000)
    return a + 1


@workflow
def wf() -> int:
    return t1(a=3)
```

### Expected behavior

The status should transition to `aborted`

We do nothing if CRD doesn't exist when deleting the workflow [here](https://github.com/flyteorg/flyte/blob/7d788cbe1e3f395d0e6ad862c4fb310458d09dbf/flyteadmin/pkg/workflowengine/impl/k8s_executor.go#L97-L100).

We should make sure the workflow status is `aborted` before returning `nil`

```golang
func (e K8sWorkflowExecutor) Abort(ctx context.Context, data interfaces.AbortData) error {
	target, err := e.executionCluster.GetTarget(ctx, &executioncluster.ExecutionTargetSpec{
		TargetID: data.Cluster,
	})
	if err != nil {
		return errors.NewFlyteAdminErrorf(codes.Internal, err.Error())
	}
	err = target.FlyteClient.FlyteworkflowV1alpha1().FlyteWorkflows(data.Namespace).Delete(ctx, data.ExecutionID.GetName(), v1.DeleteOptions{
		PropagationPolicy: &deletePropagationBackground,
	})
	// An IsNotFound error indicates the resource is already deleted.
	if err != nil && !k8_api_err.IsNotFound(err) {
		return errors.NewFlyteAdminErrorf(codes.Internal, "failed to terminate execution: %v with err %v", data.ExecutionID, err)
	}
        // TODO: make sure the workflow status is aborted here. If it's not, we should update the status.
	return nil
}
```

- [X] Yes

- [X] Yes

andresgomezfrr:
## Tracking issue
Potentially closes #5547 and #5786

To make sure Workflows are able to transition from `Aborting` to `Aborted` if flytepropeller is not able to do it.

As proposed in #5547 we're emitting an event after sending a Delete call to Flyte k8s cluster

## How was this patch tested?

With unit tests and staging environment

### DIFF
--- a/flyteadmin/pkg/rpc/adminservice/base.go
+++ b/flyteadmin/pkg/rpc/adminservice/base.go
@@ -77,7 +77,13 @@ func NewAdminServer(ctx context.Context, pluginRegistry *plugins.Registry, confi
 		repo)
 	workflowBuilder := workflowengineImpl.NewFlyteWorkflowBuilder(
 		adminScope.NewSubScope("builder").NewSubScope("flytepropeller"))
-	workflowExecutor := workflowengineImpl.NewK8sWorkflowExecutor(configuration, execCluster, workflowBuilder, dataStorageClient)
+
+	executionEventWriter := eventWriter.NewWorkflowExecutionEventWriter(repo, applicationConfiguration.GetAsyncEventsBufferSize())
+	go func() {
+		executionEventWriter.Run()
+	}()
+	
+	workflowExecutor := workflowengineImpl.NewK8sWorkflowExecutor(configuration, execCluster, workflowBuilder, dataStorageClient, executionEventWriter)
 	logger.Info(ctx, "Successfully created a workflow executor engine")
 	pluginRegistry.RegisterDefault(plugins.PluginIDWorkflowExecutor, workflowExecutor)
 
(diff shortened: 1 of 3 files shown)


pvditt (commented):
I lean against adding this eventing in admin. I feel as though this updating of status should be handled in propeller.

[FlyteWorkflow CRDs get finalizers set by default](https://github.com/flyteorg/flyte/blob/66ff1528893813763b1202763b3850ae27e86f48/flytepropeller/pkg/controller/handler.go#L128). In your scenario, it seems the delete occurs before propeller is able to set finalizers on the FlyteWorkflow. Let me  check if there'd be any issues setting this in the same loop that the resource is created - either in the propeller side admin launcher or just set it by default in admin.

I think getting the finalizer set would be a better path forward.

RRap0so:
Closing in favour of https://github.com/flyteorg/flyte/pull/5788

Issue #5786, opened 2024-09-30
[BUG] Aborting Sub-Workflows filling up the admin-launcher Cache


RRap0so:
### Describe the bug

We've been suffering with an endless growth of FlyteAdmin GetExecution calls coming from the Propeller admin-launcher.

We believe this is caused by Sub-Workflows getting stuck in `Aborting` Execution and never being evicted from the cache.. We saw this was already reported in https://github.com/flyteorg/flyte/issues/5547

With this bug the api calls to FlyteAdmin growth until FlytePropeller crashes/restarted and then it starts again.

We also see the WorkQueue growing exponentially and this is again caused by those Objects which the cache thinks it's items that need to be processed when in Fact the Main Flyte Execution was already aborted, but since it's state never goes out of `Aborting` the cache never sets a Terminal state and the items keep being added into the WorkQueues 

![image](https://github.com/user-attachments/assets/d2904f94-7ad2-4d1a-a79a-dd3fbc8f32fd)

### Expected behavior

This happens with a very specific situation but it's a very usual use-case for us:
```
-Parent Execution
---- SubWorkflow 1 (a normal flyte task) (generates a sub execution)
---- Subworkflow 2 (a flyte plugin task) (fails in  <1 sec)
```

What we see is that the Sub Execution CRD is getting deleted before is being picked up by FlytePropeller since the `SubWorkflow 2` being a flyte plugin task is very fast on checking what it needs to do and failing the parent workflow. In turn the parent workflow sends the `Kill` signal to Flyteadmin and flyteadmin deletes the CRD. This happens quite fast and SubWorkflow 1 wasn't even picked up by the propeller, making it stay in the `Aborting` limbo state. 

The CRD does not stay in the k8s cluster because since it was never picked up by Propeller `finalizer` was never added to the CRD.

It only impacts cache but that's taxing overall our Flyte System.

### Additional context to reproduce

Create a Workflow with two Subworkflows one that is a usual Flyte Task another that is a Plugin tasks which can run quite fast. The plugin task should fail creating a cascading Node failure to `SubWorkflow 1`

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #5788, opened 2024-10-01
Add finalizer to avoid premature CRD Deletion

RRap0so:
## Tracking issue

Closes https://github.com/flyteorg/flyte/issues/5786

To make sure Workflows aren't prematurely deleted by Flyteadmin and FlytePropeller can properly transitions the states to `Aborted`

## How was this patch tested?

Unit Tested

Tested in a staging environment manually.

### DIFF
--- a/flyteadmin/pkg/workflowengine/impl/prepare_execution.go
+++ b/flyteadmin/pkg/workflowengine/impl/prepare_execution.go
@@ -129,6 +129,9 @@ func PrepareFlyteWorkflow(data interfaces.ExecutionData, flyteWorkflow *v1alpha1
 	acceptAtWrapper := v1.NewTime(data.ExecutionParameters.AcceptedAt)
 	flyteWorkflow.AcceptedAt = &acceptAtWrapper
 
+	// Add finalizer
+	flyteWorkflow.Finalizers = append(flyteWorkflow.Finalizers, "flyte-finalizer")
+
 	// add permissions from auth and security context. Adding permissions from auth would be removed once all clients
 	// have migrated over to security context
 	addPermissions(data.ExecutionParameters.ExecutionConfig.SecurityContext,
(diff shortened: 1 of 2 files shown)


pvditt:
concern if propeller never picks up the FlyteWorkflow CRD then the CRD won't be able to get garbage collected.

flyteadmin/pkg/workflowengine/impl/prepare_execution.go
```diff
@@ -129,6 +129,9 @@ func PrepareFlyteWorkflow(data interfaces.ExecutionData, flyteWorkflow *v1alpha1
 	acceptAtWrapper := v1.NewTime(data.ExecutionParameters.AcceptedAt)
 	flyteWorkflow.AcceptedAt = &acceptAtWrapper
 
+	// Add finalizer
```
EngHabu, line 132: Doesn't this already happen here?
		SetFinalizerIfEmpty(mutableW, FinalizerKey)
https://github.com/unionai/flyte/blob/14eaa16a43bff3ada4e93c6c6421c6b35471bf30/flytepropeller/pkg/controller/handler.go#L128

pvditt: @EngHabu I believe in their scenario, the CRD is getting deleted before propeller has had the chance to evaluate (and set the finalizer for) the external workflow

EngHabu (approved):
Pasting here from a private channel chat for anybody who comes across this:
image

I will merge after @pvditt approves.

@pvditt What's the concern with not getting deleted? if propeller is permanently down and can't act on the object then it'll be around, right?

pvditt:
@EngHabu This simplest route is probably the best route. My concern was probably an unfounded one - the scenario would be a FlyteWorkflow CRD gets created that all propeller instances cannot access and then any TerminateExecution request wouldn't be able to delete the CRD as finalizers would be set. If this is a possibility then we can just update the TerminateExecuteRequest to include an option to ensure the CRD is deleted instead of having users utilize kubectl commands to clear a potential dangling CRD.

pvditt (approved):

PR #5789, opened 2024-10-01
Use annotations instead of delete CRDs when terminate execution

andresgomezfrr:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/5786
Alternative to: https://github.com/flyteorg/flyte/pull/5788

To make sure that the propeller can handle the CRD before they are deleted if a terminated execution is called very fast.

**New behavior:**
Instead of propagating the terminate execution to the propeller deleting the CRD, the admin now updates a new aborted annotation, the propeller will check this annotation instead of the `deletedTimestamp`. If the annotation is there the propeller will send the `ABORT` event to the admin and remove the finalizer from the CRD. The propeller garbage-collector will clean up this CRD when it is executed.

Unit test + manual test in a staging environment with real flyte execution.

### DIFF
--- a/flytepropeller/pkg/controller/finalizer.go
+++ b/flytepropeller/pkg/controller/finalizer.go
@@ -1,8 +1,9 @@
 package controller
 
 import (
-	"github.com/flyteorg/flyte/flytepropeller/pkg/apis/flyteworkflow/v1alpha1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/flyteorg/flyte/flytepropeller/pkg/apis/flyteworkflow/v1alpha1"
 )
 
 const FinalizerKey = "flyte-finalizer"


flytepropeller/pkg/controller/handler.go
```diff
@@ -103,7 +110,8 @@ func (p *Propeller) TryMutateWorkflow(ctx context.Context, originalW *v1alpha1.F
 	ctx = contextutils.WithResourceVersion(ctx, mutableW.GetResourceVersion())
 
 	maxRetries := uint32(p.cfg.MaxWorkflowRetries)
-	if IsDeleted(mutableW) || (mutableW.Status.FailedAttempts > maxRetries) {
+
+	if IsAborted(mutableW) || (mutableW.Status.FailedAttempts > maxRetries) {
```
pvditt, line 114: we could still have the case that the CRD gets deleted by some other process and should still account for that here.

flytepropeller/pkg/controller/handler.go
```diff
@@ -89,6 +91,11 @@ func SetDefinitionVersionIfEmpty(wf *v1alpha1.FlyteWorkflow, version v1alpha1.Wo
 	}
 }
 
+func IsAborted(w *v1alpha1.FlyteWorkflow) bool {
```
pvditt, line 94: can this be moved to the same place IsDeleted is defined?

flyteadmin/pkg/workflowengine/impl/k8s_executor.go
```diff
@@ -94,13 +93,27 @@ func (e K8sWorkflowExecutor) Abort(ctx context.Context, data interfaces.AbortDat
...
+	} else if k8_api_err.IsNotFound(err) {
+		logger.Infof(ctx, "Trying to abort an execution [%+v] that is not found in cluster: %s, skipping...", data.ExecutionID, target.ID)
+		return nil
+	}
+
+	if w.ObjectMeta.Annotations == nil {
+		w.ObjectMeta.Annotations = make(map[string]string)
+	}
+
+	w.ObjectMeta.Annotations[AbortedWorkflowAnnotation] = "true"
```
pvditt, line 109: concern here is if there's a case that this delete is required to delete dangling CRDs (not picked up by propeller)

andresgomezfrr:
I think that the propeller GC should remove that workflows, right?

We checked in our staging and they were cleaned up

flytepropeller/pkg/controller/handler.go
andresgomezfrr: mmmmh could be? Looking inside the if, the function there looks like is only for aborting case

flytepropeller/pkg/controller/handler.go
andresgomezfrr: Sure I will do it, tomorrow

flyteadmin/pkg/workflowengine/impl/k8s_executor.go
andresgomezfrr: I guess.. that the dangling CRDs should be removed by propeller GC, right?

flytepropeller/pkg/controller/handler.go
andresgomezfrr: done!

flyteadmin/pkg/workflowengine/impl/k8s_executor.go
andresgomezfrr: It is the same case as completed workflows/failed workflows.

flytepropeller/pkg/controller/handler.go
andresgomezfrr: I added back the isDeleted(...) to manage this case here too, now it is backward compatible, wdyt?

flyteadmin/pkg/workflowengine/impl/k8s_executor.go
andresgomezfrr: >concern here is if there's a case that this delete is required to delete dangling CRDs (not picked up by propeller) 

If the concern is about deleting CRDs if the propeller is down.. what do you think about having a flag in the terminate to force the delete from the admin? Similar to kubectl delete --force which force the deletion of a CRD.

PR #5616, opened 2024-08-01
Add FLYTE_INTERNAL_POD_NAME environment variable that holds the pod name

bgedik:
## Tracking issue
Part of RFC https://github.com/flyteorg/flyte/blob/b3add894fd879d6b6c11e37ba292ae04bf60afeb/rfc/system/5598-deterministic-errors-distributed-training.md#flytepropeller-backend

## Why are the changes needed?
Deterministic error propagation, see: https://github.com/flyteorg/flyte/blob/b3add894fd879d6b6c11e37ba292ae04bf60afeb/rfc/system/5598-deterministic-errors-distributed-training.md#flytepropeller-backend

## What changes were proposed in this pull request?

Add FLYTE_INTERNAL_POD_NAME as a env variable to pod specs.

## How was this patch tested?

Updates to unit tests

Manual run 

image

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [X] All commits are signed-off.

## Related PRs

None

## Docs link

None

Issue #5665, opened 2024-08-16
[Housekeeping] Shorten injected environment variables 


eapolinario:
### Describe the issue

The flyte-specific environment variables [injected](https://github.com/flyteorg/flyte/blob/4e7a1c53621977b8116711410231e75a6505a512/flyteplugins/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go#L51-L132) during the pod manufacturing process by propeller share a long prefix, namely `FLYTE_INTERNAL_`. 

Flytekit has had [support for shorter names](https://github.com/flyteorg/flytekit/blob/master/flytekit/bin/entrypoint.py#L221-L230) since version 0.32.0. Migrating to shorter names in the backend should be painless.

In terms of benefits of shorter environment names, it's well known that different systems (e.g. Windows and older versions of bash) have tight limits, sometimes as low as a few KB for all environment variables (including `PATH`), so having these long (and wasteful) names only adds insult to the injury.

### What if we do not do this?

Different systems deal differently with a list of environment variables that cross the system's threshold, ranging from silently truncating environment variables to crashing processes. If we don't do this we are inadvertently contributing to the problem in the long run.

### Related component(s)

flytekit, flytepropeller

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #5795, opened 2024-10-01
Add multi file error aggregation strategy

bgedik:
## Tracking issue
Part of RFC https://github.com/flyteorg/flyte/pull/5598

Deterministic error propagation, see: https://github.com/flyteorg/flyte/pull/5598

An error aggregation strategy was introduced for the remove file output reader. 

This PR is self-complete in the sense that it switches the PyTorch plugin to use the earliest timestamp error file strategy. However, the plugin won't be writing multiple files yet, as the flytekit side changes to populate multiple error files is not implemented.  Since the error reading code for earliest timestamp strategy is backward compatible, it works with a single error file as well.

My plan was the following in terms of order:
* Get this PR in, start using the new error retriever for PyTorch
* Get the entry point changes in, for creating multiple error files when environment variables are present 

## How was this patch tested?

Unit testing.

## Related PRs

* https://github.com/flyteorg/flyte/pull/5741
* https://github.com/flyteorg/flyte/pull/5616
* https://github.com/flyteorg/flytekit/pull/2797


bgedik:
cc @eapolinario @wild-endeavor

wild-endeavor (commented):
Thank you for working on this!  

Few points/questions...
* This PR is only one part of the story right?  Something still needs to direct all the plugins to write different error messages.
* Can we maybe mark the new proto fields as still a wip?  If they don't work out for some reason we can mark those slots as deprecated and at least no one else will try to use them in the meantime. Typically we try to not merge idl changes until things are working end to end, esp for messages that live in the blob store forever, but i think it's low risk in this case.
* How will this pattern be extended into Agent plugins?  The new option in the proto will need to be routed through the webapi plugin right?

What frontend UI changes are needed to make the end to end story work if any?

looks good from my end, but would like one of @eapolinario or @pingsutw to also take a look.

bgedik:
> Thank you for working on this!
> 
> Few points/questions...
> 
> * This PR is only one part of the story right?  Something still needs to direct all the plugins to write different error messages.
> * Can we maybe mark the new proto fields as still a wip?  If they don't work out for some reason we can mark those slots as deprecated and at least no one else will try to use them in the meantime. Typically we try to not merge idl changes until things are working end to end, esp for messages that live in the blob store forever, but i think it's low risk in this case.
> * How will this pattern be extended into Agent plugins?  The new option in the proto will need to be routed through the webapi plugin right?
> 
> What frontend UI changes are needed to make the end to end story work if any?
> 
> looks good from my end, but would like one of @eapolinario or @pingsutw to also take a look.

This PR is self-complete in the sense that it switches the PyTorch plugin to use the earliest timestamp error file strategy. However, the plugin won't be writing multiple files yet, as the environment variables instructing it to do so are not set and the flytekit side changes to populate multiple error files is not implemented. Since the error reading code for earliest timestamp strategy is backward compatible, it works with a single error file as well.

My plan was the following in terms of order:

Get this PR in, start using the new error retriever for PyTorch
* Get the entry point changes in, for creating multiple error files when environment variables are present (with unit tests)
* Populate the environment variables for the end to end function (at this point, I'll validate with manual tests)

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
```diff
@@ -3,75 +3,243 @@ package ioutils
...
+				}
+				errorFileAndDocs = append(errorFileAndDocs, ErrorFileAndDocument{errorFilePath: errorFilePath, errorDoc: errorDoc})
+			}
+		}
+	}
+
+	var earliestTimestamp int64 = math.MaxInt64
+	earliestExecutionError := io.ExecutionError{}
+	for _, errorFileAndDoc := range errorFileAndDocs {
+		timestamp := errorFileAndDoc.errorDoc.Error.GetTimetsamp()
```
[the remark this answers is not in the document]
bgedik: Uggh. Will fix.

wild-endeavor:
cc https://github.com/flyteorg/flytekit/pull/2797 which is the next step in the process.

wild-endeavor (approved):

flyteidl/protos/flyteidl/core/errors.proto
```diff
@@ -25,6 +25,12 @@ message ContainerError {
 
     // Defines the origin of the error (system, user, unknown).
     ExecutionError.ErrorKind origin = 4;
+
+    // Timestamp of the error
+    int64 timestamp = 5;
+
+    // Worker that generated the error
+    string worker = 6;
```
EngHabu, line 33: Haven't looking into the rest of the PR yet. I assume this is referring to a distributed pytorch plugin worker in case of pytorch plugin?

flyteidl/protos/flyteidl/core/execution.proto
```diff
@@ -73,6 +73,8 @@ message ExecutionError {
         SYSTEM = 2;
     }
     ErrorKind kind = 4;
+    // Worker that generated the error
+    string worker = 5;
```
EngHabu, line 77: Why no timestamp?

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
```diff
@@ -3,75 +3,243 @@ package ioutils
...
+			if strings.HasSuffix(errorFilePath.String(), errorFileExtension) {
+				metadata, err := e.store.Head(ctx, errorFilePath)
+				if err != nil {
+					return false, errors.Wrapf(err, "failed to read error file @[%s]", errorFilePath)
+				}
+				err = e.validatePayloadSize(metadata)
+				if err != nil {
+					return false, errors.Wrapf(err, "error file @[%s] failed payload size validation", errorFilePath)
+				}
+				hasError = true
```
EngHabu, line 157: Shouldn't we return immediately here as soon as we found at least one error?

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
```diff
@@ -3,75 +3,243 @@ package ioutils
...
+			return io.ExecutionError{}, errors.Wrapf(err, "failed to list error files @[%s]", e.errorDirPath)
+		}
+		for _, errorFilePath := range errorFilePaths {
+			if strings.HasSuffix(errorFilePath.String(), errorFileExtension) {
+				errorDoc := &core.ErrorDocument{}
+				err := e.store.ReadProtobuf(ctx, errorFilePath, errorDoc)
+				if err != nil {
+					return io.ExecutionError{}, errors.Wrapf(err, "failed to read error file @[%s]", errorFilePath.String())
+				}
+				errorFileAndDocs = append(errorFileAndDocs, ErrorFileAndDocument{errorFilePath: errorFilePath, errorDoc: errorDoc})
```
EngHabu, line 190: Instead of appending all of them here, can we just keep the oldest around... so by the time the loop is done you have a single document?

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
```diff
@@ -3,75 +3,243 @@ package ioutils
...
+
+func (s *SingleFileErrorRetriever) GetError(ctx context.Context) (io.ExecutionError, error) {
+	errorDoc := &core.ErrorDocument{}
+	err := s.store.ReadProtobuf(ctx, s.errorFilePath, errorDoc)
+	if err != nil {
+		if storage.IsNotFound(err) {
+			return io.ExecutionError{
+				IsRecoverable: true,
+				ExecutionError: &core.ExecutionError{
+					Code:    "ErrorFileNotFound",
```
EngHabu, line 99: Move to a const?

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
```diff
@@ -3,75 +3,243 @@ package ioutils
...
+}
+
+func (e *EarliestFileErrorRetriever) parseErrorFilename() (errorFilePathPrefix storage.DataReference, errorFileExtension string, err error) {
+	// If the canonical error file name is error.pb, we expect multiple error files
+	// to have name error<suffix>.pb
+	pieces := strings.Split(e.canonicalErrorFilename, ".")
+	if len(pieces) != 2 {
+		err = errors.Errorf("expected canonical error filename to have a single dot (.), got %d", len(pieces))
+		return
+	}
```
EngHabu, line 124: We should probably do this once in the constructor and store the pieces in the struct... they will never change after

flyteplugins/go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -45,6 +55,8 @@ type PluginProperties struct {
 	// override that behavior unless the resource that gets created for this plugin does not consume resources (cluster's
 	// cpu/memory... etc. or external resources) once the plugin's Plugin.GetTaskPhase() returns a terminal phase.
 	DisableDeleteResourceOnFinalize bool
+	// Specifies how errors are aggregated
+	ErrorAggregationStrategy ErrorAggregationStrategy
```
EngHabu, line 59: Why is this a property of the plugin? Isn't the EarliestError Strategy already backward compatible? 
I think if there is anything a plugin needs to optionally configure, it's the pattern of the error file names

flytestdlib/storage/stow_store.go
```diff
@@ -275,11 +275,11 @@ func (s *StowStore) List(ctx context.Context, reference DataReference, maxItems
 	} else {
 		stowCursor = cursor.customPosition
 	}
-	items, stowCursor, err := container.Items(k, stowCursor, maxItems)
+	items, stowCursor, err := container.Items(key, stowCursor, maxItems)
 	if err == nil {
 		results := make([]DataReference, len(items))
 		for index, item := range items {
-			results[index] = DataReference(item.URL().String())
+			results[index] = DataReference(fmt.Sprintf("%s://%s/%s", scheme, containerName, item.URL().String()))
```
EngHabu, line 282: Hmmm, is this backward compatible?

EngHabu (commented):
Thank you for your contribution! I've a few questions/suggestions below

flyteidl/protos/flyteidl/core/errors.proto
bgedik: The RFC made this a generic field. In the case of PyTorch plugin, it is one of the PyTorch worker pods, yes. But for other plugins, it could mean something else.

flyteidl/protos/flyteidl/core/execution.proto
bgedik: Sure, we can add that. This type is used for the resolved error. The RFC called for including the worker, I think there is no harm in adding the timestamp as well.

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
bgedik: That was my first instinct, but wanted to run validatePayloadSize for all files to keep it consistent with the existing logic.

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
bgedik: Sure. Will update.

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
bgedik: Just moved this from existing code, but makes sense to move into a constant. Will do.

flyteplugins/go/tasks/pluginmachinery/k8s/plugin.go
bgedik: It is backward compatible, yes. But the RFC called for error strategies, so different ones can be implemented for different plugins. Are you thinking of making it the default? 

@fg91  What do you think about this?

flytestdlib/storage/stow_store.go
bgedik: Yes. This API is what I introduced a few weeks back, for this line of work. Not used by anyone yet.

flyteidl/protos/flyteidl/core/execution.proto
bgedik: Addressed.

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
bgedik: Addressed.

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
bgedik: Addressed.

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
bgedik: Addressed. The reason I have not done this in the first place was that, it impacts many call sites as the constructor would have to return an error. Had to bite the bullet and just make that larger change. Please review.

flyteplugins/go/tasks/pluginmachinery/k8s/plugin.go
EngHabu: It's not clear to me what those other implementations may look like... I usually like to wait for a second/third option before adding an enum/config... If we don't know what implementations we may want to add here, I would say yes let's make it the default and allow plugins to configure the file name pattern...

wild-endeavor (approved):

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
wild-endeavor: ?

bgedik: @wild-endeavor Rather than collecting them to a list and picking from the list the one with the smallest timestamp, we now do it in one loop.

flyteplugins/go/tasks/pluginmachinery/k8s/plugin.go
bgedik: Let's break this into 2 revision requests:

1. Do not have an error aggregation strategy enum and make the new implementation, which is backward compatible, the default: I have no objections to this, but then I have the least amount of knowledge about the Flyte ecosystem here. The RFC called out for error aggregation strategies, in anticipation of more plugins making use of this new customization option. Also, not changing the default and using the multi-error file strategy where it is needed (PyTorch plugin for now) seems a bit safer then starting to adopt it for all. I need more feedback here from the involved parties. Pinging @fg91 and @eapolinario as they participated in the RFC.

2. Configuring file name patterns: What benefits do you see here? It will further complicate flytekit side error file creation. If there is a motivation, we can do it.

fg91:
@wild-endeavor 

> * How will this pattern be extended into Agent plugins?  The new option in the proto will need to be routed through the webapi plugin right?

The [RFC](https://github.com/flyteorg/flyte/pull/5598) (not merged) currently foresees this to be applicable for K8s plugins in flyteplugins. In my mind, if an agent plugin creates multiple "workers in some external system", it should aggregate the errors of those workers*. But I admittedly know too little about agents at this point. If you disagree, let's discuss this in the RFC/in the next contrib sync?

\*Some plugins in flyteplugins can't do that because they manage CRDs (e.g. PyTorchJob) and not the underlying pods and the respective controllers for e.g. PyTorchJob don't do any error aggregation across pods.

> What frontend UI changes are needed to make the end to end story work if any?

The [RFC](https://github.com/flyteorg/flyte/pull/5598/files) says this:

> With these changes, flyteadmin's `/api/v1/executions/<project>/<domain>/<execution id>` endpoint, which today provides the error message to the UI, then also provides the information which worker experienced the root cause error. `flyteconsole` needs to be modified to show this information.

flyteplugins/go/tasks/pluginmachinery/k8s/plugin.go
fg91: > It's not clear to me what those other implementations may look like... I usually like to wait for a second/third option before adding an enum/config...

In the [RFC](https://github.com/flyteorg/flyte/pull/5598/files) (not accepted yet so all of this is up for discussion), we propose that the currently existing logic with a single `error.pb` file is denoted a strategy as well with enum index `0` (default behaviour). "Earliest" would then be index `1`.

---

> Why is this a property of the plugin? Isn't the EarliestError Strategy already backward compatible?

The RFC currently states this about backwards compatibility:

> We propose that the new `MultiErrorFileRemoteFileOutputReader` falls back to reading the `error.pb` (behaviour of the default `RemoteFileOutputReader`) if no `error-<pod-name>.pb` files are found in order to solve the problem of backwards compatibility:

> * If flytekit uses a version that supports multiple error files but the backend does not yet, `pyflyte-execute` will not upload multiple error files for distributed tasks since the `FLYTE_INTERNAL_DIST_ERROR_STRATEGY` environment variable will not be set.
> * If flytekit uses an older version that does not support multiple error files while the backend does, a single error file will be uploaded despite `FLYTE_INTERNAL_DIST_ERROR_STRATEGY` being set. The output reader will, however, fall back to reading the single `error.pb`.

I think plugins need to configure in their properties whether they make use of the default single `error.pb` strategy (`0` in the enum) or the "earliest of multiple" strategy (`1` in the enum) despite this backwards compatibility because most tasks will need the default strategy. Always first checking if multiple errors files exist to then fall back to the single one will make a lot of unnecessary requests to blob storage since for certain plugins (in particular normal python task!) we already know that only a single file will be written.

I'd only use this fallback in plugins that use multiple error files to account for version mismatches between flytekit and backend.

---

> I think if there is anything a plugin needs to optionally configure, it's the pattern of the error file names

In an earlier version of the RFC we included configurable file name patterns but for simplicity decided to propose to add an `/errors` "directory" (being fully aware that buckets are not a filesystem) to the raw output prefix bucket uri. Any files in this error bucket would be considered potential error files.

---

Please let me know if you have any questions or disagree about this @EngHabu. As said above, the RFC is still open and I'm happy to discuss and change it.

flyteidl/protos/flyteidl/core/execution.proto
```diff
@@ -73,6 +74,10 @@ message ExecutionError {
         SYSTEM = 2;
     }
     ErrorKind kind = 4;
+    // Timestamp of the error
+    google.protobuf.Timestamp timestamp = 5;
```
fg91, line 78: I saw that you added this after @EngHabu asked why we don't add the timestamp here as well.
I don't mind adding the time stamp here in principle but why do we need it?

In the UI we don't really care about showing the timestamp I'd say, only the name of the worker.

bgedik: I thought it won't hurt to report the timestamp as well.

flyteplugins/go/tasks/pluginmachinery/k8s/plugin.go
bgedik: > Always first checking if multiple errors files exist to then fall back to the single one will make a lot of unnecessary requests to blob storage since for certain plugins (in particular normal python task!) we already know that only a single file will be written.

The overhead will be a single additional list call, which will return the single error.pb file.

flyteplugins/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
```diff
@@ -17,6 +17,9 @@ import (
 
 const (
 	flyteExecutionURL = "FLYTE_EXECUTION_URL"
+
+	FlyteInternalWorkerNameEnvVarKey        = "_F_WN"  // "FLYTE_INTERNAL_WORKER_NAME"
```
fg91, line 21: Nit:
I'd prefer to be verbose and explicit and specify the full `FLYTE_INTERNAL_WORKER_NAME` and `FLYTE_INTERNAL_DIST_ERROR_STRATEGY` here. 

Reasons:

* Many `FLYTE_INTERNAL_...` env vars are already defined explicitly below.
* Seeing an env var called `_F_WN` in a pod is very untelling.

I'd prefer to also use the full name for `Name: "_F_PN",` below which was added in a previous PR.

fg91, line 21:
```suggestion
	flyteInternalWorkerNameEnvVarKey        = "_F_WN"  // "FLYTE_INTERNAL_WORKER_NAME"
```

bgedik: @fg91  See https://github.com/flyteorg/flyte/issues/5665

fg91: Nit: the var above (and most others in flyte) uses lower camel case instead of pascal case.

flyteplugins/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
```diff
@@ -17,6 +17,9 @@ import (
...
+	FlyteInternalDistErrorStrategyEnvVarKey = "_F_DES" // "FLYTE_INTERNAL_DIST_ERROR_STRATEGY"
```
fg91, line 22:
```suggestion
	flyteInternalDistErrorStrategyEnvVarKey = "_F_DES" // "FLYTE_INTERNAL_DIST_ERROR_STRATEGY"
```

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
```diff
@@ -3,77 +3,232 @@ package ioutils
...
 
 	"github.com/flyteorg/flyte/flyteidl/gen/pb-go/flyteidl/core"
 	"github.com/flyteorg/flyte/flyteplugins/go/tasks/pluginmachinery/io"
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/pluginmachinery/k8s"
 	"github.com/flyteorg/flyte/flytestdlib/storage"
 )
 
-type RemoteFileOutputReader struct {
-	outPath        io.OutputFilePaths
+type ErrorRetriever interface {
```
fg91, line 18: `RemoveFileOutputReader` already implements the `OutputReader` interface. Why introduce a new one?

flyteplugins/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
bgedik: This can't be done, right? These are used by the plugin, which is in a different package.

flyteplugins/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
bgedik: This can't be done, right? These are used by the plugin, which is in a different package.

flyteplugins/go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -30,6 +30,16 @@ type PluginEntry struct {
...
 
+type ErrorAggregationStrategy int
+
+const (
+	// Single error file from a single container
+	DefaultErrorAggregationStrategy ErrorAggregationStrategy = iota
+
+	// Earliest error from potentially multiple error files
+	EarliestErrorAggregationStrategy
+)
```
fg91, line 41: Nit (non blocking):
In the RFC we currently propose:

```go
type ErrorAggregationStrategy int

const (
	// Single error file from a single container
	Default ErrorAggregationStrategy = iota

	// Earliest error from potentially multiple error files
	Earliest
```

I personally prefer to not include the type in the variable name.

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
bgedik: ErrorRetriever is an internal interface used within the `RemoteFileOutputReader`. Remote file reader reads outputs + errors. The error retriever deals with the error reading aspect, based on the strategy.

flyteplugins/go/tasks/pluginmachinery/k8s/plugin.go
bgedik: I'm having difficulty adapting to this convention, because Golang's 'poor' support for enums. If we don't have the type in the variable name, the use site will look like `k8s.Earliest` instead of `k8s.EarliestErrorAggregationStrategy` :(

flyteplugins/go/tasks/plugins/array/outputs_test.go
```diff
@@ -353,6 +353,7 @@ func TestAssembleFinalOutputs(t *testing.T) {
 		ow := &mocks2.OutputWriter{}
 		ow.OnGetOutputPrefixPath().Return("/prefix/")
 		ow.OnGetOutputPath().Return("/prefix/outputs.pb")
+		ow.OnGetErrorPath().Return("/location/prefix/error.pb")
```
fg91, line 356: Tiny nit (non blocking):
Why `/location` before `/prefix` like in the line above? Currently the `error.pb` and the `outputs.pb` are in the same bucket uri.

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
bgedik:
```
type RemoteFileOutputReader struct {
	outPath        io.OutputFilePaths
	store          storage.ComposedProtobufStore
	maxPayloadSize int64
	errorRetriever ErrorRetriever
}
```

flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
```diff
@@ -99,6 +103,25 @@ func (pytorchOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx
 			return nil, flyteerr.Errorf(flyteerr.BadTaskSpecification, "Unable to create worker replica spec: [%v]", err.Error())
 		}
 
+		updateEnvVars := func(container *apiv1.Container) {
+			if container.Env == nil {
+				container.Env = make([]apiv1.EnvVar, 0, 2)
+			}
+			container.Env = append(container.Env, apiv1.EnvVar{
```
fg91, line 110: In a previous PR you added a pod name env var, see [here](https://github.com/flyteorg/flyte/pull/5616/files#diff-a668df54af25ccb5cdc7b6d1c073ffa3b71aa22b6ba34cba3a3dc8fcb5bba7f9).

I feel like only one of the two should be injected.

flyteplugins/go/tasks/plugins/array/outputs_test.go
bgedik: Sure, let me fix that. It was copied from some other example.

flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
bgedik: Will remove the pod one.

flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
```diff
@@ -99,6 +103,25 @@ func (pytorchOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx
...
+				Name: pluginsK8s.FlyteInternalWorkerNameEnvVarKey,
+				ValueFrom: &apiv1.EnvVarSource{
+					FieldRef: &apiv1.ObjectFieldSelector{
+						FieldPath: "metadata.name",
+					},
+				},
+			})
+			container.Env = append(container.Env, apiv1.EnvVar{
+				Name:  pluginsK8s.FlyteInternalDistErrorStrategyEnvVarKey,
+				Value: "1",
```
fg91, line 120: Is there any way we can use the enum value from the properties here so that everyone understands where this comes from instead of a string "1"?

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -500,13 +500,18 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
...
-
 				// checkpoint paths are not computed here because this function is only called when writing
 				// existing cached outputs. if this functionality changes this will need to be revisited.
 				outputPaths := ioutils.NewCheckpointRemoteFilePaths(ctx, nCtx.DataStore(), subOutputDir, ioutils.NewRawOutputPaths(ctx, subDataDir), "")
-				reader := ioutils.NewRemoteFileOutputReader(ctx, nCtx.DataStore(), outputPaths, 0)
-
-				gatherOutputsRequest.reader = &reader
+				reader, err := ioutils.NewRemoteFileOutputReader(ctx, nCtx.DataStore(), outputPaths, 0)
+				if err != nil {
+					gatherOutputsRequest.responseChannel <- struct {
```
fg91, line 508: I have to admit I don't understand this change, could you please explain?

flyteplugins/go/tasks/plugins/array/outputs_test.go
bgedik: Fixed

flytepropeller/pkg/controller/nodes/cache.go
```diff
@@ -216,7 +216,10 @@ func (n *nodeExecutor) WriteCatalogCache(ctx context.Context, nCtx interfaces.No
 		catalogKey.Identifier.Domain, catalogKey.Identifier.Name, catalogKey.Identifier.Version)
 
 	outputPaths := ioutils.NewReadOnlyOutputFilePaths(ctx, nCtx.DataStore(), nCtx.NodeStatus().GetOutputDir())
-	outputReader := ioutils.NewRemoteFileOutputReader(ctx, nCtx.DataStore(), outputPaths, 0)
+	outputReader, err := ioutils.NewRemoteFileOutputReader(ctx, nCtx.DataStore(), outputPaths, 0)
+	if err != nil {
+		return catalog.NewStatus(core.CatalogCacheStatus_CACHE_DISABLED, nil), errors.Wrapf(err, "failed to initialize the remote output file reader")
```
fg91, line 221: Is my understanding correct that when passing strategy `0`, we can't get an error?

flytepropeller/pkg/controller/nodes/array/handler.go
bgedik: `NewRemoteFileOutputReader` has to return an error now, if the error path is mis-structured, because it uses some of the parsing methods for `DataReference` that can potentially return an error.

The original constructor was not returning an error. So with the new structure, the error case needs to be handled. I don't know what the `gatherOutputsRequest.responseChannel <- struct` thingie does, but if you go up a little in the file, you'll see multiple examples of doing exactly that (repeated) when there is an unexpected error. So I just followed the convention already used in the file.

flytepropeller/pkg/controller/nodes/cache.go
bgedik: That is true.

fg91: Do you think it would be possible to not pass the strategy to `NewRemoteFileOutputReader` given that below you use a new function `ioutils.NewRemoteFileOutputReaderWithErrorAggregationStrategy`?

Avoiding this would allow us to not modify this file and several others at all where currently no errors can occur when constructing the output reader.

flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
bgedik: Let me do that in a different PR, as it requires more approvals for files involving Spark tests, etc.

flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
bgedik: Do you mean the textual representation? As in "Earliest"?

flytepropeller/pkg/controller/nodes/cache.go
bgedik: We can do that if you feel uncomfortable with these changes, but I felt like the no error nature of the `NewRemoteFileOutputReader` was more of an oversight.

flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
bgedik: Ok, let me make that change.

flytepropeller/pkg/controller/nodes/array/handler.go
bgedik: Addressed.

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
bgedik: Make it into a private interface `errorRetriever`

flyteplugins/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
wild-endeavor: let's keep the names short. these aren't meant to be user-facing anyways right?  just contributor facing.

flyteplugins/go/tasks/pluginmachinery/k8s/plugin.go
wild-endeavor: i'm okay either way.

flyteplugins/go/tasks/pluginmachinery/k8s/plugin.go
fg91: Yes, a single additional call. But pytorch (or other distributed tasks) generally are not the majority of tasks and for a normal python task this is never required. I'd argue we should avoid this unnecessary call if we already know it's void.

flyteidl/protos/flyteidl/core/execution.proto
fg91: Ok, sounds good 👍

flyteplugins/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
fg91: Didn't see https://github.com/flyteorg/flyte/issues/5665. Ok for me 👍

flyteplugins/go/tasks/pluginmachinery/k8s/plugin.go
fg91: I didn't consider this would lead to `k8s.Earliest`, enums in go are truly sad, let's do it as you proposed 👍

flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
fg91: Feel free to tag me as reviewer in the PR.

flyteplugins/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
fg91: TIL that capitalization controls visibility of variables in go, never mind, sorry.

flyteplugins/go/tasks/pluginmachinery/ioutils/remote_file_output_reader.go
```diff
@@ -3,77 +3,235 @@ package ioutils
...
 
 	"github.com/flyteorg/flyte/flyteidl/gen/pb-go/flyteidl/core"
 	"github.com/flyteorg/flyte/flyteplugins/go/tasks/pluginmachinery/io"
+	"github.com/flyteorg/flyte/flyteplugins/go/tasks/pluginmachinery/k8s"
 	"github.com/flyteorg/flyte/flytestdlib/storage"
 )
 
-type RemoteFileOutputReader struct {
-	outPath        io.OutputFilePaths
+type errorRetriever interface {
```
fg91, line 18: `RemoteFileOutputReader` already implements the `OutputReader` interface, see [here](https://github.com/flyteorg/flyte/blob/1e54d21c4d4ee74245f799a57b4bb8a5534e8368/flyteplugins/go/tasks/pluginmachinery/io/iface.go#L32).

I would prefer if this could be refactored in a way that doesn't require introducing another interface for this.

fg91: `IsError` also seems to serve the same purpose as `HasError`, `ReadError` the same purpose as `GetError`.

bgedik: @fg91  `RemoteFileOutputReader` still only implements `OutputReader` and not the `errorRetriever`.

An output reader has 2 functionalities: output reading and error reading. I refactored `RemoteFileOutputReader` so that it has an error retriever member, which can be either `singleFileErrorRetriever` or `earliestFileErrorRetriever`. Both `singleFileErrorRetriever` and `earliestFileErrorRetriever` implement the `errorRetriever` interface. It is the standard strategy pattern.

It looks like you are thinking of having a `MultiFileRemoteFileOutputReader` that extends from `RemoteFileOutputReader`. I don't quite like that (and also commented on the RFC about this sometime back), as we are creating a new output reader variety for each error reading strategy, instead of creating new error reader varieties.

Here is what I can do to avoid defining a new interface: I'll refactor `OutputReader` interface to inherit from an `ErrorReader` interface, so we won't introduce something new and use `ErrorReader` interface for the `errorRetriever` member (may be call it `errorReader`).

Let me do this and if you are still not comfortable with it, we can perhaps chat about it.

bgedik: See https://github.com/flyteorg/flyte/pull/5795/commits/0a12e95f4768a57f9770b3f36ec7849827acfe5a

wild-endeavor: yeah agree with that, no to MultiFileRemoteFileOutputReader i think.  just the error side is enough.

wild-endeavor (approved):

wild-endeavor (dismissed):

wild-endeavor (approved):

PR #5806, opened 2024-10-04
Feature: Edit Project Description

robert-ulbrich-mercedes-benz:

This change allows to set the description of a Flyte project. It could hold the responsible team for a Flyte project including the contact data or some other information. 

* A new config map file was introduced including keys in the Helm chart were updated for the `flyte-core` Helm chart so they can hold a name and a description for the project
* A new structure was introduced that holds the project information to be handed over to the database repository
* For the `flyte-binary` and the single binary approach the structure of the Helm chart was changed

## How was this patch tested?

* The implementation was tested by running the regular set up process for the single binary approach
* Also the regular `flyteadmin migrate seed-projects` was tested with the adapted `flyte-single-binary-local.yaml` config file

## Docs link

The program was tested solely for our own use cases, which might differ from yours

Robert Ulbrich [robert.ulbrich@mercedes-benz.com](mailto:robert.ulrbich@mercedes-benz.com), Mercedes-Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)


robert-ulbrich-mercedes-benz:
Again a question from my side: I have edited some lines of code that never had any test. now I am expected to set up some tests. For most of the files it would be not really meaningful to provide an unit test. Is there a way to skip the circumvent the code coverage check?

charts/flyte-core/values.yaml
```diff
@@ -43,11 +43,6 @@ flyteadmin:
       memory: 50Mi
   # -- Default regex string for searching configuration files
   configPath: /etc/flyte/config/*.yaml
-  # -- Initial projects to create
```
EngHabu, line 46: Won't this break backwards compatibility with old values files?

EngHabu (commented):
Thank you for your contribution! Let's think about backwards compatibility and merge your change

charts/flyte-core/values.yaml
robert-ulbrich-mercedes-benz: If a user updates the HELM chart and the values set for the customer remain, it will not do any harm, because the projects were already created before. Projects created once are not deleted, if they do not show up on this list. 

So only if the user wants to create new projects, he or she needs to adapt the new values for the initial projects

robert-ulbrich-mercedes-benz:
Thius merge request can be closed as there is another pull request that solved it: https://github.com/flyteorg/flyte/pull/5864

davidmirror-ops:
@robert-ulbrich-mercedes-benz thank you!

Issue #3220, opened 2023-01-09
[Core feature] Kafka Authentication


pingsutw:
### Motivation: Why do you think this is important?

In flyteadmin, we send a cloud event to Kafka. However, we don't have any authentication for Kafka right now. we should expose auth config, use `Sarama` api to read the config, and send the request to Kafka.

### Goal: What should the final outcome look like, ideally?

CloudEvent Config
```yaml
cloudevents:
    enable: true
    kafka:
      brokers: <kafka_ip>
      version:  1.2.1
      security_ protocol: SASL_SSL
      sasl_config: <config for SASL>
```

### Describe alternatives you've considered 

NA

###  Propose: Link/Inline OR Additional context 

Slack Thread:
- https://flyte-org.slack.com/archives/CP2HDHKE1/p1667307468058299
- https://flyte-org.slack.com/archives/CP2HDHKE1/p1667412056307629
- https://flyte-org.slack.com/archives/CP2HDHKE1/p1670577684352189
- https://flyte-org.slack.com/archives/CP2HDHKE1/p1667412056307629

[Authentication Methods Overview](https://docs.confluent.io/platform/current/kafka/overview-authentication-methods.html#authentication-methods-overview)

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

pingsutw:
cc @eapolinario

PR #5814, opened 2024-10-06
Add basic SASL and TLS support for Kafka cloud events

Sovietaced:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/3220

These changes are needed to use cloud events with kafka brokers with password or certificate auth enabled. 

This change expands the `KafkaConfig` struct to include some SASL fields for password based auth as well as a `TLSConfig` struct for certificate based auth.  It then provides a helper function to load these into the Sarama config. Lastly it adds helm support for loading the cloud events config from a secret since password based auth will contain sensitive fields. 

## How was this patch tested?
This was tested on our production Flyte clusters.

### DIFF
--- a/charts/flyte-core/values.yaml
+++ b/charts/flyte-core/values.yaml
@@ -945,6 +945,9 @@ external_events:
 # Cloud events are used to send events (unprocessed, as Admin see them) in cloud event format to
 # an SNS topic (or gcp equivalent)
 cloud_events:
+  # -- The name of the secret to use to alternatively load in cloud events configuration via a secret. Useful when the
+  # configuration contains secrets.
+  secretName: ""
   enable: false
   type: aws
   aws:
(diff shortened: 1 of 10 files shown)


charts/flyte-core/values.yaml
```diff
@@ -945,6 +945,9 @@ external_events:
 # Cloud events are used to send events (unprocessed, as Admin see them) in cloud event format to
 # an SNS topic (or gcp equivalent)
 cloud_events:
+  # -- The name of the secret to use to alternatively load in cloud events configuration via a secret. Useful when the
+  # configuration contains secrets.
+  secretName: ""
```
pingsutw, line 950: It's for the password, right? If so, could we read the password from the secret?
we did the similar thing for db [here](https://github.com/flyteorg/flyte/blob/fa274e1fa59d0500a8d2d5926d170ca7cbef37ac/datacatalog/pkg/runtime/application_config_provider.go#L29-L36)

Sovietaced: Yeah it is for the password and I was thinking this would work for any future cloud events types that have secrets in the future. I'll look into what you linked

pingsutw (approved):

PR #5816, opened 2024-10-06
Update ray.go not to fail when going suspend state.

aminmaghsodi:
We have a Queue to schedule our ray jobs, so we need it to wait in Suspended state of k8s
Now workflow fails, exactly when job state goes to 'Suspended'

### DIFF
--- a/flyteplugins/go/tasks/plugins/k8s/ray/ray_test.go
+++ b/flyteplugins/go/tasks/plugins/k8s/ray/ray_test.go
@@ -755,8 +755,8 @@ func TestGetTaskPhase(t *testing.T) {
 		{rayv1.JobDeploymentStatusRunning, pluginsCore.PhaseRunning, false},
 		{rayv1.JobDeploymentStatusComplete, pluginsCore.PhaseSuccess, false},
 		{rayv1.JobDeploymentStatusFailed, pluginsCore.PhasePermanentFailure, false},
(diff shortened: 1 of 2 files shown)


pingsutw (dismissed):
LGTM, thanks. just curious, which gang scheduler are you using? do you need to change anything in Flyte to make your gang scheduler work with Ray task?

flyteplugins/go/tasks/plugins/k8s/ray/ray_test.go
```diff
@@ -755,7 +755,8 @@ func TestGetTaskPhase(t *testing.T) {
 		{rayv1.JobDeploymentStatusRunning, pluginsCore.PhaseRunning, false},
 		{rayv1.JobDeploymentStatusComplete, pluginsCore.PhaseSuccess, false},
 		{rayv1.JobDeploymentStatusFailed, pluginsCore.PhasePermanentFailure, false},
-		{rayv1.JobDeploymentStatusSuspended, pluginsCore.PhaseUndefined, true},
+		{rayv1.JobDeploymentStatusSuspended, pluginsCore.PhaseQueued, true},
+		{rayv1.JobDeploymentStatusSuspending, pluginsCore.PhaseUndefined, true},
```
pingsutw, line 759: @aminmaghsodi could we map it to `pluginsCore.PhaseQueued` as well?

pingsutw (approved):

PR #5821, opened 2024-10-07
Use serviceAccount for Ray from the propeller configuration 

### LINKED ISSUE #5820: [BUG] Add serviceAccountName override for Ray
### Describe the bug

Currently, the serviceAccount pulls from the securityContext of the k8s cluster that the FlytePropeller is deployed at.
Instead, have the option to configure the propeller to use a specific serviceAccountName on k8s.

### Expected behavior

When the SA is set to something other than the default value, then pull the serviceAccountName from the propeller config

_No response_

### Screenshots

_No response_

- [X] Yes

- [X] Yes

peterghaddad:
## Tracking issue
Closes #5820

Currently, the serviceAccount pulls from the securityContext of the k8s cluster that the FlytePropeller is deployed at.
Instead, have the option to configure the propeller to use a specific serviceAccountName on k8s when the configuration value is set to a value other than "default".

### DIFF
--- a/flyteplugins/go/tasks/plugins/k8s/ray/ray.go
+++ b/flyteplugins/go/tasks/plugins/k8s/ray/ray.go
@@ -190,7 +190,7 @@ func constructRayJob(taskCtx pluginsCore.TaskExecutionContext, rayJob plugins.Ra
 	}
 
 	serviceAccountName := flytek8s.GetServiceAccountNameFromTaskExecutionMetadata(taskCtx.TaskExecutionMetadata())
-	if len(serviceAccountName) == 0 {
+	if len(serviceAccountName) == 0 || cfg.ServiceAccount != "default" {
 		serviceAccountName = cfg.ServiceAccount
 	}
 


peterghaddad:
In `flyteplugins/go/tasks/plugins/k8s/ray/ray.go`, I noticed that `submitterPodTemplate := buildSubmitterPodTemplate(headPodSpec, objectMeta, taskCtx)` uses the `headPodSpec` defined early in the source. [See here.](https://github.com/flyteorg/flyte/blob/master/flyteplugins/go/tasks/plugins/k8s/ray/ray.go#L121) Even when the configuration is changed it doesn't propagate down when calling `podSpec.ServiceAccountName = cfg.ServiceAccount`

Should we be using the built out headGroupSpec instead of the one originally defined? [See here.](https://github.com/flyteorg/flyte/blob/master/flyteplugins/go/tasks/plugins/k8s/ray/ray.go#L209)

@eapolinario any thoughts here? We also observed that modifying the serviceAccount through the `podTemplate` in the Flyte sdk does not propagate.

eapolinario:
> Even when the configuration is changed it doesn't propagate down when calling `podSpec.ServiceAccountName = cfg.ServiceAccount`

The issue is that the call to [GetServiceAccountNameFromTaskExecutionMetadata](https://github.com/flyteorg/flyte/blob/ffd72a09cb09079beccec5908ebd45b430ff81c5/flyteplugins/go/tasks/pluginmachinery/flytek8s/utils.go#L74-L76) always gets the security context from task execution metadata, which has the `run_as` set.

So even though we set the service account in https://github.com/flyteorg/flyte/blob/ffd72a09cb09079beccec5908ebd45b430ff81c5/flyteplugins/go/tasks/plugins/k8s/ray/ray.go#L119 this ends up being overridden in the call to `constructRayJob`.
 
As you're proposing in this PR, the crux of the change should be around https://github.com/flyteorg/flyte/blob/ffd72a09cb09079beccec5908ebd45b430ff81c5/flyteplugins/go/tasks/plugins/k8s/ray/ray.go#L192-L195
What if we used only used the service account name from the config if that set at all (i.e. instead of relying on setting its default value to `"default"` make that `""`) and if that's not set then we call `serviceAccountName = flytek8s.GetServiceAccountNameFromTaskExecutionMetadata(taskCtx.TaskExecutionMetadata())`.

@peterghaddad , wdyt?

Also, cc: @pingsutw

peterghaddad:
> > Even when the configuration is changed it doesn't propagate down when calling `podSpec.ServiceAccountName = cfg.ServiceAccount`
> 
> The issue is that the call to [GetServiceAccountNameFromTaskExecutionMetadata](https://github.com/flyteorg/flyte/blob/ffd72a09cb09079beccec5908ebd45b430ff81c5/flyteplugins/go/tasks/pluginmachinery/flytek8s/utils.go#L74-L76) always gets the security context from task execution metadata, which has the `run_as` set.
> 
> So even though we set the service account in
> 
> https://github.com/flyteorg/flyte/blob/ffd72a09cb09079beccec5908ebd45b430ff81c5/flyteplugins/go/tasks/plugins/k8s/ray/ray.go#L119
> 
> this ends up being overridden in the call to `constructRayJob`.
> As you're proposing in this PR, the crux of the change should be around
> 
> https://github.com/flyteorg/flyte/blob/ffd72a09cb09079beccec5908ebd45b430ff81c5/flyteplugins/go/tasks/plugins/k8s/ray/ray.go#L192-L195
> 
> 
> What if we used only used the service account name from the config if that set at all (i.e. instead of relying on setting its default value to `"default"` make that `""`) and if that's not set then we call `serviceAccountName = flytek8s.GetServiceAccountNameFromTaskExecutionMetadata(taskCtx.TaskExecutionMetadata())`.
> @peterghaddad , wdyt?
> 
> Also, cc: @pingsutw

I like this approach. I'll submit a PR for this change.

Issue #5832, opened 2024-10-10
[Core feature] Allow Flyteadmin to force terminate an execution


RRap0so:
### Motivation: Why do you think this is important?

With the work done in https://github.com/flyteorg/flyte/issues/5786 it was raised that a Workflow Execution can stay dangling if  Flyte Propeller is having problems.

Flyteadmin should support a force flag so it knows that it should remove the execution instead of waiting to be garbage collected by FlytePropeller. This should make management of Workflows easier than having to go to the kubernetes cluster and removing them using kubectl.

On the Flyteadmin side this is now covered by the following [PR](https://github.com/flyteorg/flyte/pull/5833). Additional PRs are necessary to implement this features in the Flyte clients (flytectl, flyteconsole).

### Goal: What should the final outcome look like, ideally?

Both flyteconsole and flytectl are able to pass a force flag to Flyteadmin indicating that Flyteadmin should remove the execution/CRD from the kubernetes cluster.

### Describe alternatives you've considered 

N/A

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #5833, opened 2024-10-10
Rafaelraposo/force abort impl (#4)

RRap0so:
## Tracking issue
Related to #5832 

Add a force flag on the server side of Flyte to be able to terminate executions if FlytePropeller is not running in the kubernetes cluster.

This adds the force bool into `ExecutionTerminateRequest`. This in turn will be passed to the `k8s_executor`. If it's a Forced termination, it will remove the finalizer from the execution CRD and delete.

If it's not forced it keeps doing just a deletion (like it was before)

## How was this patch tested?

Unit tested.

## Related PRs

TBD

## Docs link

TBD


wild-endeavor:
@EngHabu and @hamersaw had some concerns around this I think?  Finalizers are there to help ensure that any downstream resources are cleaned up properly right?  Will this still happen properly if the finalizers are forcefully removed?  What's the downside of doing this?

pvditt:
@wild-endeavor this is a follow up to https://github.com/flyteorg/flyte/pull/5788 that added finalizers to flyteworkflows on creation as opposed to waiting for propeller to pick them up and add finalizers. This was to handle a scenario where flyteworkflow resources were getting deleted prior to getting finalizers set causing issues with subworklow status getting stuck.

I'll take a look at this PR this evening.

flyteadmin/pkg/workflowengine/impl/k8s_executor.go
```diff
@@ -94,9 +95,35 @@ func (e K8sWorkflowExecutor) Abort(ctx context.Context, data interfaces.AbortDat
...
+	if data.Force {
+		logger.Debugf(ctx, "Force deleting execution [%+v] in cluster: %s", data.ExecutionID, target.ID)
+
+		// Remove finalizers if any
+		if w.Finalizers != nil || w.ObjectMeta.Finalizers != nil {
+			w.Finalizers = []string{}
+		}
+
+		// Write the updated workflow
+		if _, err := target.FlyteClient.FlyteworkflowV1alpha1().FlyteWorkflows(data.Namespace).Update(ctx, w, v1.UpdateOptions{}); err != nil {
```
pvditt, line 117: with the force delete, don't we still want the flyteworkflow to get deleted after the finalizers are removed?

flyteadmin/pkg/workflowengine/impl/k8s_executor.go
```diff
@@ -94,9 +95,35 @@ func (e K8sWorkflowExecutor) Abort(ctx context.Context, data interfaces.AbortDat
 	if err != nil {
 		return errors.NewFlyteAdminErrorf(codes.Internal, err.Error())
 	}
-	err = target.FlyteClient.FlyteworkflowV1alpha1().FlyteWorkflows(data.Namespace).Delete(ctx, data.ExecutionID.GetName(), v1.DeleteOptions{
-		PropagationPolicy: &deletePropagationBackground,
-	})
+
+	w, err := target.FlyteClient.FlyteworkflowV1alpha1().FlyteWorkflows(data.Namespace).Get(ctx, data.ExecutionID.GetName(), v1.GetOptions{})
```
pvditt, line 99: could you put this fetch under if data.Force so we don't do this fetch if we're just deleting without removing finalizers first

flyteadmin/pkg/workflowengine/impl/k8s_executor.go
```diff
@@ -94,9 +95,35 @@ func (e K8sWorkflowExecutor) Abort(ctx context.Context, data interfaces.AbortDat
 	if err != nil {
 		return errors.NewFlyteAdminErrorf(codes.Internal, err.Error())
 	}
-	err = target.FlyteClient.FlyteworkflowV1alpha1().FlyteWorkflows(data.Namespace).Delete(ctx, data.ExecutionID.GetName(), v1.DeleteOptions{
```
pvditt, line 97: I think we still want to re-add this after the if data.force{... code block

flytectl/cmd/delete/delete.go
```diff
@@ -30,9 +32,18 @@ func RemoteDeleteCommand() *cobra.Command {
 		Short: deleteCmdShort,
 		Long:  deleteCmdLong,
 	}
+	var force bool
+	deleteCmd.PersistentFlags().BoolVarP(&force, "force", "f", false, "Force deletion without confirmation")
```
pvditt, line 36: I'm unfamiliar if there are any examples, but would be nice if just needed to include the force flag. 

Command could be:

 flytectl delete execution kxd1i72850  -d development  -p flytesnacks --force

hamersaw:
> Finalizers are there to help ensure that any downstream resources are cleaned up properly right? Will this still happen properly if the finalizers are forcefully removed? 
Only comment is that we need to check that all k8s resources that Flyte starts are cleaned up as part of removing finalizers and deleting the CR. IIUC this is meant to be used on already terminal Flyte workflows, but there are no safeguards here - a user could "force" delete an active execution and that may orphan Pods. Am I understanding this correctly?

EngHabu:
I thought about this more, I think it's an ok change because the finalizer isn't what guarantees cleanup, it's the OwnerReference and the propagation delete policy. The finalizer is there to allow propeller to act on the delete request for the workflow CRD.
My issue though is that propeller adds finalizers on the created resources as well, if we are removing the finalizer because propeller isn't running anymore, the workflow CRD will still hangaround because pods (and other resources) will have finalizers...

flyteadmin/pkg/workflowengine/impl/k8s_executor.go
RRap0so: Done.

flyteadmin/pkg/workflowengine/impl/k8s_executor.go
RRap0so: Of course, good spot.

Issue #4464, opened 2023-11-20
[Docs] document the process of debugging the mpi operator plugin installation


samhita-alla:
### Description

we have documentation on installing the MPI operator and running workflows with it. it's essential that we also address troubleshooting steps. need to review [this slack thread](https://discuss.flyte.org/t/13593667/hi-community-i-m-getting-an-unknown-status-on-every-workflow) to identify debugging steps and document them accordingly.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

10sharmashivam:
Hey, I’d like to take this issue and start working on documenting the debugging process for the MPI operator plugin installation.

davidmirror-ops:
@10sharmashivam sure. Please let us know soon if you have questions. Thanks!

10sharmashivam:
Hi @davidmirror-ops,

I went through MPI operator installation steps by step, and also through slack conversation mentioned in description. During my review, I noticed two potential causes for worker pods failing to start: resource allocation (CPU/memory) and scheduling issues (gang scheduling).

Would it be useful to include troubleshooting steps for both scenarios? I was thinking of covering both the resource request adjustments (as mentioned in the Slack thread: requests=Resources(cpu="1", mem="2000Mi")) and the gang scheduling solution.

Let me know if this level of detail would be helpful, or if I should focus on just one of these aspects.

davidmirror-ops:
@10sharmashivam thanks for going through the process. From reading the long Slack thread, I think we should refrain from generalizing but probably remind users about common/possible issues:

1. Request enough resources for the MPI Tasks. How much is enough? That depends on the task but at least we should remind users here how to request resources from the task decorator
2. If using a custom image, remind users or point them to the docs about the different [registration patterns](https://docs.flyte.org/en/latest/user_guide/flyte_fundamentals/registering_workflows.html#registration-patterns). This user, in particular, was using the "production flow" described in that document
3. I wouldn't mention gang scheduling (which is becoming an overloaded term and some could argue MPI is not gang or even co-scheduling), and wouldn't hint at a particular solution as this is not mentioned in the reference thread.

PR #5857, opened 2024-10-18
DOC MPI Operator troubleshooting steps added

10sharmashivam:
## Tracking issue

References #4464 

These changes provide important troubleshooting steps for users setting up the MPI operator for distributed training jobs on Flyte. The steps cover common issues like insufficient resource allocation for worker pods and clarify the correct workflow registration method, especially when using custom images. This will help users avoid common pitfalls and streamline the MPI operator setup process.

This pull request adds troubleshooting steps to the MPI operator installation documentation. Specifically, the following issues are addressed:

	•	Insufficient Resources for Worker Pods: Steps to ensure sufficient CPU and memory requests are set.
	•	Workflow Registration Method Errors: Clarification on using the correct registration method when working with custom images, with a link to the relevant Flyte documentation.

## How was this patch tested?

This is a documentation update based on Flyte Slack's conversation about MPI operator installation issues.

### DIFF
--- a/docs/deployment/plugins/k8s/index.rst
+++ b/docs/deployment/plugins/k8s/index.rst
@@ -1035,3 +1035,39 @@ Wait for the upgrade to complete. You can check the status of the deployment pod
   kubectl get pods -n flyte
 
 Once all the components are up and running, go to the `examples section <https://docs.flyte.org/en/latest/flytesnacks/integrations.html#native-backend-plugins>`__ to learn more about how to use Flyte backend plugins.
+
+Troubleshooting MPI Operator Installation
+-----------------------------------------
+
+This section covers common issues encountered during the setup of the MPI operator for distributed training jobs on Flyte.
+
+**Worker Pods Failing to Start (Insufficient Resources)**
+
+MPI worker pods may fail to start or exhibit scheduling issues, leading to job timeouts or failures. This often occurs due to resource constraints (CPU, memory, or GPU) in the cluster.
+
+1. Adjust Resource Requests:
+Ensure that each worker pod has sufficient resources. You can adjust the resource requests in your task definition:
+
+.. code-block:: bash
+
+      requests=Resources(cpu="<your_cpu_request>", mem="<your_mem_request>")
+
+Modify the CPU and memory values according to your cluster's available resources. This helps prevent pod scheduling failures caused by resource constraints.
+
+2. Check Pod Logs for Errors:
+If the worker pods still fail to start, check the logs for any related errors:
+
+.. code-block:: bash
+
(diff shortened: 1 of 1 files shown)


docs/deployment/plugins/k8s/index.rst
```diff
@@ -1035,3 +1035,39 @@ Wait for the upgrade to complete. You can check the status of the deployment pod
   kubectl get pods -n flyte
 
 Once all the components are up and running, go to the `examples section <https://docs.flyte.org/en/latest/flytesnacks/integrations.html#native-backend-plugins>`__ to learn more about how to use Flyte backend plugins.
+
+Troubleshooting MPI Operator Installation
```
davidmirror-ops, line 1039: I'd prefer this to be a heading under a new section called "Common issues" or something along those lines.
Then use the same structure the above section follow to have one tab for each plugin. For now it would be only "MPI", but that would leave the structure for others to add troubleshooting steps for other plugins as well. Let me know if there are questions

10sharmashivam: Thank you for the feedback! I’ve updated the documentation by moving the troubleshooting steps under a new section titled "Troubleshooting Plugin Deployments" and organized the troubleshooting steps for MPI under a dedicated group tab, which can later be used for each plugin. Please let me know if any further adjustments are needed.

davidmirror-ops: Super sorry about this but looking closer, I think putting this section on the[ MPI guide itself ](https://github.com/flyteorg/flytesnacks/blob/master/examples/kfmpi_plugin/README.md)would make it easier to discover and use, especially because the issues on the Slack thread are not about the plugin deployment itself but on using it. Let me know if that works for you and how I can help.

10sharmashivam: Thank you for your feedback on my previous PR. I’ve made the suggested changes and submitted a new PR that includes the troubleshooting steps for the MPI operator, which have been shifted to the MPI Guide itself. You can find the new PR here: [[Link to my new PR](https://github.com/flyteorg/flytesnacks/pull/1756)].

I appreciate your guidance, and I’m looking forward to your thoughts on the updates!

Just wanted to know one thing, the slack conversation also mentions the Horovod installation issues. Should I address that as well, along with other two issues which I already addressed? If yes, where should I add them- [MPI guide](https://docs.flyte.org/en/latest/flytesnacks/examples/kfmpi_plugin/index.html) or in [Distributed training using Horovod](https://docs.flyte.org/en/latest/flytesnacks/examples/kfmpi_plugin/mpi_mnist.html).

PR #5860, opened 2024-10-18
Update propeller sharding docs - types needs to be capitalized

cpaulik:

Using the lowercase version of e.g. `project` the following error is raised by flytepropeller-manager

```
level=error msg="\n\n1 error(s) decoding:\n\n* cannot parse 'shard.type' as int: strconv.ParseInt: parsing \"project\": invalid syntax"
```
## How was this patch tested?

Tested uppercase `Project` in our Flyte deployment. I did not test the uppercase `Domain` option. That one is just an assumption based on the need to uppercase `Hash` and `Project`.


davidmirror-ops (approved):
Thank you!

PR #5870, opened 2024-10-19
Remove duplicate recovery interceptor

Sovietaced:
## Tracking issue

While working on an RFC I noticed that there is a duplicate recovery interceptor installed inside the Flyte Admin server. We ideally want to use the custom one which ties into metrics.


flyteadmin/pkg/server/service.go
```diff
@@ -136,7 +134,6 @@ func newGRPCServer(ctx context.Context, pluginRegistry *plugins.Registry, cfg *c
 	)
 
 	serverOpts := []grpc.ServerOption{
-		// recovery interceptor should always be first in order to handle any panics in the middleware or server
```
Sovietaced, line 139: unnecessary comment

wild-endeavor (approved):

Issue #4622, opened 2023-12-19
[Feature request] Make Flyte's All Scope configurable


givanovexpe:
### Describe the bug

In Flyte Admin we have a hardcoded OAuth2 token scope called "all". It's a remnant from the initial implementation of OAuth2 in flyte when there was a plan to intrioduce some lightweight RBAC in flyte. "All" scope was meant to be given to admin users. However RBAC was never implemented, but the "all" scope remained. It requires creating a custom authorization server in Okta, and this cannot always be done.

Here is a tech overview of where we use it. We configure the PluginRegistry with Blanket Authorization:
https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/server/service.go#L80

In the BlanketAuthorization function we check if the identity context has ScopeAll
https://github.com/flyteorg/flyte/blob/master/flyteadmin/auth/interceptor.go#L21

which is currently set to “all”
https://github.com/flyteorg/flyte/blob/master/flyteadmin/auth/constants.go#L24

For the userauth path (i.e. from the flyteconsole) we autoinject the “all” scope in the context created from the id_token, i.e the console/admin app using the flyteconsole okta app with openid,profile scopes.

https://github.com/flyteorg/flyte/blob/master/flyteadmin/auth/token.go#L134
https://github.com/flyteorg/flyte/blob/master/flyteadmin/auth/handlers.go#L283

But for the identity context created from access_token we don’t, so unless we have the “all” scope returned from okta as we have hardcoded the check in the admin, we will fail the client with “authenticated user doesn’t have required scope”

https://github.com/flyteorg/flyte/blob/master/flyteadmin/auth/handlers.go#L276
https://github.com/flyteorg/flyte/blob/master/flyteadmin/auth/token.go#L75

### Expected behavior

The all scope should be dropped altogether (since there is no RBAC defined anyway) or should be made configurable, i.e. the value can be something else, for instance "profile". Thus we can connect Flyte with Okta for appAuth.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #5871, opened 2024-10-20
[RFC] Flyte Admin RBAC + Project/Domain Isolation

Sovietaced:
## Tracking issue
Related to https://github.com/flyteorg/flyte/issues/5189
Related to https://github.com/flyteorg/flyte/issues/4622


eapolinario:
cc: @robert-ulbrich-mercedes-benz

rfc/system/0000-flyteadmin-rbac.md
```diff
@@ -0,0 +1,203 @@
...
+* Filtering resources at the database works natively with any pagination
+* Filtering resources at the database is the most performant since the database engine is fast and ends up returning a smaller data set over the network
+* Filtering resources at the database plays nicely with logic to handle records that aren't found and does not leak data about which resources may or may not be present.
+
+We propose to have utility functions for two different workflows: 
+1. Creating Resources
+   * This is a dedicated utility function used in repository code to create resource since you cannot add a WHERE clause filter for records that don't exist yet :) 
+   * ```go
+     func (r *ExecutionRepo) Create(ctx context.Context, input models.Execution, executionTagModel []*models.ExecutionTag) error {
+         if err := util.AuthorizeResourceCreation(ctx, input.Project, input.Domain); err != nil {
```
katrogan, line 147: This is interesting, I like the middleware approach because it lets you intercept new service methods for free. In this proposal, how do you ensure that new DB methods always call the authorization util?

rfc/system/0000-flyteadmin-rbac.md
```diff
@@ -0,0 +1,203 @@
...
+authorization:
+  policies: # policies is a list instead of a map since viper map keys are case insensitive
+    - name: admin
+      rules: # list of rules
+        - name: "allow all" # name / description of the rule
+          methodPattern: ".*" # regex pattern to match on gRPC method
+    - name: read-only
+      rules:
+        - name: "read everything"
+          methodPattern: "Get.*|List.*"
```
katrogan, line 83: Curious about your thoughts on brittleness for regex parsing here. We mostly try to follow restful gRPC method names but we have a few notable exceptions like `TerminateExecution`, `RecoverExecution`, `RelaunchExecution`. If upsteam OSS Flyte merges a change that adds a new syntax, is it the responsibility of whoever deploys Flyte at their company and manages these policies to update the regexes appropriately?

rfc/system/0000-flyteadmin-rbac.md
```diff
@@ -0,0 +1,203 @@
...
+      rules:
+        - name: "r/w for the mapping project in dev only"
+          methodPattern: ".*"
+          project: mapping # access to only the mapping project
+          domain: development # access to only the development domain within the mapping project
+    - name: ci
+      rules:
+        - name: "r/w for every project in production"
+          methodPattern: ".*"
+          domain: production # you can wildcard project and declare domain level access
```
katrogan, line 94: does the same work for wild carding domain if you want, for example, mapping team write access?

rfc/system/0000-flyteadmin-rbac.md
```diff
@@ -0,0 +1,203 @@
...
+const (
+    ProjectTargetResourceScope = 0
+    DomainTargetResourceScope = 1
+)
+```
+
+#### Authorization Utils + DB Layer
+
+The final piece of the puzzle is what performs resource level authorization and filtering. Historically, I have found 
+that the best (albeit challenging) way to do this is at the database layer for a few reasons:
```
katrogan, line 137: This proposal makes sense, but there's some notable instances where we could have unauthorized side effects

specifically, for create execution, a user who has read access may be able to read the respective launch plan, we then create the CRD and only then insert the model in the DB (which gets rejected) here: https://github.com/flyteorg/flyte/blob/5f695895ed6a7fa45d980023ab874591184b0fc1/flyteadmin/pkg/manager/impl/execution_manager.go#L1027-L1062

so in this scenario, we create a rogue Flyte CRD that never gets committed to the db (operation fails with permission denied) that propeller will attempt to process and send event updates for, leading to potential system weirdness

rfc/system/0000-flyteadmin-rbac.md
```diff
@@ -0,0 +1,203 @@
...
+will make requests to the repository layer. The repository layer will leverage some utility functions to 1) authorize 
+whether resources can be created in a project/domain and 2) generate some SQL filters to be injected into queries that 
+read/update/delete resources. 
+
+Ultimately the authorization interceptor will provide RPC level authorization and set context for later use. The changes to the repository layer 
+will provide resource level isolation within the target RPC. Below is a breakdown of the three primary components of the design. 
+
+#### Authorization Config 
+The authorization config will be used to configure the authorization behavior. The first thing it describe is a way to
+resolve roles from the user's identity context. Ideally this should be flexible enough to resolve a role from different
```
katrogan, line 42: for a [multi-cluster set-up](https://docs.flyte.org/en/latest/deployment/deployment/multicluster.html#scaling-beyond-kubernetes), where the control plane (admin) and data plane (propeller) reside on separate clusters, propeller authenticates with a set of client credentials against flyteadmin to send execution events and create executions (child workflows, e.g. calling a launch plan in code)

if I'm understanding correctly, in this case we could use the userID of the propeller credentials to permit deployment-wide RW access? this isn't quite userID, but I think your suggestion can very well support applicationID using this approach. My understanding is that it's a bit more tricky to add scopes & claims to applications rather than user identities

rfc/system/0000-flyteadmin-rbac.md
Sovietaced: It is cooperative so you cannot ensure it. The implementation I have adds it for every resource possible IIRC so it would probably have to be enforced through code review

rfc/system/0000-flyteadmin-rbac.md
Sovietaced: Using something besides `.*` on the method pattern is opt in so yeah I think it would be up to the company to update their regexes and I'd imagine that is the behavior they would want. Ideally they validate the ux in non-prod and roll out to prod. 

I'm also open to supporting different matching types like prefix and exact matches.

rfc/system/0000-flyteadmin-rbac.md
Sovietaced: Yes absolutely

rfc/system/0000-flyteadmin-rbac.md
Sovietaced: That is a good point, I wasn't aware of that interaction. 

I think there are a couple ways to work around this. 
1. Bring authorization for creation into the application layer (in addition to the db layer I suppose). 
2. Rework the code such that the execution is written to the database (maybe partially) prior to creating the CRD. And then writing the rest to the DB after the CRD is created if necessary. 

I am being a bit naive with the complexity to approach number 2 but I do think this is a general problem with distributed systems. For example, the way the code is written as is makes the system susceptible to ghost CRDs since the system could crash before writing to the DB. I also understand this issue is probably much more likely to happen with authorization errors though :)

The architectural issue I see with this type of behavior boils down to how business logic does what is effectively a distributed transaction across other services (submitting CRDs in this case) and writing to the database. The most robust way to attempt this sort of behavior that I've seen is to always write to the DB first (this is the source of truth) and then opportunistically do the rest of the distributed transaction. For cases where the server crashes after writing to the DB you'll basically need some anti entropy background task to handle edge cases.

rfc/system/0000-flyteadmin-rbac.md
Sovietaced: We use a multi cluster setup and admittedly our implementation of this proposal actually has a separate config field for allow listing certain user IDs, which in our case are Okta client Ids for the propeller Okta app. This allow list gives them full access to all APIs and all projects/domains. This was more of an incremental thing I added when I realized very quickly that authorization broke propeller :)

When I was writing the proposal I realized that I could probably do away with user ID allow lists in our implementation and support role resolution more generally by allowing users to pick out which values from the token they want to potentially match with policies. So yeah, I think for Flyte workloads you'd just turn on the user ID or application ID resolution strategy and create a role like so, not having to deal with scopes at all.  

```
 - name: 0oahjhk34aUxGnWcZ0h7 # the names can even include things like okta app IDs
      rules:
        - name: "flyte propeller"
          methodPattern: ".*"
```   

I guess to be clear, the role resolution strategies configured will ultimately generate a set of strings and then try and match those strings against the policy map keys. 

But I'm open to doing something special for the Flyte workloads

katrogan: this proposal sounds like it unblocks propeller access! I'm curious how this works for supporting dynamically provisioned app credentials (besides just propeller, but for things like CICD, etc) since I believe Okta doesn't allow you to attach claims to applications unless you use custom claims in the authorization server which is a bit more involved

rfc/system/0000-flyteadmin-rbac.md
katrogan: For sure, it's definitely a problem now even without authz checks failing. I recall we waffled on this approach initially because we didn't want to commit the entry to the db until after the CRD was created (since the workflowengine packages gracefully handles already exists on retries, but falsely implying an execution was created in the reversed case was more confusing)

if we do bring authorization to the application layer (even for these exceptional cases) I'm a bit concerned we're now depending on code review to enforce authorization checks in both the db and application levels

davidmirror-ops:
2024-11-7 Contributor's sync notes: Jason introduced the proposal and expects to work on the implementation as a feature gate.

chicco785:
What's the current status of the proposal?

chicco785:
> > What's the current status of the proposal?
> 
> I am working through an initial pull request to add this functionality.

That's cool and I am happy to have a look! What about the TSC evaluation phases? I have seen other posts where RBAC was dismissed...

Sovietaced:
> That's cool and I am happy to have a look! What about the TSC evaluation phases? I have seen other posts where RBAC was dismissed...

I'm not familiar with the TSC evaluation phase but everyone is eager for this to get over the line.

Sovietaced:
I have completed an initial implementation. Please take a look: https://github.com/flyteorg/flyte/pull/6190

Sovietaced:
While the PR awaits review a few folks have checked it out in their forks and have got it working so feel free to try it out.

davidmirror-ops:
Acceptance of this proposal is contingent on the consolidation of SIG-RBAC with the scope proposed [here](https://github.com/flyteorg/community/blob/main/sig-rbac/charter.md)

testinfected:
Hi, I was wondering what's missing to get forward and merge the PR? We absolutely would love this feature, which is required for us to adopt flyte.

testinfected:
> Hi, I was wondering what's missing to get forward and merge the PR? We absolutely would love this feature, which is required for us to adopt flyte.

@davidmirror-ops ?

Sovietaced:
Closing this as there hasn't been any movement and development is focused on Flyte 2.

Issue #5908, opened 2024-10-24
[Core feature] Tuple & NamedTuple Support


Mecoli1219:
### Motivation: Why do you think this is important?

This Issue aggregates all PR related to support Tuple & NamedTuple.

### Goal: What should the final outcome look like, ideally?

Support Tuple & NamedTuple type in flyte system.

### Describe alternatives you've considered 

- [x] RFC
- [x] 1st version
  - [x] 1st version IDL
  - [x] 1st version Flyte
  - [x] 1st version Flytekit
- [ ] Try removing LiteralTypeForLiteral to reduce the complexity of FlyteIDL.
- [ ] 2nd version

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #5909, opened 2024-10-24
Remove `LiteralTypeForLiteral` by creating `IsInstance` function

Mecoli1219:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5908

The current Tuple IDL design is complicated by this function since if we want to guess the `LiteralType` from `Literal`, we have to store the name of the tuple and the name of each field inside the tuple in the `Literal`. 

However, guessing the `LiteralType` is not necessary for Flyte since whenever we create a new literal, there must be a target type (maybe inputs or outputs, but they will always have a defined type). Also, almost all use cases of `LiteralTypeForLiteral` are followed by a function `AreTypesCastable`, we could simply combine these two functions into a function determining whether a Literal could be instantiated by a LiteralType. It is really similar to `isinstance()` function in Python.

- [x] I created an `IsInstance()` for the flytepropeller to check whether a Literal could be represented by a LiteralType.
- [x] Remove `LiteralTypeForLiteral` and update all necessary code in FlyteAdmin & FlytePropeller.
- [x] Update the unit tests
  - [x] Change unit tests for `LiteralTypeForLiteral` to `IsInstance`
  - [x] Update the error message for other cases.

## How was this patch tested?
- [x] Unit testing
- [x] Built and run some Flytekit test

## Additional Concerns
<s>Someone suggested that we shouldn't merge this PR before the release</s> 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR implements a significant change in Flyte's type checking system by replacing the deprecated LiteralTypeForLiteral with a new IsInstance function. The changes involve refactoring the array handler's type checking mechanism, fixing a bug in blob validation by adding dimensionality checks, and simplifying type checking logic while adding safety checks for map key existence.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 5
</div>


Future-Outlier (changes requested):
Can you test every test case be affected in flytekit remote?
and provide screenshot?
I think this function is not well tested by unit test and integration test.

flytepropeller/pkg/compiler/validators/utils.go
```diff
@@ -170,130 +169,250 @@ func UnionDistinctVariableMaps(m1, m2 map[string]*core.Variable) (map[string]*co
...
+
+type collectionInstanceChecker struct {
+	literalType *core.LiteralType
+}
+
+func (t collectionInstanceChecker) isInstance(lit *core.Literal) bool {
+	if _, ok := lit.GetValue().(*core.Literal_Collection); !ok {
+		return false
+	}
+	for _, x := range lit.GetCollection().Literals {
```
pingsutw, line 242: If we add the tuple literal type but still represent the literal as a collection, how do we check if this collection is a tuple or not

Mecoli1219: The type of Literal for tuple will be Collection, but the LiteralType won't. We'll have to create another new `InstanceChecker` (perhaps `tupleInstanceChecker`) to deal with that.

flytepropeller/pkg/compiler/validators/utils.go
```diff
@@ -170,130 +169,250 @@ func UnionDistinctVariableMaps(m1, m2 map[string]*core.Variable) (map[string]*co
...
-	return nil
 }
 
-func GetTagForType(x *core.LiteralType) string {
-	if x.GetStructure() == nil {
-		return ""
+func IsInstance(lit *core.Literal, t *core.LiteralType) bool {
+	instanceChecker := getInstanceChecker(t)
+
+	if lit.GetOffloadedMetadata() != nil {
```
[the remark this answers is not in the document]
wild-endeavor: sorry incomplete review... will get back to this later.

Mecoli1219:
@Future-Outlier I tested the IsInstance with the flytekit's unit tests and integration tests. I couldn't run eager in my environment, and it is not related to this update. Besides that, all test works successfully.
image
image

Mecoli1219:
Updated `flytekit` Integration test:
image
unit test:
image
image

pingsutw (dismissed):

Mecoli1219:
@wild-endeavor @eapolinario This PR is ready to review, thanks! 
@pingsutw I think there is nothing related to coding changed since your last review. Only `go.mod` & `go.sum` is updated.

flytepropeller/pkg/compiler/validators/utils.go
```diff
@@ -170,125 +169,244 @@ func UnionDistinctVariableMaps(m1, m2 map[string]*core.Variable) (map[string]*co
...
-	case *core.Literal_OffloadedMetadata:
-		return l.GetOffloadedMetadata().GetInferredType()
 	}
-	return nil
+}
+
+func IsInstance(lit *core.Literal, t *core.LiteralType) bool {
+	instanceChecker := getInstanceChecker(t)
+
+	if lit.GetOffloadedMetadata() != nil {
```
wild-endeavor, line 406: let's make this just return true... if it's offloaded, just return true.

If you just return true, it might be a false positive.  That means the offloaded literal might actually be of the incorrect type.  But this should be fine... it'll result in a flytekit error at runtime.  that should be fine.

As of this PR, the call-sites for this new function are:
image

All these should be either irrelevant for offloaded, or just result in a runtime error.

cc @EngHabu

flytepropeller/pkg/compiler/validators/utils.go
```diff
@@ -170,125 +169,244 @@ func UnionDistinctVariableMaps(m1, m2 map[string]*core.Variable) (map[string]*co
...
+	if blobType == nil {
+		return false
+	}
+
+	// Empty blobs should match any blob.
+	if blobType.GetFormat() == "" || t.literalType.GetBlob().GetFormat() == "" {
+		return true
+	}
+
+	return blobType.GetFormat() == t.literalType.GetBlob().GetFormat()
```
wild-endeavor, line 278: I think we should also add a check to make sure `dimensionality` also matches.  (i.e. FlyteFile should not be an instance of FlyteDirectory)

flytepropeller/pkg/controller/nodes/array/handler.go
```diff
@@ -600,7 +598,7 @@ func (a *arrayNodeHandler) Handle(ctx context.Context, nCtx interfaces.NodeExecu
 				// use the OffloadLargeLiteralKey to create  {OffloadLargeLiteralKey}_offloaded_metadata.pb file in the datastore.
 				// Update the url in the outputLiteral with the offloaded url and also update the size of the literal.
 				offloadedOutputFile := v1alpha1.GetOutputsLiteralMetadataFile(outputLiteralKey, nCtx.NodeStatus().GetOutputDir())
-				if err := common.OffloadLargeLiteral(ctx, nCtx.DataStore(), offloadedOutputFile, outputLiteral, a.literalOffloadingConfig); err != nil {
+				if err := common.OffloadLargeLiteral(ctx, nCtx.DataStore(), offloadedOutputFile, outputLiteral, outputLiteralTypes[outputLiteralKey], a.literalOffloadingConfig); err != nil {
```
wild-endeavor, line 601: Can we change `common.OffloadLargeLiteral` to remove the inferredType arg?  I don't think we need it.  Then we can remove this `outputLiteralTypes` map completely.

flytepropeller/pkg/compiler/validators/utils.go
```diff
@@ -170,125 +169,244 @@ func UnionDistinctVariableMaps(m1, m2 map[string]*core.Variable) (map[string]*co
...
-		err := fmt.Errorf("got unknown literal type: [%v].\n"+
-			"Suggested solution: Please update all your Flyte deployment images to the latest version and try again", lt)
-		return err
-	}
-	if lt.GetCollectionType() != nil {
-		return ValidateLiteralType(lt.GetCollectionType())
+func (t unionInstanceChecker) isInstance(lit *core.Literal) bool {
+	unionType := t.literalType.GetUnionType()
+
+	if u := lit.GetScalar().GetUnion().GetType(); u != nil {
```
wild-endeavor, line 341: @eapolinario i want to look at this carefully.

wild-endeavor (commented):
Minor comments...
Just to be clear, we are doing this because for tuples, we really would like to _not_ have to create a new `Literal` (but we will create a new `LiteralType`).  If we're not going to create a new literal, then LiteralTypeForLiteral will never return that new LiteralType.

This function call was always felt a bit unnecessary as well.

@EngHabu - do you remember any other reasons for wanting to remove this function?

flytepropeller/pkg/compiler/validators/utils.go
wild-endeavor: talked to enghabu about this - ignore this comment for now, and let's keep this function as you have it.  we'll get rid of the inferred type in the future.

flytepropeller/pkg/controller/nodes/array/handler.go
Mecoli1219: I think I have to keep it for this PR, but we can remove it after the `inferredType` arg in `common.OffloadLargeLiteral()` is removed.

Sovietaced:
Closing this stale PR since development has moved towards Flyte 2.

PR #5911, opened 2024-10-24
Set HttpOnly and Secure flags in session cookies

eapolinario:
Setting these 2 fields is standard practice. All modern browsers implement them. 

We set `HttpOnly` and `Secure` flags in all cookies produced by Flyte. Notice that those are generated only if auth is enabled.

More information about those flags:
- https://owasp.org/www-community/HttpOnly: The HttpOnly option prevents JavaScript code from accessing the cookie. This can provide some protection from cookie theft if an attacker successfully exploits a Cross-Site Scripting (XSS) vulnerability.
- https://cwe.mitre.org/data/definitions/614.html: The Secure cookie flag is an option that can be used to prevent the browser from sending the cookie over a plaintext connection. This can prevent an attacker from attempting session hijacking through Man-in-the-Middle (MitM) attacks.

~Currently we [allow the use of auth without TLS](https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/server/service.go#L343-L375), but I'm wondering if we should remove that case (or disallow it explicitly).~
edit: This is a common setup. We now have a separate config to control whether cookies have the `Secure` header set. This is supposed to be used only for testing as it potentially exposes users who serve flyteconsole with TLS disabled and auth enabled to aforementioned attacks.

### DIFF
--- a/flyteadmin/auth/cookie.go
+++ b/flyteadmin/auth/cookie.go
@@ -70,7 +70,7 @@ func NewSecureCookie(cookieName, value string, hashKey, blockKey []byte, domain 
 		Domain:   domain,
 		SameSite: sameSiteMode,
 		HttpOnly: true,
-		Secure:   config.GetConfig().Security.Secure,
+		Secure:   !config.GetConfig().Security.InsecureCookieHeader,
 	}, nil
 }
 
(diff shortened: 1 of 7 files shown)


Sovietaced:
> Currently we [allow the use of auth without TLS](https://github.com/flyteorg/flyte/blob/master/flyteadmin/pkg/server/service.go#L343-L375), but I'm wondering if we should remove that case (or disallow it explicitly).

I feel like it’s pretty normal to run plaintext from an ingress to an origin on the same node. At least that’s what we do.

wild-endeavor (approved):

PR #5915, opened 2024-10-24
draft of idea

0yukali0:
## Tracking issue
Related to https://github.com/flyteorg/flyte/issues/5575

### DIFF
--- a/rfc/system/5575-supporting-yunikorn-and-kueue.md
+++ b/rfc/system/5575-supporting-yunikorn-and-kueue.md
@@ -1,0 +1,127 @@
+# [Newbie] Supporting Yunikorn and Kueue
+
+**Authors:**
+
+- @yuteng
+
+## 1 Executive Summary
+
+Providing k8s resource management, gang scheduling and preemption for flyte apllications by 3rd software including Apache Yunikorn and Kueue.
+
+## 2 Motivation
+
+Flyte support multi-tenant and many k8s plugins.
+
+Kueue and Yunikorn support gang scheduling and preemption.
+Gang scheduling gurantees some avialable k8s crd services including spark, ray service with sufficient resource and preemption make sure high priority task execute immediately.
+
+Flyte doesn't maintain a resource management for multi-tenant which Yunikorn can solve it with hierarchy resource queues.
+
+## 3 Proposed Implementation
+
+```yaml
+queueconfig:
+  general:
+  - org: org1
+    users: "*" 
+    priorityclass: priority-default
+    acl:
+      - jobs: "ray,dask"
+        users: "user1,user2"
+        priorityclass: priority-ray
+        gangscheduling: true
+      - jobs: "spark"
+        users: "user1,user3"
+        priorityclass: priority-spark
+        gangscheduling: true
+    namespace:
+      default: "ns1"
+      namespaces: "ns1|ns2"
+  - org: org2
+    users: "user4, user5"
+    acl:
+      - jobs: "*"
+        users: "*"
+    namespace:
+      namespaces: "*"
+```
+
+// TODO how this yaml maps yunikorn configuration 
+// TODO how this yaml maps kueue CRD
+
+A SchedulerConfigManager maintains config from mentioned yaml.
+It patch labels or annotations of k8s resources after they pass rules in config.
+
+```go
+func (e *PluginManager) launchResource(ctx context.Context, tCtx pluginsCore.TaskExecutionContext) (pluginsCore.Transition, error) {
+  o, err := e.plugin.BuildResource(ctx, k8sTaskCtx)
+	if err != nil {
+		return pluginsCore.UnknownTransition, err
+	}
+  if err := e.SchedulerConfigManager.Label(0); err != nil {
+    return pluginsCore.UnknownTransition, err
+  }
+}
+```
+When batchscheduler in flyte is yunikorn, some examples are like following.
+For exmaple, this appoarch submit a Rayjob owned by user1 in org1 to "root.org1.ray".
+A spark application in ns1 submitted by user4 in org1 is in "root.org1.ns1".
+In the other hand, results of these examples are "org1-ray" and "org1-ns1" when adopting Kueue.
+
+## 4 Metrics & Dashboards
+
+1. Yunikorn scheduler add applications to a specific queue based on thier user info, queue name for any application type.
+2. Yunikorn and Kueue provides gang scheduling based annotations For Ray and spark.
+3. Preemption behavior meets user-defined configuration in yunikorn.
+
+## 5 Drawbacks
+
+This appoarch doens't provide a way to keep consistent between the accuate resource quato of each each group and  configuration in scheduler.
+
+## 6 Alternatives
+
+## 7 Potential Impact and Dependencies
+
+## 8 Unresolved questions
+
+## 9 Conclusion
+
+Yunikorn and Kueue support gang scheduling to run all neccesary pods at same time when required resource are available.
+Yunikorn provide preemption which calculate the priority of application based on its priority class an priority score of the queue where it submitted in order to trigger high-prioirty or emergency application immediately. 
+Yunikorn hierachy queue includes grarateed resources setting and acls.
+
+## 10 RFC Process Guide, remove this section when done
+
+*By writing an RFC, you're giving insight to your team on the direction you're taking. There may not be a right or better decision in many cases, but we will likely learn from it. By authoring, you're making a decision on where you want us to go and are looking for feedback on this direction from your team members, but ultimately the decision is yours.*
+
+This document is a:
+
+- thinking exercise, prototype with words.
+- historical record, its value may decrease over time.
+- way to broadcast information.
+- mechanism to build trust.
+- tool to empower.
+- communication channel.
+
+This document is not:
+
+- a request for permission.
+- the most up to date representation of any process or system
+
+**Checklist:**
+
+- [ ]  Copy template
+- [ ]  Draft RFC (think of it as a wireframe)
+- [ ]  Share as WIP with folks you trust to gut-check
+- [ ]  Send pull request when comfortable
+- [ ]  Label accordingly
+- [ ]  Assign reviewers
+- [ ]  Merge PR
+
+**Recommendations**
+
+- Tag RFC title with [WIP] if you're still ironing out details.
+- Tag RFC title with [Newbie] if you're trying out something experimental or you're not entirely convinced of what you're proposing.
+- Tag RFC title with [RR] if you'd like to schedule a review request to discuss the RFC.
+- If there are areas that you're not convinced on, tag people who you consider may know about this and ask for their input.
+- If you have doubts, ask on [#feature-discussions](https://slack.com/app_redirect?channel=CPQ3ZFQ84&team=TN89P6GGK) for help moving something forward.


kumare3:
My preference is to use kueue or anyone that has min changes to codebase and does not need too many modifications to the pods

rfc/system/5575-supporting-yunikorn-and-kueue.md
```diff
@@ -0,0 +1,131 @@
+# [Newbie] Supporting Yunikorn and Kueue
+
+**Authors:**
+
+- @yuteng
+
+## 1 Executive Summary
+
+Providing kubernetes (k8s) resource management, gang scheduling and preemption for flyte applications by third-party software, including Apache Yunikorn and Kueue.
```
fg91, line 9: Could you please explain what preemption means here compared to what preemption means in the context of spot instances on e.g. AWS or GCP?

rfc/system/5575-supporting-yunikorn-and-kueue.md
```diff
@@ -0,0 +1,131 @@
...
+
+```yaml
+queueconfig:
+  scheduler: yunikorn
+  jobs:
+    - type: "ray"
+      gangscheduling: "placeholderTimeoutInSeconds=60 gangSchedulingStyle=hard"
+    - type: "spark"
+      gangscheduling: "placeholderTimeoutInSeconds=30 gangSchedulingStyle=hard"
+```
```
fg91, line 30: Is this list complete or an example? I.e. will this also work for plugins like kubeflow pytorch, tf, mpi or dask, ...?

rfc/system/5575-supporting-yunikorn-and-kueue.md
```diff
@@ -0,0 +1,131 @@
...
+
+Mentioned configuration indicates what queues exist for an org.
```
fg91, line 32: Could you please explain what an org is in this context? It's not the same as [this](https://github.com/flyteorg/flyte/blob/ab67b4e5f95bae98814bb7b1ba0e3d5a3893544d/flyteidl/protos/flyteidl/admin/common.proto#L110) `org` right?

rfc/system/5575-supporting-yunikorn-and-kueue.md
```diff
@@ -0,0 +1,131 @@
...
+Hierarchical queues will be structured as follows.
+root.org1.ray、root.org1.spark and root.org1.default".
+
+ResourceFlavor allocates resource based on labels which indicates that category-based resource allocation by organization label is available.
```
fg91, line 36: Could you please explain how the the reseource flavor will be determined? Is there a way to derive this from the task decorator args `@task(resources=..., accelerator=...)?

It would be really nice if tasks that need e.g. an A100 GPU were automatically not in the same queue as tasks that need 2 x T4 GPUs. We're using kubeflow pytorch jobs with [scheduler plugins' gang scheduling](https://www.kubeflow.org/docs/components/training/user-guides/job-scheduling/#scheduler-plugins-with-coscheduling) and have observed jobs being starved that the cluster would have had resources for because other jobs couldn't be scheduled but had a higher priority.

rfc/system/5575-supporting-yunikorn-and-kueue.md
```diff
@@ -0,0 +1,131 @@
...
+Thus, a clusterQueue including multiple resources represents the total acessaible resource for an organization.  
+| clusterQueue | localQueue |
+| --- | --- |
+| Org | ray、spark、default |
+A tenant can submit organization-specific tasks to queues such as org.ray, org.spark and org.default to track which job types are submittable. 
+
+
+A SchedulerConfigManager maintains config from mentioned yaml.
```
fg91, line 44: `SchedulerConfigManager` would be a go struct or are you suggesting a new backend service?

rfc/system/5575-supporting-yunikorn-and-kueue.md
```diff
@@ -0,0 +1,131 @@
...
+It patches labels or annotations on k8s resources after they pass rules specified in the configuration.
+
+```go
```
fg91, line 47: Why not have a single interface and two implementations of the same interface for yunikorn and kueue?

rfc/system/5575-supporting-yunikorn-and-kueue.md
```diff
@@ -0,0 +1,131 @@
...
+  // 2.QueueName
+  // 3.ApplicationID
+}
+
+func PatchPodSpec(target *v1.PodSpec, labels map[string]string) error {
+  // Get Metaobject from target
+  // Add label is the specific label doesn't exist
+}
+```
+
```
fg91, line 74: Do any additional k8s resources have to be created for the queues or does a queue exist as soon as a pod has an annotation with a new queue name?

rfc/system/5575-supporting-yunikorn-and-kueue.md
```diff
@@ -0,0 +1,131 @@
...
+
+Creat a scheduler plugin according to the queueconfig.scheduler.
+Its basic responsibility validate whether submitted application is accepted. 
+When a Yunikorn scheduler plugin created, it will create applicationID and queue name.
+in the other hand, a Kueue scheduler plugin constructs labels including localQueueName, preemption.
+
+```go
+func (e *PluginManager) launchResource(ctx context.Context, tCtx pluginsCore.TaskExecutionContext) (pluginsCore.Transition, error) {
```
fg91, line 82: Maybe `addObjectMetadata` which is called by `launchResource` would be a better place to inject the required metadata.

rfc/system/5575-supporting-yunikorn-and-kueue.md
fg91: I would prefer in `func (e *PluginManager) launchResource(` to not call queue or yunikorn specific code, see snippet below, but just a general interface whose implementation depends on the propeller config.

rfc/system/5575-supporting-yunikorn-and-kueue.md
fg91: I don't see this in any of the code snippets below.

rfc/system/5575-supporting-yunikorn-and-kueue.md
```diff
@@ -0,0 +1,131 @@
...
+  o, err := e.plugin.BuildResource(ctx, k8sTaskCtx)
+	if err != nil {
+		return pluginsCore.UnknownTransition, err
+	}
+  if o, err = e.SchedulerPlugin.MutateResourceForKueue(o); err == nil {
```
fg91, line 87: We should only mutate the resource if the plugin manager manages a plugin which the user configured a queue for, right? How will this matching be done? Just comparing this name string

```yaml
queueconfig:
  scheduler: yunikorn
  jobs:
    - type: "ray"
```

to the name of the plugin?

rfc/system/5575-supporting-yunikorn-and-kueue.md
```diff
@@ -0,0 +1,131 @@
...
+Its basic responsibility validate whether submitted application is accepted. 
+When a Yunikorn scheduler plugin created, it will create applicationID and queue name.
+in the other hand, a Kueue scheduler plugin constructs labels including localQueueName, preemption.
+
+```go
+func (e *PluginManager) launchResource(ctx context.Context, tCtx pluginsCore.TaskExecutionContext) (pluginsCore.Transition, error) {
+  o, err := e.plugin.BuildResource(ctx, k8sTaskCtx)
+	if err != nil {
+		return pluginsCore.UnknownTransition, err
+	}
```
fg91, line 86: Would be nice to not have `Kueue` specific code here but a general interface, see [this](https://github.com/flyteorg/flyte/pull/5915/files#r1829818809) comment.

rfc/system/5575-supporting-yunikorn-and-kueue.md
```diff
@@ -0,0 +1,131 @@
...
+  if o, err = e.SchedulerPlugin.MutateResourceForKueue(o); err == nil {
+      return pluginsCore.UnknownTransition, err
+    }
+  } else {
+     return pluginsCore.UnknownTransition, err
+  }
+}
+```
+When batchscheduler in flyte is yunikorn, some examples are like following.
+For example, this appoarch submits a Ray job owned by user1 in org1 to "root.org1.ray".
```
fg91, line 96: Where does flytepropeller know the user from?

rfc/system/5575-supporting-yunikorn-and-kueue.md
```diff
@@ -0,0 +1,131 @@
...
+Flyte support Spark, Ray and Kubeflow CRDs including Pytorch and TFjobs.
+The Spark and Ray operators have supported Yunikorn gang scheduling since task group calculation were implemented in these operators.
+Taskgroup calculation implementation in pods aspect in flyte or kubeflow is required for supporting kubeflow CRDs.
+In the other hand, Kueue currently doesn't support Spark CRD.
+| Operator | Yunikorn | Kueue |
+| --- | --- | --- |
+| Spark | v | x |
+| Ray | v | v |
+| Kubeflow | x | v |
+
```
fg91, line 123: From what I understand, one only needs to add labels/annotations on the worker pods. Can't we do this purely from flyte by modifying the pod template spec of the respective CRD? What do the operators have to do in addition to that?

rfc/system/5575-supporting-yunikorn-and-kueue.md
```diff
@@ -0,0 +1,131 @@
...
+## 8 Unresolved questions
+
+
+## 9 Conclusion
+
+Yunikorn and Kueue support gang scheduling allowing all necassary pods to run sumultaneously when required resource are available.
+Yunikorn provides preemption calculating the priority of applications based on thier priority class and priority score of the queue where they are submitted, in order to trigger high-prioirty or emergency application immediately. 
+Yunikorn's hierarchical queue includes grarateed resources settings and ACLs.
```
fg91, line 131: Nit: Could you please run the doc through a spelling checker? 🙇

rfc/system/5575-supporting-yunikorn-and-kueue.md
0yukali0: This is an example.
Admin can set default configuration about gang scheduling for CRD in flyte k8s plugins.

rfc/system/5575-supporting-yunikorn-and-kueue.md
0yukali0: Yes, applying [Kueue CRD](https://kueue.sigs.k8s.io/docs/concepts/cluster_queue/)s describe the quota a queue when adopting Kueue.

In the other hand, queues are configured by setting [Yunikorn configuration](https://yunikorn.apache.org/docs/user_guide/queue_config).

rfc/system/5575-supporting-yunikorn-and-kueue.md
0yukali0: No, need to create Kueue CRDs first.
A [cluster queue](https://kueue.sigs.k8s.io/docs/concepts/cluster_queue/) defines resource quota which property is defined by [resource flavors](https://kueue.sigs.k8s.io/docs/concepts/resource_flavor/).
I think creating resource flavors to categorizing resources under a cluster queue is available solution.
```
apiVersion: kueue.x-k8s.io/v1beta1
kind: ResourceFlavor
metadata:
  name: "spot-t4"
spec:
  nodeLabels:
    cloud.google.com/gke-accelerator: nvidia-tesla-t4
  nodeTaints:
  - effect: NoSchedule
    key: cloud.google.com/gke-accelerator: nvidia-tesla-t4
    value: "true"
  tolerations:
  - key: "spot-taint"
    operator: "Exists"
    effect: "NoSchedule"
    
   apiVersion: kueue.x-k8s.io/v1beta1
kind: ClusterQueue
metadata:
  name: "cluster-queue"
spec:
  namespaceSelector: {} # match all.
  resourceGroups:
  - coveredResources: ["cpu", "memory", "nvidia.com/gpu"]
    flavors:
    - name: "spot-t4"
      resources:
      - name: "cpu"
        nominalQuota: 9
      - name: "memory"
        nominalQuota: 36Gi
      - name: "nvidia.com/gpuu"
        nominalQuota: 50
    - name: "spot-a100"
      resources:
      - name: "cpu"
        nominalQuota: 18
      - name: "memory"
        nominalQuota: 72Gi
      - name: "nvidia.com/gpu"
        nominalQuota: 100
```
In the other hand,  kueue preemption requires Kueue [WorkloadPriorityClass ](https://kueue.sigs.k8s.io/docs/concepts/workload_priority_class/)and patching job with label. 
The plugin received the preemption label and then it should patch it to pods belonging same job

rfc/system/5575-supporting-yunikorn-and-kueue.md
0yukali0: Agree, i updated the document..

rfc/system/5575-supporting-yunikorn-and-kueue.md
0yukali0: Yes, [current progress](https://github.com/flyteorg/flyte/pull/5594) fetch the pod templates from CRDs and patch label on them.
If operators implement the mechanism to patch group label for their CRD to support gang scheduling in the future, we can start to remove the code for generating group labels to reduce the maintaining overhead.

davidmirror-ops:
2025-01-15 Contributor's meetup notes: this was tested at Union and found it to be hard to support. Consider closing

PR #5931, opened 2024-10-28
blob-storage-usage doc update

### LINKED ISSUE #4881: [Docs] Provide documentation on where blob storage is accessed and which keys are being used
### Description

There are several different places where various Flyte services or tools require blob storage access, and for folks new to the project, it would be nice to have a few notes on the internals of how API calls happen. The initial question I raised in Slack at https://flyte-org.slack.com/archives/CP2HDHKE1/p1707508706335559 was 

```
Is there a good graph of the sequence of API calls between `pyflyte run --remote` / flyteadmin and blob storage endpoints? I'm mostly trying to have a good picture around when `CreateUploadLocation` gets called /used, etc
```

Some of this is documented in https://docs.flyte.org/en/latest/concepts/data_management.html, but it would be nice to extend that to include:

- `pyflyte run --remote` calls FlyteAdmin, calls blob storage to get a signed url and returns it to the client so that it can upload
- when inputs / outputs are downloaded / uploaded and using what creds for native Flyte tasks
- when inputs / outputs are downloaded / uploaded for raw container tasks (i.e. by copilot)
- default config vs. other config mechanisms

While Flyte has a good default approach of using IRSA in AWS, that might not always be the best fit -- so it's helpful to know some of the specifics around blob storage access patterns so that the right solution can be employed.

- [X] Yes

- [X] Yes

Ritesh-Udgata:
# Building the Flyte docs locally

## Prerequisites

* `conda` (We recommend Miniconda installed with an [official installer](https://docs.conda.io/projects/miniconda/en/latest/index.html#latest-miniconda-installer-links))

* [`conda-lock`](https://github.com/conda/conda-lock)

## Set up the build environment

In the `flyteorg/flyte` root directory do:

```bash
$ conda-lock install --name monodocs-env monodocs-environment.lock.yaml
$ conda activate monodocs-env
$ pip install ./flyteidl
```

This creates a new environment called `monodocs-env` with all the dependencies needed to build the docs. You can choose a different environment name if you like.

## Building the docs

In the `flyteorg/flyte` root directory make sure you have activated the `monodocs-env` (or whatever you called it) environment and do:

```bash
$ make docs
```

The resulting `html` files will be in `docs/_build/html`.

Closes #4881

### DIFF
--- a/docs/blob-storage.md
+++ b/docs/blob-storage.md
@@ -1,0 +1,19 @@
+Flyte’s interaction with blob storage occurs through several key components and API calls, especially when executing workflows remotely. Here's an overview to help you understand how `pyflyte run --remote` interacts with blob storage and where keys or credentials are involved.
+
+1. Blob Storage Usage in `pyflyte run --remote`:
+   - When you execute a workflow remotely with `pyflyte run --remote`, the client communicates with FlyteAdmin to get a signed URL for uploading the workflow's inputs to blob storage. This URL is then used by the client to upload data directly to storage.
+
+2. CreateUploadLocation API Call:
+   - FlyteAdmin handles the `CreateUploadLocation` call, generating pre-signed URLs that allow users or services to upload inputs to storage without exposing sensitive keys directly. This ensures controlled and secure access to storage resources.
+
+3. Input/Output Handling in Native and Raw Container Tasks:
+   - For native Flyte tasks, inputs and outputs are uploaded/downloaded through FlyteAdmin using appropriate credentials configured on the platform.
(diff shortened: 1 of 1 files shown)


davidmirror-ops:
@Ritesh-Udgata thank you. Can you clarify if this is connected to an issue?

ppiegaze (approved):

Ritesh-Udgata:
I have added sign off along with gpg signature to pass the DCO check. Am i eligible for swags ? If so could you reopen the form ? I am asking because this PR was approved two weeks ago approximately.

Issue #5666, opened 2024-08-16
[Core feature] Decouple submitterPod resources from ray task pod_template


jpoler:
### Motivation: Why do you think this is important?

Currently the ray plugin uses the pod_template provided to the task as the basis for all pod specs:
- The RayCluster head
- RayCluster workers
- The ray job submit kubernetes Job

This is a pain point when the RayCluster head and workers are intended to be scheduled on GPU nodes. I do not want to waste an entire GPU node for the submitter.

### Goal: What should the final outcome look like, ideally?

It is not possible to configure RayCluster pod templates and the submitter pod template separately. If it were, it would be possible to schedule the submitter with appropriately minimal resource requests and leave out other configurations that have nothing to do with the submitter pod (for example in my use case only the ray head/worker need the GPU, shared memory volume mount, service account, etc.)

I found #4170, which looks like it was trying to address this issue, but it hasn't seen any progress since October 2023. At a high level the approach it takes makes sense to me, where the pod_template provided to the task configures the resources for the submitter job, and then the ray head/worker have new config fields to configure their resources explicitly. In my opinion this change looks like it is headed in the right direction, but would be improved with a slight adaptation where it allows for the user to provide the entire pod template alongside resources. Otherwise it won't be possible to do things on the ray head/worker like configure volume mounts and env vars, etc.

### Describe alternatives you've considered 

I don't see an alternative to adding separate config parameters for separate pod specs. It doesn't seem like a good idea to hard-code the submitter pod spec for minimal resource requests (e.g. just a small request/limit for CPU and memory), because there very well could be a use case where someone wants a GPU for the submitter. It wouldn't make a lot of sense to preclude that use-case IMO.

I do see this PR that adds a `Resource` config to 

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

pingsutw:
How about adding a submitter spec in the ray cluster [proto](https://github.com/flyteorg/flyte/blob/e9413c0ccb6672a0129e7e357bbf8a92ae0a3411/flyteidl/protos/flyteidl/plugins/ray.proto#L24)? and add [resource](https://github.com/flyteorg/flyte/blob/e9413c0ccb6672a0129e7e357bbf8a92ae0a3411/flyteidl/protos/flyteidl/plugins/kubeflow/pytorch.proto#L46) to the head, worker, and submitter spec.

kumare3:
[...]

PR #5933, opened 2024-10-29
Decouple ray submitter, worker, and head resources

Sovietaced:
## Tracking issue
Related to https://github.com/flyteorg/flyte/issues/5666

If you try to use the Ray plugin for GPU related workloads the submitter pod ends up consuming as many GPUs as the heads/worker nodes. 

This pull request introduces resources fields in the worker and head node specs. It then uses the task resources for the submitter pod. 

## How was this patch tested?
Unit tests

## Related PRs

flytekit PR will come

## Docs link

TBA


EngHabu (dismissed):
Thank you for this!

flyteidl/protos/flyteidl/plugins/ray.proto
```diff
@@ -35,6 +37,8 @@ message HeadGroupSpec {
   // Optional. RayStartParams are the params of the start command: address, object-store-memory.
   // Refer to https://docs.ray.io/en/latest/ray-core/package-ref.html#ray-start
   map<string, string> ray_start_params = 1;
+  // Resource specification for ray head pod
+  core.Resources resources = 2;
```
pingsutw, line 41: I'm considering whether we should use [CommonReplicaSpec](https://github.com/flyteorg/flyte/blob/13b3d82c94fb61047a49aa2093a2b17fb6a7c792/flyteidl/protos/flyteidl/plugins/common.proto#L15) here since someone might also want to set different image for the head node.

Sovietaced: I was considering similar ideas, having some sort of message to group these kinds of configurations but just figured the abstraction will come in time when use cases come along and the plugin evolves.  The `CommonReplicaSpec` could work but also includes fields like `replicas` that don't make sense in this context since there is only supposed to be a single head pod.

Sovietaced: I think if we wanted true flexibility we'd have some sort of `PodTemplateSpec` message since that would align with the kuberay CRDs. We'd just have to figure out how to merge those configurations with what Flyte is setting in the plugin. I'm willing to go down that route though..

pingsutw: cc @EngHabu @hamersaw any thoughts on this

Sovietaced: We could start this off as 

```
message RayPodTemplateSpec {
  RayPodSpec pod_spec = 1;
}

message RayPodSpec {
  repeated RayContainer containers = 1;
}

message RayContainer {
  core.Resources resources = 1;
}
```

and add functionality as necessary.

EngHabu: Ray supports a full PodTemplateSpec, we should not piece meal this. Just expose the full PodTemplate in the idl (maybe just reference [this](https://github.com/unionai/flyte/blob/f448a0358d8706a09b65b96543134f629327d755/flyteidl/protos/flyteidl/core/tasks.proto#L165) and let the SDK fill it in as needed. It's fine to expose a nicer interface in the SDK that takes the existing Resources object... etc. and populate the template... IMHO

Sovietaced: @EngHabu Looks like you linked to a private repo but if the line numbers are the same as the open source stuff I think it's the k8s pod spec which makes sense.. I'll update the PR

flyteplugins/go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -567,7 +589,7 @@ func getEventInfoForRayJob(logConfig logs.LogConfig, pluginContext k8s.PluginCon
 	// RayJob CRD does not include the name of the worker or head pod for now
 	logOutput, err := logPlugin.GetTaskLogs(input)
 	if err != nil {
-		return nil, fmt.Errorf("failed to generate task logs. Error: %w", err)
+		return nil, fmt.Errorf("failed to genermergeCustomPodSpecor: %w", err)
```
eapolinario, line 592: nit: fix the typo.

flyteidl/protos/flyteidl/plugins/ray.proto
```diff
@@ -37,8 +37,8 @@ message HeadGroupSpec {
   // Optional. RayStartParams are the params of the start command: address, object-store-memory.
   // Refer to https://docs.ray.io/en/latest/ray-core/package-ref.html#ray-start
   map<string, string> ray_start_params = 1;
-  // Resource specification for ray head pod
-  core.Resources resources = 2;
+  // Pod Spec for the ray head pod
+  core.K8sPod k8s_pod = 2;
```
eapolinario, line 41: let's not reuse field numbers. Mark the previous field as deprecated and leave a comment so that IDEs show that explicitly (e.g. https://github.com/flyteorg/flyte/blob/master/flyteidl/protos/flyteidl/admin/task_execution.proto#L79-L81)

Sovietaced: These are net new fields, if you mean it reused between the two commits in the pull request I'll just squash my changes into a single commit.

Sovietaced: Fixed

eapolinario (approved):
Awesome. Thank you!

PR #5813, opened 2024-10-05
Minor auto_refresh cleanup

Sovietaced:
## Why are the changes needed?

This pull request is some follow up work from https://github.com/flyteorg/flyte/pull/5767 to clean up auto refresh so that it has a constructor that returns a pointer to the implementation instead of an interface. This gives the user of the constructor more flexibility (ie. for intra-package testing) while still adhering to the auto_refresh interface. 

## What changes were proposed in this pull request?

The default auto_refresh cache implementation was updated to the `InMemoryAutoRefresh` and moved into its own file. 
Some additional fields were added to improve assertions given the nature of the asynchronous enqueue/syncing that occurs with the auto refresh cache. 

Lastly, the new constructor for the in memory auto refresh cache was updated to use the options pattern so that test-only  values or non-default options only have to be added in those situations. 

The preexisting helper functions to create refresh caches have been left in place for backwards compatibility. 

## Check all the applicable boxes 

- [x] I updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] All commits are signed-off. 
 <div id='description'>
<h3>Summary by Bito</h3>
Refactoring of auto-refresh cache implementation with core functionality moved to InMemoryAutoRefresh type. Introduces flexible constructor pattern using options while maintaining backward compatibility. Adds new fields for improved testing of async operations. Changes are structural without modifying core functionality.
<br>
<br>
<b>Unit tests added</b>: False
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 3
</div>

PR #5940, opened 2024-10-30
Add option to disable synchronization on creation for auto refresh cache

Sovietaced:
We have an interesting usage pattern of the auto refresh cache. 
- New items can be added to the cache up to 1k QPS. 
- Items are synchronized in batches of 500 items.
- Items are synchronized once every second. 

When items are synchronized in single item batches upon creation it disrupts our batch synchronization strategy and puts a lot of load on the server where task state is being read from. You could tune the rate limiter but it will also affect both single item batch and regular batch synchronization. So the solution we found was to add this option which just doesn't try to add a single item batch when items are created in the cache. 

Adds an option to the auto refresh cache `SyncOnCreate` which defaults to true for backwards compatibility. 

## How was this patch tested?

We run this in production but I also added unit tests. 

## Related PRs

Built on top of https://github.com/flyteorg/flyte/pull/5813
 
 <div id='description'>
<h3>Summary by Bito</h3>
Introduces 'SyncOnCreate' configuration option for auto-refresh cache system to control immediate synchronization behavior. The feature allows disabling immediate item sync on creation, defaulting to true for backward compatibility. This optimization targets high-load scenarios by enabling batch synchronization instead of individual syncs. Implementation includes configuration changes and corresponding test coverage.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>


pvditt (approved):
Thanks for adding this. I think it's fine to add this without introducing an additional config to toggle syncing on create.

PR #5942, opened 2024-10-30
Binary IDL Attribute Access for Map Task

Future-Outlier:
### Tracking Issue
[Flyte Issue #5318](https://github.com/flyteorg/flyte/issues/5318)

### Why are the changes needed?
This update ensures that the following example workflow functions as expected:

```python
@dataclass
class MyDataClass:
    my_ints: list[int]

@flytekit.task
def print_int(my_int: int):
    print(f"my_int: {my_int}")

@flytekit.workflow
def my_workflow(my_data_class: MyDataClass = MyDataClass(my_ints=[1, 2, 5, 10])):
    flytekit.map_task(print_int)(my_int=my_data_class.my_ints)
```

### What changes are proposed in this pull request?
The primary change is to ensure that a `collection literal` is created when the resolved value is a `list`, allowing map tasks to function properly with collections.

#### Updated Workflow Lifecycle
**Non-Map Task (List handling)**:
1. **Before**:  
   `Python value -> binary IDL -> resolved value -> binary IDL -> list transformer -> Python value`
2. **After**:  
   `Python value -> binary IDL -> resolved value -> binary IDL in literal collection -> list transformer -> expected Python type transformer -> Python value`

**Map Task**:
1. **Before**:  
   The array node handler didn’t receive a collection literal[1], so it couldn’t determine the length of the literal value.
2. **After**:  
   `Python value -> binary IDL -> resolved value -> binary IDL in literal collection -> ArrayNodeMapTask (Flytekit) -> expected Python type transformer -> Python value`

#### References
- [The array node handler didn’t receive a collection literal Array Node Handler)](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/nodes/array/handler.go#L215-L225)
- [ArrayNodeMapTask (Flytekit)](https://github.com/flyteorg/flytekit/blob/master/flytekit/core/array_node_map_task.py#L231-L256)

### Testing
This patch has been verified using unit tests and remote execution in Flytekit.

### Screenshots

| Image 1 | Image 2 |
|---|---|

### Checklist
- [x] Documentation has been updated where necessary.
- [x] All new and existing tests pass.

### DIFF
--- a/flytepropeller/pkg/controller/nodes/attr_path_resolver.go
+++ b/flytepropeller/pkg/controller/nodes/attr_path_resolver.go
@@ -46,8 +46,7 @@ func resolveAttrPathInPromise(ctx context.Context, datastore *storage.DataStore,
 			}
 			currVal = currVal.GetCollection().GetLiterals()[attr.GetIntValue()]
 			index++
-		// scalar is always the leaf, so we can break here
-		case *core.Literal_Scalar:
+		default:
 			break
 		}
 	}
@@ -107,9 +106,7 @@ func resolveAttrPathInPbStruct(nodeID string, st *structpb.Struct, bindAttrPath 
 }
 
 // resolveAttrPathInBinary resolves the binary idl object (e.g. dataclass, pydantic basemodel) with attribute path
-func resolveAttrPathInBinary(nodeID string, binaryIDL *core.Binary, bindAttrPath []*core.PromiseAttribute) (*core.
-	Literal,
-	error) {
+func resolveAttrPathInBinary(nodeID string, binaryIDL *core.Binary, bindAttrPath []*core.PromiseAttribute) (*core.Literal, error) {
 
 	binaryBytes := binaryIDL.GetValue()
 	serializationFormat := binaryIDL.GetTag()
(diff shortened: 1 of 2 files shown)


flytepropeller/pkg/controller/nodes/attr_path_resolver.go
```diff
@@ -46,8 +46,7 @@ func resolveAttrPathInPromise(ctx context.Context, datastore *storage.DataStore,
 			}
 			currVal = currVal.GetCollection().GetLiterals()[attr.GetIntValue()]
 			index++
-		// scalar is always the leaf, so we can break here
-		case *core.Literal_Scalar:
+		default:
 			break
```
Future-Outlier, line 50: this is the right logic, we should stop the iteration if this is not a map, collection, or offloaded literal.

Future-Outlier:
In this PR ([flyteorg/flyte#5942](https://github.com/flyteorg/flyte/pull/5942)), I’m considering whether the propeller might crash if the resolved value is a collection containing over 1000 dataclass instances. The current approach would serialize it as a binary collection and pass it to the Flytekit array node task:

**Current Approach:**

```
[1000+ dataclass instances] -> serialized as binary collection -> Flytekit array node task
```

[Reference Code](https://github.com/flyteorg/flyte/pull/5942/files#diff-ee7f936e440a7e043b3bc7acb4ea255ba991dea8f3144d24ab276c3a292de018R168-R186)

---

**Alternative Approach:**

To reduce potential memory pressure, we could consider an alternative that serializes the collection directly as binary, then processes it as a Python value in Flytekit:

```
[1000+ dataclass instances] -> serialize directly as binary -> Flytekit Python value -> process using these steps
```

[Alternative Code](https://github.com/flyteorg/flytekit/pull/2876/files#diff-50b831faf1906e528057ba3242680e26a34d7665bb37b46658815bba835e52ccR254-R259)

This alternative may reduce the load on the propeller by avoiding the need to handle a large nested structure directly, potentially improving stability and performance in cases with large collections.
but we will need to deserialize lots of dataclass (waste resource)

**My Thoughts:**
I prefer the first one, which is this PR, since Byron's attribute access did similar things before
(the recursion function can be seen as a for loop)
https://github.com/flyteorg/flyte/blob/856e606419a3756db8c4d8c0968bee80fbb445b4/flytepropeller/pkg/controller/nodes/attr_path_resolver.go#L212-L253

wild-endeavor (approved):
i think this is correct. I'm a little worried about unions, but i think we should revisit that after the other issues are sorted if there is an issue there.

PR #5944, opened 2024-10-31
Fix: Make appProtocols optional in flyteadmin and flyteconsole services in helm chart

fg91:

https://github.com/flyteorg/flyte/pull/5240, among other things, set `appProtocol` values of `TCP` for the ports of the flyteadmin and flyteconsole services in the flyte-core helm chart.

[This](https://github.com/flyteorg/flyte/pull/5240#issuecomment-2400342871) comment by @ddl-ebrown explains why this was necessary in their deployment which uses an ingress in combination with the istio service mesh:

> [...] the ingress may be configured in a way that
it sends TLS traffic to internal Flyte services. Istio will use port
names to determine traffic - and may therefore assume the appProtocol
of http, even though traffic from ingress -> flyteadmin is actually
https. This misconfiguration prevents any traffic from flowing
through the ingress to the service.
Flyteadmin http and grcp ports are accessible using http and
grpc values for appProtocol respectively within the cluster (i.e.
from a user workspace), but as soon as traffic travels between the
ingress and the service those settings will not work. The most
"compatible" setting is tcp which works for any network stream.

---

In my deployment (which also uses istio but not with an nginx ingress but instead with a GCE ingress and an istio ingress gateway, see [here](https://github.com/flyteorg/flytekit/tree/master/plugins/flytekit-identity-aware-proxy#configuring-your-flyte-deployment-to-use-iap) for deplyoment guide), this particular change is problematic:

```yaml
# flyteadmin service
    - name: grpc
      port: 81
      protocol: TCP
      # intentionally set to TCP instead of grpc
      appProtocol: TCP
      targetPort: 8089
```

https://github.com/flyteorg/flyte/pull/5240 intentionally sets the `appProtocol` for the `grpc` port to `TCP` over `grpc`. In my deployment this prevents traffic flowing from the istio ingress gateway to the flyteadmin pod.

This worked before the `appProtocol` was set (assumably because the protocol was derived from the name of the port) and it does also work if I change the protocol to `grpc`.

I currently cannot upgrade my deployment without manually patching the service after deployment. Because of this I'd like to make the `appProtocol` values in the helm chart optional.

Because the app protocols used to not be set since they are not required for the default nginx ingress/no service mesh deployment the helm chart provides, I propose to disable the app protocol fields by default.

This would mean that you, @ddl-ebrown, need to set the flags to true when upgrading to a helm chart version that contains this change.

If somebody has the opinion that the default should be true, I'm absolutely happy to discuss this.

## How was this patch tested?
The app protocol values used to not be set and are not needed for the default nginx ingress deployment offered by the helm chart.

### DIFF
--- a/charts/flyte-core/README.md
+++ b/charts/flyte-core/README.md
@@ -191,7 +191,7 @@ helm install gateway bitnami/contour -n flyte
 | flyteadmin.resources | object | `{"limits":{"cpu":"250m","ephemeral-storage":"100Mi","memory":"500Mi"},"requests":{"cpu":"10m","ephemeral-storage":"50Mi","memory":"50Mi"}}` | Default resources requests and limits for Flyteadmin deployment |
 | flyteadmin.secrets | object | `{}` |  |
 | flyteadmin.securityContext | object | `{"fsGroup":65534,"fsGroupChangePolicy":"Always","runAsNonRoot":true,"runAsUser":1001,"seLinuxOptions":{"type":"spc_t"}}` | Sets securityContext for flyteadmin pod(s). |
-| flyteadmin.service | object | `{"additionalPorts":[],"annotations":{"projectcontour.io/upstream-protocol.h2c":"grpc"},"loadBalancerSourceRanges":[],"type":"ClusterIP"}` | Service settings for Flyteadmin |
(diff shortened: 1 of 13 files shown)


fg91:
@ddl-ebrown could you please take a look at whether these changes work for you?

davidmirror-ops:
@fg91 Agree with making it configurable and disabled by default as long as the idea is to maintain a lean but customizable base chart. 
We'd need to communicate this as a breaking change for Istio users who rely on it.
Please look at the conflicts (maybe you'll have to rebase and make helm again) and I'll approve. Thank you

davidmirror-ops (approved):

PR #4726, opened 2024-01-14
Feat: Allow controlling in which task phases log links are shown

fg91:
## Why are the changes needed?

Currently, all task log links appear in the Flyte UI once the respective pod(s) are running and they are never removed.

In this PR I add config options to log links to allow showing certain log links during the queueing/initializing phase or to hide them once the task terminates.

This enables for instance the following user stories:

* As a platform engineer I want to show platform users log links already during the queuing phase of a task that help the user understand why a task pod might not be scheduleable. (Especially for task plugins relying on CRDs, the status updates shown in the Flyte UI don’t include information such as `“FailedTriggerScaleUp”`).
* As a platform engineer I want to hide certain log links once tasks complete as the log links are not valid anymore. Examples are Flyte-interactive VSCode links or links to the “live pod view" in the GKE "workloads" section.

## What changes were proposed in this pull request?

I propose to add two new options to task logs `flyteidl/protos/flyteidl/core/execution.proto`:

```
message TaskLog {
    …
    bool ShowWhilePending = 5;
    bool HideOnceFinished = 6;
}
```

This example …

```yaml
  logs:
    ...
    templates:
    - displayName: ShowPending
      showWhilePending: true
      templateUris:
    - displayName: ShowPendingHideFinished
      hideOnceFinished: true
      showWhilePending: true
      templateUris:
```

… results in the following UX:

image

## How was this patch tested?

I ran propeller and admin with the proposed changes. I ensured that the new behaviour works for all K8s plugins (pod, kubeflow, spark, dask, ray) as well as map tasks. For each plugin, screenshots are included in comments further below.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] All commits are signed-off.

## Related PRs

I added further tests in https://github.com/flyteorg/flyte/pull/5200 which ensure all plugins send new events to admin in the queuing phase if the “reason” changes. Currently this is implemented properly only for the pod plugin and had to be fixed for this PR.

PR #5945, opened 2024-10-31
Fix: avoid log spam for log links generated during the pod's pending phase

fg91:

In https://github.com/flyteorg/flyte/pull/4726 I introduced the option to configure the lifecycle of log links as is documented [here](https://docs.flyte.org/en/latest/user_guide/productionizing/configuring_logging_links_in_the_ui.html#configure-lifetime-of-logging-links). In particular, log links can be configured to be already shown before a task starts running or be configured to disappear after a task stops.

@Tom-Newton [noted](https://github.com/flyteorg/flyte/pull/4726/files#r1819110678) that their flytepropeller logs are spammed with error messages because the log link generation logic for pod tasks checks whether container statuses exist and if not logs this as an error. Prior to https://github.com/flyteorg/flyte/pull/4726 it made sense to always log this as an error because log links were only generated in the "Running" phase of the task when the pod should have container statuses. Now that we optionally generate log links already in the pending phase, it is not unexpected that the container statuses are empty.

Log the respective message with debug log level if the pod is in the pending phase. Keep logging with error log level in other pod phases.

### DIFF
--- a/flyteplugins/go/tasks/logs/logging_utils.go
+++ b/flyteplugins/go/tasks/logs/logging_utils.go
@@ -31,7 +31,14 @@ func GetLogsForContainerInPod(ctx context.Context, logPlugin tasklog.Plugin, tas
 
 	containerID := v1.ContainerStatus{}.ContainerID
 	if uint32(len(pod.Status.ContainerStatuses)) <= index {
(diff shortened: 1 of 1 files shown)


Tom-Newton (approved):
Thanks

PR #5949, opened 2024-10-31
Hide generated launch plans starting with .flytegen in the UI

troychiu:
Currently, the launch plan page will show generated launch plans, which make users hard to find their launch plan. We should hide them like we do for workflows. 
Screenshot 2024-11-05 at 4 03 13 PM

1. In ListNamedEntities, if no filter provided, filter by name starting with `.flytegen`. I did this instead of filtering by resource state since I would like to avoid a db migration.
2. Allow launchplan type entities being updated.

## How was this patch tested?
1. unit tests
2. integration tests with sandbox

### Screenshots
Screenshot 2024-11-05 at 4 04 12 PM

### DIFF
--- a/flyteadmin/pkg/common/filters.go
+++ b/flyteadmin/pkg/common/filters.go
@@ -22,6 +22,7 @@ type GormQueryExpr struct {
 // Complete set of filters available for database queries.
 const (
 	Contains FilterExpression = iota
+	NotContains
 	GreaterThan
 	GreaterThanOrEqual
 	LessThan
@@ -37,6 +38,8 @@ const (
 	joinArgsFormat          = "%s.%s"
 	containsQuery           = "%s LIKE ?"
 	containsArgs            = "%%%s%%"
+	notContainsQuery        = "%s NOT LIKE ?"
+	notContainsArgs         = "%%%s%%"
 	greaterThanQuery        = "%s > ?"
 	greaterThanOrEqualQuery = "%s >= ?"
 	lessThanQuery           = "%s < ?"
@@ -50,6 +53,7 @@ const (
 // Set of available filters which exclusively accept a single argument value.
 var singleValueFilters = map[FilterExpression]bool{
 	Contains:           true,
+	NotContains:        true,
 	GreaterThan:        true,
 	GreaterThanOrEqual: true,
 	LessThan:           true,
@@ -68,6 +72,7 @@ const EqualExpression = "eq"
 
 var filterNameMappings = map[string]FilterExpression{
 	"contains":      Contains,
+	"not_contains":  NotContains,
 	"gt":            GreaterThan,
 	"gte":           GreaterThanOrEqual,
 	"lt":            LessThan,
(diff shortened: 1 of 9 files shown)


flyteadmin/pkg/common/filters.go
```diff
@@ -37,6 +38,8 @@ const (
 	joinArgsFormat          = "%s.%s"
 	containsQuery           = "%s LIKE ?"
 	containsArgs            = "%%%s%%"
+	notContainsQuery        = "%s NOT LIKE ?"
```
katrogan, line 41: nit: not contains read a bit weird, why not 'NotLike' here? Then we could even support regex or wild carding with this too

katrogan (commented):
Lgtm but let's wait on FE to confirm expected usages

flyteadmin/pkg/manager/impl/validation/named_entity_validator.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/flyteorg/flyte/flyteidl/gen/pb-go/flyteidl/core"
 )
 
-var archivableResourceTypes = sets.NewInt32(int32(core.ResourceType_WORKFLOW), int32(core.ResourceType_TASK))
+var archivableResourceTypes = sets.NewInt32(int32(core.ResourceType_WORKFLOW), int32(core.ResourceType_TASK), int32(core.ResourceType_LAUNCH_PLAN))
```
katrogan, line 13: can you verify with this change if this resolves https://linear.app/unionai/issue/COR-2094/support-archiving-launchplans?

we'll need to possibly make uctl/flytectl changes and definitely UI ones but it would be good to update the ticket with this context

flyteadmin/pkg/common/filters.go
troychiu: good idea. Then we can filter by start with. Updated.

flyteadmin/pkg/manager/impl/validation/named_entity_validator.go
troychiu: drop a comment there. Thank you!

flyteadmin/pkg/manager/impl/named_entity_manager.go
```diff
@@ -75,12 +68,8 @@ func (m *NamedEntityManager) GetNamedEntity(ctx context.Context, request *admin.
 	return util.GetNamedEntity(ctx, m.db, request.ResourceType, request.Id)
 }
 
-func (m *NamedEntityManager) getQueryFilters(referenceEntity core.ResourceType, requestFilters string) ([]common.InlineFilter, error) {
+func (m *NamedEntityManager) getQueryFilters(requestFilters string) ([]common.InlineFilter, error) {
 	filters := make([]common.InlineFilter, 0)
-	if referenceEntity == core.ResourceType_WORKFLOW {
```
katrogan, line 80: nice, love cleaning up this code (it's also very confusing we always force a filter on workflow state even if you filter on state down below

katrogan (approved):
thank you! do you mind updating the fully supported set of filter functions documentation here too: https://github.com/flyteorg/flyte/blob/96c467e14691eb3e956d3b21c05fb724c240fe97/docs/user_guide/concepts/control_plane/admin.rst#adding-request-filters

flyteadmin/pkg/manager/impl/named_entity_manager.go
troychiu: Yes that's confusing!

pingsutw (approved):

Issue #5937, opened 2024-10-29
[BUG] github.com/flyteorg/flyte/flytectl/cmd/sandbox	unit test is flaky 


Sovietaced:
### Describe the bug

Its not clear why but this unit test seems to flake when running on CI. 
```
FAIL	github.com/flyteorg/flyte/flytectl/cmd/sandbox	0.086s
```

https://github.com/flyteorg/flyte/actions/runs/11580285322/job/32238460143?pr=5814

### Expected behavior

I expect the unit test to pass. 

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

nkwangleiGIT:
seems this testcase was successful at the latest attempt, it's not clear why it failed here, and not able to reproduce it locally.
do we have any log to check for the root cause in the CI? or we can add one for later investigation.

Sovietaced:
> seems this testcase was successful at the latest attempt, it's not clear why it failed here, and not able to reproduce it locally.
> 
> do we have any log to check for the root cause in the CI? or we can add one for later investigation.

I don't but I have a hunch the flakiness is related to transient issues with the docker calls and the GitHub runner.

eapolinario:
This is related to https://github.com/flyteorg/flyte/pull/5325 in that that PR didn't replace all instances of the `testutils.Setup` with the corresponding teardown.

Here's a PR with the fix: https://github.com/flyteorg/flyte/pull/5950

PR #5950, opened 2024-11-01
Fix remaining misuses of capturing the default file descriptors in flytectl unit tests

eapolinario:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/5937

https://github.com/flyteorg/flyte/pull/5325 didn't fix all the misuses of the capturing of stdout+stderr in unit tests. This PR fixes the remaining cases.

~We add the teardown as a defer statement to all uses of `testutils.Setup`. 
Notice that in most cases we use barebones tests, which don't have a central place for setup/teardowns, so even though there's a comment explaining that `Teardown` has to be invoked, there's no automated way, at least to my knowledge, to do that.~

edit: as [suggested](https://github.com/flyteorg/flyte/pull/5950#issuecomment-2452695075) by @Sovietaced , we can leverage the setup call to hookup the call to cleanup (to restore stdout and stderr).

## How was this patch tested?
Confirmed that tests ran successfully Ran unit tests in a loop a few times:
```
❯ for i in {1..100}; do
  go test -count=1 -cover ./... -race || break;
done
```

### DIFF
--- a/flytectl/cmd/compile/compile_test.go
+++ b/flytectl/cmd/compile/compile_test.go
@@ -6,7 +6,7 @@ import (
 
 	config "github.com/flyteorg/flyte/flytectl/cmd/config/subcommand/compile"
 	cmdCore "github.com/flyteorg/flyte/flytectl/cmd/core"
-	u "github.com/flyteorg/flyte/flytectl/cmd/testutils"
+	"github.com/flyteorg/flyte/flytectl/cmd/testutils"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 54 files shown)


Sovietaced:
I'm not super familiar with this style of unit tests but I feel that passing `*testing.T` to the setup function and then using `t.Cleanup(...)` is less error prone because it doesn't rely on anyone remembering to use `defer teardown()`.

Sovietaced (approved):

PR #5967, opened 2024-11-06
Add option to install CRD as a part of chart install in flyte-binary

marrrcin:
Due to strict RBAC policies in some k8s clusters, it's impossible to create `ClusterRole` with wide set of permissions for service account used in `flyte-binary` deployment. A namespace-scoped `Role` could be used, but it requires:
* creating own `Role` and `RoleBinding` (possible by creating a sub-chart)
* disabling installation of `FlyteWorkflow` CRD at runtime, after Helm install

More context: https://flyte-org.slack.com/archives/C01P3B761A6/p1729865751682139

1. Added CRD for FlyteWorkflow into flyte-binary chart
2. Added flag in values to configure propeller to skip CRD installation
3. Added flag in values to enable CRD installation at `helm install`

## How was this patch tested?

1. Created new k3d cluster
5. Installed modified Helm chart 2 times:
3.1. with `helm install`-level CRDs creation enabled and propeller-level CRDs creation disabled.
2.2. with `helm install`-level CRDs creation disabled and propeller-level CRDs creation enabled.
6. Both cases worked.

### Setup process
See above.

## Related PRs

N/A

## Docs link

N/A


chrisrun:
+1

charts/flyte-binary/values.yaml
```diff
@@ -375,6 +378,10 @@ rbac:
   # extraRules Add additional rules to the ClusterRole
   extraRules: []
 
+crds:
```
davidmirror-ops, line 381: Not sure I follow, is this section still needed? Similar to how flyte-core handles this, we could leave the `configuration.propeller.createCrd` key as both condition for the flyteworkflow CRD Helm template and as the input for the `create-flyteworkflow-crd` key in the configmap.

marrrcin: Changed.
The original intent was to be able to get rid of CRDs installation completely, but that might be too confusing.

charts/flyte-binary/templates/crds/flyteworkflow.yaml
```diff
@@ -0,0 +1,32 @@
+{{- if not .Values.configuration.propeller.createCRDs }}
```
davidmirror-ops, line 1:
```suggestion
{{- if  .Values.configuration.propeller.createCRDs }}
```
I think this should be the logic?

davidmirror-ops (changes requested):

marrrcin:
@davidmirror-ops pushed! :)

charts/flyte-binary/templates/crds/flyteworkflow.yaml
marrrcin: No, the logic is correct. If `propeller` is set to create CRDs at runtime, then we don't want to install them as a part of helm chart installation.

davidmirror-ops: You're right. Thanks for your patience, this helped me understand better how the CRD install operation works.

davidmirror-ops (approved):
I just installed flyte-binary using this chart and can confirm the expected behavior about the CRD.
Thank you!

marrrcin:
Awesome, thanks!

Issue #5907, opened 2024-10-24
[Core feature] Summary of Interactive Mode for Jupyter Notebook 


Mecoli1219:
### Motivation: Why do you think this is important?

This issue aggregates all PRs related to Jupyter Notebook interactive mode. 

### Goal: What should the final outcome look like, ideally?

```
remote = FlyteRemote(Config.auto(), default_project="flytesnacks", default_domain="development", interactive_mode_enabled=True)

out = remote.execute(wf, inputs={})
```

### Describe alternatives you've considered 

- [x] Support Interactive Mode for Jupyter Notebook
- [x] Automatic assign version
  - [ ] Need update as described in https://github.com/flyteorg/flyte/issues/5881
- [x] Restrict Dynamic mode
  - [ ] Perhaps support Dynamic mode?
- [x] Restrict different Python version
- [x] Blog post: https://hackmd.io/@1E0FEh2MS_OpXGUAjEFIOQ/ByTTT821Jl
- [ ] Some new ideas waiting to be discussed
  - [ ] <s>New interface like FlyteFuture? (Supporting `task.remote()`)</s>

###  Propose: Link/Inline OR Additional context 

Check out the blog post: https://hackmd.io/@1E0FEh2MS_OpXGUAjEFIOQ/ByTTT821Jl

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

eapolinario:
Thank you, this is one of the best features released recently in Flyte!

PR #5997, opened 2024-11-12
[WIP] [Docs] Add usage for jupyter notebook interactive mode

Mecoli1219:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5907

## Screenshots
image

## Docs link
https://flyte--5997.org.readthedocs.build/en/5997/user_guide/flyte_fundamentals/running_and_scheduling_workflows.html


Mecoli1219:
@samhita-alla @davidmirror-ops How does this look? (I am using a hacky way now. I'm unsure if there is any other appropriate method to render a cell from Jupyter Notebook.)

image

Future-Outlier:
Michael, you are 絕活哥

samhita-alla:
@Mecoli1219 thanks so much for working on this! have you tried using https://docs.readthedocs.io/en/stable/guides/jupyter.html#using-notebooks-in-other-formats since `myst-nb` is already enabled as a sphinx extension?

Mecoli1219:
@samhita-alla I realize the link you provided renders the entire jupytext format file as a single sphinx page. What I'm specifically looking for is rendering just one cell from an ipynb file, ideally with a link to the flytesnacks example. (Previously this worked with `rli`/`remoteliteralinclude`, but since we've switched to `literalinclude`, we lost the linking capability).

Thinking about it now, I may be overcomplicating this. We could simply write similar example code directly in our docs like in my screenshot.

What are your thoughts on the `interactive_mode` documentation - should we include the complete example, or would it be better to just show the basic usage (like in my screenshot) and perhaps provide links to flytesnacks for more detailed examples?

davidmirror-ops:
@Mecoli1219 I'd vote for rendering the notebook as a page if that's simpler. The benefit is that the notebook in flytesnacks would be "self contained" with code and instructions in a single place.

Mecoli1219:
Closed as the alternative PR is created: https://github.com/flyteorg/flyte/pull/6036

PR #6023, opened 2024-11-18
chore: using domain-qualified finalizers

### LINKED ISSUE #6019: [Housekeeping] Use domain-qualified finalizers
### Describe the issue

After upgrading our cluster to Kubernetes 1.29 we started seeing warnings like the following in Flyte admin logs:
```
W1115 11:54:31.278411      14 warnings.go:70] metadata.finalizers: "flyte-finalizer": prefer a domain-qualified finalizer name to avoid accidental conflicts with other finalizer writers
```
The change was introduced in https://github.com/kubernetes/kubernetes/pull/119508 so it's an external change we should adapt to. The [current finalizer](https://github.com/flyteorg/flyte/blob/d1a723e001556e1fb301e2d7e61735dda0b41558/flytepropeller/pkg/controller/finalizer.go#L5) is not domain-qualified and besides all the finalizer-managing logic in that file has become obsolete.

Given this, we should change the finalizer to a proper one and also take the opportunity to get rid of that code and leverage the finalizer logic in https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/controller/controllerutil/controllerutil.go, which we would have maintained and tested for free.

Not using domain-qualified finalizers causes noisy warning messages in the logs for now but such finalizers might be required in future Kubernetes versions, which would mean all of Flyte finalizer logic would stop working.

### Related component(s)

Flyteadmin, flytepropeller.

- [X] Yes

- [X] Yes

trutx:
## Tracking issue

Closes #6019 

Switching to domain-qualified finalizers. Kubernetes introduced a warning in https://github.com/kubernetes/kubernetes/pull/119508 so using old finalizers is harmless today, but updating them for the sake of clean Flyte admin logs and in advance of possible future enforcements of such finalizers.

* Switch to a global domain-qualified finalizer: from `flyte-finalizer` to `flyte.lyft.com/finalizer`
* Switch the `k8s` plugin finalizer from `flyte/flytek8s` to `flyte.lyft.com/finalizer-k8s`
* Switch the `array` plugin finalizer from `flyte/array` to `flyte.lyft.com/finalizer-array`
* Remove the `finalizers.go` and `finalizers_test.go` files and start leveraging the finalizer goodies in the upstream `controllerutil` package
* Keep the _removal_ of the old finalizer for backwards compatibility. It will no longer be added. This part should be eventually removed
* Stop removing _all_ finalizers and only remove Flyte's instead. This allows users to add their own finalizers

## How was this patch tested?

Unit tests were modified to check for the presence/absence of the new finalizer. Some new tests were added and some existing tests were fixed so that `clearFinalizer()` func was actually run.


flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager_test.go
```diff
@@ -1015,15 +1036,16 @@ func TestFinalize(t *testing.T) {
...
+				Namespace:  tctx.TaskExecutionMetadata().GetNamespace(),
+				Finalizers: []string{finalizer},
 			},
 		}
 
 		assert.NoError(t, fakeKubeClient.GetClient().Create(ctx, o))
 
 		p.OnBuildIdentityResource(ctx, tctx.TaskExecutionMetadata()).Return(o, nil)
-		pluginManager := PluginManager{plugin: &p, kubeClient: fakeKubeClient}
+		pluginManager := PluginManager{plugin: &p, kubeClient: fakeKubeClient, updateBackoffRetries: 5}
```
trutx, line 1048: This `updateBackoffRetries` parameter ends up being `e.updateBackoffRetries` in
```
	retryBackoff := wait.Backoff{
		Duration: time.Duration(e.updateBaseBackoffDuration) * time.Millisecond,
		Factor:   2.0,
		Jitter:   0.1,
		Steps:    e.updateBackoffRetries,
	}
```
If this is unset (which it was in tests), then `Steps: 0` which means `_ = wait.ExponentialBackoff(retryBackoff, func() (bool, error) {` inside `Finalize()` times out almost immediately, which effectively means `e.clearFinalizer()` is never called and the finalizer is never removed. The default setting is 5 but I think it's too dangerous a) to allow users to tweak these backoff settings, or b) to use an exponential backoff at all.

Thoughts?

eapolinario: The [original](https://github.com/flyteorg/flyte/pull/5673) context of this change had to do with informers getting stale info in the case of array nodes. 

cc: @pvditt 

> to allow users to tweak these backoff settings

Do you mean we should validate that it's a strictly positive value, right?

> to use an exponential backoff at all.

Can you expand on that?

trutx:
> Do you mean we should validate that it's a strictly positive value, right?

That, or completely remove the option to set an arbitrary value there. I can picture situations where a too short backoff expires when the k8s client inside the backoff func tries to talk to a busy/lagged/high network latency k8s apiserver.

> Can you expand on that?

What I meant is we have to decide first if it's ok to leave to-be-deleted objects in the cluster with a finalizer nobody is going to remove, or not. If that's ok we don't need a backoff, we just need to try once and we have to make sure no ultrafast backoff gives up before the one try has even been performed.

But OTOH, if we don't want to leave any garbage behind maybe we should retry infinitely instead. I am up for exponentially spacing the retries not to hammer the apiserver, but I think such retries should keep happening until the removal finally succeeds.

However I think all of this is outside the scope of this PR :D

pingsutw:
> But OTOH, if we don't want to leave any garbage behind maybe we should retry infinitely instead. I am up for exponentially spacing the retries not to hammer the apiserver, but I think such retries should keep happening until the removal finally succeeds.

+1 to retry with backoff. We’ve noticed some pods staying in the namespace indefinitely, and we’ve had to manually remove them, which is not great.

pingsutw (commented):
LGTM overall, @pvditt mind taking another look as well

pvditt:
I'm in favor of keeping the current retries w/ backoff as is for now. If that fails after the hardcoded values, it should bubble up in an error. 

We did run into issues as @pingsutw mentioned, but that was w/ an external service re-adding finalizers after propeller cleared them. 

Thanks for cleaning up a lot of the finalizers logic - I agree that propeller should only remove flyte specific finalizers. Changes look great.

pvditt (approved):

Issue #552, opened 2020-10-14
[Backend][Plugin] Create a fail fast plugin


katrogan:
**Why would this plugin be helpful to the Flyte community**
In certain cases it would be useful to leverage [configurable plugins](https://github.com/lyft/flyte/issues/499) as a whitelist mechanism without necessarily having a fallback implementation. In this case a fail-fast plugin implementation would be useful to have as the default plugin implementation for a task type and whitelisted projects/domains could be added for the actual plugin handler.

**Type of Plugin**
- [ ] Web Service (e.g. AWS Sagemaker, GCP DataFlow, Qubole etc...)
- [ ] Kubernetes Operator (e.g. TfOperator, SparkOperator, FlinkK8sOperator, etc...)
- [x] Customized Plugin using native kubernetes constructs
- [ ] Other

**Can you help us with the implementation?**
- [x] Yes
- [ ] No

**Additional context**
...

katrogan:
cc @kumare3 @wild-endeavor

kumare3:
@katrogan can we create this plugin in plugin machinery and instead of no-op its actually a fail fast?

katrogan:
@kumare3 do you mean implement the core pluginmachinery plugin interface https://github.com/lyft/flyteplugins/blob/master/go/tasks/pluginmachinery/core/plugin.go#L36?

kumare3:
@katrogan yes, and keep it as a plugin that is available with pluginmachinery

PR #6045, opened 2024-11-21
Revert "[COR-2297/] Fix nested offloaded type validation (#552) (#5996)"

eapolinario:
This reverts commit f20b8aa082820c56c0c721670b54148afa7d36a4.

## Tracking issue
Related to https://github.com/flyteorg/flyte/pull/5103

In https://github.com/flyteorg/flytekit/pull/2872 we're handling the case of offloading literals in map tasks single executions explicitly. The impact of this is that we can simplify the handling of this case in the backend, invalidating https://github.com/flyteorg/flyte/pull/5996.

Revert https://github.com/flyteorg/flyte/pull/5996.

## How was this patch tested?
Confirmed that the original test workflow works:
```
import logging
from typing import List
from flytekit import map_task, task, workflow,LaunchPlan

logging.basicConfig(level=logging.DEBUG)
logger = logging.getLogger("flytekit")
logger.setLevel(logging.DEBUG)

@task(cache=True, cache_version="1.1")
def my_30mb_task(i: str) -> str:
    return f"Hello world {i}" * 30 * 100 * 1024

@task(cache=True, cache_version="1.1")
def generate_strs(count: int) -> List[str]:
    return ["a"] * count

@workflow
def my_30mb_wf(mbs: int) -> List[str]:
  strs = generate_strs(count=mbs)
  return map_task(my_30mb_task)(i=strs)

@workflow
def big_inputs_wf(input: List[str]):
   noop()

@task(cache=True, cache_version="1.1")
def noop():
    ...

big_inputs_wf_lp = LaunchPlan.get_or_create(name="big_inputs_wf_lp", workflow=big_inputs_wf)

@workflow
def ref_wf(mbs: int):
  big_inputs_wf_lp(input=my_30mb_wf(mbs))
```

### DIFF
--- a/flytepropeller/pkg/compiler/validators/utils.go
+++ b/flytepropeller/pkg/compiler/validators/utils.go
@@ -276,13 +276,13 @@ func LiteralTypeForLiteral(l *core.Literal) *core.LiteralType {
 	case *core.Literal_Collection:
 		return &core.LiteralType{
 			Type: &core.LiteralType_CollectionType{
-				CollectionType: literalTypeForLiterals(l.GetCollection().Literals),
+				CollectionType: literalTypeForLiterals(l.GetCollection().GetLiterals()),
 			},
 		}
 	case *core.Literal_Map:
 		return &core.LiteralType{
 			Type: &core.LiteralType_MapValueType{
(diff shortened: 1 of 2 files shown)


wild-endeavor (approved):
+1 this is just a pure git revert

PR #6054, opened 2024-11-27
Support ArrayNode subNode timeouts

hamersaw:
## Tracking issue
N/A

Currently, timeouts are not supported on ArryaNode subNodes. If there is a timeout set in the task decorator, that is applied only to the overall ArrayNode and not individual subNodes. This change ensures subNodes are executed with timeouts as a standalone execution would be.

We store a `deltaTimestamp` to track `startedAt` for subNode executions. This delta is computed from the overall ArrayNode start time. This is stored using a `bitarray` (just like other ArrayNode subNode metadata), with a preset 18 bits of precision. This means for 5k subNodes there will be approximately 90kb added to the FlyteWorkflow CR - reducing scalability by a small amount. The 18 bits gives us around 3 days of delta time, meaning subNodes start times can be tracked up to 3 days after the overall ArrayNode starts. If this proves to be too imprecise, we can increase this value and incur further reductions to scale.

It is important to note **that we do not have a task phase for `TIMED_OUT`**. So in the UI this will display as `FAILED`. This is a restriction of the current eventing scheme for map tasks.

## How was this patch tested?
This was tested locally under a variety of failure scenarios (ex. retryable failures).

### Setup process
```
from flytekit import map_task, task, workflow
from typing import List

import time
from flytekit.exceptions.user import FlyteRecoverableException


@task(retries=3, timeout=timedelta(minutes=1))
def say_hello(name: str) -> str:
    #time.sleep(120)

    time.sleep(10)
    raise FlyteRecoverableException(f"testing failure for {name}")

    return f"hello {name}"


@workflow
def say_hello_wf(name: List[str]) -> List[str]:
    return map_task(say_hello)(name=name)
```

### Screenshots
N/A

## Related PRs
N/A

## Docs link
N/A

### DIFF
--- a/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/iface.go
+++ b/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/iface.go
@@ -290,6 +290,7 @@ type ExecutableArrayNodeStatus interface {
 	GetSubNodeTaskPhases() bitarray.CompactArray
 	GetSubNodeRetryAttempts() bitarray.CompactArray
 	GetSubNodeSystemFailures() bitarray.CompactArray
+	GetSubNodeDeltaTimestamps() bitarray.CompactArray
 	GetTaskPhaseVersion() uint32
 }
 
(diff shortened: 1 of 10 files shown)


pvditt (approved):
changes look great. This reminded me of this issue https://linear.app/unionai/issue/COR-2212/[bug]-setting-task-metadata-for-the-parent-array-node-causes-incorrect where the timeout gets applied to the parent node. Let me this change up in flytekit this evening - would probably be nice to have this fixed before this subnode timeout gets released.

hamersaw:
> changes look great. This reminded me of this issue [https://linear.app/unionai/issue/COR-2212/[bug]-setting-task-metadata-for-the-parent-array-node-causes-incorrect](https://linear.app/unionai/issue/COR-2212/%5Bbug%5D-setting-task-metadata-for-the-parent-array-node-causes-incorrect) where the timeout gets applied to the parent node. I'll get a small PR up to fix this on the flytekit end.

This example workflow in ^^ works here. I think the timeout gets applied to the parent and subNodes. I know I looked through flytekit and convinced myself of it before implementing this, but can't seem to find the line now.

PR #6058, opened 2024-11-29
Flyteadmin digest comparison should rely on database semantics

### LINKED ISSUE #4780: [Housekeeping] Flyteadmin digest comparison should rely on database semantics
### Describe the issue

In order to compare an already registered task (and workflow, and launch plan) when an identical version is registered, we [get](https://github.com/flyteorg/flyte/blob/8f516e839074d0c1013ab58650a1a0fec575d990/flyteadmin/pkg/manager/impl/task_manager.go#L91-L104) the existing task from the db and manually compare its digest. This is prone to race conditions. It would be better to wrap this in a transaction where we attempt to upsert the task only on a matching digest

Slightly more error prone method for determining already exists task structure

### Related component(s)

flyteadmin

- [X] Yes

- [X] Yes

popojk:
## Tracking issue
Closes #4780 

In current TaskManager CreateTask [code](https://github.com/flyteorg/flyte/blob/8f516e839074d0c1013ab58650a1a0fec575d990/flyteadmin/pkg/manager/impl/task_manager.go#L91-L120), FlyteAdmin checks if a task with the same ID already exists in the database. If it does, FlyteAdmin verifies whether the registered task has a different digest compared to the existing task. If no task with the same ID is found in the database, FlyteAdmin proceeds to create the task in the database.

However, the current approach may lead to a race condition that prevents the digest comparison for two identical tasks from occurring. For example, consider two identical tasks (tasks with the same ID and digest), A and B, being registered to FlyteAdmin simultaneously. It is likely that the digest check will be skipped because the existing task is not yet present in the database. Consequently, one task will be created in the database, and the other will fail due to a primary key conflict. (Refer to the diagram below for a better understanding.)
 
截圖 2024-11-29 下午2 36 48

1.Do digest check in a transactional way:

The procedure of creating task should be 1. create task -> 2. if task id exists already(pramary key conflict) -> 3. do digest check. The pseudocode could look like
```
in the transaction:
try:
  create a task with given primary key
except:
  primary key already exists
  get existing entry with identical primary key
    if digest of existing == new entry's digest -> return `NewTaskExistsIdenticalStructureError`
    else -> return `NewTaskExistsDifferentStructureError`
```
In this way we can make sure that task digest will be checked even though 2 identical task registered at the same time frame. Refer to the diagram below for a better understanding.

截圖 2024-11-22 下午5 28 50

2.Write Task to DB before write Description in TaskRepo Create method:

In current TaskRepo [Create method](https://github.com/flyteorg/flyte/blob/8f516e839074d0c1013ab58650a1a0fec575d990/flyteadmin/pkg/repositories/gormimpl/task_repo.go#L23-L47), task description is created before task. However, if TaskManger catches primary key conflict error from task description creation and try to get existing task in DB for digest check, a task not found error could possibly occurred as task is not yet created in DB, which does not make sense for user. In this PR it is proposed to write Task to DB before write Description in TaskRepo Create method.

## How was this patch tested?

Set up a simple workflow with 2 tasks

截圖 2024-11-29 下午3 23 27

Write a shell script to request task registration 10 times at the same time to simulate hi concurrency situation. It is expected that each task will be registered successfully once only, otherwise the response message should shown AlreadyExists.

截圖 2024-11-29 下午3 31 06

The result show each task only registered once as we expected

截圖 2024-11-28 上午11 55 20
截圖 2024-11-28 上午11 55 46

Then, we make a shell script to register 2 groups of tasks with same ID but different digest at the same time. It is expected that TaskExistsDifferentStructureError will shown in the response

截圖 2024-11-29 下午3 40 48

The error shown as expected

截圖 2024-11-29 上午11 51 21

 <div id='description'>
<h3>Summary by Bito</h3>
This PR implements a transactional approach to resolve race conditions in FlyteAdmin's task registration process. The changes modify task creation workflow by handling primary key conflicts and performing digest checks. The implementation reorders task and description creation sequence in the database for consistent error handling.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/flyteadmin/pkg/repositories/gormimpl/task_repo.go
+++ b/flyteadmin/pkg/repositories/gormimpl/task_repo.go
@@ -30,12 +30,12 @@ func (r *TaskRepo) Create(ctx context.Context, input models.Task, descriptionEnt
 			}
 			return nil
 		}
-		tx := r.db.WithContext(ctx).Omit("id").Create(descriptionEntity)
+		tx := r.db.WithContext(ctx).Omit("id").Create(&input)
 		if tx.Error != nil {
 			return r.errorTransformer.ToFlyteAdminError(tx.Error)
 		}
 
-		tx = r.db.WithContext(ctx).Omit("id").Create(&input)
+		tx = r.db.WithContext(ctx).Omit("id").Create(descriptionEntity)
 		if tx.Error != nil {
 			return r.errorTransformer.ToFlyteAdminError(tx.Error)
 		}
--- a/flyteadmin/pkg/manager/impl/task_manager.go
+++ b/flyteadmin/pkg/manager/impl/task_manager.go
@@ -88,19 +88,7 @@ func (t *TaskManager) CreateTask(
 		logger.Errorf(ctx, "failed to compute task digest with err %v", err)
 		return nil, err
 	}
-	// See if a task exists and confirm whether it's an identical task or one that with a separate definition.
-	existingTaskModel, err := util.GetTaskModel(ctx, t.db, request.GetSpec().GetTemplate().GetId())
-	if err == nil {
-		if bytes.Equal(taskDigest, existingTaskModel.Digest) {
-			return nil, errors.NewTaskExistsIdenticalStructureError(ctx, request)
-		}
-		existingTask, transformerErr := transformers.FromTaskModel(*existingTaskModel)
-		if transformerErr != nil {
-			logger.Errorf(ctx, "failed to transform task from task model")
-			return nil, transformerErr
-		}
-		return nil, errors.NewTaskExistsDifferentStructureError(ctx, request, existingTask.GetClosure().GetCompiledTask(), compiledTask)
-	}
+	// Create Task in DB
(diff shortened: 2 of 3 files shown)


flyteadmin/pkg/repositories/gormimpl/task_repo.go
```diff
@@ -30,12 +30,12 @@ func (r *TaskRepo) Create(ctx context.Context, input models.Task, descriptionEnt
 			}
 			return nil
 		}
-		tx := r.db.WithContext(ctx).Omit("id").Create(descriptionEntity)
+		tx := r.db.WithContext(ctx).Omit("id").Create(&input)
```
katrogan, line 33: I think this change is technically not necessary since this is all wrapped in a transaction, and if any insert fails then the whole transaction should be rolled back.

katrogan (approved):
Nice! thank you so much for taking on these changes

for the testing:
> Then, we make a shell script to register 2 groups of tasks with same ID but different digest at the same time. 
did you modify the task definition to force a different task digest? I didn't quite follow from the description

eapolinario (approved):

PR #6098, opened 2024-12-10
Fix flytestdlib's stowStore.List for google cloud storage

fg91:

For RFC https://github.com/flyteorg/flyte/pull/5598, flytepropeller was given the ability to list error files in the so-called *raw output prefix* bucket of an execution with the goal of identifying which worker pod in a failed distributed task experienced the first error.

For this purpose, the `StowStore` in flytestdlib was given a `List` function.

For google cloud storage buckets, the listing and subsequent access of the error files currently does not work: When listing a bucket `gs://some-bucket/...`, one receives items in the form `google://storage.googleapis.com/download/storage/v1/b/some-bucket/...` which then cannot be found by the stow store for the `gs://` prefix.
This PR fixes this.

Don't use `item.URL()` but construct the url from the protocol, the container name (bucket name), and the item name.

## How was this patch tested?

Tested in a GCP deployment and adapted unit test.

## Related PRs

* https://github.com/flyteorg/flyte/pull/5741
* https://github.com/flyteorg/flyte/pull/5795

### DIFF
--- a/flytestdlib/storage/stow_store.go
+++ b/flytestdlib/storage/stow_store.go
@@ -263,7 +263,7 @@ func (s *StowStore) Head(ctx context.Context, reference DataReference) (Metadata
 }
 
 func (s *StowStore) List(ctx context.Context, reference DataReference, maxItems int, cursor Cursor) ([]DataReference, Cursor, error) {
-	_, containerName, key, err := reference.Split()
+	protocol, containerName, key, err := reference.Split()
 	if err != nil {
 		s.metrics.BadReference.Inc(ctx)
 		return nil, NewCursorAtEnd(), err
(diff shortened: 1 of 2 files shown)


flytestdlib/storage/stow_store.go
```diff
@@ -291,7 +291,17 @@ func (s *StowStore) List(ctx context.Context, reference DataReference, maxItems
...
+			// bucket `gs://some-bucket/...` will result in items with URLs like
+			// `google://storage.googleapis.com/download/storage/v1/b/some-bucket/...`
+			// which subsequently cannot be found by the stow store.
+
+			u := url.URL{
+				Scheme: protocol,
+				Host:   containerName,
+				Path:   item.Name(),
+			}
+			results[index] = DataReference(u.String())
```
eapolinario, line 304: Can we fix this in our fork of stow? It looks like we change the scheme [here](https://github.com/flyteorg/stow/blame/632adef717d32cc01727f9ec5b064c0505bbffb9/google/item.go#L84), which is used in gcs implementation of [convertToStowItem](https://github.com/flyteorg/stow/blob/master/google/container.go#L152-L176).

fg91: Will do, overlooked we use a fork of stow 👌

fg91:
> Can we fix this in our fork of stow? It looks like we change the scheme [here](https://github.com/flyteorg/stow/blame/632adef717d32cc01727f9ec5b064c0505bbffb9/google/item.go#L84), which is used in gcs implementation of [convertToStowItem](https://github.com/flyteorg/stow/blob/master/google/container.go#L152-L176).

Closing this PR in favour of https://github.com/flyteorg/stow/pull/17 @eapolinario

PR #6103, opened 2024-12-11
Fix: Make distributed error aggregation opt-in

fg91:

For RFC https://github.com/flyteorg/flyte/pull/5598, flytepropeller was given the ability to list error files in the so-called raw output prefix bucket of an execution with the goal of identifying which worker pod in a failed distributed task experienced the first error.

In GCP, listing the error files requires the `"storage.objects.list"` permission which so far wasn't given to propeller. I added this permission to the Flyte propeller custom role [here](https://github.com/unionai-oss/deploy-flyte/pull/40).

That being said, because this feature is therefore not backwards compatible, I propose to make it opt-in.

If you agree with this, I'll make another PR to document this feature and how to activate it [here](https://docs.flyte.org/en/latest/flytesnacks/examples/kfpytorch_plugin/pytorch_mnist.html) and/or [here](https://docs.flyte.org/en/latest/deployment/plugins/k8s/index.html#install-the-kubernetes-operator).

Only search for multiple error files from the different workers of a distributed task as proposed in RFC https://github.com/flyteorg/flyte/pull/5598 if actively enabled in the flytepropeller config in order to not strictly require the addition of the `"storage.objects.list"` permission.

## How was this patch tested?

Ran flytepropeller with/without the flag enabled locally for a GKE based deployment and adapted unit tests.

### DIFF
--- a/flyteplugins/go/tasks/pluginmachinery/flytek8s/config/config.go
+++ b/flyteplugins/go/tasks/pluginmachinery/flytek8s/config/config.go
@@ -218,6 +218,8 @@ type K8sPluginConfig struct {
 
 	// Extended resources that should be added to the tolerations automatically.
 	AddTolerationsForExtendedResources []string `json:"add-tolerations-for-extended-resources" pflag:",Name of the extended resources for which tolerations should be added."`
+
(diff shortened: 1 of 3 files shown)


eapolinario (approved):
Thank you. This makes sense, please tag me on the docs PR as well.

fg91:
> Thank you. This makes sense, please tag me on the docs PR as well.

https://github.com/flyteorg/flytesnacks/pull/1776

PR #6145, opened 2025-01-06
Add labels and annotations to with_overrides

pryce-turner:

Improvement to allow setting labels and annotations dynamically at run time for things like cost allocation

Adds labels and annotations to with_overrides()

## How was this patch tested?

TBD

## Docs link

TBD

### DIFF
--- a/flyteidl/protos/flyteidl/core/workflow.proto
+++ b/flyteidl/protos/flyteidl/core/workflow.proto
@@ -12,6 +12,7 @@ import "flyteidl/core/literals.proto";
 import "flyteidl/core/tasks.proto";
 import "flyteidl/core/types.proto";
 import "flyteidl/core/security.proto";
+import "flyteidl/admin/common.proto";
 import "google/protobuf/duration.proto";
 import "google/protobuf/wrappers.proto";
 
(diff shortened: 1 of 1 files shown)


flyteidl/protos/flyteidl/core/workflow.proto
```diff
@@ -199,6 +200,10 @@ message NodeMetadata {
     oneof cache_serializable_value {
         bool cache_serializable = 9;
     }
+
+    Labels labels = 10;
+
+    Annotations annotations = 11;
```
thomasjpfan, line 206:
```suggestion
    flyteidl.admin.Labels labels = 10;

    flyteidl.admin.Annotations annotations = 11;
```

thomasjpfan (commented):
To generate the flyteidl files from a proto change:

```bash
cd flyteidl
make generate
```

Then build flyte binary:

https://docs.flyte.org/en/latest/community/contribute/contribute_code.html#requirements

flyteidl/protos/flyteidl/core/workflow.proto
pryce-turner: How come RetryStrategy can be inferred from `literals` but Labels and Annotations can't be inferred from `common`?

PR #6159, opened 2025-01-10
Fix: Upgrade stow to fix listing error files

fg91:

For RFC https://github.com/flyteorg/flyte/pull/5598, flytepropeller was given the ability to list error files in the so-called *raw output prefix* bucket of an execution with the goal of identifying which worker pod in a failed distributed task experienced the first error.

For this purpose, the `StowStore` in flytestdlib was given a `List` function.

For google cloud storage buckets, the listing and subsequent access of the error files currently does not work: When listing a bucket `gs://some-bucket/...`, one receives items in the form `google://storage.googleapis.com/download/storage/v1/b/some-bucket/...` which then cannot be found by the stow store for the `gs://` prefix.

This was fixed in https://github.com/flyteorg/stow/pull/17. This PR upgrades the stow dependency

 
 <div id='description'>
<h3>Summary by Bito</h3>
Dependency upgrade of stow from v0.3.10 to v0.3.11 across Flyte components to resolve GCS bucket listing issues. The update fixes incompatible URL format problems in Google Cloud Storage buckets. This change supports RFC #5598 implementation for improved failure detection in distributed tasks.
<br>
<br>
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 3
</div>


pingsutw (approved):
Thank you!

PR #6178, opened 2025-01-17
set viper delimiter to :: to support . in config keys

### LINKED ISSUE #6166: [BUG] Parsing YAML for k8s DefaultAnnotations misunderstands "." in annotation keys to mean a nested dict.
### Describe the bug

In the FlytePropellor config ([docs](https://docs.flyte.org/en/latest/deployment/configuration/generated/flytepropeller_config.html)), there appears to be a bug in how the YAML file is parsed. Dots in keys are misunderstood as nested dicts. This is relevant, because e.g. Google Kubernetes Engine we might want to set an annotation:
`cluster-autoscaler.kubernetes.io/safe-to-evict: "false"`, which includes dots.

Consider the following config
```
  k8s:
    plugins:
      k8s:
        default-annotations: 
          test.annotation: "true"
```

This raises the following flytepropellor error:
```
time="2025-01-14T10:33:00Z" level=error msg="\n\n\n1 error(s) decoding:\n\n* 'default-annotations[test]' expected type 'string', got unconvertible type 'map[string]interface {}', value: 'map[annotation:true]'"
```
This suggests that the YAML is parsed into `{"default-annotations": {"test": {"annotation": "true}}}` instead of `{"default-annotations": {"test.annotation": "true}}`, as we'd want.
### Expected behavior

I'd expect the above config to lead to a default annotation `test.annotation: "true"`, but I get a FlytePropellor error.

Using flyte-core 1.13.3

### Screenshots

_No response_

- [X] Yes

- [X] Yes

pvditt:
## Tracking issue
closes: https://github.com/flyteorg/flyte/issues/6166

parsing configs breaks when "." is part of a key

Update key delimiter to "::" when parsing configs

Note both "." and "::" in keys are valid yaml, but "." seems more likely to be utilized than "::"

## How was this patch tested?
- updated unit test

### Labels

### DIFF
--- a/flytestdlib/config/tests/accessor_test.go
+++ b/flytestdlib/config/tests/accessor_test.go
@@ -22,6 +22,7 @@ import (
 	k8sRand "k8s.io/apimachinery/pkg/util/rand"
 
 	"github.com/flyteorg/flyte/flytestdlib/config"
+	"github.com/flyteorg/flyte/flytestdlib/config/viper"
 	"github.com/flyteorg/flyte/flytestdlib/internal/utils"
 )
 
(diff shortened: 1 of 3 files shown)


fg91:
@pvditt I ran into this viper limitation today too. Is this PR something we could revive?

machichima:
Hi @fg91 ,
I created a new PR for this: https://github.com/flyteorg/flyte/pull/7369. PTAL if you have time!
Thanks

PR #6201, opened 2025-01-28
Enable 0 workers replicaSpec for pytorchjob

punkerpunker:

Currently, PyTorchJob allows not to specify Worker in case only one worker (Master) is needed for the job. Current functionality of the plugin doesn't allow to do so, however, in some cases we still want to achieve a Master-only PyTorchJob.

Made a change so `pytorch` plugin is no longer complains about a 0-replica worker, but instead - if replicas=0 is passed, plugin just doesn't render it. 

I understand this isn't a best design for the moment, but that is a least-invasive approach to achieve it without changing e.g `pyflyte` validation behavior (if e.g we accept that worker=None could be passed to the PyTorch `task_config`. 

## How was this patch tested?

We are running the fork code in our production environment

 <div id='description'>
<h3>Summary by Bito</h3>
Enhanced PyTorch operator plugin to support master-only configurations by removing the requirement for worker replicas. Modified replica specification logic to make worker specs optional. Updated test suite to validate both zero-worker and single-worker configurations.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
+++ b/flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
@@ -145,16 +145,22 @@ func (p pytorchOperatorResourceHandler) BuildResource(ctx context.Context, taskC
 			"Invalid TaskSpecification, unsupported task template version [%v] key", taskTemplate.GetTaskTypeVersion())
 	}
 
+	jobSpec := kubeflowv1.PyTorchJobSpec{}
 	if *workerReplicaSpec.Replicas <= 0 {
-		return nil, fmt.Errorf("number of workers must be greater than 0")
-	}
-
-	jobSpec := kubeflowv1.PyTorchJobSpec{
-		PyTorchReplicaSpecs: map[commonOp.ReplicaType]*commonOp.ReplicaSpec{
-			kubeflowv1.PyTorchJobReplicaTypeMaster: masterReplicaSpec,
-			kubeflowv1.PyTorchJobReplicaTypeWorker: workerReplicaSpec,
-		},
-		RunPolicy: runPolicy,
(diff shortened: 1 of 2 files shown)


flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
```diff
@@ -145,16 +145,17 @@ func (p pytorchOperatorResourceHandler) BuildResource(ctx context.Context, taskC
 			"Invalid TaskSpecification, unsupported task template version [%v] key", taskTemplate.GetTaskTypeVersion())
 	}
 
-	if *workerReplicaSpec.Replicas <= 0 {
-		return nil, fmt.Errorf("number of workers must be greater than 0")
+	jobSpec := kubeflowv1.PyTorchJobSpec{}
```
pingsutw, line 148: Could we add a small unit test?

punkerpunker: Hey @pingsutw, I've removed `TestBuildResourcePytorchV1WithZeroWorker` as it was intended to fail in the case corrected with this PR. Replaced it with the `TestBuildResourcePytorchV1WithDifferentWorkersNumber` test, thank you.

pingsutw (approved):

flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
pingsutw: One unit test is failing. otherwise, lgtm.

punkerpunker: Should be fixed now, thanks.

pingsutw (approved):

Issue #642, opened 2020-12-16
[Feature][UI] Show link to k8s dashboard on Execution Details page.


schottra:
Administrators debugging issues with Flyte infrastructure would benefit from quick access to the k8s dashboard in order to access more information.
This link is cluster and region-specific, but typically each region will run its own instance of `flytconsole`. So, the UI just needs to be able to construct a link by substituting the `project`, `domain`, and `cluster` values from an Execution into a format string.

**Flyte component**
- [ ] Overall
- [ ] Flyte Setup and Installation scripts
- [ ] Flyte Documentation
- [ ] Flyte communication (slack/email etc)
- [ ] FlytePropeller
- [ ] FlyteIDL (Flyte specification language)
- [ ] Flytekit (Python SDK)
- [ ] FlyteAdmin (Control Plane service)
- [ ] FlytePlugins
- [ ] DataCatalog 
- [ ] FlyteStdlib (common libraries)
- [x] FlyteConsole (UI)
- [ ] Other

**[Optional] Propose: Link/Inline**
* Flyteconsole will add an additional environment variable specifying the k8s dashboard format string, with placeholders for the `project`, `domain`, and `cluster` values. If this value is present in the client-side bundle, we will render an additional link on the ExecutionDetails page to the k8s dashboard for the cluster.

kumare3:
@schottra how do we know where the cluster is hosted at - I mean the endpoint? 
At Lyft we use the cluster id in a dns configuration, but that may not be true always. This seems to be a configuration that should be driven from FlyteAdmin. @anandswaminathan ?

schottra:
@kumare3 Platform administrators in any given deployment know what the url is and that would be provided as an environment variable to flyteconsole. My understanding is that the only parts which change regularly are the cluster/project/domain. 
So, for example, we may provide an environment variable of the form:
```
K8S_DASHBOARD_FORMAT_STRING=https://dashboard.ingress.{{cluster}}.us-east-1.mycompany.com/#!/pod?namespace={{project}}-{{domain}}
```
Flyteconsole would read this env variable and do a string format to insert the correct values fro cluster/project/domain.

I agree that a more robust solution would be to add the dashboard link as a field on an entity returned by Admin and have administrators configure Admin with the correct values. That may require more effort than is available at the moment. :-)

kumare3:
@schottra thank you for the detailed explanation. So just a couple things
1. Absence of this configuration removes the link from the UI?
[...]

PR #6228, opened 2025-02-06
Return NotFound error when decks uri is empty (#642)

pmahindrakar-oss:
A deck url not found in node closure shouldn't result in an internal error and instead should return NotFound grpc code

Changing the returned grpc code for the API

## How was this patch tested?

Tested by verifying the UI calls are no longer returning 500 and instead returning 404 
 <div id='description'>
<h3>Summary by Bito</h3>
The PR implements improved error handling in the data proxy service, changing error codes from Internal to NotFound for missing deck URLs. The test suite underwent significant refactoring with the introduction of CreateDownloadLinkDependencies struct and setup function. Additional test cases were implemented to validate error code behavior for empty deck URLs.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/flyteadmin/dataproxy/service.go
+++ b/flyteadmin/dataproxy/service.go
@@ -179,7 +179,7 @@ func (s Service) CreateDownloadLink(ctx context.Context, req *service.CreateDown
 	}
 
 	if len(nativeURL) == 0 {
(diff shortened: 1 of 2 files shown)


Future-Outlier (approved):
thank you

PR #6245, opened 2025-02-14
RFC: Simplify `flytekit.Resources`

granthamtaylor:
Defining resources for tasks is error prone and counter-intuitive due to the nuances around `requests` and `limits`.

This RFC discusses an alternative solution to simplify the definition of resources, remove some footguns, and smoothen flytekit altogether.

### DIFF
--- a/rfc/system/0000-simplify-resources.md
+++ b/rfc/system/0000-simplify-resources.md
@@ -1,0 +1,61 @@
+# [RFC Template] Simplify Resources
+
+**Authors:**
+
+- @granthamtaylor
+
+## 1 Executive Summary
+
+Defining resources for tasks is error prone and counter-intuitive due to the nuances around `requests` and `limits`.
+
+This RFC discusses an alternative solution to simplify the definition of resources, remove some footguns, and smoothen flytekit altogether.
+
+## 2 Motivation
+
+Some of the footguns around `requests` and `limits` are non-deterministic and not intuitive, thus can escape detection during development but have significant and detrimental impact in production.
+
+For example, the possibility of setting `requests < limits` for `mem` can lead to a pod becoming overloaded. For `cpu`, doing so can lead to CPU availability, for some multi-processing applications, to be non-deterministic, thus resulting in unpredictable performance degradation.
+
+Additionally, setting `requests < limits` for `gpu` and `ephemeral_storage` should not be allowed.
+
+Having `limits` and `requests` be configured under two distinct instances of `flytekit.Resources` makes validation challenging.
+
+Lastly, the implementation of `limits` and `requests` as two distinct instances of `flytekit.Resources` makes for an awkward experience of defining new resource configurations (`accelerator`, `shared_memory`, `processor`, etc)
+
+## 3 Proposed Implementation
+
+- Deprecate `limits` and `requests` for every entity that requires them (`task`, `dynamic`, `eager`, etc)
+    - For `cpu`, the value in `Resources` should define the `cpu` `request`, because CPU allocation can be non-deterministic if `limits > requests`
+    - For `mem`, the value in `Resources` should define the `mem` `request`, because memory allocation can be non-deterministic if `limits > requests`, potentially resulting in OOM errors.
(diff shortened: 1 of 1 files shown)


granthamtaylor:
> This RFC would be a big loss for me. I love having control over reqs/lims directly. 
> 
> The suggestion in the RFC could be a new golden path sure if it's supposed to "smoothen" things, 
> 
> while keeping the classic requests/limits setting for power users and people think they know what they are doing with a warning in code and documentation that it's for advanced users. 

Hi! Thank you for sharing. Perhaps, going forward, we can allow power users to define the request and limits separately in the form of a tuple `("1Gi", "2Gi")`. I think that, ideally, this is preferable over the current pattern of two separate arguments ("requests" and "limits") in that, as a single argument it'll be easy to validate that within `Resources`. Additionally, I think this will help keep new users from hurting themselves. 

Is it just `mem` and `cpu` that you want to have complete control over?

rfc/system/0000-simplify-resources.md
```diff
@@ -0,0 +1,61 @@
...
+
+Defining resources for tasks is error prone and counter-intuitive due to the nuances around `requests` and `limits`.
+
+This RFC discusses an alternative solution to simplify the definition of resources, remove some footguns, and smoothen flytekit altogether.
+
+## 2 Motivation
+
+Some of the footguns around `requests` and `limits` are non-deterministic and not intuitive, thus can escape detection during development but have significant and detrimental impact in production.
+
+For example, the possibility of setting `requests < limits` for `mem` can lead to a pod becoming overloaded. For `cpu`, doing so can lead to CPU availability, for some multi-processing applications, to be non-deterministic, thus resulting in unpredictable performance degradation.
```
katrogan, line 17: for the memory case in particular, when a `requests != limits` the pod is considered burstable (quality of service designation) and more likely to be evicted when the node is under memory pressure

rfc/system/0000-simplify-resources.md
```diff
@@ -0,0 +1,61 @@
...
+
+*What are the main metrics we should be measuring? For example, when interacting with an external system, it might be the external system latency. When adding a new table, how fast would it fill up?*
+
+## 5 Drawbacks
+
+We are losing some flexibility here for power users. However, generally speaking, being able to set both `requests` and `limits` for CPU / memory is a well known footgun.
+
+## 6 Alternatives
+
+Should any resource type (`CPU`) need both `limits` and `requests`, we should allow definition of them via a `tuple` (IE `(400m, 600m)`).
```
katrogan, line 45: +1 to having an escape hatch but prioritizing the common use-case

katrogan:
@cjidboon94 that's really helpful feedback, do you mind expanding in which circumstances you prefer to set request != limits

fg91:
I agree that the UX of the task arguments `@task(requests=..., limits=..., accelerator=..., shared_memory=)` is definitely very Kubernetes-centric:

* separate requests and limits directly translate to pod requests/limits
* accelerator not part of requests/limits because it translates to a node selector/affinity and toleration
* shared memory not part of requests limits because it translates to a volume+mount

I also agree that for ML engineers/Data Scientists who don't know/care about K8s, the UX can feel a bit awkward. Especially the intricacies of requests vs. limits.

We maintain a decorator wrapping the flytekit task decorator which derives the limits from the requests in a way that makes sense for us. Our users never use the limits arg.

At the same time, I feel for platform engineers it absolutely makes sense to have full control over both requests and limits.

To summarize:
From an ML engineer's perspective, I would favor:

* Aggregation of cpu, memory, gpu, shared memory, ephemeral storage, ... into a single struct
* Deriving sane defaults for the limits derived from the specified requests so that I as an ML engineer don't have to think about K8s intricacies.

From a platform engineer's perspective:
* I would like to be able to override the requests/limits derived by default from the specified resources.


---

All of this being said, this is a very breaking change which needs to be considered carefullly.

davidmirror-ops:
@granthamtaylor and team. What is the status of this proposal? Do we still want to pursue it?

davidmirror-ops:
Closing as this is expected to be part of an upcoming update to the SDK

Issue #6254, opened 2025-02-17
[BUG] External command only gets called once


andrei-trandafir:
### Describe the bug

For services that use the `ExternalCommand` auth type with the admin server, the command execution only happens once when the flyteidl client is instantiated.

This behaviour was introduced in issue [#5606](https://github.com/flyteorg/flyte/issues/5606) (PR #5686) when optimisations were made to reduce the number of calls to the admin authentication metadata endpoints. These changes also resulted in putting the creation of the token source provider for the client behind a `Once` synchronization ([link](https://github.com/flyteorg/flyte/pull/5686/files#diff-f86785e90480135a48cec68aa96b032633a62eb515207c57a8db7804a5a9b238R122)).

As the external command gets executed in the token source provider ([link](https://github.com/flyteorg/flyte/blob/release-v1.13/flyteidl/clients/go/admin/token_source_provider.go#L122)), this only gets executed once, resulting in errors if the provided token expires or needs to be refreshed for any reason.

This bug broke our propeller instance when upgrading from version 1.11.0 to 1.14.1 as our external command returns a token that expires afters 45 minutes and results in a `Unauthenticated` error being returned from the admin instance.

### Expected behavior

The external command should be re-executed whenever authentication fails. This ensures that if the external command expires or depends on some external configuration that might change, the command is re-run and authentication can continue.

This can be implemented by creating a custom `externalCommandTokenSource` that runs the command instead of the token source provider. Given that the token would be saved into the token cache, this command wouldn't be called until the token would expire, effectively providing the same experience for deployments that have non-expiring tokens.

This approach keeps the same behaviour for other authentication methods and the same behaviour of caching the authentication metadata that was originally introduced in the PR mentioned earlier.

### Additional context to reproduce

1. Deploy a flyte cluster where the propeller uses `ExternalCommand` authentication where the command returns a token that will expire.
2. Wait until the token expires

### Are you sure this issue hasn't been raised already?

- [x] Yes

### Have you read the Code of Conduct?

- [x] Yes

wild-endeavor:
@andrei-trandafir just to be clear this is what you're referring to correct? https://github.com/flyteorg/flyte/pull/6271/files

wild-endeavor:
Oh sorry... I completely missed that.  I'm not used to not seeing the pr badge.  Mind adding some unit tests to your PR and testing it?

[...]

PR #6255, opened 2025-02-18
Move external command execution from token source provider to token source

andrei-trandafir:
## Tracking issue

#6254 

External command doesn't get called again if authentication starts failing due to the returned token.

Create a new externalCommandTokenSource that runs the external command instead of the token source provider.
This results in:
- The external command is called as part of the `MaterialiseCredentials` function instead of the `getTokenSourceAndMetadata` function that is synchronized (in the `flyteidl/go/client/admin/auth_interceptor.go`).
- If the token in the token cache results in an `Unauthenticated` response from the flyte admin, the external command is called again.

## How was this patch tested?

- Deployed the change to a flytepropeller instance using `ExternalCommand` auth that returns an expiring token
- Waited for the expiry of the token and checked whether authentication still works

## Related PRs

The issue resulted from the following PR: #5686

### DIFF
--- a/flyteidl/clients/go/admin/token_source_provider.go
+++ b/flyteidl/clients/go/admin/token_source_provider.go
@@ -120,15 +120,24 @@ func NewExternalTokenSourceProvider(command []string) (TokenSourceProvider, erro
 }
 
 func (e ExternalTokenSourceProvider) GetTokenSource(ctx context.Context) (oauth2.TokenSource, error) {
-	output, err := externalprocess.Execute(e.command)
+	return &externalCommandTokenSource{
+		command: e.command,
+	}, nil
+}
+
+type externalCommandTokenSource struct {
+	command []string
+}
+
+func (s *externalCommandTokenSource) Token() (*oauth2.Token, error) {
+	output, err := externalprocess.Execute(s.command)
 	if err != nil {
 		return nil, err
 	}
-
-	return oauth2.StaticTokenSource(&oauth2.Token{
+	return &oauth2.Token{
 		AccessToken: strings.Trim(string(output), "\t \n"),
 		TokenType:   "bearer",
-	}), nil
+	}, nil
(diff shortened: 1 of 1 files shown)


flyteidl/clients/go/admin/token_source_provider.go
```diff
@@ -120,15 +120,24 @@ func NewExternalTokenSourceProvider(command []string) (TokenSourceProvider, erro
...
 func (e ExternalTokenSourceProvider) GetTokenSource(ctx context.Context) (oauth2.TokenSource, error) {
-	output, err := externalprocess.Execute(e.command)
+	return &externalCommandTokenSource{
+		command: e.command,
+	}, nil
+}
+
+type externalCommandTokenSource struct {
+	command []string
+}
```
eapolinario, line 130: Do we really need to define `externalCommandTokenSource`? We could implement `Token` in `ExternalTokenSourceProvider`, right?

eapolinario (commented):
Looks reasonable, I have one small comment. Also, can you add a test to auth_interceptor_test.go?

flyteidl/clients/go/admin/token_source_provider.go
andrei-trandafir: We could implement `Token()` in `ExternalTokenSourceProvider`, but I feel it would potentially feel a bit confusing to follow.
I don't personally mind either way.

wild-endeavor:
@andrei-trandafir would you mind adding a unit test?  is that simple to do?

andrei-trandafir:
@wild-endeavor 
I've added a unit test to the token source for the external command.
I've tried adding a test to the auth_interceptor.go checking if in the case that we're using ExternalCommand, the external command is called during the `MaterializeCredentials`, but I haven't managed to get it working. Not sure if it's possible

PR #6267, opened 2025-02-22
[WPI] Feat: add batch delete execution_phase api

taieeuu:
### Labels


flyteadmin/pkg/rpc/adminservice/execution.go
```diff
@@ -152,3 +152,17 @@ func (m *AdminService) TerminateExecution(
 	m.Metrics.executionEndpointMetrics.terminate.Success()
 	return response, nil
 }
+
+func (m *AdminService) DeleteExecutionPhase(
+	ctx context.Context, request *admin.ExecutionPhaseDeleteRequest) (*admin.ExecutionPhaseDeleteResponse, error) {
+	var response *admin.ExecutionPhaseDeleteResponse
+	var err error
+	m.Metrics.executionEndpointMetrics.terminate.Time(func() {
+		response, err = m.ExecutionManager.DeleteExecutionPhase(ctx, request)
```
Future-Outlier, line 161: do you know where to see the metrics?
is it stored in database or a metric server something like that?

flyteidl/protos/flyteidl/admin/execution.proto
```diff
@@ -430,3 +430,13 @@ message WorkflowExecutionGetMetricsResponse {
...
+
+message ExecutionPhaseDeleteRequest {
+    core.WorkflowExecutionIdentifier id = 1;
+    core.WorkflowExecution.Phase phase = 2;
+}
+  
+message ExecutionPhaseDeleteResponse {
+    string message = 1;
+}
+
```
Future-Outlier, line 442:
```suggestion

```

flyteadmin/pkg/rpc/adminservice/execution.go
taieeuu: The data is stored in Prometheus, but 'flyte demo start' does not launch Prometheus.

flyteidl/protos/flyteidl/admin/execution.proto
```diff
@@ -430,3 +430,13 @@ message WorkflowExecutionGetMetricsResponse {
     // hierarchical structure using Flyte entity references.
     core.Span span = 1;
 }
+
+message ExecutionPhaseDeleteRequest {
```
katrogan, line 434: rather than requiring this deletion to be based on phase attributes only (what if one day we want to support timestamp filters instead?) can we make this more generic and re-use the ResourceListRequest (https://github.com/flyteorg/flyte/blob/a2331bd69b9f9ac38ce4f292c11d2374d93960db/flyteidl/protos/flyteidl/admin/common.proto#L203-L225) which includes out of the box support for filtering by phase, timestamp and other fields?

katrogan: something like


```
rpc DeleteExecutions (flyteidl.admin. ResourceListRequest) returns (flyteidl.admin.ExecutionDeleteResponse) {
		option (google.api.http) = {
      delete: "/api/v1/delete_execution/{id.project}/{id.domain}"
		};
		option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
		  description: "Delete all executions matching request filters"
		};
	};
}
```

katrogan:
before we proceed with deleting items from the db can we maybe write up an issue to address
1. what we're trying to solve (mass terminate executions)
2. why the [current api](https://github.com/flyteorg/flyte/blob/master/flyteidl/protos/flyteidl/service/admin.proto#L313-L322) is insufficient
3. and if the current api works, how we can use the remote client in flytekit SDK to allow us to batch terminate executions and accepts a list of input parameters (created at window, current phase filters, etc)

I think as it stands now this PR could orphan executions unintentionally if we're not careful or make it really easy to delete from the DB which is a departure from how flyteadmin currently records entities. Flyteadmin allows us to archive or tombstone various entities but delete is a scarier tool

Issue #3696, opened 2023-05-18
[Core feature] Add execution id as an available log parameter for log links UI


btang-stripe:
### Motivation: Why do you think this is important?

We want to integrate other products with Flyte and be able to link to the logs. Specifically we'd like to link to wandb which requires an identifier created at runtime. Using the Flyte execution id as a wand group id would allow us to create the log link

### Goal: What should the final outcome look like, ideally?

the execution id of the workflow should be available as a [log parameter](https://docs.flyte.org/projects/cookbook/en/latest/auto/deployment/configure_logging_links.html)

### Describe alternatives you've considered 

Providing the ability to pass arbitrary strings as log parameters would be ideal, then we could pull the actual wandb run id instead. This, however, would not be as straight forward as passing the Flyte execution id.

###  Propose: Link/Inline OR Additional context 

Currently the plugin machinery does not have the [execution id](https://github.com/flyteorg/flyteplugins/blob/9a2bbbaf2f3ac9e38222ba7755ac4bda1ac2fdb7/go/tasks/pluginmachinery/tasklog/template.go#LL91C28-L91C39). Talking with @kumare3 this should be updated [here](https://github.com/flyteorg/flyteplugins/blob/9a2bbbaf2f3ac9e38222ba7755ac4bda1ac2fdb7/go/tasks/pluginmachinery/core/exec_metadata.go#L36) and [here](https://github.com/flyteorg/flyteplugins/blob/9a2bbbaf2f3ac9e38222ba7755ac4bda1ac2fdb7/go/tasks/pluginmachinery/tasklog/plugin.go#L7)

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

kumare3:
Thank you for opening the issue. I really like the idea of adding execution id to the log parameters so that it can be used to power things like weights and biases directly from Flyte UI. 
So this would be used for groupID, what about the run-id / experiment id itself?

hamersaw:
@fg91 if I recall you discussed some deeper integration with wandb - mind summarizing here?

fg91:
> @fg91 if I recall you discussed some deeper integration with wandb - mind summarizing here?

I wouldn't make it specific to wandb. I described our use case [here](https://github.com/flyteorg/flyte/discussions/3666) in the rfc incubator:

> We would like to be able to add links to experiment tracking servers like Mlflow, Wandb, ...
[...]

PR #6285, opened 2025-02-27
Doc: Document usage of dynamic log links

fg91:

Allowing to configure custom log links for specific tasks has been a widely requested feature, see e.g. [here](https://github.com/flyteorg/flyte/discussions/3666) or [here](https://github.com/flyteorg/flyte/issues/3696#issuecomment-1557992297).

So-called "dynamic log links" which were introduced for this purpose are now used for instance by the wandb or vscode plugins.

How these dynamic log links can be used is not yet documented in the [Configuring logging links in the UI](https://docs.flyte.org/en/latest/user_guide/productionizing/configuring_logging_links_in_the_ui.html) page, this PR adds an explanation.

### DIFF
--- a/docs/user_guide/productionizing/configuring_logging_links_in_the_ui.md
+++ b/docs/user_guide/productionizing/configuring_logging_links_in_the_ui.md
@@ -104,6 +104,95 @@ task_logs:
             - "https://..."
 ```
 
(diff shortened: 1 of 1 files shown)


eapolinario (approved):
Thank you so much.

PR #6318, opened 2025-03-08
Fix: fastcache should not cache lookup on failed node

fg91:

See [here](https://github.com/flyteorg/flyte/pull/4524#issuecomment-2633583071) for bug description.

> Looking at [this line](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/nodes/executor.go#L980-L980) (the check you linked), the easy fix is to add && p.GetPhase() != handler.EPhaseFailed as you suggest. I'm wondering if we can consolidate these to p.GetPhase() == EPhaseQueued"? because the preExecute` function returns Queued whenever the inputs correctly resolve. Anything else would not require a cache lookup. ([source](https://github.com/flyteorg/flyte/pull/4524#issuecomment-2638104081))

@hamersaw let's go for the `p.GetPhase() != handler.EPhaseFailed` because I'm convinced this is a correct fix and I'm not 100% sure consolidating to ` p.GetPhase() == EPhaseQueued` is.

## Related PRs

[](https://github.com/flyteorg/flyte/pull/4524)

### DIFF
--- a/flytepropeller/pkg/controller/nodes/executor.go
+++ b/flytepropeller/pkg/controller/nodes/executor.go
@@ -977,7 +977,7 @@ func (c *nodeExecutor) handleNotYetStartedNode(ctx context.Context, dag executor
 	}
 
 	var cacheStatus *catalog.Status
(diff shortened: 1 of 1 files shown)


hamersaw (approved):
Sounds good to me. Thanks for the PR!

Issue #6238, opened 2025-02-10
[Core feature] Allow setting metadata (labels&annotations) individually on all K8s task types


fg91:
### Motivation: Why do you think this is important?

Today, as a Flyte user, I have the following options to set labels/annotations on the pods/CRD objects of K8s tasks in a flyte workflow execution:

1. Set via pod template:

    ```py
    def task(
        pod_template=PodTemplate(annotations=..., labels=...)
    )
    ```

    This sets labels/annotations on the pods of individual tasks.
    For distributed tasks (like pytorch, ray, ...) this sets the metadata not on the CRD object but its pod template spec.

2. Set via `pyflyte run --labels ... --annotations ...`

    This applies the metadata to *all* K8s objects in a flyte workflow execution, including task pods *and* task CRD objects. However, this mechanism doesn't work on an individual task level.

As a Flyte user, I would like to be able to specify specific labels/annotations for individual k8s task CRD objects like pytorch jobs, ray job, ... (the same way I already can today for pods via the pod template):

```py
@task(
    task_config=PyTorch(
        num_workers=...,
        ...
        # Proposed addition:
        metadata=ObjectMeta(
            annotations={"kueue.x-k8s.io/queue-name": "queue-name"},
            labels={...}
        )
    )
)
```

I propose to use the same syntax/flyteidl type for all K8s (non-pod) plugins like `Elastic`, `TfJob`, `MpiJob`, `RayJobConfig`, ...

---

In my concrete case, I would like to have this feature in order to leverage Kueue to gang schedule worker pods for distributed pytorch training tasks (e.g. as documented [here](https://cloud.google.com/kubernetes-engine/docs/how-to/provisioningrequest)).
This requires setting a queue name annotation on the underlying PytorchJob CRD object.

There have been previous asks from the community to enable such a feature/integration:

* Attempts to integrate [Yunikorn and Kueue more deeply into flytepropeller](https://github.com/flyteorg/flyte/pull/5915) which weren't accepted though.

    In contrast, the feature I propose allows users to choose to use Kueue while it isn't a Flyte-Kueue integration. Instead it is a very general feature that could be used for any other integration as well that makes use of annotations/labels to select workloads.

[...]

PR #6319, opened 2025-03-08
Feat: Add JobObjectMeta to plugins flyteidl

fg91:
## Tracking issue

Related to #6238

### Labels


flyteidl/protos/flyteidl/plugins/common.proto
```diff
@@ -25,3 +26,10 @@ message CommonReplicaSpec {
   // RestartPolicy determines whether pods will be restarted when they exit
   RestartPolicy restart_policy = 4;
 }
+
+// Object metadata applied to the CRD object underlying a task execution
+message JobObjectMeta {
```
kumare3, line 31: Why is this a new object? We have labels and annotations right?

Why not part of task template

fg91: You are right, I should reuse [`message K8sObjectMetadata {`](https://github.com/flyteorg/flyte/blob/a2331bd69b9f9ac38ce4f292c11d2374d93960db/flyteidl/protos/flyteidl/core/tasks.proto#L354) :thinking:

fg91: > Why not part of task template

For normal python function tasks you can set custom pod labels/annotations via the pod template arg of the task decorator.

As described in https://github.com/flyteorg/flyte/issues/6238 I'm looking to add a way to allow setting labels/annotations to CRD objects of plugins like kubeflow, dask, spark, ray, ... (because I want to use Kueue which requires an annotation on the job):

```
@task(
    task_config=PyTorch(
        num_workers=...,
        ...
        # Proposed addition:
        metadata=ObjectMeta(
            annotations={"kueue.x-k8s.io/queue-name": "queue-name"},
            labels={...}
        )
    )
)
```

Pod template doesn't work because this sets it on the worker pods, not the CRD object.

Does it sound reasonable to you to add the existing `K8sObjectMetadata` to the distributed job proto messages like `DistributedPyTorchTrainingTask` (see [here](https://github.com/flyteorg/flyte/blob/a2331bd69b9f9ac38ce4f292c11d2374d93960db/flyteidl/protos/flyteidl/plugins/kubeflow/pytorch.proto#L21))?

fg91:
Closing, will be implemented differently, `@task(labels=..., annpotations=...)`, see original issue.

PR #6336, opened 2025-03-12
fix: really assign string scp claim

vlada-dudr:
because before this change string formatted claims were ignored.

## How was this patch tested?

I did run it against keycloak, which was forced to send `scp` claim as string. I.e. `"scp": "all"`

### Labels

- **fixed**: String formatted scp claim now works 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR fixes a critical bug in the authentication module where string formatted scp claims were incorrectly handled. The fix corrects assignment logic in the claims verifier and updates test cases to verify proper scope value processing. The implementation was verified against keycloak to ensure robustness.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/flyteadmin/auth/authzserver/claims_verifier_test.go
+++ b/flyteadmin/auth/authzserver/claims_verifier_test.go
@@ -79,12 +79,12 @@ func Test_verifyClaims(t *testing.T) {
 		identityCtx, err := verifyClaims(sets.NewString("https://myserver", "https://myserver2"),
 			map[string]interface{}{
 				"aud": []string{"https://myserver"},
-				"scp": "all",
+				"scp": "my-scope",
 			})
 
 		assert.NoError(t, err)
 		assert.Equal(t, "https://myserver", identityCtx.Audience())
-		assert.Equal(t, sets.NewString("all"), identityCtx.Scopes())
+		assert.Equal(t, sets.NewString("my-scope"), identityCtx.Scopes())
 	})
 	t.Run("unknown scope", func(t *testing.T) {
 		identityCtx, err := verifyClaims(sets.NewString("https://myserver", "https://myserver2"),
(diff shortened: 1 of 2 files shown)


eapolinario (commented):
Good catch. Can you fix [unit test](https://github.com/flyteorg/flyte/blob/0fbd3467d9c42a80f407483ff453328429a1cacd/flyteadmin/auth/authzserver/claims_verifier_test.go#L80-L83)?

vlada-dudr:
Like this?

flyteadmin/auth/authzserver/claims_verifier_test.go
```diff
@@ -78,6 +78,7 @@ func Test_verifyClaims(t *testing.T) {
 	t.Run("String scope", func(t *testing.T) {
 		identityCtx, err := verifyClaims(sets.NewString("https://myserver", "https://myserver2"),
 			map[string]interface{}{
+				"client_id": "my-client",
 				"aud": []string{"https://myserver"},
 				"scp": "all",
```
Sovietaced, line 83: Maybe use something other than the default `all` scope so its clear that it was not injected.

Sovietaced:
Also make sure to run `make lint-fix` under /flyteadmin to resolve lint issues

vlada-dudr:
@Sovietaced I did `make lint-fix`, but I don't see any issues I introduced - output is same as if I run it without my commit. I also rebased to current master.

Sovietaced:
> @Sovietaced I did `make lint-fix`, but I don't see any issues I introduced - output is same as if I run it without my commit. I also rebased to current master.

I think the gci linter is upset about the order of keys in the map that you modified.

flyteadmin/auth/authzserver/claims_verifier_test.go
vlada-dudr: Hi, so I removed `client_id` and changed `scp` to something else.

Sovietaced (approved):

eapolinario (approved):

PR #6262, opened 2025-02-21
Fix issues with injected pod template init containers

Sovietaced:
## Tracking issue
 Closes #6229 

## Why are the changes needed?
We discovered a bug where init containers from pod templates were being injected into tasks which had no init containers. This is because of an out of order merge of the task pod spec onto the pod template spec. 

### Labels

Please add one or more of the following labels to categorize your PR:
- **added**: For new features.
- **changed**: For changes in existing functionality.
- **deprecated**: For soon-to-be-removed features.
- **removed**: For features being removed.
- **fixed**: For any bug fixed.
- **security**: In case of vulnerabilities

This is important to improve the readability of release notes.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

PR #6363, opened 2025-03-22
[FIX] Error when using different primary container name in driver and executor

machichima:

There's a bug in Spark after this PR: https://github.com/flyteorg/flyte/pull/6262. For example, When setting primary container name in driver pod named "driver-primary", the error of `invalid TaskSpecification, container [driver-primary] not defined` is raised.

image

Use `MergeBasePodSpecOntoTemplate` that enables setting `primaryContainerName` and `primaryInitContainerName`.

## How was this patch tested?

Unittest is added to check if setting primary container name that's not  "primary" in driver and executor, spark can perform as expected.

### Labels

### DIFF
--- a/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -746,20 +746,20 @@ func MergeBasePodSpecOntoTemplate(templatePodSpec *v1.PodSpec, basePodSpec *v1.P
 					return nil, err
 				}
 			}
-		}
+		} else {
+			// Check for any name matching template containers
+			for _, templateContainer := range templatePodSpec.Containers {
+				if templateContainer.Name != container.Name {
+					continue
+				}
 
-		// Check for any name matching template containers
-		for _, templateContainer := range templatePodSpec.Containers {
-			if templateContainer.Name != container.Name {
-				continue
-			}
-
-			if mergedContainer == nil {
-				mergedContainer = &templateContainer
-			} else {
-				err := mergo.Merge(mergedContainer, templateContainer, mergo.WithOverride, mergo.WithAppendSlice)
-				if err != nil {
-					return nil, err
+				if mergedContainer == nil {
+					mergedContainer = &templateContainer
+				} else {
+					err := mergo.Merge(mergedContainer, templateContainer, mergo.WithOverride, mergo.WithAppendSlice)
+					if err != nil {
+						return nil, err
+					}
 				}
 			}
 		}
--- a/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -35,7 +35,7 @@ const unsignedSIGKILL = 247
 const defaultContainerTemplateName = "default"
 const defaultInitContainerTemplateName = "default-init"
 const primaryContainerTemplateName = "primary"
-const primaryInitContainerTemplateName = "primary-init"
+const PrimaryInitContainerTemplateName = "primary-init"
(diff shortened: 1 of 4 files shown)


flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -746,20 +746,20 @@ func MergeBasePodSpecOntoTemplate(templatePodSpec *v1.PodSpec, basePodSpec *v1.P
...
 				}
 			}
-		}
-
-		// Check for any name matching template containers
-		for _, templateContainer := range templatePodSpec.Containers {
-			if templateContainer.Name != container.Name {
-				continue
-			}
+		} else {
```
Sovietaced, line 749: This change seems okay to me. Can we have the same behavior for init containers to keep it consistent?

flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -35,7 +35,7 @@ const unsignedSIGKILL = 247
 const defaultContainerTemplateName = "default"
 const defaultInitContainerTemplateName = "default-init"
 const primaryContainerTemplateName = "primary"
-const primaryInitContainerTemplateName = "primary-init"
+const PrimaryInitContainerTemplateName = "primary-init"
```
Sovietaced, line 38: These are reserved container names, I don't think they should be exported and used in the spark plugin logic

Sovietaced:
> There's a bug in Spark after this PR: #6262. For example, When setting primary container name in driver pod named "driver-primary", the error of `invalid TaskSpecification, container [driver-primary] not defined` is raised.

I'm not sure I understand the issue. The way that the overlay is intended to work is that it matches on container names and does not honor any magic container names. 

Are you unable to specify the custom driver and selector pod specs with container names that match the primary driver or executor containers? If those container names are not deterministic and you wanted to support some magic values like `primary-driver` or `primary-executor` you could just handle that in the plugin itself and dynamically rename the custom pod specs and use the overlay method. Or the plugin could be modified so that the containers for the driver and executor are deterministic like we do in the Ray plugin (ray-head, ray-worker). 

What are you actually trying to do with the custom pod specs?

Sovietaced:
> This PR is actually fixing that if user set primary_container_name different than "primary", than we should then use the name they set as the primary container name.

Why do you need to set the primary container name in the pod template at all? If the goal is to customize the driver or executor pods all you need to do is pass the pod spec with a container name that matches the target container name.

machichima:
> Why do you need to set the primary container name in the pod template at all? If the goal is to customize the driver or executor pods all you need to do is pass the pod spec with a container name that matches the target container name.

Yes, we don’t necessarily need to set the `primary_container_name` here. However, since Spark already uses a `PodTemplate` for specifying the `driver_pod` and `executor_pod`, which provides a field for users to specify the `primary_container_name`, I think we could allow users to set it. If they don’t specify it, it would simply default to 'primary', and user can pass pod spec with a container name 'primary'.

It’s more of an additional feature, not strictly necessary, but an option for flexibility.

Sovietaced:
> Yes, we don’t necessarily need to set the `primary_container_name` here. However, since Spark already uses a `PodTemplate` for specifying the `driver_pod` and `executor_pod`, which provides a field for users to specify the `primary_container_name`, I think we could allow users to set it. If they don’t specify it, it would simply default to 'primary', and user can pass pod spec with a container name 'primary'.

That makes me think, what is the value in being able to specify the `primary_container_name` at all. I'm not sure it's relevant in this use case. 

But still I'm curious what is your end goal with being able to configure it? What do you expect it to do that simply configuring a container in the pod spec to match the existing driver or executor container name won't do?

Sovietaced:
What do you think @pingsutw

PR #6401, opened 2025-04-07
Allow setting `s3.endpoint` in flyte-core

dwo:

It is not currently possible to use an S3-compatible service in `flyte-core`.

Allowing setting the endpoint of an S3-compatible service in the `flyte-core` Helm chart as it is in `flyte-binary`.

## How was this patch tested?

I deployed this with the endpoint set, and confirmed that the value made it into the appropriate configmaps.

### Labels

- **added**: Ability to configure flyte-core Helm chart S3-compatible endpoint

## Related PRs

- N/A

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR adds configurable S3 endpoint support to the flyte-core Helm chart and updates multiple Kubernetes manifest files by replacing old secret values with new secure tokens across YAML files. The implementation includes conditional endpoint settings, updated documentation, and improved secret management with updated haSharedSecret and checksum values.
<br>
<br>
<br>
<br>
</div>


charts/flyte-core/values.yaml
```diff
@@ -1,3 +1,4 @@
+---
```
dwo, line 1:
```suggestion
```

Sovietaced:
You may need to run `make helm` and commit the generated changes.

Sovietaced:
Our company uses an S3 compatible backend VAST and we are able to set these values but I realized that we pass them in through a secret. So I'll add an issue to add support for that

Sovietaced:
I think this is reasonable but I'll let this sit for the maintainers to take a look.

PR #6419, opened 2025-04-17
Add support for configuring storage from a secret to avoid exposing secrets

### LINKED ISSUE #6394: [Core feature] `authType=accesskey` has no means to provide the secret securely

flyte-core: If you need to use `authType=accesskey` for `storage`, the only option is to pass the secret part of the access key in your Helm values (`storage.s3.secretKey`) and to have it inlined into various `ConfigMap`s.

This stands in contrast to how the database password is handled with `common.databaseSecret` where you can have your password mounted from the specified secret to `/etc/db/pass.txt` or similar.

Ideally the secret part of storage access keys should not be passed through Helm values or appear in a `ConfigMap`, it should be configurable via a `Secret`.

I tried to check if it is possible to specify `initContainers` for the various components of `flyte-core`, but this is not the case. This could allow one to get a secret into the environment variables of an init container and then substitute values in the mounted config.

_No response_

- [x] Yes

- [x] Yes

Sovietaced:
## Tracking issue
Closes #6394

It is not possible to configure an S3 storage backend without leaking the secret key in the `values.yaml` file. Although this requires folks to include non-secrets in the k8s secret it is flexible enough  to support all the types of storage (s3, gcs, azure, etc). 

Adds an option to load in the storage configuration YAML via a secret. It loads the secret onto the filesystem alongside the base configuration YAML with a projected volume. 

ie. 

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: flyte-storage
data:
  storage.yaml: ....
```

## How was this patch tested?

Tested in our production environment.

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR implements support for configuring storage with secrets to enhance credential security. Kubernetes templates were updated to use projected volumes that conditionally include secret data instead of direct configMap usage. Docker manifest files were revised with updated secret values and checksums for better security compliance. These changes have been validated in production environments.
<br>
<br>
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 3
</div>


dwo (approved):
Implementation LGTM from end-user perspective 👍

pingsutw (approved):
lgtm, thanks. Mind rebasing the PR

pingsutw (approved):

PR #6421, opened 2025-04-19
Feat: Apply labels and annotations set via task decorator to pods and CR objects

### LINKED ISSUE #6238: [Core feature] Allow setting metadata (labels&annotations) individually on all K8s task types

fg91:
## Tracking issue
Closes #6238

Today, as a Flyte user, I have the following options to set labels/annotations on the pods/CRD objects of K8s tasks in a flyte workflow execution:

1. Set via pod template:

    ```py
    def task(
        pod_template=PodTemplate(annotations=..., labels=...)
    )
    ```

    This sets labels/annotations on the pods of individual tasks.
    For distributed tasks (like pytorch, ray, ...) this sets the metadata not on the CRD object but its pod template spec.

2. Set via `pyflyte run --labels ... --annotations ...`

    This applies the metadata to *all* K8s objects in a flyte workflow execution, including task pods *and* task CRD objects. However, this mechanism doesn't work on an individual task level.

As a Flyte user, I would like to be able to specify specific labels/annotations for individual k8s task CRD objects like pytorch jobs, ray job, ... (the same way I already can today for pods via the pod template). For this aim, the following arguments are added to the task decorator:

```py
@task(
    labels=...,
    annotations=...,
)
def test_task():
    ...
```

These labels and annotations are added to the object metadata of the object executing the task (Pod or CRD object).

This PR adds a `metadata` field of type `K8sObjectMetadata` to the `TaskMetadata` proto message and applies this additional metadata in the plugin manager's `addObjectMetadata` function.

## How was this patch tested?

* Executed flyteadmin and flytpropeller with these changes and the corresponding changes in flytekit.
* Adapted unit tests

## Related PRs

https://github.com/flyteorg/flytekit/pull/3243
 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request adds a new metadata field to the TaskMetadata proto, enabling users to apply custom labels and annotations to Kubernetes objects. The enhancement spans multiple language implementations including Go, TypeScript, Python, and Rust, providing more granular control over metadata for both pods and custom resources.
<br>
<br>
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 5
</div>


flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager_test.go
```diff
@@ -1008,6 +1011,106 @@ func TestPluginManager_AddObjectMetadata(t *testing.T) {
...
+			"task_template_annotation": "task_template_annotation_val",
+		}, o.GetAnnotations())
+		assert.Equal(t, map[string]string{
+			"task_template_label": "task_template_label_val",
+			"l1":                  "lv1",
+		}, o.GetLabels())
+	})
+
+	t.Run("Task template K8s metadata overwrites object metadata", func(t *testing.T) {
+		o := &v1.Pod{
```
fg91, line 1046: Labels/Annotations set via `@task(pod_template=PodTemplate(...))` are overwritten by `@task(labels=..., annotations=...)`.

flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager_test.go
```diff
@@ -1008,6 +1011,106 @@ func TestPluginManager_AddObjectMetadata(t *testing.T) {
...
+			"task_template_annotation": "task_template_annotation_val",
+		}, o.GetAnnotations())
+		assert.Equal(t, map[string]string{
+			"task_template_label": "task_template_label_val",
+			"l1":                  "lv1",
+		}, o.GetLabels())
+	})
+
+	t.Run("Task template K8s metadata overwritten by task execution metadata", func(t *testing.T) {
+		o := &v1.Pod{}
```
fg91, line 1086: Labels/annotations set via `@task(labels=..., ...)` are overwritten by `pyflyte run --labels ... --annotations ...`.

Sovietaced (approved):
This seems reasonable to me. Labels/annotations set at the launch plan would apply to all the CRs since they get inherited by the workflow CR. So this should let you deviate on a per task basis.

PR #6433, opened 2025-04-30
flyte-core: Default propeller `rawoutput-prefix` to use `storage.bucketName`

dwo:

We only configured `storage.bucketName` in our `flyte-core` deployment and found our workflows throwing permissions errors when passing `pandas.DataFrame` objects between tasks. We set `S3FS_LOGGING_LEVEL=DEBUG` and observed `my-s3-bucket` in the errors.

It would have saved us from having to figure this out if the propeller `rawoutput-prefix` was re-using `storage.bucketName`: they are set to the same value in the defaults so this change is a no-op to the default, while still allowing the flexibility to override where needed.

`configmap.core.propeller.rawoutput-prefix` is modified to re-use `{{ .Values.storage.bucketName }}` by default.

## How was this patch tested?

I rendered the templates locally to confirm that the values changed as I expected:
```
helm template --set storage.bucketName=foo . | grep rawoutput-prefix
      rawoutput-prefix: s3://foo/
```

### Labels

- **changed**: flyte and flyte-core chart's `rawoutput-prefix` in propeller config now re-uses the value of `storage.bucketName` by default

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR updates the propeller configuration in flyte charts to use dynamic references to storage.bucketName for rawoutput-prefix, resolving potential permission issues. It also updates docker manifest files with new secret and checksum values to maintain consistency across environments.
<br>
<br>
<br>
<br>
</div>


Sovietaced (approved):
Took me a little time to understand how this works. There is a `tpl` function called when this is loaded into a config map, which allows referencing other values.

dwo:
@Sovietaced ah thanks for confirming: I saw that some of the other charts were doing something similar and didn't question it as much as I probably should have 😅

PR #6436, opened 2025-05-01
Only enqueue workflows if flytepropeller is leader

### LINKED ISSUE #6435: [BUG] Flyte Propeller Followers/Replicas Enqueue Workflows And Fill Up Queue
### Describe the bug

Flyte Propeller Followers/Replicas that have not won the leader election enqueue workflows but do not process them. This leads to a controllers memory leak and likely fills the queue with many stale workflows for if/when the follower/replica does become the leader. This current behavior isn't very confidence inspiring for a hot standby. 

Below is an image of the queue depth increasing on the follower. 

Image

### Expected behavior

I would expect that the follower/replica wouldn't enqueue workflows or would process them but throw them away.

_No response_

### Screenshots

_No response_

- [x] Yes

- [x] Yes

Sovietaced:
## Tracking issue
Closes #6435 

Without these changes followers/replicas in highly available deployments will enqueue workflows until they saturate the work queue. Once the work queue is saturated these workflows may become very stale. Once a follower does become a leader it will need to churn through these stale workflows that might not even exist anymore.

Added a thread safe bool for whether the propeller is leader and check it before enqueueing workflows. 

## How was this patch tested?
Not yet but will test in our development environment.

 <div id='description'>
<h3>Summary by Bito</h3>
This PR implements leader-only workflow enqueuing in HA flytepropeller deployments using an atomic boolean flag, preventing followers from adding stale workflows to the queue. It enhances system reliability and consistency during leadership transitions, and refines workflow handling. The changes also improve logging of workflow phases for failed executions to facilitate better monitoring and debugging.
<br>
<br>
<br>
<br>
</div>

### DIFF
--- a/flytepropeller/pkg/controller/controller.go
+++ b/flytepropeller/pkg/controller/controller.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"runtime/pprof"
+	"sync/atomic"
 	"time"
 
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
@@ -93,6 +94,7 @@ type Controller struct {
 	leaderElector  *leaderelection.LeaderElector
 	levelMonitor   *ResourceLevelMonitor
 	executionStats *workflowstore.ExecutionStatsMonitor
+	leader         atomic.Bool
 }
 
 // Run either as a leader -if configured- or as a standalone process.
@@ -133,6 +135,7 @@ func (c *Controller) run(ctx context.Context) error {
 
 // Called from leader elector -if configured- to start running as the leader.
 func (c *Controller) onStartedLeading(ctx context.Context) {
+	c.leader.Store(true)
 	backgroundCtx, cancelNow := context.WithCancel(ctx)
 	logger.Infof(ctx, "Acquired leader lease.")
 	go func() {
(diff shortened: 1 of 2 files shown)


Sovietaced:
The end2end test is failing but I've tested this in a development environment without issues hmm.

flytepropeller/pkg/controller/controller.go
```diff
@@ -157,6 +161,10 @@ func (c *Controller) enqueueFlyteWorkflow(obj interface{}) {
 		return
 	}
 	key := wf.GetK8sWorkflowID()
+	if !c.leader.Load() {
```
machichima, line 164: If the leader elector is not configured, it seems like `c.leader` will never be set to `true`, and this part will ignore all workflows and never enqueue them

flytepropeller/pkg/controller/controller.go
```diff
@@ -142,6 +145,7 @@ func (c *Controller) onStartedLeading(ctx context.Context) {
 	}()
 
 	<-backgroundCtx.Done()
+
 	logger.Infof(ctx, "Lost leader lease.")
```
machichima, line 149: Do we need to set `c.leader` to `False` here to prevent it from keep enqueuing workflow while it is not leader anymore?

Sovietaced:
@machichima this one wasn't ready for review yet since I was trying to work out the build failure. I'll circle back later.

PR #6448, opened 2025-05-12
Set MaxConcurrentStreams for flyteadmin grpc server

pmahindrakar-oss:
## Tracking issue
N/A (internal stability enhancement)

Under high load, FlyteAdmin currently accepts all gRPC requests without constraint due to the lack of a MaxConcurrentStreams configuration. This can lead to unbounded in-flight requests, resulting in excessive memory usage, DB connection exhaustion, increased latency, and eventual cancellations or INTERNAL errors.

Adding a MaxConcurrentStreams limit allows the gRPC server to enforce backpressure at the transport layer, ensuring that only a bounded number of requests are actively processed at a time. This helps stabilize system behavior under bursty or sustained traffic conditions, particularly for high-volume APIs coming from propeller

Introduced MaxConcurrentStreams as a configurable field in the FlyteAdmin gRPC server configuration.

Updated the grpc.NewServer initialization to apply this limit when set.

Default behavior remains unchanged (unbounded) unless explicitly configured.

## How was this patch tested?
Trivial change as it keeps the old behavior if unset.
Unit test for config changes

### DIFF
--- a/flyteadmin/pkg/config/config.go
+++ b/flyteadmin/pkg/config/config.go
@@ -51,6 +51,7 @@ type GrpcConfig struct {
 	ServerReflection         bool `json:"serverReflection" pflag:",Enable GRPC Server Reflection"`
 	MaxMessageSizeBytes      int  `json:"maxMessageSizeBytes" pflag:",The max size in bytes for incoming gRPC messages"`
 	EnableGrpcLatencyMetrics bool `json:"enableGrpcLatencyMetrics" pflag:",Enable grpc latency metrics. Note Histograms metrics can be expensive on Prometheus servers."`
+	MaxConcurrentStreams     int  `json:"maxConcurrentStreams" pflag:",Limit on the number of concurrent streams to each ServerTransport."`
 }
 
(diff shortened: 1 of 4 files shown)


Sovietaced (approved):

pmahindrakar-oss:
Thanks @Sovietaced for the review
You're right.MaxConcurrentStreams applies per HTTP/2 connection, so its effectiveness as a backpressure mechanism is scoped to each individual client connection.
In practice, this means:
If FlyteAdmin is receiving requests from many Propeller replicas or through multiple load-balanced client connections , the global concurrency is not bounded and it scales with the number of client connections.
So it’s not a global rate limiter or concurrency cap, but it’s still useful in preventing per-connection overload, especially from bursty or misbehaving clients.
That said, this change is still beneficial because:
It establishes per-connection protection, which reduces the risk of individual clients (or load balancer streams)overwhelming the server.
We can add more comprehensive rate control via interceptors or other means.

Sovietaced (approved):

PR #4655, opened 2023-12-30
[Feature] Datacatalog cache deletion

pvditt:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/2867_

## Why are the changes needed?

Flyte doesn't support the deletion of cached task executions. Adding cache eviction/cleanup gives Flyte users more control over their data helping with things such as GDPR compliance + reducing their cache size.

Opted to support just single task executions as invalidating an entire workflow's cache can run into timeout issues going down the nesting of nodes.

## What changes were proposed in this pull request?

- add flyteadmin endpoint to support the deletion of single task executions
- add catalog endpoint to support the deletion of single task executions

## How was this patch tested?

1) unit tests

2) Run workflow that caches task executions
```
grpcurl -plaintext -d '{"task_execution_id": {"node_execution_id": {"execution_id": {"project": "flytesnacks", "domain": "development", "name": "fbcf25e4303104b4b81e"}, "node_id": "n0"}, "task_id": {"project": "flytesnacks", "domain": "development", "name": "basics.cache_test.basic_cache", "version": "98XuDRSQW4hVWKYQdQfd1Q=="}, "retry_attempt": 0}}' localhost:8089 flyteidl.service.CacheService.EvictTaskExecutionCache
```

### Setup process

- if want to verify from console, need to pull down: https://github.com/flyteorg/flyteconsole/pull/835 and then manually update flyteconsole/packages/flyteidl-types/node_modules/@flyteorg/flyteidl/gen/pb-js/flyteidl.d.ts
```
        enum CatalogCacheStatus {
            CACHE_DISABLED = 0,
            CACHE_MISS = 1,
            CACHE_HIT = 2,
            CACHE_POPULATED = 3,
            CACHE_LOOKUP_FAILURE = 4,
            CACHE_PUT_FAILURE = 5,
            CACHE_SKIPPED = 6,
            CACHE_EVICTED = 7
        }
```
(NOTE - a follow up PR will be to get that change merged into console since console doesn't support CACHE_EVICT CatalogCacheStatus)

- can also verify by checking artifact_data and artifacts tables

## Check all the applicable boxes 

- [x] I updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] All commits are signed-off.

PR #6451, opened 2025-05-13
Add support for cache TTL

### LINKED ISSUE #3305: is it possible to add cache expiration feature?
The idea is add the possibility to define a cache expiration period
to avoid confusion, is seems the best solution is to consider the cache expiration based on the task definition of the reading process

API:
add a new cache_expiration parameter to flytekit.task

`@task(cache=True, cache_version='v1', cache_expiration=timedelta(days=7))`

will define a task that will only use the cache, if the cache is less then a week old

### Discussed in https://github.com/flyteorg/flyte/discussions/3238

<div type='discussions-op-text'>

<sup>Originally posted by **ophiry** January 16, 2023</sup>
that is - include in the task decorator a parameter which will control for how long the cache will be considered valid
</div>

Sovietaced:
## Tracking issue
Closes #3305 

Flyte data may be stored in S3 buckets that have retention policies. If those retention policies delete the underlying data in S3 the entries in Flyte data catalog are stale and point to nonexistent buckets. This results in errors at task runtime in flytekit.  

This pull request modifies Flyte data catalog such that artifacts in the catalog can expire and no longer be valid.  The general approach is to allow for multiple rows with the same artifact key but to only retain one non-expired row ideally. I chose to allow for multiple rows to exist so that cleanup of expired artifacts could be done asynchronously. Also, there are multiple tables that hold references to the artifact table and ensuring these dependents were also expired seemed challenging. 

#### Data Model
1. A new column is added to the artifacts table, `expires_at`. This dictates when the artifact is no longer valid. 
2. The composite primary key for artifacts was broadened to include `created_at` such that there can be multiple rows with the same artifact key (ideally with one and only one non-expired entry if insertions are done carefully and atomically).
3. The compose primary kettle for tag was breaded to include `artifact_id` such that multiple tags can exist with the same artifact key but can only point to one artifact. 

#### Data Layer
1. The data layer was updated to filter expired artifacts in get/list operations, this propagates to tags as well. 
2. The data layer was updated such that non-expired artifacts are searched before creating new ones. 

#### Manager Layer
1. The TTL is used with a pluggable clock to generated the `expires_at` value for records.

#### Propeller
Updated propeller to plumb the cache TTL from the task metadata to Flyte data catalog API calls.

#### Future Updates
At a high level I'd like to follow up on this work with a background job of some sort that queries the DB for expired cache entries and then thoroughly cleans them up by deleting the contents if possible and then later removing the rows from Postgres. Similar to what is done in https://github.com/flyteorg/flyte/pull/4655

I realize flytekit changes will need to follow up here as well but I think those will be relatively trivial and this change is entirely backwards compatible. 

## How was this patch tested?

Unit tests, for now. 

### Labels

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the Flyte data catalog by introducing support for cache TTL, allowing artifacts to expire and preventing stale entries that could lead to runtime errors. Key modifications include the addition of an 'expires_at' column in the artifacts table and updates to the data layer for filtering expired artifacts. The implementation is designed to be backward compatible and includes unit tests to validate the new functionality.

</div>


datacatalog/pkg/repositories/gormimpl/artifact.go
```diff
@@ -126,57 +148,46 @@ func (h *artifactRepo) Update(ctx context.Context, artifact models.Artifact) err
 	timer := h.repoMetrics.UpdateDuration.Start(ctx)
 	defer timer.Stop()
 
-	tx := h.db.WithContext(ctx).Begin()
-	defer func() {
-		if r := recover(); r != nil {
-			tx.Rollback()
+	err := h.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
```
Sovietaced, line 151: Mostly refactoring to use the nicer transaction syntax

datacatalog/pkg/repositories/errors/postgres.go
```diff
@@ -40,9 +40,14 @@ func (p *postgresErrorTransformer) fromGormError(err error) error {
 }
 
 func (p *postgresErrorTransformer) ToDataCatalogError(err error) error {
+	var dce catalogErrors.DataCatalogError
+	if errors.As(err, &dce) {
```
Sovietaced, line 44: A little improvement so we can liberally call `ToDatacatalogError` on things

kumare3:
From reading you description it sees expired at is a better primary key member

Sovietaced:
> From reading you description it sees expired at is a better primary key member

Yeah that would work as well. Just need to widen that composite key with something.

datacatalog/pkg/manager/impl/artifact_manager.go
```diff
@@ -379,6 +392,15 @@ func (m *artifactManager) UpdateArtifact(ctx context.Context, request *datacatal
 
 	// update artifact in DB, also replaces/upserts associated artifact data
 	artifactModel.ArtifactData = artifactDataModels
+
+	// Reset TTL on the artifact since all the data is fresh.
+	if request.GetTtl() != nil {
```
ntate-stackav, line 397: This might be problematic depending on how the object storage engine implements the lifecycle rule for the object. It isn't required that the TTL be reset if a `/PUT` request is issued for a given key. For example, S3 triggers off of the `creationTimestamp`.

Sovietaced:
Closing this as stale.

PR #6457, opened 2025-05-17
Support Volcano scheduler integration by enabling PodGroup creation for podTask in Propeller

zhuo-zhi:

Support [Volcano scheduler](https://volcano.sh/en/docs/schduler_introduction/) integration by enabling PodGroup creation for podTask in Propeller. Volcano is a Kubernetes-native batch scheduler optimized for high-performance, AI/ML, and big data workloads.

Before this change, all podTasks associated with a FlyteWorkflow shared a single PodGroup, which was automatically created just once by the Volcano controller when the first podTask was launched. Subsequent podTasks reused this PodGroup, which introduced limitations. Specifically, different podTasks could not be scheduled with different Volcano queues or job priorities, since the shared PodGroup enforced a single queue and priority for all.

This change enables Propeller to create individual Volcano PodGroups for each podTask, allowing them to specify distinct queues and job priorities for scheduling.

1. The plugin manager for the k8s pod plugin creates Volcano PodGroups before creating the corresponding pods for podTasks. The PodGroup will have the same name as the pod, ensuring unique mapping between PodGroup and Pod. It also adds the PodGroup name to the pod’s annotations, ensuring that the Volcano PodGroup controller [does not create the PodGroup again](https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/podgroup/pg_controller.go#L168).
2. Each PodGroup inherits the queue and priority from its Pod.
3. Introduced a feature flag `enable-create-pod-group-for-pod` in the config to control whether this feature is enabled. The default value is false to ensure backward compatibility.
4. Regarding PodGroup garbage collection, the PodGroup is deleted by Kubernetes' GC. Since the PodGroup's owner reference points to a FlyteWorkflow, the deletion process follows a cascading effect. When the FlyteWorkflow is deleted, Kubernetes GC removes the associated PodGroup due to the absence of a living owner reference.

## How was this patch tested?
Unit tests added
Used internally at Linkedin.

### Labels

### DIFF
--- a/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go
+++ b/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go
@@ -41,6 +41,7 @@ import (
 	"github.com/flyteorg/flyte/flytestdlib/logger"
 	"github.com/flyteorg/flyte/flytestdlib/promutils"
 	"github.com/flyteorg/flyte/flytestdlib/promutils/labeled"
+	volcanov1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
 )
 
 const (
@@ -196,6 +197,93 @@ func (e *PluginManager) getPodEffectiveResourceLimits(ctx context.Context, pod *
 	return podRequestedResources
 }
 
+// getRequestResource will get the actual request resource.
+// It will first check if there is a defined request resource, if not, it will use the limit resource as the request.
+func getRequestResource(resources v1.ResourceRequirements) v1.ResourceList {
+	merged := v1.ResourceList{}
+	for name, req := range resources.Requests {
+		merged[name] = req
+	}
+	// If resources don't have requests, it defaults limit if that is explicitly specified.
+	for name, lim := range resources.Limits {
+		if _, ok := merged[name]; !ok {
+			merged[name] = lim
+		}
+	}
+	return merged
+}
+
+// createPodGroupForPod will build the associated volcano podgroup for a pod.
+func createPodGroupForPod(taskCtx pluginsCore.TaskExecutionMetadata, pod *v1.Pod) *volcanov1beta1.PodGroup {
+	// minResources is a concept in Volcano. A PodGroup will be enqueued when the required minResources
+	// are available in the cluster. Enqueue is a pre-filtering step before resource allocation,
+	// so strict computation is not necessary as long as the PodGroup meets the criteria for enqueuing.
+	//
+	// The minResources value is determined by:
+	// - The the total resources requested by all main containers.
+	// - The highest individual resource request among the init containers.
+	minResources := v1.ResourceList{}
+	for _, c := range pod.Spec.Containers {
+		requestResources := getRequestResource(c.Resources)
+		for name, quantity := range requestResources {
+			if q, ok := minResources[name]; !ok {
+				minResources[name] = quantity.DeepCopy()
+			} else {
+				q.Add(quantity)
+				minResources[name] = q
+			}
+		}
+	}
+	// InitContainers are run sequentially before other containers start, so the highest
+	// init container resource is compared against the sum of app containers to determine
+	// the effective usage.
+	for _, c := range pod.Spec.InitContainers {
+		requestResources := getRequestResource(c.Resources)
+		for name, quantity := range requestResources {
+			if value, ok := minResources[name]; !ok || quantity.Cmp(value) > 0 {
+				minResources[name] = quantity.DeepCopy()
+			}
+		}
+	}
+	podGroup := &volcanov1beta1.PodGroup{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: volcanov1beta1.SchemeGroupVersion.String(),
+			Kind:       "PodGroup",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            pod.Name,
+			Namespace:       pod.Namespace,
+			OwnerReferences: []metav1.OwnerReference{taskCtx.GetOwnerReference()},
+			Annotations:     map[string]string{},
+			Labels:          map[string]string{},
+		},
+		Spec: volcanov1beta1.PodGroupSpec{
+			MinMember:         1,
+			PriorityClassName: pod.Spec.PriorityClassName,
+			MinResources:      &minResources,
+		},
+	}
+	if pod.Labels != nil {
+		for k, v := range pod.Labels {
+			podGroup.Labels[k] = v
+		}
+	}
+	if pod.Annotations != nil {
+		for k, v := range pod.Annotations {
+			podGroup.Annotations[k] = v
+		}
+	}
+	if pod.Annotations == nil {
+		pod.Annotations = make(map[string]string)
+	}
+	if queueName, ok := pod.Annotations[volcanov1beta1.QueueNameAnnotationKey]; ok {
+		podGroup.Spec.Queue = queueName
+	}
+	// This is to prevent volcano from creating the podgroup for the pod
+	pod.Annotations[volcanov1beta1.KubeGroupNameAnnotationKey] = podGroup.Name
+	return podGroup
+}
+
 func (e *PluginManager) launchResource(ctx context.Context, tCtx pluginsCore.TaskExecutionContext) (pluginsCore.Transition, error) {
 	tmpl, err := tCtx.TaskReader().Read(ctx)
 	if err != nil {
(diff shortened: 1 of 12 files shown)


zhuo-zhi:
Hi reviewers, could you please help me add the `added` label to this PR? It seems I don’t have the access. Thank you!

Sovietaced:
Without looking to deeply at this I feel that the plugin manager code probably shouldn't be doing exceptional work for a third party Kubernetes scheduler. My intuition is that you would want to create a new plugin for Volcano and configure your Propeller instance to use that as the default plugin.

zhuo-zhi:
> Without looking to deeply at this I feel that the plugin manager code probably shouldn't be doing exceptional work for a third party Kubernetes scheduler. My intuition is that you would want to create a new plugin for Volcano and configure your Propeller instance to use that as the default plugin.

@Sovietaced Thanks for the review.

All the changes introduced in this PR are gated behind a configuration flag, which is disabled by default. This ensures that the plugin manager will not perform this functionality unless explicitly enabled.

This feature is primarily designed to support the current k8s Pod plugin, and more broadly, any future plugins that launch Pod resources. If I understand correctly, the current code structure [assumes each plugin is responsible for managing only one k8s object](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go#L212). This assumption is why a new plugin doesn't work because additional logic is required in the plugin manager to create PodGroups before launching Pods.

For other plugins such as TFJob, PyTorchJob, and MPIJob, their corresponding operators are already integrated with Volcano and can create PodGroups on their own.

flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go
```diff
@@ -232,6 +320,19 @@ func (e *PluginManager) launchResource(ctx context.Context, tCtx pluginsCore.Tas
...
+				if !k8serrors.IsAlreadyExists(err) {
+					reason := k8serrors.ReasonForError(err)
+					return pluginsCore.UnknownTransition, errors.Wrapf(stdErrors.ErrorCode(reason), err, "failed to create volcano podgroup for pod")
+				}
+				// The error is IsAlreadyExists
+				logger.Warnf(ctx, "PodGroup [%s] is already exists. Err: %v", podGroup.Name, err)
+			}
+		}
+
 		err = backOffHandler.Handle(ctx, func() error {
```
Sovietaced, line 336: What about this case?

Sovietaced:
> All the changes introduced in this PR are gated behind a configuration flag, which is disabled by default. This ensures that the plugin manager will not perform this functionality unless explicitly enabled.

I understand. My main concern is that core flyte propeller internals shouldn't have references to third party elements like volcano. You're using a feature flag to branch the logic in the generic plugin manager code to do something very specific for volcano. 

I will wait for more of the Flyte core maintainers to look at this but I'm thinking it might make more sense to rearchitect Flyte to allow plugins to build multiple resources.

flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go
zhuo-zhi: As before, this error will still be handled by [the logic a few lines below](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go#L242).

zhuo-zhi:
> > All the changes introduced in this PR are gated behind a configuration flag, which is disabled by default. This ensures that the plugin manager will not perform this functionality unless explicitly enabled.
> 
> I understand. My main concern is that core flyte propeller internals shouldn't have references to third party elements like volcano. You're using a feature flag to branch the logic in the generic plugin manager code to do something very specific for volcano.
> 
> I will wait for more of the Flyte core maintainers to look at this but I'm thinking it might make more sense to rearchitect Flyte to allow plugins to build multiple resources.

I see your point about making the plugin manager code more generic. I'm not sure how much effort it would take, or whether there are concrete plans to rearchitect the Flyte plugin to support building multiple resources. In the meantime, would it be possible to include this change in an earlier release so users can opt in earlier if needed? It could also help validate the effectiveness of a potential future rearchitecture. More feedback from the Flyte core maintainers would also be greatly appreciated.

davidmirror-ops:
@zhuo-zhi can we add docs [here](https://github.com/unionai/docs/blob/main/content/deployment/flyte-plugins/kubernetes-plugins.md) on how to prepare the K8s cluster for this integration?

zhuo-zhi:
> @zhuo-zhi can we add docs [here](https://github.com/unionai/docs/blob/main/content/deployment/flyte-plugins/kubernetes-plugins.md) on how to prepare the K8s cluster for this integration?

@davidmirror-ops Thanks for the review. I've created a separate PR to document the preparation steps: https://github.com/unionai/docs/pull/372

flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go
```diff
@@ -196,6 +197,93 @@ func (e *PluginManager) getPodEffectiveResourceLimits(ctx context.Context, pod *
 	return podRequestedResources
 }
 
+// getRequestResource will get the actual request resource.
```
fg91, line 200: I feel this module is not the right place to add this.

fg91:
> > All the changes introduced in this PR are gated behind a configuration flag, which is disabled by default. This ensures that the plugin manager will not perform this functionality unless explicitly enabled.
> 
> I understand. My main concern is that core flyte propeller internals shouldn't have references to third party elements like volcano. You're using a feature flag to branch the logic in the generic plugin manager code to do something very specific for volcano.
> 
> I will wait for more of the Flyte core maintainers to look at this but I'm thinking it might make more sense to rearchitect Flyte to allow plugins to build multiple resources.

I agree with everything @Sovietaced said, I also feel we cannot have such specialized logic in the plugin manager itself, even with a feature flag :/

zhuo-zhi:
> > > All the changes introduced in this PR are gated behind a configuration flag, which is disabled by default. This ensures that the plugin manager will not perform this functionality unless explicitly enabled.
> > 
> > 
> > I understand. My main concern is that core flyte propeller internals shouldn't have references to third party elements like volcano. You're using a feature flag to branch the logic in the generic plugin manager code to do something very specific for volcano.
> > I will wait for more of the Flyte core maintainers to look at this but I'm thinking it might make more sense to rearchitect Flyte to allow plugins to build multiple resources.
> 
> I agree with everything @Sovietaced said, I also feel we cannot have such specialized logic in the plugin manager itself, even with a feature flag :/

@fg91 @Sovietaced Thanks for the insights. I understand your concern. I’m wondering if there are any actionable steps you’d suggest I take to help move the PR forward?

Rearchitecting Flyte to support plugins that build multiple resources appears to be a significant change, as it would likely require refactoring all existing plugins along with the plugin manager. If that’s the only viable path for integration, would it be possible for me to get some support on this?

Sovietaced:
> @fg91 @Sovietaced Thanks for the insights. I understand your concern. I’m wondering if there are any actionable steps you’d suggest I take to help move the PR forward?

I would defer to the Union folks or people that understand the Propeller architecture deeply.

fg91:
I agree that re-architecting the plugin interface to allow for "extra resources" would be a bigger effort. I don't see a clean way to integrate the creation of resources like PodGroup into Flyte without this though.

The adapted [plugin interface](https://github.com/flyteorg/flyte/blob/c710fb60db62793510a60d03ffa15a671267ae89/flyteplugins/go/tasks/pluginmachinery/k8s/plugin.go#L106) could in theory look like this:

```go
type Plugin interface {
	// Defines a func to create a query object (typically just object and type meta portions) that's used to query k8s
	// resources.
	BuildIdentityResource(ctx context.Context, taskCtx pluginsCore.TaskExecutionMetadata) (client.Object, error)

	// Defines a func to create the full resource object that will be posted to k8s.
	BuildResource(ctx context.Context, taskCtx pluginsCore.TaskExecutionContext) (client.Object, error)

	// Analyses the k8s resource and reports the status as TaskPhase. This call is expected to be relatively fast,
	// any operations that might take a long time (limits are configured system-wide) should be offloaded to the
	// background.
	GetTaskPhase(ctx context.Context, pluginContext PluginContext, resource client.Object) (pluginsCore.PhaseInfo, error)

	// Properties desired by the plugin
	GetProperties() PluginProperties

        // New
        // Defines a func to create query objects for the extra resources e.g. for aborting
        BuildExtraIdentityResources(...) ([]client.Object, error)
        
        // Defines a func to create the full extra resource objects that will be posted to k8s.
        BuildExtraResources(...) ([]client.Object, error)
}
```

Which extra resources would be created would depend on the user configuration for the plugin returned by 
 `GetProperties()`.

Flytepropeller would not do any state monitoring for those extra resources, just create them along the creation of the main task resource. And it would delete them on abort. In theory we could omit the `BuildExtraIdentityResources` if propeller injected an owner reference to the main task object into the extra resources so they would get cleaned up automatically if the main task object was deleted. That would leave us with the additional `BuildExtraResources`.

I don't think the existing plugins would create that much work because their respective `BuildExtraResources` functions would just do nothing initially. But the creation of the resources (and their deletion or injection of owner refs) would have to be implemented in the k8s plugin.

All this is just brainstorming though, something like this would definitely require an RFC (I could help with the process) and it also definitely would need early input from Union folks like @eapolinario @wild-endeavor, @kumare3.

---
Alternative approach:

If your main goal was not *officially integrating* Flyte with Volcano but you/your organization being able to *use Flyte pod tasks with Volcano*, I think a much lower hanging fruit (albeit not an official integration into flyte) would be to handle the PodGroup creation in a (mutating) admission webhook that gets triggered when flytepropeller creates the pods. You could deploy such a service along side Flyte. I've done something like this in the past for other resources.

kumare3:
Cc @troychiu this was the one

PR #6309, opened 2025-03-05
Add workflow concurrency controller

thomasjhuang:
### Labels

Please add one or more of the following labels to categorize your PR:
- **added**: For new features.
- **changed**: For changes in existing functionality.
- **deprecated**: For soon-to-be-removed features.
- **removed**: For features being removed.
- **fixed**: For any bug fixed.
- **security**: In case of vulnerabilities

This is important to improve the readability of release notes.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

PR #6475, opened 2025-05-29
Add workflow concurrency control

thomasjhuang:
## Tracking issue
This solves #5659  and also closes out #6309. The related flytekit change is here - https://github.com/flyteorg/flytekit/pull/3267. Docs are added in this PR - https://github.com/unionai/docs/pull/417 

This work is to provide workflow concurrency control at a project level. It is a barebones implementation with the guarantee that at most `x` many workflows will be running, but it cannot guarantee that `x` many workflows are always running. We can control concurrency across all versions, and users specify the controls via `LaunchPlan` instantiation:
```
concurrency_limited_lp = LaunchPlan.get_or_create(
    name="my_concurrent_lp",
    workflow=my_workflow,
    concurrency=ConcurrencyPolicy(
        max_concurrency=3,
        behavior=ConcurrencyLimitBehavior.SKIP,
    ),
)
```

The primary mechanism is fairly straightforward - whenever we attempt to launch an execution, make a db query to check for running executions given the NamedEntityIdentifier triplet (project/domain/wf_name), and if the running executions is above the threshold for concurrency (`max_concurrency`) then we immediately fail to create the execution. 

## How was this patch tested?
Unit tests added to execution_manager_test.go, but namely this was internally tested at LinkedIn since we are porting this feature externally. **More testing is in progress on local sandbox.**

### Labels
- **added**: `ConcurrencyPolicy` and logic surrounding workflow concurrency management, as well as db migration to add an index on `executions` table for `execution_phase`. 

### Screenshots

I was able to confirm locally that we receive the correct error message when we are running multiple workflows concurrently at limits of 1 and 2. Within the same project, running a workflow without limits works. 

```
Error: rpc error: code = ResourceExhausted desc = Concurrency limit (1) reached for launch plan concurrency_
limit_1. Skipping execution.
{"json":{},"level":"error","msg":"rpc error: code = ResourceExhausted desc = Concurrency limit (1) reached f
or launch plan concurrency_limit_1. Skipping execution.","ts":"2025-07-07T23:12:38-07:00"}
```

## Related PRs
 #5659 and #6309, will also add in flytekit PR as reference.

## Docs link
- Will add in docs 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request introduces a feature for managing workflow concurrency at the project level, allowing users to set limits on concurrent executions via a ConcurrencyPolicy. It includes logic to check running executions, enhances logging, and updates the database schema to support these changes, addressing issues #5659 and #6309.

</div>


Sovietaced:
Can you run `make generate` in `flyteidl` dir and `make lint-fix` in `flyteadmin` dir.

davidmirror-ops:
@thomasjhuang thanks for your contribution. Also please check out the failing DCO check.
Do you plan to set a specific phase when an execution is skipped? That could be a bit beyond the scope of this PR though but it'd be useful.
Also please add some [docs ](https://github.com/unionai/docs)so users know how to leverage this in launchplans.

kumare3:
Taking a look

kumare3:
@troychiu can you please take a first look, you looked into this recently right?

troychiu:
I don't really have context but would love to take a look.

kumare3:
@troychiu sorry wrong one. I meant the volcano PR. I can take a look at this one

thomasjhuang:
Okay I've added a few things - docs, unit test, and ran `make generate`. `make lint-fix` didn't indicate any issues on code changes that I've introduced here. Is there a reason why the checks are all pending? Probably ready for review now. Please also check out the related flytekit change - https://github.com/flyteorg/flytekit/pull/3267. 

@davidmirror-ops @kumare3 @Sovietaced

davidmirror-ops:
@thomasjhuang thanks for this amazing contribution. Sorry about this but could you port the docs update to the new repo? https://github.com/unionai/docs I'll take point on updating this repo's `docs` folder to indicate it's no longer the content that ends up rendered in the docs site

flyteadmin/pkg/manager/impl/execution_manager.go
```diff
@@ -1156,6 +1168,97 @@ func (m *ExecutionManager) createExecutionModel(
...
 
+func checkLaunchPlanConcurrency(ctx context.Context, launchPlan *admin.LaunchPlan, executionRepo repositoryInterfaces.ExecutionRepoInterface, metrics executionSystemMetrics) error {
+	lpID := launchPlan.GetId()
+	lpProject := lpID.GetProject()
+	lpDomain := lpID.GetDomain()
+	lpName := lpID.GetName()
+	ctxForTimer := contextutils.WithProjectDomain(ctx, lpProject, lpDomain)
+	ctxForTimer = contextutils.WithLaunchPlanID(ctxForTimer, lpName)
+	timer := metrics.ConcurrencyCheckDuration.Start(ctxForTimer)
+	defer timer.Stop()
```
EngHabu: nit
```suggestion
	defer metrics.ConcurrencyCheckDuration.Start(ctxForTimer).Stop()
```

flyteadmin/pkg/manager/impl/execution_manager.go
```diff
@@ -1156,6 +1168,97 @@ func (m *ExecutionManager) createExecutionModel(
...
+		SELECT
+			COUNT(executions.id)
+		FROM
+			executions
+		INNER JOIN
+			launch_plans ON executions.launch_plan_id = launch_plans.id
+		WHERE
+			executions.project = '${lpProject}'
+			AND executions.domain = '${lpDomain}'
+			AND launch_plans.name = '${lpName}'
```
EngHabu, line 1221: Should it also filter on launch_plans.project/domain? If you confirmed it uses the right index, then nvm

flyteidl/protos/flyteidl/admin/launch_plan.proto
```diff
@@ -144,6 +144,22 @@ message LaunchPlanSpec {
...
+    // A setting for the maximum number of concurrent workflows
+    int32 max = 1;
+
+    // Descriptor for workflow execution behavior after the concurrency limit is hit
+    ConcurrencyLimitBehavior behavior = 2;
+}
+
+enum ConcurrencyLimitBehavior {
+    // A workflow that will be skipped because the limit has been hit.
+    SKIP = 0;
```
EngHabu: I know the repository doesn't strictly adhere to this (unfortunately) but I try to keep new code compliant. Checkout the style guide (scroll to enums) here: https://protobuf.dev/programming-guides/style/
```suggestion
enum ConcurrencyLimitBehavior {
    CONCURRENCY_LIMIT_BEHAVIOR_UNSPECIFIED = 0;
    
    // A workflow that will be skipped because the limit has been hit.
    CONCURRENCY_LIMIT_BEHAVIOR_SKIP = 1;
```

flyteadmin/pkg/manager/impl/execution_manager.go
```diff
@@ -1156,6 +1168,97 @@ func (m *ExecutionManager) createExecutionModel(
...
+	lpID := launchPlan.GetId()
+	lpProject := lpID.GetProject()
+	lpDomain := lpID.GetDomain()
+	lpName := lpID.GetName()
+	ctxForTimer := contextutils.WithProjectDomain(ctx, lpProject, lpDomain)
+	ctxForTimer = contextutils.WithLaunchPlanID(ctxForTimer, lpName)
+	timer := metrics.ConcurrencyCheckDuration.Start(ctxForTimer)
+	defer timer.Stop()
+
+	logger.Infof(ctx, "Checking concurrency limits for launch plan %v with policy %+v", lpID, launchPlan.GetSpec().GetConcurrencyPolicy())
```
Sovietaced: I think this should be a debug IMO.

flyteadmin/pkg/manager/impl/execution_manager.go
```diff
@@ -1156,6 +1168,97 @@ func (m *ExecutionManager) createExecutionModel(
...
+
+	// Active phases based on the workflow enum
+	activePhases := []string{
```
Sovietaced: Move this out of the method and reuse to reduce allocations since it's a static slice.

flyteadmin/pkg/manager/impl/execution_manager.go
```diff
@@ -1156,6 +1168,97 @@ func (m *ExecutionManager) createExecutionModel(
...
+		core.WorkflowExecution_QUEUED.String(),
+		core.WorkflowExecution_RUNNING.String(),
+	}
+
+	projectFilter, err := common.NewSingleValueFilter(common.Execution, common.Equal, "project", lpProject)
+	if err != nil {
+		logger.Errorf(ctx, "Failed to create project filter for concurrency check: %v", err)
```
Sovietaced: nit: I'm not a fan of logging AND returning/throwing errors. If each call site logs and returns an error you get logs filled with cascading logs, each with different levels of context around what happened. So I prefer doing one or the other, so logging the error when it's finally handled. 

I find it better to wrap the errors with some context so when you finally log it you effectively understand all the context on one spot.  

```go
return fmt.Errorf("creating project filter for concurrency check: %w, err)
```

flyteadmin/pkg/manager/impl/execution_manager.go
```diff
@@ -1097,6 +1099,16 @@ func (m *ExecutionManager) launchExecution(
 		Namespace:             namespace,
 	}
 
+	// Check ConcurrencyPolicy for LaunchPlan, this is a means to limit the concurrency of a launch plan across versions
+	// NOTE: There's a potential race condition here. Multiple concurrent requests
```
Sovietaced, line 1108: Yeah this is a classic problem that you can really only solve with an external lock of some sort (ie. pg advisory, or redis)

flyteadmin/pkg/manager/impl/execution_manager.go
```diff
@@ -1156,6 +1168,97 @@ func (m *ExecutionManager) createExecutionModel(
...
+			common.LaunchPlan: true,
+		},
+	})
+
+	if err != nil {
+		logger.Errorf(ctx, "Failed to count active executions using JOIN for launch plan %s.%s.%s: %v", lpProject, lpDomain, lpName, err)
+		// We still proceed to log the count as 0 and potentially hit the concurrency limit if Max is 0.
+	}
+
+	logger.Infof(ctx, "Found %d active executions for launch plan %s.%s.%s (any version)", count, lpProject, lpDomain, lpName)
```
Sovietaced: nit: should be debug before merging IMO

flyteadmin/pkg/manager/impl/execution_manager.go
```diff
@@ -1156,6 +1168,97 @@ func (m *ExecutionManager) createExecutionModel(
...
+
+	// Check against the policy limit
+	if count >= int64(launchPlan.GetSpec().GetConcurrencyPolicy().GetMax()) {
+		logger.Warningf(ctx, "Concurrency limit (%d) reached for launch plan %v (active: %d)", launchPlan.GetSpec().GetConcurrencyPolicy().GetMax(), lpID, count)
+		if launchPlan.GetSpec().GetConcurrencyPolicy().GetBehavior() == admin.ConcurrencyLimitBehavior_SKIP {
```
Sovietaced: Would be slightly more future proof to use a switch statement on the enum and panic or error in the default case to ensure new values are always handled correctly.

flyteadmin/pkg/manager/impl/execution_manager.go
```diff
@@ -1156,6 +1168,97 @@ func (m *ExecutionManager) createExecutionModel(
...
+			metrics.ConcurrencyLimitHits.WithLabelValues(lpProject, lpDomain, lpName).Inc()
+			logger.Infof(ctx, "Skipping execution creation for launch plan %v due to concurrency limit", lpID)
```
Sovietaced: I'd keep this as a warn and maybe remove the one above.

flyteidl/protos/flyteidl/admin/launch_plan.proto
Sovietaced: I agree

flyteidl/protos/flyteidl/admin/launch_plan.proto
```diff
@@ -144,6 +144,22 @@ message LaunchPlanSpec {
     // ClusterAssignment controls how to select an available cluster on which executions of this LaunchPlan should run.
     // This can be overwritten at execution creation level.
     ClusterAssignment cluster_assignment = 23;
+
+    // Concurrency setting for a given workflow to determine concurrency limits and behavior when the limit is breached
+    ConcurrencyPolicy concurrency_policy = 25;
```
Sovietaced: Should this be 24? I wonder if you have something already using 24 in your fork :P

Sovietaced (approved):

Sovietaced (changes requested):
I'm looking at porting a version of this to our fork and one thing I'm looking at is how this affects Flyte scheduler. Flyte scheduler is hardcoded to try and retry the execution creation up to 30 times if it fails, unless the gRPC status code in the error is `codes.AlreadyExists`. 

I think the scheduler code will need to be updated to either give up on `ResourceExhausted` or  we'll need to a way to articulate this case through some other means.

flyteidl/protos/flyteidl/admin/launch_plan.proto
thomasjhuang: Got it, accepted suggestion

thomasjhuang:
> @thomasjhuang thanks for this amazing contribution. Sorry about this but could you port the docs update to the new repo? https://github.com/unionai/docs I'll take point on updating this repo's `docs` folder to indicate it's no longer the content that ends up rendered in the docs site

Got it, I've opened a PR here - https://github.com/unionai/docs/pull/417

flyteadmin/pkg/manager/impl/execution_manager.go
thomasjhuang: Yea, we avoided external locking here otherwise it would become pretty complex.

flyteadmin/pkg/manager/impl/execution_manager.go
thomasjhuang: Done

flyteadmin/pkg/manager/impl/execution_manager.go
thomasjhuang: Done

flyteadmin/pkg/manager/impl/execution_manager.go
thomasjhuang: Makes sense, updated this as well

flyteadmin/pkg/manager/impl/execution_manager.go
thomasjhuang: Done

flyteadmin/pkg/manager/impl/execution_manager.go
thomasjhuang: Adding launch_plans.project/domain probably isn't needed bc the join already ensures we only see launch plans from the same project/domain as the filtered executions. I had done a MySQL `explain` and it does use the newly created index correctly.

flyteadmin/pkg/manager/impl/execution_manager.go
thomasjhuang: Done

flyteidl/protos/flyteidl/admin/launch_plan.proto
thomasjhuang: Ah good catch, changed to 24

flyteadmin/pkg/manager/impl/execution_manager.go
thomasjhuang: Got it, moved to switch case

thomasjhuang:
> I'm looking at porting a version of this to our fork and one thing I'm looking at is how this affects Flyte scheduler. Flyte scheduler is hardcoded to try and retry the execution creation up to 30 times if it fails, unless the gRPC status code in the error is `codes.AlreadyExists`.
> 
> I think the scheduler code will need to be updated to either give up on `ResourceExhausted` or we'll need to a way to articulate this case through richer error details.

Makes sense - this change will increase `ResourceExhausted` errors noticeably, especially if measured as a metric and it can be misleading since it isn't really an error. Some changes on flytescheduler are probably needed. Although not implemented yet, I intend to also add the execution phase `SKIPPED` so that the UI can indicate skipped executions rather than throw resource error.

Sovietaced:
We have an end to end test currently validating this functionality and it seems like there might be a correctness error in the logic that looks for previous executions. 

```go
func TestFlyte_WorkflowConcurrencyLimits(t *testing.T) {

	lp := "e2e_singleton_workflow"
	client, err := config.Flyte.GetAdminClient(config.TestCtx)
	require.NoError(t, err, "getting flyte admin client")

	latestLaunchPlan := config.Flyte.FindLatestLaunchPlan(config.TestCtx, t, lp)

	t.Logf("Found most recent launch plan with version [%s]", latestLaunchPlan.GetId().GetVersion())

	_, err = client.AdminClient().CreateExecution(config.TestCtx, &pbadmin.ExecutionCreateRequest{
		Project: config.Flyte.Project,
		Domain:  config.Flyte.Domain,
		Spec: &pbadmin.ExecutionSpec{
			LaunchPlan: &pbcore.Identifier{
				ResourceType: pbcore.ResourceType_LAUNCH_PLAN,
				Project:      config.Flyte.Project,
				Domain:       config.Flyte.Domain,
				Name:         lp,
				Version:      latestLaunchPlan.GetId().GetVersion(),
			},
		},
	})
	require.NoError(t, err, "creating execution")

	// Creating a second execution should fail while the first is non-terminal
	_, err = client.AdminClient().CreateExecution(config.TestCtx, &pbadmin.ExecutionCreateRequest{
		Project: config.Flyte.Project,
		Domain:  config.Flyte.Domain,
		Spec: &pbadmin.ExecutionSpec{
			LaunchPlan: &pbcore.Identifier{
				ResourceType: pbcore.ResourceType_LAUNCH_PLAN,
				Project:      config.Flyte.Project,
				Domain:       config.Flyte.Domain,
				Name:         lp,
				Version:      latestLaunchPlan.GetId().GetVersion(),
			},
		},
	})
	require.Error(t, err, "creating execution")
	s, ok := status.FromError(err)
	require.True(t, ok, "should be a grpc status error")
	require.Equal(t, codes.ResourceExhausted, s.Code())
}
```

This is failing in our production environment where there is more load and I'm wondering if the state filtering isn't quite right.

flyteadmin/pkg/manager/impl/execution_manager.go
```diff
@@ -47,6 +47,11 @@ import (
 
 const childContainerQueueKey = "child_queue"
 
+var activeExecutionPhases = []string{
+	core.WorkflowExecution_QUEUED.String(),
```
Sovietaced, line 51: I think this needs `core.WorkflowExecution_UNDEFINED` but I will test in our local environment shortly.

thomasjhuang: So if a workflow is Undefined we want to count that as a running phase? On our end, we actually had to reduce this list to just queued + running because when I had a bigger list including "accepting", "succeeding", "aborting" there were sometimes platform level bugs where a workflow was stuck in "aborting" or other states and then the concurrency limit would actually prevent the schedule from running at all. I'm not too strongly against adding Undefined here though.

Sovietaced: `UNDEFINED` is the default state so it shouldn't be problematic. 

I'm specifically seeing it in our production environment where there can be a large delta of time between submitting a workflow and it getting picked up by propeller, which would put it into a queued state.

flyteadmin/pkg/manager/impl/execution_manager_test.go
```diff
@@ -4088,6 +4088,290 @@ func TestListExecutions_LegacyModel(t *testing.T) {
...
+								if field == "execution_project" {
+									if val, ok := gormQueryExpr.Args.(string); ok && val == launchPlanID.GetProject() {
+										foundProject = true
+									}
+								} else if field == "execution_domain" {
+									if val, ok := gormQueryExpr.Args.(string); ok && val == launchPlanID.GetDomain() {
+										foundDomain = true
+									}
+								} else if field == "phase" {
+									if phaseList, ok := gormQueryExpr.Args.([]string); ok && len(phaseList) > 0 {
```
Sovietaced: Can this be updated to exact match the length of the phase list? Would make the unit test more robust to catching changes.

flyteadmin/pkg/manager/impl/execution_manager.go
Sovietaced: Alright so we added this into our environment and it did cause problems. There seem to be some cases where old workflows can be stuck in an undefined state. We're still trying to root cause.

Sovietaced: I root caused it and filed an issue here: https://github.com/flyteorg/flyte/issues/6556

So I'm okay with omitting UNDEFINED until I fix this other issue.

Sovietaced:
> Makes sense - this change will increase `ResourceExhausted` errors noticeably, especially if measured as a metric and it can be misleading since it isn't really an error. Some changes on flytescheduler are probably needed. Although not implemented yet, I intend to also add the execution phase `SKIPPED` so that the UI can indicate skipped executions rather than throw resource error.

For this version can we at least treat it as non-retryable? That's what we're doing and it seems to be ok.

flyteadmin/pkg/manager/impl/execution_manager_test.go
thomasjhuang: Got it, will add this

Sovietaced (approved):

Sovietaced:
@EngHabu seemed happy with this. We've been using a variation of it in production for the past couple weeks so I think its safe to land this.

popojk:
> > I'm looking at porting a version of this to our fork and one thing I'm looking at is how this affects Flyte scheduler. Flyte scheduler is hardcoded to try and retry the execution creation up to 30 times if it fails, unless the gRPC status code in the error is `codes.AlreadyExists`.
> > I think the scheduler code will need to be updated to either give up on `ResourceExhausted` or we'll need to a way to articulate this case through richer error details.
> 
> Makes sense - this change will increase `ResourceExhausted` errors noticeably, especially if measured as a metric and it can be misleading since it isn't really an error. Some changes on flytescheduler are probably needed. Although not implemented yet, I intend to also add the execution phase `SKIPPED` so that the UI can indicate skipped executions rather than throw resource error.

Hi @thomasjhuang @Sovietaced . Since I’m currently working on some scheduler related issues, I can go ahead and open an issue for this and submit a PR.

PR #6521, opened 2025-07-03
Fix workflow equality check

### LINKED ISSUE #6520: [BUG] Flyte Propeller Redundant Workflow Detection isBroken
### Flyte & Flytekit version

Flyte v1.15.X

### Describe the bug

Flyte Propeller has [logic to avoid writes to etcD when the in-memory workflow status is unchanged](https://github.com/stackav-third-party/flyte/blob/main/flytepropeller/pkg/controller/handler.go#L321-L325). Unfortunately it does not work because the task node status always seem to be dirty even when there is no diff. This is because the task node status id unconditionally set to dirty when `setX` calls are made to it. 

There may also be a problem with comparing unserialized values that will never be present from workflows read from etcD.

### Expected behavior

I would expect Flyte Propeller to avoid writes to etcD when there is no diff in the workflow status. 

_No response_

### Screenshots

_No response_

- [x] Yes

- [x] Yes

Sovietaced:
## Tracking issue
Closes #6520 

The [core propeller logic has a branch which compares the workflow read from etcD to the workflow that may have been mutated by the propeller state machine logic](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/controller/handler.go#L320-L326).The intention is that if the workflow status is the same it does not bother writing the workflow to etcD. The comparator logic for this operation includes bugs where two workflow that have the same state are considered different and thus trigger an etcD write (or at least an API call). As such, a shockingly high amount of our workflow update are redundant, wasted API calls (~60%). 

Screenshot 2025-07-03 at 1 47 34 PM

Screenshot 2025-07-03 at 1 47 50 PM

This pull request proposes to only set the `dirty` flags in TaskNodeStatus when a value is set that differs from the existing value. This should prevent workflows from being marked dirty during regular workflow processing, like from this [call site](https://github.com/flyteorg/flyte/blob/e3cabef618b167399f1ffe3ab4549584d2da527f/flytepropeller/pkg/controller/nodes/transformers.go#L256-L267).  

For posterity, I also eliminated fields now serialized to etcD from the `.Equals` comparator.

Depending on the running workflows this pull request has an opportunity to massively reduce k8s API and etcD load at scale. Previously k8s API and etcD load would scale linearly with running workflows due to how workflows are regularly queued and processed. With this change we have been able to run thousands of concurrent workflows with minimal k8s and etcD load (4-5 operations per second). 

## How was this patch tested?
Tested in our non-production environment where it has eliminated redundant workflow updates. Will be landing in our production environment soon.

No more redundant updates in non-prod
Screenshot 2025-07-03 at 1 58 56 PM

Also eliminates many wasted API calls to update the workflow. 

Screenshot 2025-07-03 at 2 06 51 PM

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request optimizes the workflow equality check in the Flyte propeller by implementing an Equals method for the WorkflowNodeStatus struct and ensuring 'dirty' flags in TaskNodeStatus are only set when values differ. This reduces unnecessary updates and API calls while streamlining the equality comparator by removing obsolete fields, enhancing overall performance. Tests have been added to validate the new functionality, contributing to performance improvements.

</div>

### DIFF
--- a/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/iface.go
+++ b/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/iface.go
@@ -21,7 +21,6 @@ import (
 
 //go:generate mockery --all --with-expecter
 
-type CustomState map[string]interface{}
 type WorkflowID = string
 type TaskID = string
 type NodeID = string
--- a/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
+++ b/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
@@ -198,6 +198,25 @@ func (in *WorkflowNodeStatus) SetWorkflowNodePhase(phase WorkflowNodePhase) {
 	}
 }
 
+func (in *WorkflowNodeStatus) Equals(other *WorkflowNodeStatus) bool {
+	if in == nil && other == nil {
+		return true
+	}
+	if in == nil || other == nil {
+		return false
+	}
+
+	if in.ExecutionError != nil && other.ExecutionError != nil {
+		if !proto.Equal(in.ExecutionError, other.ExecutionError) {
+			return false
+		}
+	} else if in.ExecutionError != other.ExecutionError {
+		return false
+	}
+
+	return in.Phase == other.Phase
+}
+
 type GateNodePhase int
 
 const (
@@ -408,7 +427,6 @@ type NodeStatus struct {
 	SubNodeStatus map[NodeID]*NodeStatus `json:"subNodeStatus,omitempty"`
 	// We can store the outputs at this layer
 
-	// TODO not used delete
 	WorkflowNodeStatus *WorkflowNodeStatus `json:"workflowNodeStatus,omitempty"`
 
 	TaskNodeStatus    *TaskNodeStatus    `json:",omitempty"`
@@ -952,6 +970,10 @@ func (in *NodeStatus) Equals(other *NodeStatus) bool {
 		if !v.Equals(otherV) {
 			return false
 		}
+	}
+
+	if !in.WorkflowNodeStatus.Equals(other.WorkflowNodeStatus) {
+		return false
 	}
 
 	if in.Error != nil && other.Error != nil {
--- a/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
+++ b/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/golang/protobuf/proto"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/flyteorg/flyte/flyteidl/gen/pb-go/flyteidl/core"
@@ -30,7 +31,7 @@ func (in *MutableStruct) ResetDirty() {
 	in.isDirty = false
 }
 
-func (in MutableStruct) IsDirty() bool {
+func (in *MutableStruct) IsDirty() bool {
 	return in.isDirty
 }
 
@@ -157,7 +158,8 @@ func (in *DynamicNodeStatus) Equals(o *DynamicNodeStatus) bool {
 	if in == nil || o == nil {
 		return false
 	}
-	return in.Phase == o.Phase && in.Reason == o.Reason
+
+	return in.Phase == o.Phase && in.Reason == o.Reason && in.IsFailurePermanent == o.IsFailurePermanent && in.Error.Equals(o.Error)
 }
 
 type WorkflowNodePhase int
@@ -217,6 +219,18 @@ func (in *GateNodeStatus) SetGateNodePhase(phase GateNodePhase) {
 		in.SetDirty()
 		in.Phase = phase
 	}
+}
+
+func (in *GateNodeStatus) Equals(other *GateNodeStatus) bool {
+	if in == nil && other == nil {
+		return true
+	}
+
+	if in == nil || other == nil {
+		return false
+	}
+
+	return in.Phase == other.Phase
 }
 
 type ArrayNodePhase int
@@ -328,7 +342,7 @@ func (in *ArrayNodeStatus) SetTaskPhaseVersion(taskPhaseVersion uint32) {
 	}
 }
 
-func (in *ArrayNodeStatus) DeepCopyInto(out *ArrayNodeStatus) {
+func (in *ArrayNodeStatus) deepCopyInto(out *ArrayNodeStatus) {
 	*out = *in
 	out.MutableStruct = in.MutableStruct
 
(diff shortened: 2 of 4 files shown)


hamersaw (commented):
This is fantastic! I'm going to pull locally to run with OTEL tracing on just to double check performance. Once I get that done I'm happy to approve.

Just a note on etcd performance - I always wanted to explore [protobuf APIs](https://kubernetes.io/docs/reference/using-api/api-concepts/#protobuf-encoding-compatibility) for k8s calls too. I suspect it would speed up creating Pods at scale while reducing etcd load. It doesn't seem available for CRDs, but I know some other projects have used aggregated APIs (ex. [cozystack](https://github.com/aenix-io/cozystack)) to use protobuf APIs over their CRDs. My initial benchmarks here showed significant reductions in CRD serialization size; which, if etcd stored in protobuf rather than json, would certainly help scale significantly.

pvditt:
As we use isDirty to check if a node has already been evaluated in a given propeller loop, would it make more sense to remove the isDirty check on the Equals check as opposed to limiting when isDirty is set?

hamersaw:
> As we use isDirty to check if a node has already been evaluated in a given propeller loop, would it make more sense to remove the isDirty check on the Equals check as opposed to limiting when isDirty is set?

An example for this, if we have a workflow like:
```
@task
def foo(x: int): int
    pass
    
@task
def bar(x: int, y:int)
    pass
    
@workflow
def wf(a: int, b:int)
    o1 = foo(x=a)
    o2 = foo(x=b)
    bar(x=o1, y=o2)
```
because bar is downstream of both calls to `foo` we could traverse the node twice within a propeller evaluation loop depending on how `isDirty` is set. The problem is that propeller expects nodes to have a single change within each evaluation loop, multiple changes can cause inconsistent behavior.

I think I've convinced myself that with this PR we are still updating `isDirty` if there is any change, which would cause a node to only evaluate once in each propeller loop. So the worst case scenario is that we evaluate the node mutliple times in the same loop only if there is no change (is this true?). I think I agree that it would be more explicit to just remove the `isDirty` from the equality check (I don't think that will cause any errors?).

Sovietaced:
> I think I've convinced myself that with this PR we are still updating `isDirty` if there is any change, which would cause a node to only evaluate once in each propeller loop. So the worst case scenario is that we evaluate the node mutliple times in the same loop only if there is no change (is this true?). I think I agree that it would be more explicit to just remove the `isDirty` from the equality check (I don't think that will cause any errors?).

I'm okay with removing `isDirty` from the equality check.

I am still confused as to why `isDirty` is implemented with diff checking for ArrayNodeStatus, DynamicNodeStatus, and GateNodeStatus but not for NodeStatus and TaskNodeStatus. My impression is that the bug was related to inconsistency there but maybe the diff checking was intentional. 

If `isDirty` is used to check if a node has already been evaluated should we remove diff checking for the node types that implement it?

kumare3:
I remember there was a case in which etcd would report back and update even if we don't see any changes (could be missed checks) the rev. Number will be updated

Sovietaced:
> I remember there was a case in which etcd would report back and update even if we don't see any changes (could be missed checks) the rev. Number will be updated

Yeah, this equality logic is a bit fragile. For context, we discovered this while implementing a custom version cache and we were comparing serialized workflow statuses to prevent redundant writes. Although doing double serialization is a potential waste of resources its a pretty canonical/robust way to understand if there is gonna be a diff in etcD or not.

Sovietaced:
@hamersaw @pvditt I have updated the pull request just to fix the equals logic. It includes test coverage that was missing previously. I have it running in our non-production environment and I'm seeing similar results so it looks promising.

flytepropeller/pkg/apis/flyteworkflow/v1alpha1/iface.go
```diff
@@ -21,7 +21,6 @@ import (
 
 //go:generate mockery --all --with-expecter
 
-type CustomState map[string]interface{}
```
Sovietaced, line 24: Removed because unused

flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -30,7 +31,7 @@ func (in *MutableStruct) ResetDirty() {
 	in.isDirty = false
 }
 
-func (in MutableStruct) IsDirty() bool {
+func (in *MutableStruct) IsDirty() bool {
```
Sovietaced, line 34: There were misc warnings about mixing pointer and non-pointer receiver functions so I fixed that up.

flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -157,7 +158,8 @@ func (in *DynamicNodeStatus) Equals(o *DynamicNodeStatus) bool {
 	if in == nil || o == nil {
 		return false
 	}
-	return in.Phase == o.Phase && in.Reason == o.Reason
+
+	return in.Phase == o.Phase && in.Reason == o.Reason && in.IsFailurePermanent == o.IsFailurePermanent && in.Error.Equals(o.Error)
```
Sovietaced, line 162: Updated this to include all the fields that are serialized.

flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -219,6 +221,18 @@ func (in *GateNodeStatus) SetGateNodePhase(phase GateNodePhase) {
 	}
 }
 
+func (in *GateNodeStatus) Equals(other *GateNodeStatus) bool {
```
Sovietaced, line 243: This didn't exist so I added it.

flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -328,7 +342,7 @@ func (in *ArrayNodeStatus) SetTaskPhaseVersion(taskPhaseVersion uint32) {
 	}
 }
 
-func (in *ArrayNodeStatus) DeepCopyInto(out *ArrayNodeStatus) {
+func (in *ArrayNodeStatus) deepCopyInto(out *ArrayNodeStatus) {
```
Sovietaced, line 364: These don't need to be exported so I fixed that.

flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -345,30 +359,53 @@ func (in *ArrayNodeStatus) DeepCopy() *ArrayNodeStatus {
 	}
 
 	out := &ArrayNodeStatus{}
-	in.DeepCopyInto(out)
+	in.deepCopyInto(out)
 	return out
 }
 
+func (in *ArrayNodeStatus) Equals(other *ArrayNodeStatus) bool {
```
Sovietaced, line 385: This didn't exist so I added it.

flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -383,6 +420,9 @@ type NodeStatus struct {
 
 	// Not Persisted
 	DataReferenceConstructor storage.ReferenceConstructor `json:"-"`
+	DataDir                  DataReference                `json:"-"`
```
Sovietaced, line 441: Moved these down to be more clear that they are not persisted/serialized.

flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -817,43 +857,77 @@ func (in *NodeStatus) SetOutputDir(d DataReference) {
...
+	if in.QueuedAt != nil && other.QueuedAt != nil {
+		if *in.QueuedAt != *other.QueuedAt {
+			return false
+		}
+	} else if in.QueuedAt != other.QueuedAt {
 		return false
 	}
 
-	if in.Phase != other.Phase {
+	if in.StartedAt != nil && other.StartedAt != nil {
```
Sovietaced, line 904: I added everything that is serialized here.

flytepropeller/pkg/apis/flyteworkflow/v1alpha1/node_status.go
```diff
@@ -894,33 +976,6 @@ func (in *NodeStatus) GetExecutionError() *core.ExecutionError {
 	return nil
 }
 
-// THIS IS NOT AUTO GENERATED
-func (in *CustomState) DeepCopyInto(out *CustomState) {
```
Sovietaced, line 898: Removed because unused

Sovietaced:
Had a regression due to the workflow node status not being compared so just added that. Also remove the comment that claims it is unused.

pvditt (approved):
this is really great. Thank you

PR #6523, opened 2025-07-04
refactor: Adjust CoPilot init and Pod status logic

pingsutw:
Containers task sometimes fails because

1. Propeller fails to read the output because it tries to get the output before the sidecar completes.
2. The sidecar is terminated before it can upload the file because the main container finishes before the signal watcher is created. As a result, the sidecar doesn’t handle the SIGTERM signal properly and gets killed by Kubernetes

* Modified `AddCoPilotToPod` to prepend the sidecar container to the list of init containers, ensuring the sidecar starts before the downloader. This guarantees the signal watcher is active before the main container completes. 

* Updated `DemystifyPodStatus` to check the status of init containers in addition to regular containers. This ensures that tasks are only marked as succeeded when both containers and sidecars have completed successfully. 

* Adjusted `assertPodHasCoPilot` to include init containers in the validation loop, ensuring volume mounts are correctly asserted for all containers.

* Added assertions in `TestAddCoPilotToPod` to verify the order and naming of init containers, ensuring the sidecar and downloader are correctly configured.

## How was this patch tested?

```python
from flytekit import ContainerTask, kwtypes


greeting_task = ContainerTask(
    name="test",
    image="alpine:latest",
    input_data_dir="/var/inputs",
    output_data_dir="/var/outputs",
    inputs=kwtypes(name=str),
    outputs=kwtypes(greeting=str),
    command=["/bin/sh", "-c", "sleep 100 | echo 'Hello, my name is {{.inputs.name}}.' | tee -a /var/outputs/greeting"],
)
```

### Screenshots
Screenshot 2025-07-03 at 5 01 46 PM

### DIFF
--- a/flyte-single-binary-local.yaml
+++ b/flyte-single-binary-local.yaml
@@ -56,6 +56,8 @@ plugins:
     cloudwatch-enabled: false
     stackdriver-enabled: false
   k8s:
+    co-pilot:
+      image: pingsutw/copilot:test
     default-env-vars:
       - FLYTE_AWS_ENDPOINT: http://flyte-sandbox-minio.flyte:9000
       - FLYTE_AWS_ACCESS_KEY_ID: minio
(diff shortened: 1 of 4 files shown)


Sovietaced (approved):
Other option is to have the main container block until the sidecar is running. If the downloader is ever removed this solution might break but that seems unlikely since the downloader is necessary for the container tasks.

PR #6526, opened 2025-07-09
Enqueue correct work item in node execution context

### LINKED ISSUE #6525: [BUG] Node Execution Context Enqueues Wrong Workflow Identifier
### Flyte & Flytekit version

Flyte v1.15.3

### Describe the bug

We see logspam (~48k/day) about nonsensical workflows that cannot be found: `Workflow namespace[]/name[<project>:<domain>:<wf_name>] not found, may be deleted.` The namespace for the workflow identifier is always empty and can't be found. 

### Expected behavior

I would expect to never seen workflows enqueued with an empty namespace and invalid identifier.

_No response_

### Screenshots

_No response_

- [x] Yes

- [x] Yes

Sovietaced:
## Tracking issue
Closes #6525 

Previously this code would attempt to enqueue the broad workflow identifier as opposed to the execution namespace & execution ID. As a result the logs would be spammed with the following WARNING log being spammed: `Workflow namespace[]/name[<project>:<domain>:<wf_name>] not found, may be deleted` where <> were substituted for real values. The namespace was always empty and the execution ID was incorrect.  This behavior likely wasn't a major issue due to the eventually consistent nature of the shared informer and how it regularly queues workflows for processing.

This pull request updates the code to enqueue the Flyte workflow k8s resource name & namespace which is what Flyte propeller expects since it operates on the flyteworkflows custom resource. This should make Flyte propeller process enqueued workflows as originally intended, resulting in faster results and no more log spam.

1. Enqueue workflows using the Flyte workflow k8s resource name and namespace. 
2. Remove the warning log in `passthrough` since we're returning an error. Generally we should log and also return errors, we should log errors at the latest point where we handle them so the error is printed only once.

## How was this patch tested?
1. Unit test
2. In our production environment

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the workflow enqueueing mechanism in the Flyte propeller by ensuring the correct Kubernetes resource name and namespace are used, resolving a logging issue that caused spam warnings, which reduces warning logs and improves processing speed. Unit tests have been updated to validate these changes.

</div>

### DIFF
--- a/flytepropeller/pkg/controller/nodes/node_exec_context.go
+++ b/flytepropeller/pkg/controller/nodes/node_exec_context.go
@@ -276,11 +276,7 @@ func (c *nodeExecutor) BuildNodeExecutionContext(ctx context.Context, executionC
 	}
 
 	workflowEnqueuer := func() error {
-		workflowIdentifier := types.NamespacedName{
-			Name:      executionContext.GetName(),
-			Namespace: executionContext.GetNamespace(),
-		}
-		c.enqueueWorkflow(workflowIdentifier.String())
+		c.enqueueWorkflow(executionContext.GetK8sWorkflowID().String())
 		return nil
 	}
 


kumare3:
Are you suggesting that GetID returns a wrong name

Sovietaced:
> Are you suggesting that GetID returns a wrong name

Yes, I could track down how that value is being constructed if it’s supposed to be different but the other times workflows are re-enqueued like this it’s a different code path without this issue.

Sovietaced:
> Are you suggesting that GetID returns a wrong name

To be more clear. The namespace would always parse into empty string and it doesn’t resolve to a valid execution ID. So propeller fails to look up the workflow from the k8s api when the value is enqueued.

Sovietaced:
> Are you suggesting that GetID returns a wrong name

Dug into this a big more. GetID comes from the workflow spec. 

The workflow spec is created here: https://github.com/flyteorg/flyte/blob/c19f4a1b8bec3e572f6b49b1f90c3066ea9f704d/flytepropeller/pkg/compiler/transformers/k8s/workflow.go#L123-L136

And that ID is generated here: https://github.com/flyteorg/flyte/blob/c19f4a1b8bec3e572f6b49b1f90c3066ea9f704d/flytepropeller/pkg/compiler/transformers/k8s/workflow.go#L45-L74

Which aligns with the PR description where I said the value being enqueued had no namespace and had an execution ID of `<project>:<domain>:<wf_name>`. Which is not an acceptable value to be enqueued into the work queue.

flytepropeller/pkg/controller/nodes/node_exec_context.go
```diff
@@ -276,7 +276,11 @@ func (c *nodeExecutor) BuildNodeExecutionContext(ctx context.Context, executionC
 	}
 
 	workflowEnqueuer := func() error {
-		c.enqueueWorkflow(executionContext.GetID())
+		workflowIdentifier := types.NamespacedName{
+			Name:      executionContext.GetName(),
+			Namespace: executionContext.GetNamespace(),
+		}
+		c.enqueueWorkflow(workflowIdentifier.String())
```
machichima: I think we can directly use `executionContext.GetK8sWorkflowID().String()` here, which is also used in the other enqueuing code path:

https://github.com/flyteorg/flyte/blob/66ea07d83e4c4b7560128841a651179dfe7d9902/flytepropeller/pkg/controller/controller.go#L152-L162

This will give the id format we are expecting

image

machichima:
Just for reference, I found that the `EnqueueOwnerFunc()` (which is the `workflowEnqueuer` function here) is called in the following places. So I think this bug occurs when we use dynamic workflow, sub-workflow, or refresh task

image

flytepropeller/pkg/controller/nodes/node_exec_context.go
Sovietaced: Good find!

Issue #6529, opened 2025-07-10
[BUG] PodTemplate Resources state pollution between workflow runs


punkerpunker:
### Flyte & Flytekit version

`main` branch

### Summary
The PodTemplateStore cache returns direct references to cached objects instead of copies, causing resource configurations from one workflow to leak into subsequent workflows. This results in workflows without inline resources unexpectedly inheriting resource specifications from previously executed workflows.

### Description
When multiple workflows execute using the same pod template, the second and subsequent workflows may inherit resource configurations (CPU, memory, GPU, etc.) from earlier workflows, even when no inline resources are specified. This happens because the PodTemplateStore.LoadOrDefault() method returns a pointer to the cached pod template object, which gets mutated during resource processing.

### Root Cause
In flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_template_store.go, the LoadOrDefault() method returns a direct reference to the cached object:
```
func (p *PodTemplateStore) LoadOrDefault(namespace string, podTemplateName string) *v1.PodTemplate {
    // ...
    if podTemplate, ok := podTemplates.Load(namespace); ok {
        return podTemplate.(*v1.PodTemplate)  // ← BUG: Returns pointer to cached object
    }
    // ...
}
```

### Expected behavior

flytepropeller doesn't preserve resources different between runs

### Additional context to reproduce

1. Execute a workflow (Workflow A) that specifies inline resources (e.g., high CPU/memory requirements)
2. Execute a different workflow (Workflow B) using the same pod template but without inline resources
3. Observe that Workflow B unexpectedly uses the resource specifications from Workflow A
4. Restart flytepropeller to clear the cache
5. Execute Workflow B again - it now works correctly until another workflow with inline resources executes

### Are you sure this issue hasn't been raised already?

- [x] Yes

### Have you read the Code of Conduct?

- [x] Yes

Sovietaced:
This should land in Flyte 1.16

PR #6483, opened 2025-06-03
Do not override runtime podTemplate resources with config defaults if they're set

punkerpunker:
## Tracking issue
https://github.com/flyteorg/flyte/issues/6482

## Why are the changes needed?

Pod template resources for CPU and memory were being overwritten by task resource defaults, while other resources like `nvidia.com/gpu` and `rdma/infiniband` were preserved correctly. This broke the expected resource priority order for users relying on PodTemplates. 

In our case, we're using a multi-cluster deployment of Flyte, and our infrastructure is very heterogeneous in terms of resources, storage and network. We want to make sure our "infra" related code (such as PodTemplate) is separated from the Flyte workflow definitions, and we expect that limits for each env for the steps are set within "infra" code. This is especially handy where we have a lot of tasks that are occupying a node fully, regardless of the amount of resources there are (e.g in case there is 10 CPUs in nodes - we assign 9 CPUs, in case there is 200 CPUs - we assign 199, and so on).

**The problem:** When users specify resources in PodTemplates like this:
```yaml
resources:
  requests:
    cpu: "55"              # ❌ Was being overwritten by task defaults
    memory: "1837Gi"       # ❌ Was being overwritten by task defaults  
    nvidia.com/gpu: "120"  # ✅ Was preserved correctly
    rdma/infiniband: "63"  # ✅ Was preserved correctly
```

Only GPU and RDMA resources were respected, while CPU and memory were replaced with platform defaults, causing inconsistent behavior and breaking user expectations.

## What changes were proposed in this pull request?

This PR implements proper resource priority handling by fetching pod template resources early and merging them with the correct precedence order:

1. **Early Pod Template Fetching**: Modified `ApplyFlytePodConfiguration` to fetch the base pod template before applying customizations, enabling access to pod template resources during container customization.

2. **Smart Resource Extraction**: Added `ExtractContainerResourcesFromPodTemplate()` function with intelligent fallback logic:
   - Exact container name match → "primary" → "default" → empty

3. **Enhanced Merging Logic**: 
   - Added `AddFlyteCustomizationsToContainerWithPodTemplate()` for pod template-aware resource handling
   - Added `MergeResourcesIfMissing()` to merge resources only when not already set (preserving higher priority)
   - Maintained backward compatibility with existing `AddFlyteCustomizationsToContainer()`

4. **Correct Priority Implementation**: Now properly follows the intended resource priority order:
   1. **Resource overrides** (highest priority)
   2. **Container inline resources** 
[...]

PR #6530, opened 2025-07-10
Fix PodTemplate resource state pollution between workflow runs

punkerpunker:
## Tracking issue
https://github.com/flyteorg/flyte/issues/6529

Without that, in cases where `podTemplate` is involved we could get unexpected resource overrides from previous runs. 

I changed LoadOrDefault() to return a deep copy of the cached pod template, this ensures that each workflow gets its own independent copy of the pod template, preventing any mutations from affecting other workflows

## How was this patch tested?
I've tested it in the Flyte multi-cluster environment of ours (devel)

## Related PRs
This is a follow-up for https://github.com/flyteorg/flyte/pull/6483 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request resolves resource state pollution in PodTemplate management within the Flyte framework by ensuring each workflow has an independent PodTemplate copy. This change is vital for preserving workflow execution integrity, particularly in multi-cluster setups.

</div>

### DIFF
--- a/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_template_store.go
+++ b/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_template_store.go
@@ -41,15 +41,16 @@ func (p *PodTemplateStore) Delete(podTemplate *v1.PodTemplate) {
 
 // LoadOrDefault returns the PodTemplate with the specified name in the given namespace. If one
 // does not exist it attempts to retrieve the one associated with the defaultNamespace.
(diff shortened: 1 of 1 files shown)


punkerpunker:
cc @davidmirror-ops

davidmirror-ops (approved):

kumare3:
Good callout

Sovietaced (approved):

Sovietaced:
I find it strange that propeller would be modifying a pod template but this change seems safe enough that I'm not going to bother tracking down the code :)

Issue #6533, opened 2025-07-14
[BUG] Full support for Namespace-Scoped Installation


diranged:
### Flyte & Flytekit version

Flyte Version: `v1.15.3`

### Describe the bug

I work within our cloud/kubernetes engineering team and we're working with our ML/Data team to help them get Flyte running. Our team runs dozens of Kubernetes clusters for the organization, and we provide developers with tightly managed Namespaces to deploy their applications. I am running into some challenges though with the way the `flyte-core` chart is setup - it has several places where it manages cluster-level resources such as `ClusterRoles`, `ClusterRoleBindings` and `CustomResourceDefinitions`.

To put it simply - we  do not allow developer applications to install _any Cluster-scoped resources_ - so resources like `ClusterRoles`, `ClusterRoleBindings`, `CustomResourceDefinitions` are all prohibited. In the event that we need to install a cluster-scoped resource to support a developer application, that is done from a privileged resource out-of-band from the developer application.

### Expected behavior

I expect that the `flyte-core` chart can be configured easily to operate inside of a single `Namespace` and with only _Namespace-scoped_ resources. Specifically:

### Flyte Propeller "Webhook" Service

The [flyte-core/templates/propeller/webhook.yaml](https://github.com/flyteorg/flyte/blob/v1.15.3/charts/flyte-core/templates/propeller/webhook.yaml) has hard-coded and non-configurable `ClusterRole` and `ClusterRoleBinding` resources. These should be configurable so that they can either be a `ClusterRole/ClusteRoleBinding` or `Role/RoleBinding` resources with the flip of a switch.

Additionally, the RBAC permissions should be defined in `values.yaml` so that they can be overridden. In particular the [`mutatingwebhookconfigurations`](https://github.com/flyteorg/flyte/blob/v1.15.3/charts/flyte-core/templates/propeller/webhook.yaml#L169C9-L169C38) permission would need to be removed for a namespace-scoped installation.

#### Flyte Propeller RBAC Resources

The main Propeller resources [flyte-core/templates/propeller/rbac.yaml](https://github.com/flyteorg/flyte/blob/v1.15.3/charts/flyte-core/templates/propeller/rbac.yaml) hard-coded and non-configurable `ClusterRole` and `ClusterRoleBinding` resources. These should be configurable so that they can either be a `ClusterRole/ClusteRoleBinding` or `Role/RoleBinding` resources with the flip of a switch.

[...]

Issue #3827, opened 2023-07-04
[Core feature] Enable flyte-binary to be deployed without cluster-wide permissions


davidmirror-ops:
### Motivation: Why do you think this is important?

Mutliple users (like [here](https://flyte-org.slack.com/archives/CP2HDHKE1/p1687258665151559) and [here](https://flyte-org.slack.com/archives/C05EZCLUCAF/p1688420255026849)) are struggling to deploy `flyte-binary` chart to K8s environments where they don't have cluster-wide permissions but are more or less constrained to what can be done on a single namespace.

As a way to prepare Flyte for enterprise environments where potential attack surface is a concern and the principle of least privilege is observed, creating a path for single-namespace deployment of Flyte would be useful.

### Goal: What should the final outcome look like, ideally?

Make he RBAC to be namespaced, flyte-binary/flytepropeller set to watch a single namespace, and namespace mapping for project/domains fixed to the single namespace

### Describe alternatives you've considered 

None

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #6534, opened 2025-07-15
feat(charts/flyte-core): allow configuring the RBAC rules as namespace scoped

diranged:
## Tracking issue
Related to https://github.com/flyteorg/flyte/issues/6533

## Why are the changes needed?=

This initial work should allow the Helm chart `flyte-core` to be installed in a "namespace-scoped" way that doesn't directly create any Cluster-scoped resources (`ClusterRole`, `ClusterRoleBinding`). The intention is to help support namespace-scoped installations where cluster-scoped privileges are not allowed to be granted to applications.

See https://github.com/flyteorg/flyte/issues/6533 for our reasoning and discussion. Also relates to https://github.com/flyteorg/flyte/issues/3827 - but I will update the `flyte-binary` chart in a separate PR if this PR is accepted.

## How was this patch tested?

Local testing with `helm install` as well as doing a bunch of `helm template > orig ... helm template > new && diff -b -u orig new` testing.

### Labels


davidmirror-ops (approved):
@diranged thanks so much for putting this effort in not only making this configurable but making the change "backwards compatible"

davidmirror-ops:
@diranged for the failing check, I think you need to run `make helm` in your fork and push the changes

diranged:
> @diranged for the failing check, I think you need to run `make helm` in your fork and push the changes

Thanks - I've just updated it..

davidmirror-ops:
Thanks @diranged seems like the check is still complaining about `deployment/sandbox/flyte_helm_generated.yaml` could you double check on your end the result of `make helm`?

diranged:
I'm checking now ... I suspect this has to do with a version difference in `helm` and how it handles null values (which has been changing recently). As a side recommendation, is there a reason the changes aren't self-committed back to the repo?

diranged:
Also, another side note - i notice every time I run the `make helm` we get a different value here:

```
.venvdiranged@Matts-Macbook-Pro flyte-fork (⎈|staging-analytics1:flyte) % git diff
diff --git a/docker/sandbox-bundled/manifests/complete-connector.yaml b/docker/sandbox-bundled/manifests/complete-connector.yaml
index 4623932e0..1f093966d 100644
--- a/docker/sandbox-bundled/manifests/complete-connector.yaml
+++ b/docker/sandbox-bundled/manifests/complete-connector.yaml
@@ -822,7 +822,7 @@ type: Opaque
 ---
 apiVersion: v1
 data:
-  haSharedSecret: UFUzMTlpN2E5QlBHYTlCMA==
+  haSharedSecret: YmtBU2ZOT2E4M0IwV2tqTg==
   proxyPassword: ""
   proxyUsername: ""```
```

davidmirror-ops:
> is there a reason the changes aren't self-committed back to the repo?

Not a good reason AFAICT. The action is required and defined [here](https://github.com/flyteorg/flyte/blob/a8409270adec273635d47aa05ab4f34665ecf510/Makefile#L34-L37) but probably making this a pre-commit hook or an actual action would simplify things

charts/flyte-core/values.yaml
```diff
@@ -371,6 +380,72 @@ flytepropeller:
...
+        resources:
+        - '*'
+        verbs:
+        - get
+        - list
+        - watch
+        - create
+        - update
+        - delete
+        - patch
```
machichima: nit: Could we align the format to indent all list items by two spaces?

charts/flyte-core/templates/admin/rbac.yaml
```diff
@@ -1,3 +1,34 @@
+{{- /*
+Deprecation: Replacing createClusterRole with generic rbac setting. In a future
+release the print will be replaced with fail.
+*/}}
+{{- if ne .Values.flyteadmin.serviceAccount.createClusterRole nil }}
+  {{- print "WARNING: .Values.flyteadmin.serviceAccount.createClusterRole is set (either true or false). Please use the .Values.flyteadmin.serviceAccount.rbac setting instead." }}
```
machichima: Is there a reason why we do "WARNING" here rather than a direct fail as for `clusterRole`?

charts/flyte-core/templates/admin/rbac.yaml
```diff
@@ -16,29 +47,29 @@ imagePullSecrets: {{ tpl (toYaml .) $ | nindent 2 }}
...
 metadata:
   name: {{ template "flyte.namespace" . -}}-{{- template "flyteadmin.name" . }}
   labels: {{ include "flyteadmin.labels" . | nindent 4 }}
 rules:
-- apiGroups: {{ toYaml .Values.flyteadmin.serviceAccount.clusterRole.apiGroups | nindent 2 }}
-  resources: {{ toYaml .Values.flyteadmin.serviceAccount.clusterRole.resources | nindent 2 }}
-  verbs: {{ toYaml .Values.flyteadmin.serviceAccount.clusterRole.verbs | nindent 2 }}
+- apiGroups: {{ toYaml .Values.flyteadmin.serviceAccount.rbacRules.apiGroups | nindent 2 }}
+  resources: {{ toYaml .Values.flyteadmin.serviceAccount.rbacRules.resources | nindent 2 }}
+  verbs: {{ toYaml .Values.flyteadmin.serviceAccount.rbacRules.verbs | nindent 2 }}
```
machichima: nit: can we also modify this to `nindent 4` to align the indentation format?

```yaml
  - apiGroups:
      - ""
    resources:
      - pods
```

Currently it is:

```yaml
  - apiGroups:
    - ""
    resources:
    - pods
```

charts/flyte-core/templates/admin/rbac.yaml
diranged: Fair - would you prefer I just `fail`?

diranged:
@machichima,
  Can you re-review? I'd like to get this merged.. :/

charts/flyte-core/templates/admin/rbac.yaml
```diff
@@ -1,3 +1,34 @@
...
+  {{- fail ".Values.flyteadmin.serviceAccount.createClusterRole is set (either true or false). Please use the .Values.flyteadmin.serviceAccount.rbac setting instead." }}
+{{- end }}
+
+{{- /*
+Check if both createClusterRole and rbac are set to non-null values. This will
+lead to unexpected behaviors, so we will fail here.
+*/}}
+{{- if and (ne .Values.flyteadmin.serviceAccount.createClusterRole nil) (ne .Values.flyteadmin.serviceAccount.rbac nil) }}
+  {{- fail "Both .Values.flyteadmin.serviceAccount.createClusterRole and .Values.flyteadmin.serviceAccount.rbac are set. Please only use the .Values.flyteadmin.serviceAccount.rbac setting." }}
+{{- end }}
```
machichima, line 15: While we directly fail if the `createClusterRole` is set, I think we do not need this?

charts/flyte-core/templates/admin/rbac.yaml
```diff
@@ -1,3 +1,34 @@
+{{- /*
+Deprecation: Replacing createClusterRole with generic rbac setting. In a future
+release the print will be replaced with fail.
+*/}}
```
machichima, line 4:
```suggestion
{{- /*
Deprecation: Renamed .Values.flyteadmin.serviceAccount.createClusterRole to 
.Values.flyteadmin.serviceAccount.rbac
*/}}
```

Align the deprecation message as below

charts/flyte-core/templates/admin/rbac.yaml
```diff
@@ -16,29 +47,29 @@ imagePullSecrets: {{ tpl (toYaml .) $ | nindent 2 }}
 {{- end }}
 
 ---
-{{- if .Values.flyteadmin.serviceAccount.createClusterRole }}
+{{- if (or .Values.flyteadmin.serviceAccount.rbac .Values.flyteadmin.serviceAccount.createClusterRole) }}
```
machichima, line 50:
```suggestion
{{- if .Values.flyteadmin.serviceAccount.rbac }}
```

machichima (commented):
Overall LGTM, just some small suggestions left. Please ping me after the modification and I can approve and merge this.

Thank you!

charts/flyte-core/templates/admin/rbac.yaml
diranged: Well - the intention is to explicitly inform the user that they are using a deprecated setting and get them to move to the new setting. I could do some kind of a warning - but in my experience warnings tend to not be reviewed by humans very well, so I like to just fail outright.

charts/flyte-core/templates/admin/rbac.yaml
diranged: Technically speaking - somebody could go in and set `rbac: false`, and `createClusterRole: true` - and then this wouldn't work. I'd like to leave the `or` in here until we _remove_ the `createClusterRole` setting entirely.

diranged: Technically speaking - somebody could go in and set `rbac: false`, and `createClusterRole: true` - and then this wouldn't work. I'd like to leave the `or` in here until we _remove_ the `createClusterRole` setting entirely.

diranged:
@machichima Can you check my updates and comments?

machichima (dismissed):
LGTM. Thank you!

machichima:
Hi @diranged,

Sorry just came back for merging this and found that there's a conflict. Could you please resolve it?  You can simply keep your own version when merging then do `make helm` once again.

Thank you!

Please ping me after this

Sovietaced (dismissed):
This looks good. Retains the previous default behavior. Added the `deprecated` tag just to ensure this gets noted in the release notes that a configuration path is now deprecated.

diranged:
I've rebased and pushed @machichima

PR #6538, opened 2025-07-22
Add task config in the flytectl config file

arbaobao:

If we specified the `project/domain` in the flytectl config, we don't need to add `-p flytesnack -d deployment` in every flytectl command. 

Read the flytectl config when there aren't project/domain in the  flytectl config.

## How was this patch tested?

Run flytectl command without specifying project/domain
```bash
flytectl get workflow
```
Before:

Screenshot 2025-07-23 at 12 44 13 AM

After:

Add task config in flytectl config and Run flytectl command
Screenshot 2025-07-23 at 12 47 09 AM

Screenshot 2025-07-23 at 12 46 40 AM

### Screenshots

Screenshot 2025-07-23 at 12 52 26 AM
Screenshot 2025-07-23 at 12 52 44 AM

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request introduces a new task configuration feature in the flytectl tool, enabling users to set project and domain settings in the configuration file. This enhancement streamlines command usage and minimizes the need for repetitive command-line inputs, improving the overall user experience. It includes the addition of a TaskConfig struct and updates to command handling, as well as modifications to the GetConfig function to read these values from the configuration when not explicitly provided.

</div>


flytectl/pkg/configutil/configutil.go
```diff
@@ -25,11 +25,18 @@ type DataConfig struct {
 	SecretKey string
 }
 
+type TaskConfig struct {
```
pingsutw: Should we define it in flytectl/cmd/config/config.go?

flytectl/cmd/config/config.go
```diff
@@ -14,6 +15,9 @@ var (
 	}
 
 	section = config.MustRegisterSection("root", defaultConfig)
+
+	taskConfig  = &configutil.TaskConfig{}
```
pingsutw:
```suggestion
	defaultTaskConfig  = &configutil.TaskConfig{}
```

flytectl/cmd/core/cmd.go
```diff
@@ -53,10 +53,18 @@ func generateCommandFunc(cmdEntry CommandEntry) func(cmd *cobra.Command, args []
 
 		if !cmdEntry.ProjectDomainNotRequired {
 			if config.GetConfig().Project == "" {
-				return fmt.Errorf("project and domain are required parameters")
+				if config.GetTaskConfig().Project != "" {
```
pingsutw: We can merge the config and taskConfig [here](https://github.com/flyteorg/flyte/blob/ebad67cf2c29c1edab4a0137a4140f90db9be141/flytectl/cmd/config/config.go#L46-L48) instead

flytectl/cmd/core/cmd.go
```diff
@@ -53,10 +53,18 @@ func generateCommandFunc(cmdEntry CommandEntry) func(cmd *cobra.Command, args []
...
 			}
 			if config.GetConfig().Domain == "" {
-				return fmt.Errorf("project and domain are required parameters")
+				if config.GetTaskConfig().Domain != "" {
+					config.GetConfig().Domain = config.GetTaskConfig().Domain
+				} else {
+					return fmt.Errorf("project and domain are required parameters")
+				}
 			}
 		}
```
machichima, line 62: Can we add a log here to indicate which project and domain are being used? Since these values can come from either the YAML config or parameters, logging them would help users know exactly what project and domain are being used

flytectl/cmd/core/cmd.go
arbaobao: Yes, I have adjusted the logic about merging the config.

flytectl/cmd/core/cmd.go
arbaobao: @machichima  I think it will be easier to debug. Here’s the log output — any suggestions on the output?
Screenshot 2025-07-23 at 11 53 18 PM

machichima: I think we can just put them in a single line and simplify to:

```go
fmt.Printf("Using Project: %s, Domain: %s\n", config.GetConfig().Project, config.GetConfig().Domain)
```

machichima:
I think the field name `task` is a bit ambiguous for content like "project", "domain", and "org". Also, we should clarify that this field is only used when working with `flytectl`. Would it make sense to rename it to something like `flytectl_context` / `flytectl_task_context` (or suggest a better name)?

cc @pingsutw WDYT?

pingsutw:
> I think the field name task is a bit ambiguous for content like "project", "domain", and "org".

It's intentional, we want to make it consistent with Flyte SDK

kumare3:
I agree the field name is not great, but we have shopped in v2

machichima (dismissed):
LGTM. Thank you!

flytectl/cmd/config/config.go
```diff
@@ -40,5 +49,12 @@ func (cfg Config) MustOutputFormat() printer.OutputFormat {
...
 // GetConfig will return the config
 func GetConfig() *Config {
-	return section.GetConfig().(*Config)
+	r := section.GetConfig().(*Config)
+	if taskSection.GetConfig().(*TaskConfig).Project != "" {
+		r.Project = taskSection.GetConfig().(*TaskConfig).Project
+	}
+	if taskSection.GetConfig().(*TaskConfig).Domain != "" {
+		r.Domain = taskSection.GetConfig().(*TaskConfig).Domain
+	}
```
pingsutw: `r.Project` should actually take higher precedence over `taskSection.GetConfig().(*TaskConfig).Project` since it's from the command line `flytectl --project ...`

pingsutw (approved):

flytectl/cmd/core/cmd.go
```diff
@@ -58,6 +58,7 @@ func generateCommandFunc(cmdEntry CommandEntry) func(cmd *cobra.Command, args []
 			if config.GetConfig().Domain == "" {
 				return fmt.Errorf("project and domain are required parameters")
 			}
+			fmt.Printf("Using Project: %s, Domain: %s\n", config.GetConfig().Project, config.GetConfig().Domain)
```
honnix, line 61: This print out seems to break json output. I think it should either be reverted or silenced when outputting json.

PR #6541, opened 2025-07-24
Fix Use of constant `state` value in OAuth 2.0 URL

odaysec:
https://github.com/flyteorg/flyte/blob/cfcea4dd8181e038bc4e4de2d0c0021b1eab02d9/flyteidl/clients/go/admin/pkce/auth_flow_orchestrator.go#L61-L61

fix the issue, we need to ensure that a unique, non-guessable `state` value is passed to the `AuthCodeURL` function. This can be achieved by using the `stateString` generated on line 50. Additionally, the `stateString` should be validated in the callback handler to ensure the integrity of the OAuth flow.
- Replace the empty string (`""`) in the `AuthCodeURL` function call on line 61 with the `stateString` variable.
- Ensure that the `stateString` is validated in the callback handler to confirm that it matches the value sent in the initial request.

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request fixes a critical bug in the OAuth 2.0 implementation by ensuring a unique, non-guessable `state` value is used in the `AuthCodeURL` function. It also adds validation for the `stateString` in the callback handler to enhance the security and integrity of the authentication process.

</div>

### DIFF
--- a/flyteidl/clients/go/admin/pkce/auth_flow_orchestrator.go
+++ b/flyteidl/clients/go/admin/pkce/auth_flow_orchestrator.go
@@ -58,7 +58,7 @@ func (f TokenOrchestrator) FetchTokenFromAuthFlow(ctx context.Context) (*oauth2.
 	values.Add(nonceKey, nonces)
 	values.Add(codeChallengeKey, pkceCodeChallenge)
 	values.Add(codeChallengeMethodKey, codeChallengeMethodVal)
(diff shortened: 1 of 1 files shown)


Sovietaced:
@odaysec did you perform any testing? As someone who is not an expert in OAuth how will I know this won't cause regressions for our users in the event that its not correct.

PR #6652, opened 2025-10-03
[FEAT] Add function name to TaskMetadata proto

machichima:

Short name is not always equal to function name. In the following cases they will be different:

```python
# function name: testtask / short name: rename
@env.task(short_name="rename")
async def testtask(data: str = "default string", n: int = 3) -> str:
    return data
```

Therefor we should have a separate field for `function_name`

~~Add `function_name` field into `TaskMetadata`~~

Based on the offline discussion with @pingsutw , we decided to rename `short_name` as `function_name` instead of adding a new field

### Labels


kumare3:
Isn't this the task id.Name

machichima:
> Isn't this the task id.Name

`TaskIdentifier.Name` is `environment_name.function_name`

After discuss with @pingsutw , we decided to close this and leave it as `ShortName` for now. Will either rename it or add new `FunctionName` field if needed in the future

PR #6653, opened 2025-10-03
RunService.CreateRun from_automation flags

iaroslav-ciupin:
### Labels


flyteidl2/workflow/run_service.proto
```diff
@@ -97,6 +97,9 @@ message CreateRunRequest {
 
   // The run spec to use.
   task.RunSpec run_spec = 5;
+
+  // True when run is created by some automation(schedule, webhook, etc.)
+  bool from_automation = 8;
```
kumare3: should this be in the RunSpec?

kumare3: Also should this have the automation id / or other info?

iaroslav-ciupin: I am not following why it should be in RunSpec and have other info? 
I only introduced this bool flag to correctly update `triggered_at` [timestamp in this PR](https://github.com/unionai/cloud/pull/12971)
What use cases are you referring to?

EngHabu: Why can't we use `trigger_name != nil` to decide that?

iaroslav-ciupin: please read this PR and [schedule triggers PR](https://github.com/unionai/cloud/pull/12971) descriptions for context.
we need this to differentiate 2 situations:
1. User explicitly triggers a trigger from UI/CLI <- in this case, FE/CLI will NOT specify this bool flag
2. Automation(cron,webhook,artifact) triggers a trigger in background <- in this case, all automation integration code will explicitly send `from_automation: true`
in both cases we're creating a run from a trigger (edited)

EngHabu: Got it.. I think we should add something like this instead... (as a general design pattern, booleans are very limiting.. they rarely cover all cases you will want and you will find yourself adding more and more of them...)

```suggestion
  enum Source {
      UNSPECIFIED = 0
      CLI = 1
      AUTOMATION/SYSTEM = 2
      ...
   }
   
   Source trigger_source = 8;
```

flyteidl2/workflow/run_definition.proto
```diff
@@ -410,3 +410,9 @@ message ActionSpec {
   // group this action belongs to, if applicable.
   string group = 8;
 }
+
+enum RunSource {
+  RUN_SOURCE_UNSPECIFIED = 0;
+  RUN_SOURCE_USER = 1;
```
EngHabu: Can we break this apart into CLI, CODE, UI ? I think this is going to be a useful metric to show...

iaroslav-ciupin: done

PR #6659, opened 2025-10-07
Add Connector proto

pingsutw:
## Tracking issue
NA

Add Connector proto

## How was this patch tested?

`make gen`


pingsutw:
Going to add the connector service

PR #6681, opened 2025-10-16
Implement async event recording to unblock reconciliation workers

RRap0so:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/6682

Root Cause: Cross-region traffic between FlyteAdmin and its database adds significant latency (300ms-1s) to each event write. Since event recording happened synchronously in the reconciliation loop (admin_eventsink.go:93-150), each reconciliation worker thread blocks waiting for the database transaction to complete across regions.

When FlyteAdmin experiences latency (300ms-1s per event), synchronous event recording blocks reconciliation workers. 

With hundreds or thousands of workflows, this causes:
  - 10x slowdown in reconciliation times
  - Cascading delays as workers wait for event responses
  - Potential DDOS of FlyteAdmin under high load
  - Degraded overall system performance

  Implemented asynchronous event recording to decouple event sending from reconciliation, eliminating cross-region latency from the critical path:

  Key changes:
  - Events go into a buffered channel (1000 capacity by default), processed by a background worker
  - Sink() returns in ~1ms instead of blocking for cross-region latency
  - Rate limiting on the worker side to avoid overwhelming admin/database
  - In-flight tracking with sync.Map to prevent duplicate queuing
  - Graceful shutdown drains the queue
  - Added prometheus metrics: events_queued, events_processed, events_failed, queue_depth, processing_latency_ms

  Config addition:
  - EventQueueSize parameter (default: 1000)

## How was this patch tested?

 Added comprehensive tests for async processing, rate limiting, backpressure, graceful shutdown, network failures, retries, and high volume scenarios. Also tested in a staging server triggering 1000 executions at the time.
 
### Labels

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request implements asynchronous event recording in the AdminEventSink to enhance the performance of reconciliation workers by decoupling event sending from the reconciliation process. Key features include a buffered channel for event processing, rate limiting, in-flight tracking to prevent duplicates, and new Prometheus metrics for monitoring. Comprehensive tests have been added to ensure reliability under high load conditions.

</div>

### DIFF
--- a/flytepropeller/events/admin_eventsink.go
+++ b/flytepropeller/events/admin_eventsink.go
@@ -3,10 +3,12 @@ package events
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/golang/protobuf/proto"
 	grpcRetry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
+	"github.com/prometheus/client_golang/prometheus"
 	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
 	"go.opentelemetry.io/otel/propagation"
 	"golang.org/x/time/rate"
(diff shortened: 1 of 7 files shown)


kumare3:
This is a correctness concern in v1. Since it is possible to drop these writes it is possible that ui will now show some statuses or nodes as terminated but they will be done.

RRap0so:
> This is a correctness concern in v1. Since it is possible to drop these writes it is possible that ui will now show some statuses or nodes as terminated but they will be done.

Thanks for the comment! I've explicitly tried to ensure that if there are problems with Admin for example when the queue is full, we return a ResourceExhausted error:

  https://github.com/flyteorg/flyte/pull/6681/files#diff-8003ce63d666c965c0f3843d05ffefddbc0a315aa2b007d7b44da0396ed3f786R142-R148
  
This error propagates up through the reconciliation loop and causes the node status to NOT be updated. To confirm  I've added a test to prove this behavior: https://github.com/flyteorg/flyte/pull/6681/files#diff-8003ce63d666c965c0f3843d05ffefddbc0a315aa2b007d7b44da0396ed3f786R2864-R2924

The test verifies:
  - When event recording fails (queue full), node status is NOT updated
  - Reconciliation retries and eventually succeeds
  - Node status is only updated after successful event recording

RRap0so:
Seems like more changes are needed in flyteadmin to make sure all the calls are properly async. It's doable but with a big change. Closing this for now.

PR #6685, opened 2025-10-20
fix proto

popojk:
## Tracking issue
n/a

We need to update proto for ListRuns api scope by task identifier

Update run_service.proto and regenerate

## How was this patch tested?

Test in parent PR


flyteidl2/workflow/run_service.proto
```diff
@@ -200,6 +200,9 @@ message ListRunsRequest {
 
     // List runs created from a trigger.
     common.TriggerName trigger_name = 6;
+
+    // Task identifier for filtering runs
+    task.TaskIdentifier task_id = 7;
```
squiishyy: Also would love to add TaskName here as well as a potential filter. It is the same value was underneath the hood just without the `version` field. We should built this functionality in now as we will likely want it in the future :)

popojk: @squiishyy Yup! Just added TaskName here as a filter. I will wait for this PR to be merged first and the main PR will checkout to the latest commit hash of flyte2 submodule.

Issue #6348, opened 2025-03-19
[BUG] Spark Pod Template Resources not getting applied


fiedlerNr9:
### Describe the bug

Resources defined in Flyte Pod Templates do not seem to get applied to Flyte Spark Tasks. I am able to submit the following sample code but dont see the resources getting propagated down to the driver or executor pods. Sample Workflow im using:

``` python
from operator import add

import flytekit
from flytekit import ImageSpec, Resources, task, workflow, dynamic, PodTemplate
from flytekitplugins.spark import Spark
from utils import f
from kubernetes.client import (
    V1Container,
    V1PodSpec,
)

container_image = ImageSpec(
    name="sparkfun",
    registry="ghcr.io/fiedlernr9",
    python_version="3.9",
    packages=["flytekitplugins-spark", "kubernetes"],
)

pod_template = PodTemplate(
    pod_spec=V1PodSpec(
        containers=[
            V1Container(
                # name="spark-kubernetes-executor",
                name="primary",
                resources={
                    "requests": {"cpu": "4" ,"ephemeral-storage": "10Gi"},
                    "limits": {"cpu": "4", "ephemeral-storage": "10Gi"},
                },
            )
        ]
    )
)


@task(
    task_config=Spark(
        # This configuration is applied to the Spark cluster
        spark_conf={
            "spark.driver.memory": "1000M",
            "spark.executor.memory": "4000M",
            "spark.executor.cores": "3",
            "spark.executor.instances": "5",
            "spark.driver.cores": "2",
        }
    ),
    pod_template=pod_template,
    container_image=container_image,
)
def hello_spark(partitions: int) -> float:
    print("Starting Spark with Partitions:! {}".format(partitions))

    n = 1 * partitions
    sess = flytekit.current_context().spark_session
    count = (
        sess.sparkContext.parallelize(range(1, n + 1), partitions).map(f).reduce(add)
    )
    print("Im done")
    pi_val = 4.0 * count / n
    return pi_val


@task(
    container_image=container_image,
    pod_template=pod_template,
)
def print_every_time(value_to_print: float):
    print("My printed value: {}".format(value_to_print))


@workflow
def wf(partitions: int = 1000) -> float:
    pi = hello_spark(partitions=partitions)
    print_every_time(value_to_print=pi)
    return pi
```

Describing Resources for driver:
```
Limits:                                                                                                                                                                          │
│       cpu:     1                                                                                                                                                                     │
[...]

PR #6603, opened 2025-08-25
wip

Sovietaced:
### Labels

Please add one or more of the following labels to categorize your PR:
- **added**: For new features.
- **changed**: For changes in existing functionality.
- **deprecated**: For soon-to-be-removed features.
- **removed**: For features being removed.
- **fixed**: For any bug fixed.
- **security**: In case of vulnerabilities

This is important to improve the readability of release notes.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

PR #6686, opened 2025-10-20
[Feat] Upgrade spark-operator to v2.3.0

machichima:
## Tracking issue

Related to: https://github.com/flyteorg/flyte/issues/6348

To be able to set `PriorityClassName` and `ephermal-storage`, we need to first upgrade our spark plugin

Update spark operator to version `2.3.0` and upgrade other packages to support this.

## How was this patch tested?

Ensure build and spark test passed

### Labels


kevinliao852:
It seems the new spark-operator version requires Go >= 1.24.1.
We might need to wait until [PR #6603](https://github.com/flyteorg/flyte/pull/6603)
 (Go 1.24 upgrade) gets merged before moving forward here.

Sovietaced:
> It seems the new spark-operator version requires Go >= 1.24.1.
> We might need to wait until [PR #6603](https://github.com/flyteorg/flyte/pull/6603)
> (Go 1.24 upgrade) gets merged before moving forward here.

You should now be unblocked

kevinliao852:
Thanks for the update! I’ll resume work on this task.

corleyma:
@kevinliao852 You still working on this, out of curiosity?  We're also finding a need to set executor ephemeral storage.

kevinliao852:
Yes, still working on this.

The spark-operator v2.3.0 upgrade is in progress, and we’re currently discussing the **Kubernetes libraries** upgrade with the team since it’s a dependency for moving this forward. I’ll post updates here as the discussion progresses.

corleyma:
For others following along at home: looks like k8s client version unpin [has merged](https://github.com/flyteorg/flyte/pull/6835), so that's one blocker removed.

kevinliao852:
There’s also a new update related to this:
PR flyteorg/flyte#6843 addresses the remaining operator version bump and is currently under review.
Once that lands, the operator upgrades should be complete.

Since the Kubernetes libs upgrade affects several operators and the dependencies are already updated, we can now focus on upgrading the Spark operators

kevinliao852:
We’ve identified a dependency and behavior mismatch between spark-operator and Flyte around config parsing.
spark-operator uses `github.com/spf13/viper v1.21.0`, which depends on `github.com/go-viper/mapstructure/v2 v2.4.0`.
Flyte is still on an **older Viper version** that uses the legacy mapstructure.

### Structural decoding differences

With newer Viper + mapstructure v2, a config like:

```
default-annotations:
  - annotationKey1: annotationValue1
  - annotationKey2: annotationValue2
```

is decoded as:

```
[]map[string]string
```

instead of the previously accepted:

```
map[string]string
```

which causes the following test to fail:

```
k8sConfig := flyteK8sConfig.GetK8sPluginConfig()
assert.True(t, k8sConfig.InjectFinalizer)
assert.Equal(t, map[string]string{
    "annotationKey1": "annotationValue1",
    "annotationKey2": "annotationValue2",
}, k8sConfig.DefaultAnnotations)
```

### Key normalization differences

Newer Viper applies case-insensitive key normalization, so map keys are transformed:

```
annotationKey1 -> annotationkey1
```

### In contrast, Flyte’s current setup:

> Is more permissive in decoding.
> Preserves key casing.

With the new Spark plugin version, some existing Flyte configurations may experience **breaking changes** due to the updated parsing semantics. It’s less of a version bump and more of a runtime behavior risk. It would be good to align on the expected behavior before moving forward.

Sovietaced:
> We’ve identified a dependency and behavior mismatch between spark-operator and Flyte around config parsing.
> 
> spark-operator uses `github.com/spf13/viper v1.21.0`, which depends on `github.com/go-viper/mapstructure/v2 v2.4.0`.
> 
> Flyte is still on an **older Viper version** that uses the legacy mapstructure.
> 
> 
> 
> ### Structural decoding differences
> 
> 
> 
> With newer Viper + mapstructure v2, a config like:
> 
> 
> 
> ```
> 
> default-annotations:
> 
>   - annotationKey1: annotationValue1
> 
>   - annotationKey2: annotationValue2
> 
> ```
> 
> 
> 
> is decoded as:
> 
> 
> 
> ```
> 
> []map[string]string
> 
> ```
> 
> 
> 
> instead of the previously accepted:
> 
> 
> 
> ```
> 
> map[string]string
> 
> ```
> 
> 
> 
> which causes the following test to fail:
> 
> 
> 
> ```
> 
> k8sConfig := flyteK8sConfig.GetK8sPluginConfig()
> 
> assert.True(t, k8sConfig.InjectFinalizer)
> 
> assert.Equal(t, map[string]string{
> 
>     "annotationKey1": "annotationValue1",
> 
>     "annotationKey2": "annotationValue2",
> 
> }, k8sConfig.DefaultAnnotations)
> 
> ```
> 
> 
> 
> ### Key normalization differences
> 
> 
> 
> Newer Viper applies case-insensitive key normalization, so map keys are transformed:
> 
> 
> 
> ```
> 
> annotationKey1 -> annotationkey1
> 
> ```
> 
> 
> 
> ### In contrast, Flyte’s current setup:
> 
> 
> 
> > Is more permissive in decoding.
> 
> > Preserves key casing.
> 
> 
> 
> With the new Spark plugin version, some existing Flyte configurations may experience **breaking changes** due to the updated parsing semantics. It’s less of a version bump and more of a runtime behavior risk. It would be good to align on the expected behavior before moving forward.

I ran into this issue when upgrading Go to 1.25 and was able to force viper back to the old version which would be the path of least resistance. 

https://github.com/flyteorg/flyte/pull/6771/commits/725c1c96a1d5091181a579237cf9aeeafd62eeb5

pingsutw:
Could we pin Viper to an older version, like https://github.com/machichima/flyte/blob/d4a450ac1e49583e96847b8661b3893360a94e5d/flyteidl/go.mod#L137-L139

kevinliao852:
Really appreciate the help here.

I might be missing something, and I’m happy to be corrected, but based on several experiments, my understanding is that the core issue is not Viper itself, but **mapstructure**. Newer Viper versions depend on `github.com/go-viper/mapstructure/v2`, which **Spark Operator** explicitly **requires**. Since `/v2` is a distinct Go module path, Go does not allow downgrading or redirecting it back to the legacy v1 implementation via `replace` or `exclude`.

kevinliao852:
After discussion with the team, we decided to upgrade spark-operator to v2.1.1 to address the current mapstructure / Viper dependency incompatibilities and to enable ephemeral-storage support.

In addition, @Sovietaced’s approach resolves the dependency issue by explicitly locking Viper to v1.11.0, which is compatible with spark-operator v2.1.1.

Sovietaced:
Closing due to inactivity, feel free to reopen and rebase against the latest changes and fixed issues.

PR #6691, opened 2025-10-22
Triggers enum adjustments

iaroslav-ciupin:

Embedded enum in Proto causes buf-ts to generate TypeScript enum with wrong string values("SCHEDULE" , but we need "TYPE_SCHEDULE" to match backend's Go implementation)

## How was this patch tested?

tested in https://github.com/unionai/cloud/pull/13197

### Labels


buf.yaml
```diff
@@ -21,6 +21,7 @@ lint:
       - flyteidl2/logs/dataplane/payload.proto
       - flyteidl2/secret/definition.proto
       - flyteidl2/plugins/spark.proto
+      - flyteidl2/task/common.proto
```
EngHabu, line 24: Since this is going to be a code migration anyway, can we not add an exception and properly name the enums? I added the previous ones because I didn't want to make an already huge PR even bigger..

iaroslav-ciupin: code migration is tested here https://github.com/unionai/cloud/pull/13197 but I don't want to rename enum because it will cause data format change in DB

EngHabu: How is this stored in the DB? as a JSON string? the enum value? a serialized proto?

iaroslav-ciupin: it is stored as a Postgres enum which matches existing proto enum strings

iaroslav-ciupin: here is the code of transformer https://github.com/unionai/cloud/blob/main/workflow/repositories/transformers/trigger.go#L104

EngHabu: ah ok

PR #6696, opened 2025-10-23
Extend task triggers summary

iaroslav-ciupin:
### Labels


flyteidl2/task/task_definition.proto
```diff
@@ -75,13 +75,21 @@ message TaskIdentifier {
 }
 
 message TaskTriggersSummary {
+  reserved 1;
+
+  message TriggerDetails {
```
squiishyy, line 80: Considering we have [TriggerDetails](https://github.com/flyteorg/flyte/blob/92b31876f14f9c79e318bb04e6c23efa1ad833df/flyteidl2/trigger/trigger_definition.proto#L76-L87) already defined in `trigger_definition.proto`, would love name this something else, even though it will be prefixed differently.

Maybe `TriggerSummaryDetails`?

iaroslav-ciupin: I left name Details precisely because its a different package and a different prefix: `task.TaskTriggersSummary_TriggerDetails` which doesn't collide with existing `trigger.TriggerDetails`, but keeps same meaning. Its still a "detailed information" about an individual trigger, but with a different context.
Its somewhat similar to `trigger.Trigger` and `task.TaskTrigger`(same trigger information, but in different contexts).
Why do you want to rename it?

squiishyy: My main goal was to not have them be the same despite the package name but I think you're right and we shouldn't worry about that and use the package/proto structure name to our advantage here :) 

Thank you!

PR #6718, opened 2025-11-05
Copy v2 plugins

pvditt:
Copy over plugins 

## How was this patch tested?
Ran in devbox + dogfood to test plugins along with executor overhaul 

TODO: will follow up w/ adding support for accelerated plugins and secret manager


pingsutw (approved):
nit: can we remove the Athena/BigQuery/Databricks/Snowflake plugins under plugins/webapi since we’ve deprecated them?

PR #6721, opened 2025-11-07
Typed interface migrate

machichima:

Change `TypedInterface` from using map to list, in order to preserve the arguments order.

### Labels


flytestdlib/utils/variable_utils.go
```diff
@@ -0,0 +1,36 @@
...
+			return entry.GetValue()
+		}
+	}
+	return nil
+}
+
+// SetVariable adds or updates a variable in a VariableMap.
+// If the key already exists, it updates the value. Otherwise, it appends a new entry.
+func SetVariable(vm *core.VariableMap, key string, variable *core.Variable) {
+	if vm == nil {
```
wild-endeavor: can we also check that variable is not nil?  we'll never need to add a nil variable right?

wild-endeavor (approved):

flyteidl2/core/interface.proto
```diff
@@ -23,10 +23,16 @@ message Variable {
...
+message VariableEntry {
+  string key = 1;
+  Variable value = 2;
+}
+
 // A map of Variables
 message VariableMap {
-  // Defines a map of variable names to variables.
-  map<string, Variable> variables = 1;
+  // Use repeated key value pair
```
EngHabu:
```suggestion
  // Use repeated key value pair to maintain ordering across different languages
```

PR #6722, opened 2025-11-07
Add description field in task and trigger protos

popojk:

Add description field to Task and Trigger protos

Update proto and regenerate


flyteidl2/task/environment.proto
```diff
@@ -13,4 +13,7 @@ message Environment {
     (buf.validate.field).string.min_len = 1,
     (buf.validate.field).string.max_len = 63
   ];
+
+  // Optional description of environment
+  string description = 2;
```
wild-endeavor: let's add the validation here also, we can always remove it in the future but i think safer to add for now.

flyteidl2/task/task_definition.proto
```diff
@@ -121,6 +124,20 @@ message Task {
   TaskMetadata metadata = 2 [(buf.validate.field).required = true];
 }
 
+// Link to source code used to define this entity
+message SourceCode {
+  string link = 1;
```
wild-endeavor, line 129: we can also consider sending a remote, a file path, and a line number, and letting the backend construct the actual link.  but this is okay for now.

flyteidl2/task/task_definition.proto
```diff
@@ -121,6 +124,20 @@ message Task {
...
+}
+
+message DescriptionEntity {
```
wild-endeavor: I really don't like this name but I can't think of anything better.  cc @pingsutw any ideas?

flyteidl2/task/task_definition.proto
popojk: I just left a comment in the code base to remind future me to implement what you mentioned about

flyteidl2/task/task_definition.proto
pingsutw: Documentation?

flyteidl2/task/environment.proto
popojk: I've added string length validation to every description field, thanks!

flyteidl2/task/task_definition.proto
```diff
@@ -121,6 +124,33 @@ message Task {
...
+  string remote_url = 1;
+
+  // File path relative to git repo absolute path
+  string file_path = 2;
+
+  // The line number of the flyte task in code file
+  int32 line_number = 3;
+
+  // Whether the git tree is clean
+  bool is_git_tree_clean = 4;
```
popojk: We need this field to tell if the git tree is clean. If it is not clean, the url will not be suffixed with line number otherwise the line number could be wrong.

flyteidl2/task/task_definition.proto
popojk: @wild-endeavor Is the proto name "DocumentationEntity" sounds good to you?

Issue #6703, opened 2025-10-26
Containerize components


EngHabu:
- [ ] Dockerfile for each component
- [ ] Dockerfile for manager

WangWang0226:
#take

PR #6730, opened 2025-11-11
fix(runs/docker): enable CGO for go-sqlite3 and include gen/go in bui…

WangWang0226:
---

## Tracking issue

[#6703](https://github.com/flyteorg/flyte/issues/6703)

---

The `runs` service Dockerfile build failed due to:
1. The previous `runs/Dockerfile` used an incorrect build path (`/workspace/runs/cmd/main.go` did not exist), because the entire repository was copied into `/workspace/runs`, resulting in a nested structure (`/workspace/runs/runs/cmd/main.go`). This caused a “directory not found” build error. 

2. The `go.mod` file specifies a minimum Go version of 1.24.6, but the original Dockerfile used golang:1.21-alpine as the builder image. This version mismatch caused go mod download to fail with an error stating that the Go toolchain was too old. 

3. The build context did not include `gen/go` files because of `.dockerignore` rules, leading to module import errors for `flyteidl2`.

4. Since the `go-sqlite3` driver depends on CGO, the image built with `CGO_ENABLED=0` could not initialize the SQLite database at runtime.

---

This PR fixes and improves the Docker build process for the **runs** component:

* **Updated `.dockerignore`**

  * Now excludes `gen/` by default but re-includes `gen/go/**` to ensure generated proto Go files are available in the build context.
* **Updated `runs/Dockerfile`**

  * Fixed COPY path to copy the entire repo to `/workspace` and build from `/workspace/runs`.
  * Upgraded Go base image to `golang:1.24.6-alpine` (to match `go.mod` requirement).
  * Installed `build-base` (includes `gcc`, `musl-dev`) to support CGO builds.
  * Enabled `CGO_ENABLED=1` and added build tag `"sqlite_omit_load_extension"` for security and SQLite compatibility.

These updates ensure the image builds successfully and the service can connect to SQLite without runtime errors.

---

## How was this patch tested?

1. Verified Docker build completes successfully:

   ```bash
   docker build -f runs/Dockerfile -t flyte-v2-runs:test .
   ```
2. Verified service runs correctly and connects to SQLite:

   ```bash
   docker run --rm -p "8090:8090" flyte-v2-runs:test --config /app/config.yaml
   ```
3. Confirmed health check endpoint responds successfully at `/healthz`.
    ```
    curl http://localhost:8090/healthz
    ```
All tests passed as expected.

---

### Labels

* **fixed**
* **changed**

---

### Setup process

To rebuild the image locally:

```bash
docker build -f runs/Dockerfile -t flyte-v2-runs:test .
docker run --rm -p 8090:8090 flyte-v2-runs:test --config /app/config.yaml
```

---

### Screenshots
- docker build & docker run
  image

- curl http://localhost:8090/healthz
  image

---

### Checklist

* [x] I updated the `.dockerignore` to include generated Go code.
* [x] I confirmed the new Docker image builds successfully.

---

## Related PRs

N/A

---

## Docs link

N/A

---

### DIFF
--- a/gen/rust/Cargo.lock
+++ b/gen/rust/Cargo.lock
@@ -430,9 +430,9 @@ checksum = "df3b46402a9d5adb4c86a0cf463f42e19994e3ee891101b1841f30a545cb49a9"
 
 [[package]]
 name = "hyper"
-version = "1.8.0"
+version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1744436df46f0bde35af3eda22aeaba453aada65d8f1c171cd8a5f59030bd69f"
+checksum = "2ab2d4f250c3d7b1c9fcdff1cece94ea4e2dfbec68614f7b87cb205f24ca9d11"
 dependencies = [
  "atomic-waker",
  "bytes",
@@ -466,9 +466,9 @@ dependencies = [
 
 [[package]]
 name = "hyper-util"
-version = "0.1.17"
+version = "0.1.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3c6995591a8f1380fcb4ba966a252a4b29188d51d2b89e3a252f5305be65aea8"
+checksum = "52e9a2a24dc5c6821e71a7030e1e14b7b632acac55c40e9d2e082c621261bb56"
 dependencies = [
  "bytes",
  "futures-channel",


runs/Dockerfile
```diff
@@ -1,21 +1,21 @@
...
 
-# Copy the runs service code
-COPY . ./runs/
+# Copy the entire repository
+COPY . /workspace
 
 # Build the service
 WORKDIR /workspace/runs
-RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o runs-service ./cmd/main.go
+RUN CGO_ENABLED=1 GOOS=linux go build -tags "sqlite_omit_load_extension" -a -o /workspace/runs/runs-service ./cmd/main.go
```
machichima: Curious, why setting the tag as `sqlite_omit_load_extensiion`?
And why are we removing `-installsuffix cgo`?

machichima:
```suggestion
RUN CGO_ENABLED=1 GOOS=linux go build -tags "sqlite_omit_load_extension" -a -o runs-service ./cmd/main.go
```
I think we can keep `runs-service` as we are inside `/workspace/runs` already?

runs/Dockerfile
```diff
@@ -1,21 +1,21 @@
...
 
-# Install build dependencies
-RUN apk add --no-cache git make
+# Install dependencies
+RUN apk add --no-cache git make build-base
 
 WORKDIR /workspace
 
 # Copy go mod files
 COPY ../go.mod ../go.sum ./
```
machichima: Just curious, can we do:

```
COPY go.mod go.sum ./
```

runs/Dockerfile
```diff
@@ -1,21 +1,21 @@
...
 RUN go mod download
 
-# Copy the runs service code
-COPY . ./runs/
+# Copy the entire repository
+COPY . /workspace
```
machichima:
```suggestion
COPY . ./
```

I think we can do this as we are already in `/workspace`?

machichima (commented):
Thank you for this! Left a few comment. Will try it out locally later. 
btw I saw the `docker-compose` file in `runs/` also, could you also try running the service with docker-compose?

runs/Dockerfile
popojk: Adding -a -installsuffix cgo ensures that the build cache is isolated from non-CGO packages. Could you please share the reason for removing it?

popojk (commented):
The PR looks great, thanks for your effort! Just left a question, could you PTAL?

runs/Dockerfile
WangWang0226: Thanks for reviewing! 
I did some research and below is my thought and explanation of why removing `-installsuffix cgo`. If anything looks incorrect, please let me know!

## My Understanding of Why `-installsuffix cgo` Was Needed Originally

In the old GOPATH-era build system, all compiled `.a` files were stored in the same directory:

```
$GOPATH/pkg/linux_amd64/<package>.a
```

This caused a problem, if you first built with **`CGO_ENABLED=1`** and then built again with **`CGO_ENABLED=0`**, Go would **reuse the old CGO-enabled `.a` file** instead of recompiling. The “pure Go” build would silently link against `libsqlite3.so` and then **crash** on Alpine/scratch images.

To avoid mixing these artifacts, developers used **`-installsuffix cgo`** to put CGO builds into a separate cache directory.

---

## So Why I Think `-installsuffix cgo` Is No Longer Needed

### 1. The "runs-service" always requires CGO

The `runs-service` imports:

```go
import _ "github.com/mattn/go-sqlite3"
```

`go-sqlite3` is a **CGO-only** driver.

This means even if runtime uses Postgres, compile time *always* loads the SQLite driver, therefore **CGO_ENABLED=1 is required for every valid build**.

So:

- The project will never produce CGO=0 builds

- The “CGO=0 vs CGO=1 cache mixing” scenario no longer exists

- Therefore, `-installsuffix cgo` does not serve any purpose here

The previous `CGO_ENABLED=0` setting in the Dockerfile was actually invalid and caused runtime errors when we set SQLite as database rather than `PostgreSQL` in `runs/config.yaml` : 
```
Error: failed to initialize database: failed to initialize database: Binary was compiled with 'CGO_ENABLED=0', go-sqlite3 requires cgo to work. This is a stub
```

---

### 2. Since Go 1.10, the build cache is content-addressed and automatically separates different build options

Go 1.10 introduced a new build cache located under `$GOCACHE`. Cache entries are stored in hashed directories like:

```
$GOCACHE/02/
$GOCACHE/fd/
```

The key is based on source content, build flags, compile version...etc, means that:

- CGO=0 and CGO=1 automatically produce different cache entries

- No manual directory separation is needed

- No risk of cache collisions caused by CGO differences

#### References

#####  Go 1.10 Release Notes (official)

[https://go.dev/doc/go1.10#build](https://go.dev/doc/go1.10#build)

> The go build command now detects out-of-date packages purely based on the content of source files, specified build flags, and metadata stored in the compiled packages. … The old advice to add -a … is no longer necessary.

##### `go` command documentation (Build & test caching)

[https://pkg.go.dev/cmd/go#hdr-Build_and_test_caching](https://pkg.go.dev/cmd/go#hdr-Build_and_test_caching)

> The build cache correctly accounts for changes to Go source files, compilers, compiler options, and so on.

##### Article: “Go Build Cache Mechanics”

[https://medium.com/@AlexanderObregon/go-build-cache-mechanics-6ada202c0502](https://medium.com/@AlexanderObregon/go-build-cache-mechanics-6ada202c0502)

Key points:

* Cache is **content-addressable**
* Artifacts are stored under **hashes**, not filenames
* Hash includes: source, compiler version, **build flags**

So, different build flags → different hash → completely separate cached results

---

### 3. Verification: I verified experimentally that CGO=0 and CGO=1 produce different cache entries

I ran both builds with `go build -x` and compared `$GOCACHE`.

#### Build 1: `CGO_ENABLED=0`

```
CGO_ENABLED=0 go build -x -o runs-service ./cmd/main.go
```

New entries appeared:

```
02/, ec/, 5e/, ...
```

#### Build 2: `CGO_ENABLED=1`

```
CGO_ENABLED=1 go build -x -o runs-service ./cmd/main.go
```

New, different entries appeared:

```
fd/, b9/, 94/, ...
```

The result tells us:

- **CGO=0 and CGO=1 builds do *not* share cache**
- **Go correctly treats them as independent builds**

WangWang0226: Thanks for reviewing! About `sqlite_omit_load_extension`, after reviewing go-sqlite3 and some investigation, I think this tag isn’t necessary, I’ll remove it.

For `-installsuffix cgo`, I explained the reasoning in @popojk’s comment below.

And no problem I’ll update the docker-compose file as well!

WangWang0226: Yes I'll fix it!

runs/Dockerfile
WangWang0226: Yes, it's better!

runs/Dockerfile
WangWang0226: Yeah!

runs/Dockerfile
machichima: Thank you for the explanation! While we only got one go build here, I think it's fine to remove `-installsuffix cgo`.

machichima:
@WangWang0226 feel free to re-request our review once you are done or leave comment if you have any question. Thank you!

runs/Dockerfile
popojk: Thanks for the explanation. Agree to remove `-installsuffix cgo` for now.

runs/Makefile
WangWang0226, line 1: Fix: Docker Compose v1 → v2 migration in Makefile
The Makefile previously used deprecated Compose v1 commands:
```
docker-compose up
docker-compose down
docker-compose logs
```
These have been replaced by v2 syntax:

```
docker compose up
docker compose down
docker compose logs
```
Ref: [Docker Compose Migration](https://docs.docker.com/compose/releases/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2)

All Compose v1 usages were replaced accordingly.

I also noticed the Makefile contains legacy and redundant structure (e.g., multiple Postgres startup paths, duplicated commands, mixed DB backends), but Makefile refactor is outside of the scope of this PR, so I'm thinking open a issue for this, what do you think?

runs/config-postgres.yaml
WangWang0226, line 1: ## **Fix: Incorrect Postgres host/port inside container**

### **Problem**
The previous `config-postgres.yaml` incorrectly used:

```yaml
host: "localhost"
port: 5433
```

This caused the `runs-service` container to fail connecting to Postgres because:
* Inside a container, **localhost refers to the container itself**, not the Postgres container.
* Port **5433 is the host-mapped port** (`5433:5432`), and is **not accessible from other containers**.

### **Fix (use correct Postgres DNS + internal port)**

```diff
- host: "localhost"
- port: 5433
+ host: "postgres"
+ port: 5432
```

### **Why this works**

* `postgres` resolves to the Postgres container via Docker’s internal DNS
* `5432` is the Postgres container’s internal listening port

---

## **How to Test**

1. Start Docker Compose in foreground

```
docker compose up
```

2. Verify service state and health

```
docker compose ps
```

output:

```
NAME                  IMAGE               COMMAND                  SERVICE        CREATED          STATUS                    PORTS
flyte-runs-postgres   postgres:15         "docker-entrypoint.s…"   postgres       25 seconds ago   Up 24 seconds (healthy)   0.0.0.0:5433->5432/tcp
flyte-runs-service    runs-runs-service   "./runs-service --co…"   runs-service   25 seconds ago   Up 14 seconds (healthy)   0.0.0.0:8090->8090/tcp
```

3. Check logs

  Postgres logs:
  
  ```
  docker compose logs -f postgres
  ```
  
  runs-service logs:
  
  ```
  docker compose logs -f runs-service
  ```

4. Validate Postgres readiness
- **Container internal health check**

    ```
    docker exec flyte-runs-postgres pg_isready -U postgres
    ```
    output:
    ```
    /var/run/postgresql:5432 - accepting connections
    ```

- **Container internal SQL query**

    ```
    docker exec -it flyte-runs-postgres psql -U postgres -d flyte_runs -c 'select 1;'
    ```
    output:
   ```
   ?column? 
  ----------
          1
  (1 row)
    ```

- **Host machine connection test (5433 → 5432 mapping)**

    ```
    psql -h localhost -p 5433 -U postgres -d flyte_runs -c 'select 1;'
    ```
  output:
    ```
    Password for user postgres: 
     ?column? 
    ----------
            1
    (1 row)
    ```
---

## **5. Validate runs-service health endpoint**

```
curl -f http://localhost:8090/healthz
```
output:
```
OK% 
```

runs/docker-compose.yaml
WangWang0226, line 1: ## **Fix: Removed unused env variables**

The previous `docker-compose.yaml` included:

```yaml
environment:
  - DATABASE_HOST=postgres
  - DATABASE_PORT=5432

```

These environment variables were **not used anywhere**:
- The runs-service **does not read these environment variables**
- The service **only reads the mounted config file** (`config-postgres.yaml`) via this in `docker-compose.yaml`:
    
    ```
    volumes:
      - ./config-postgres.yaml:/app/config.yaml
    command: ["--config", "config.yaml"]
    ```
Keeping unused env variables is misleading because it suggests that the service depends on them when it does not. So I removed them to avoid confusion.

runs/Makefile
popojk: Yup, I agree we open another to fix the Makefile issues that you mentioned

runs/config-postgres.yaml
```diff
@@ -10,8 +10,8 @@ runs:
 database:
   # Using PostgreSQL
   postgres:
-    host: "localhost"
-    port: 5433
+    host: "postgres"
```
popojk, line 13: Should we also fix the port name [here](http://localhost:8089), where the run service should call queue service by docker container name instead of localhost?

runs/docker-compose.yaml
```diff
@@ -28,9 +28,6 @@ services:
     depends_on:
```
popojk, line 28: As mentioned in the init [config](https://github.com/flyteorg/flyte/blob/91ba87d5e4d0a0b99706d87a568c7cbf2b94cb21/runs/config.yaml#L11-L13), right now we use SQLite instead of PostgreSQL as the db of run service. Do we need to run PostgreSQL service in docker compose now?

WangWang0226: Right now `docker-compose.yaml` is only for running the service in **PostgreSQL mode**, since it needs both the `runs-service` container and the `postgres` container.
The default configuration uses **SQLite**, which does not require any external database service and can be run with a simple `docker run` command.

Does this answers your question?

runs/config-postgres.yaml
WangWang0226: For now, I’m not sure how to properly fix the value of `queueServiceUrl: "http://localhost:8089"`. The correct solution depends on how we plan to containerize and orchestrate the entire system.

If we eventually want to run **runs / queue / executor** as separate services in containers, then we should have a **root-level docker-compose.yaml** that defines *all* of them together. In that setup, every service would run on the same default Docker network, and they would communicate using **service names + container ports**.
For example:

```
runs → queue = http://queue:8089
```

The existing `runs/docker-compose.yaml` is only a partial or subsystem example for *runs + postgres* service. The other components (queue, manager) don't have Dockerfiles yet, and the unified “root-level docker-compose” does not exist yet. So we currently don’t have a definitive place to decide what `queueServiceUrl` should be, since we don't have the **service names + container ports** of queue service.

We could change it to `http://queue:8089` as a forward-looking assumption, but without the full multi-service compose in place, it’s hard to say what the correct value should be right now. What do you think?

popojk: @WangWang0226 Apologize for late reply. Let's come back to this issue later. We will use helm chart to start all services, which will pass the queue service domain into the container(in k8s pod). However, the helm charts are not ready yet.

runs/docker-compose.yaml
popojk: OK, thanks for letting me know.

machichima (approved):
LGTM! Thank you

PR #6733, opened 2025-11-12
Fix: Avoid merging pod template twice if template container name matches task container name

fg91:

Users can configure [pod templates ](https://www.union.ai/docs/v1/flyte/deployment/flyte-configuration/configuring-podtemplates/#create-a-podtemplate-resource) that are applied to all task pods.
Currently, if the pod template has a container called `"primary"` or `"default"` and the task container uses the same name, the merging occurs twice leading to environment variables being duplicated.

This PR fixes this.

 <div id='description'>
<h3>Summary by Bito</h3>

<ul>

<li>Enhances the merging logic for pod templates to prevent duplication of environment variables when container names match, specifically for 'primary' or 'default'.</li>

<li>Addresses an issue where merging could occur multiple times under these conditions.</li>

<li>New tests have been introduced to ensure the correct behavior of the merging process, thereby improving the overall reliability of pod template handling.</li>

<li>Overall, this pull request touches on the merging logic for pod templates and introduces new tests to ensure reliability.</li>

</ul>

</div>

### DIFF
--- a/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper_test.go
+++ b/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper_test.go
@@ -2341,12 +2341,25 @@ func TestMergeBasePodSpecsOntoTemplate(t *testing.T) {
 						Env:   []v1.EnvVar{{Name: "FOO", Value: "BAR"}},
 					},
 				},
+				InitContainers: []v1.Container{
+					{
+						Name:  "default-init",
+						Image: "default-task-init-image",
+						Env:   []v1.EnvVar{{Name: "INIT_FOO", Value: "INIT_BAR"}},
+					},
+				},
 			},
 			basePodSpec: &v1.PodSpec{
 				Containers: []v1.Container{
 					{
 						Name:  "default",
 						Image: "task-image",
+					},
+				},
+				InitContainers: []v1.Container{
+					{
+						Name:  "default-init",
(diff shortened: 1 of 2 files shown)


flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper_test.go
```diff
@@ -2331,6 +2331,70 @@ func TestMergeBasePodSpecsOntoTemplate(t *testing.T) {
...
+					},
+				},
+			},
+			expectedResult: &v1.PodSpec{
+				Containers: []v1.Container{
+					{
+						Name:  "default",
+						Image: "task-image",
+						Env: []v1.EnvVar{
+							{Name: "FOO", Value: "BAR"},
```
fg91, line 2372: Without the fix, the same env var would be configured twice.

Sovietaced:
I think I briefly ran into this issue when some unit tests were running with improper container names, but I ended up just fixing the unit test container names. I think this is LGTM but I'll need to think about it a little more.

Sovietaced:
Should the same be done for init containers?

fg91:
> Should the same be done for init containers?

Good call! I tested this, the duplicated merge currently also happens for init containers. Fixed in https://github.com/flyteorg/flyte/pull/6733/commits/03cfd1c32c0eac5e0f3deb3d3c313139c2951750

PR #6742, opened 2025-11-14
Common `Filter` to take in `int` values as well

squiishyy:
### Labels

## Docs link

- `main` 
  - \#6583
    - **Common `Filter` to take in `int` values as well** :point\_left:


Makefile
```diff
@@ -95,7 +95,7 @@ buf-python:
 	@$(MAKE) sep
 
 .PHONY: buf
-buf: buf-dep buf-format buf-lint buf-rust buf-python buf-go buf-ts buf-ts-check
+buf: buf-dep buf-format buf-lint buf-rust buf-python buf-go buf-ts buf-ts-check mocks
```
squiishyy, line 98: Adding mock generation to this command as mocks gets deleted when running this, so it should also regenerate it (otherwise people will have to run another command to unblock themselves with changes)

flyteidl2/common/list.proto
```diff
@@ -76,6 +76,8 @@ message Filter {
   // e.g. name or version
   string field = 2;
 
+  // ONLY ONE of the following value fields should be set.
   // Only in the case of a VALUE_IN function, values may contain multiple entries.
   repeated string values = 3;
+  repeated int32 int_values = 4;
```
EngHabu, line 82: You can make this a OneOf without breaking the wire format (You can even change the field name unless we are using json for requests).. Is the reason you are not so you don't break the code compilation?

squiishyy: The main reason is we can't have
```
oneof values {
  repeated string strings = 3;
  repeated int32 ints = 4;
}
```

We'd have to wrap these lists in their own proto struct to get it to work :/

EngHabu: aaah yes yes I hate that, yes :(

EngHabu (approved):

Sovietaced:
Please reopen against the `main` branch

PR #6767, opened 2025-11-28
Add queue dockerfile

WangWang0226:
## Tracking issue

Related to #6703

---

The Queue service previously lacked containerization and Kubernetes deployment manifests, preventing it from running inside a cluster and from performing end-to-end integration tests alongside other Flyte components.

This PR introduces a Dockerfile and K8s deployment configuration so that:

- The Queue service can be built into a reproducible OCI image
- The service runs inside its own Namespace
- RBAC permissions for manipulating `TaskAction` CRs are explicitly defined
- Other components (e.g., Runs) can access Queue through a stable ClusterIP Service

These additions make it possible to validate the Queue service in a real Kubernetes environment.

---

### **1. Dockerfile for Queue service**

- Multi-stage build (Go builder → minimal Alpine runtime)
- `CGO_ENABLED=0` static binary for lightweight deployment
- Bundled default `config.yaml`
- Container health check exposed at `/healthz`

### **2. Kubernetes deployment manifests (Kustomize-based)**

```
queue/config/
  ├── default/        (Kustomization root)
  ├── manager/        (Deployment + Service)
  └── rbac/           (SA + Role + RoleBinding)
```

- Namespace is defined at deployment time (`flyte`)
- RBAC is namespace-scoped and grants minimal permissions to operate `TaskAction` CRs
- Deployment specifies readiness & liveness probes
- ClusterIP Service exposes port **8089** (`http://queue:8089`)

### **3. End-to-end validation**

Verified locally using kind:

- Image builds successfully
- Deployment becomes Ready
- Probes function correctly
- Queue can create/delete `TaskAction` CRs via Kubernetes API

---

#### **1. Build the Docker image**

```bash
docker build -t flyte-queue:latest -f queue/Dockerfile .

```

#### **2. Create a kind cluster**

```bash
kind create cluster --name flyte-dev
kubectl cluster-info --context kind-flyte-dev

```

#### **3. Install Executor CRDs (required for TaskAction recognition)**

```bash
kubectl apply -f executor/config/crd/bases
kubectl get crd taskactions.flyte.org

```

#### **4. Load the image into kind**

```bash
kind load docker-image flyte-queue:latest --name flyte-dev

```

#### **5. Deploy Queue using Kustomize**

```bash
kubectl create namespace flyte
kubectl apply -k queue/config/default

```

> queue/config/default/kustomization.yaml aggregates resources from rbac/ and manager/ and applies them into the flyte namespace.
> 

---

### **6. Verification**

```bash
kubectl get pods -n flyte
kubectl logs deploy/queue -n flyte -f
kubectl get svc queue -n flyte
kubectl port-forward svc/queue -n flyte 8089:8089
curl -f http://localhost:8089/healthz

```

---

### ✅ **B. Functional Verification (Creating/Deleting TaskAction CR)**

Keep the port-forward active, then run:

```bash
go run queue/testclient/main.go

```

This demonstrates:

- Queue can create and delete `TaskAction` CRs
- RBAC rules allow the required operations
- The Kubernetes API interaction logic is functioning correctly

You may also observe CR lifecycle changes:

```bash
kubectl get taskactions -n flyte -w
kubectl describe taskaction <name> -n flyte

```

---

### Cleanup

```bash
kubectl delete -k queue/config/default
kind delete cluster --name flyte-dev

```
---

## Notes / Compatibility

- Requires Executor CRDs (`executor/config/crd/bases`); without these, `TaskAction` cannot be created.
- RBAC is namespace-scoped; cluster-wide operation would require a ClusterRole/ClusterRoleBinding.
- In-cluster service discovery:
    - `http://queue:8089`
    - or full DNS: `queue.flyte.svc.cluster.local:8089`

---

## Labels

- **added**: Dockerfile for Queue service
- **added**: K8s manifests (Namespace, SA, RBAC, Deployment, Service)
- **added**: End-to-end validation for Queue → TaskAction CR
- **changed**: Queue can now run inside Kubernetes and be consumed by other services

---

## Docs link

N/A (deployment and testing documentation can be added to README in a future PR)

---

### DIFF
--- a/queue/Dockerfile
+++ b/queue/Dockerfile
@@ -29,9 +29,5 @@ COPY --from=builder /workspace/queue/config.yaml .
 
 EXPOSE 8089
 
-HEALTHCHECK --interval=30s --timeout=3s --start-period=5s --retries=3 \
-  CMD wget --no-verbose --tries=1 --spider http://localhost:8089/healthz || exit 1
-
 ENTRYPOINT ["./queue-service"]
 CMD ["--config", "config.yaml"]
-
--- a/queue/config/rbac/role.yaml
+++ b/queue/config/rbac/role.yaml
@@ -3,7 +3,13 @@ kind: Role
 metadata:
   name: queue-role
 rules:
-  - apiGroups: ["flyte.org"]
-    resources: ["taskactions"]
-    verbs: ["get", "list", "watch", "create", "delete"]
-
+- apiGroups:
+  - flyte.org
+  resources:
+  - taskactions
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - watch
--- a/queue/config/rbac/service_account.yaml
+++ b/queue/config/rbac/service_account.yaml
@@ -1,5 +1,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
+  labels:
+    app.kubernetes.io/name: queue
+    app.kubernetes.io/managed-by: kustomize
   name: queue-sa
-


queue/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+WORKDIR /app
+
+# Copy binary and default config
+COPY --from=builder /workspace/queue/queue-service .
+COPY --from=builder /workspace/queue/config.yaml .
+
+EXPOSE 8089
+
+HEALTHCHECK --interval=30s --timeout=3s --start-period=5s --retries=3 \
+  CMD wget --no-verbose --tries=1 --spider http://localhost:8089/healthz || exit 1
```
popojk: Should we remove the docker health check here as we already set health check in K8s layer?

queue/k8s/queue.yaml
```diff
@@ -0,0 +1,85 @@
+apiVersion: v1
```
popojk: Could we move this K8s yaml file into /queue/config dir? Please refer to the folder structure of executor/config.

queue/k8s/queue.yaml
```diff
@@ -0,0 +1,85 @@
...
+  namespace: flyte
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: queue-role
+  namespace: flyte
+rules:
+  - apiGroups: ["flyte.org"]
+    resources: ["taskactions", "taskactions/status"]
```
popojk: Should we remove the "taskactions/status" resource as it is not used in queue service?

queue/config/manager/manager.yaml
```diff
@@ -0,0 +1,85 @@
...
+        app: queue
+    spec:
+      serviceAccountName: queue-sa
+      containers:
+        - name: queue
+          image: flyte-queue:latest
+          imagePullPolicy: IfNotPresent
+          args: ["--config", "/app/config.yaml"]
+          ports:
+            - containerPort: 8089
```
popojk, line 22: It is ok for now. However in the future we should set separate ports for http, gRPC, and health check when we add more functionalities to queue service.

queue/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+ENTRYPOINT ["./queue-service"]
+CMD ["--config", "config.yaml"]
```
popojk: I think maybe we don't need the config.yaml here? Right now we have default config [here](https://github.com/flyteorg/flyte/blob/91ba87d5e4d0a0b99706d87a568c7cbf2b94cb21/queue/config/config.go#L11-L19). In the future we will use helm chart to overwrite the config settings if needed.

queue/Dockerfile
WangWang0226: Yes, since we only run the container in k8s, we can keep the health check in k8s layer only. I've removed the one in dockerfile.

queue/k8s/queue.yaml
WangWang0226: Thank you for reminding! I've refactored the `queue.yaml` to `/queue/config` dir following the structure of `executor/config`.

queue/k8s/queue.yaml
WangWang0226: Yes, thanks for reminding! I've removed it.

queue/config/manager/manager.yaml
WangWang0226: Thanks for the heads up! That makes sense, I’ll keep it in mind for future updates!

queue/config/manager/manager.yaml
```diff
@@ -0,0 +1,48 @@
...
+      app: queue
+  template:
+    metadata:
+      labels:
+        app: queue
+    spec:
+      serviceAccountName: queue-sa
+      containers:
+        - name: queue
+          image: flyte-queue:latest
```
machichima, line 18: just a note, we should discuss about the image name later on. But we can keep this now and modify in the future when the image name is set 👍

queue/config/rbac/role.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: queue-role
+rules:
+  - apiGroups: ["flyte.org"]
+    resources: ["taskactions"]
+    verbs: ["get", "list", "watch", "create", "delete"]
```
machichima: nit: could we make the format same as the one in `executor` rbac config?

https://github.com/flyteorg/flyte/blob/d5838d45c06345437fded9f70eadb6182bb1cb35/executor/config/rbac/role.yaml#L11-L18

queue/config/rbac/service_account.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: queue-sa
+
```
machichima: Could we also add following labels here like what we did in executor?


```yaml
app.kubernetes.io/name: queue
app.kubernetes.io/managed-by: kustomize
```

https://github.com/flyteorg/flyte/blob/d5838d45c06345437fded9f70eadb6182bb1cb35/executor/config/rbac/service_account.yaml#L4-L6

queue/Dockerfile
machichima: Sorry just to make sure if you removed it or not, as I still see `HEALTHCHECK` in `Dockerfile`

WangWang0226: Sorry my bad. I've removed it!

queue/config/manager/manager.yaml
WangWang0226: No problem, I'll keep it in mind!

queue/config/rbac/role.yaml
WangWang0226: Thanks for the reminding, I've aligned with the format as executor's `role.yaml`.

queue/config/rbac/service_account.yaml
WangWang0226: Thanks, I've fixed this!

machichima (approved):
This is awesome! Thank you for this

PR #6802, opened 2025-12-12
Prevent panic when retry attempt value in map task gets larger than bitarray was allocated for

fg91:

With a default configuration of `max-node-retries-system-failures=3` and `default-max-attempts=1`, the following workflow leads to a panic in the node array handler:

```py
@task(
    requests=Resources(cpu="50m", mem="250Mi"),
    limits=Resources(cpu="100m", mem="250Mi"),
    retries=8,
)
def test(inp: int) -> None:
    if inp == 2:
        # Go OOM on purpose or any other "retriable user error"
        a = " " * (300 * 1024 * 1024)


@workflow 
def wf1() -> None:
    map_task(test)(inp=[i for i in range(3)])
```

The panic occurs [here](https://github.com/flyteorg/flyte/blob/99f2ddd9f41022ae77604bb31f7a83a8d3942df6/flytestdlib/bitarray/compact_array.go#L34) when validating whether the current value of attempts can be tracked in a bitarray that was allocated for this purpose.

The following goes wrong:

Currently the `retries=8` set on the task decorator is *not* taken into account when determining the max value of the number of attempts: [Here](https://github.com/flyteorg/flyte/blob/99f2ddd9f41022ae77604bb31f7a83a8d3942df6/flytepropeller/pkg/controller/nodes/array/handler.go#L255) in the array node handler, for the example workflow above, `nCtx.Node().GetRetryStrategy()` is `nil`. `maxAttemptsValue` is therefore `max-node-retries-system-failures=3` + `default-max-attempts=1` equals to `maxAttemptsValue=4`. When creating the bit array to track the retry attempts [here](https://github.com/flyteorg/flyte/blob/99f2ddd9f41022ae77604bb31f7a83a8d3942df6/flytepropeller/pkg/controller/nodes/array/handler.go#L273), it is determined that 3 bits are needed to store values that cannot be higher than 4.

However, the task is actually retried 8 times. Up to attempt 7 this still works because only 3 bits are needed. But attempt 8 then panics [here](https://github.com/flyteorg/flyte/blob/99f2ddd9f41022ae77604bb31f7a83a8d3942df6/flytestdlib/bitarray/compact_array.go#L34).

When determining the max value that can be stored in the bit array, we need to account not only for `nCtx.Node().GetRetryStrategy()` but also for `nCtx.Node().GetArrayNode().GetSubNodeSpec().GetRetryStrategy()` which does contain the desired `retries=8`.

During my debugging, `nCtx.Node().GetRetryStrategy()` was actually always `nil` but I cannot say for sure this is always the case. So I propose to allocate the bit array to have space for the max of both values to be on the safe side.

## How was this patch tested?

The example above panics on latest master but works with the changes in this PR.

 <div id='description'>
<h3>Summary by Bito</h3>

<ul>

<li>Addresses a critical bug in the node array handler that could lead to a panic when the retry attempt value exceeds the allocated bit array size.</li>

<li>Ensures that both the node's retry strategy and the sub-node's retry strategy are considered when determining the maximum number of attempts.</li>

<li>Enhances the stability of workflows that involve retries, particularly in scenarios where the retry count is high.</li>

<li>Overall, this pull request addresses a critical bug in the node array handler, improving stability in retry workflows.</li>

</ul>

</div>

### DIFF
--- a/flytepropeller/pkg/controller/nodes/array/handler_test.go
+++ b/flytepropeller/pkg/controller/nodes/array/handler_test.go
@@ -1196,6 +1196,101 @@ func TestHandleArrayNodePhaseExecutingSubNodeFailures(t *testing.T) {
 	}
 }
 
+func TestHandleArrayNodePhaseExecutingSubNodeRetryBitArraySizing(t *testing.T) {
+	// Regression test: when the sub-node spec has a RetryStrategy with MinAttempts
+	// higher than config.DefaultMaxAttempts, the bitarray for tracking retry attempts
+	// must be sized to accommodate the sub-node's retry budget. Previously, before #6802,
+	// only the array node's retry strategy was considered for sizing, causing a panic when the
+	// attempt count exceeded the bitarray capacity:
+	// panic: Value [2] is too big. Max value is [1].
+	ctx := context.Background()
+
+	config.GetConfig().NodeConfig.DefaultMaxAttempts = 1
+	config.GetConfig().NodeConfig.MaxNodeRetriesOnSystemFailures = 0
+	config.GetConfig().NodeConfig.IgnoreRetryCause = false
+
+	// The task itself is configured with 3 attempts, the array node
+	// defaults to the config of 1 attempt max
+	subNodeMinAttempts := 3
+	nodeSpec := v1alpha1.NodeSpec{
+		ID: "foo",
+		ArrayNode: &v1alpha1.ArrayNodeSpec{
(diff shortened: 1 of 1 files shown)


pingsutw (approved):

fg91:
I'll try to add a unit test let's not merge yet.

fg91:
> Is unit test coverage possible?

@Sovietaced I added a unit test that follows the pattern of `TestHandleArrayNodePhaseExecutingSubNodeFailures` and which would panic without the fix.

PR #6809, opened 2025-12-18
WIP: Fix stuck workflow nodes that are starved by map tasks exceeding the max parallelism budget

fg91:

In this minimal reproducible example ... 

```py
from flytekit import task, workflow, map_task
import time

@task
def sleep(inp: int) -> None:
    time.sleep(inp)


@workflow
def wf() -> None:
    map_task(sleep)(inp=[3600 for _ in range(30)])  # array node with long running tasks. 30 is chosen to exceed default max parallelism of 25
    sleep(inp=30)  # Task that runs more quickly
```

... I observe the following and in my opinion undesirable behaviour:

* Both the map task and the node after it start running.
* The `n1` pod succeeds quickly:

    Screenshot 2025-12-18 at 11 44 04

* The `n1` node stays in `Queued` state in the Flyte UI despite the underlying pod already having succeeded:
    Screenshot 2025-12-18 at 11 43 45
* The `n1` node is updated in the Flyte UI from `Queued` to `Succeeded` only after the map task (or at least enough tasks within it) completes as well which could be hours later.

### Reason for this behaviour

* In the [`RecursiveNodeHandler` ](https://github.com/flyteorg/flyte/blob/be517bd6fd6410d234cd25046c0d496682094c4a/flytepropeller/pkg/controller/nodes/executor.go#L178C33-L178C53) which traverses the workflow graph, we check whether the current degree of parallelism has exceeded the max parallelism:

  ```go
  func (c *recursiveNodeExecutor) RecursiveNodeHandler(ctx context.Context, execContext executors.ExecutionContext, dag executors.DAGStructure, nl executors.NodeLookup, currentNode v1alpha1.ExecutableNode) ( interfaces.NodeStatus, error) {
      ...
      if IsMaxParallelismAchieved(ctx, currentNode, nodePhase, execContext) {
          ...
          return interfaces.NodeStatusRunning, nil
  ```

* For `n1`, `IsMaxParallelismAchieved` gives `true`:
  * In the example above, when evaluating `n1`, the current degree of parallelism takes a value of `31` which is bigger than the default max parallelism of `25`.
  * The degree of parallelism is increased by 1 for the array node itself [here](https://github.com/flyteorg/flyte/blob/be517bd6fd6410d234cd25046c0d496682094c4a/flytepropeller/pkg/controller/nodes/array/handler.go#L473) and 30 times for each of the tasks in the map task [here](https://github.com/flyteorg/flyte/blob/be517bd6fd6410d234cd25046c0d496682094c4a/flytepropeller/pkg/controller/nodes/task/handler.go#L819).
  * This means that the task phase of `n1` will be evaluated only once less than 25 tasks are still running within the map task `n0`.

* As a user, I would expect `n1` to be marked as succeeded immediately after the pod completes and not hours later when enough of the array node tasks complete.
* Alternatively, if there is no "parallelism budget" for `n1`, I would expect `n1` to not start at all until `n0` is done. But as a user I wouldn't expect this "mixed" behaviour with a node that is seemingly stuck for hours despite having completed.

## Discussion

The behaviour can be avoided when modifying the parallelism tracking logic to count the map task as `1` and not as `1 + 30` (in this example). I would like to discuss which of the two is the intended behaviour.

### DIFF
--- a/flytepropeller/pkg/controller/nodes/task/handler.go
+++ b/flytepropeller/pkg/controller/nodes/task/handler.go
@@ -3,7 +3,9 @@ package task
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"runtime/debug"
+	"strings"
 	"sync"
 	"time"
 
@@ -677,6 +679,20 @@ func (t Handler) Handle(ctx context.Context, nCtx interfaces.NodeExecutionContex
 	}
 
 	tCtx, err := t.newTaskExecutionContext(ctx, nCtx, p)
+
+	// Note: There is no intention of merging this as is.
+	// This is only supposed to demonstrate what would be needed to avoid the issue raised in this PR.
+	// If we agree that the desired behaviour is that the parallelism isn't incremented for each pod
+	// in a map task, we'd need a proper mechanism to detect whether we are in a map task which doesn't
+	// involve reflection and string matching.
+	ctxType := reflect.TypeOf(nCtx).String()
+	var isArrayNode bool
+	if strings.Contains(ctxType, "arrayNodeExecutionContext") {
+		isArrayNode = true
+	} else {
+		isArrayNode = false
+	}
+
 	if err != nil {
 		return handler.UnknownTransition, errors.Wrapf(errors.IllegalStateError, nCtx.NodeID(), err, "unable to create Handler execution context")
 	}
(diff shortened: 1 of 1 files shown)


pvditt:
@fg91 apologies for the significant delay. I'll take a look at this issue today

fg91:
> @fg91 apologies for the significant delay. I'll take a look at this issue today

Thank you!!

fg91:
Closing as fixed by @pvditt in https://github.com/flyteorg/flyte/pull/6929#pullrequestreview-3832358041.
Thank you!

PR #6813, opened 2025-12-23
Add task service

machichima:
## Tracking issue

Related to #6706

## Why are the changes needed?

Add task service for flyte-v2

## What changes were proposed in this pull request?

Main changes

* `runs/service/task_service.go`: Implemented the TaskService handlers for deploying, retrieving, and listing tasks.

* `runs/repository/models`:
    * `task.go`: Defined the Task GORM model.
    * `task_spec.go`: Added the TaskSpec model.

* `runs/repository/interfaces/`: Defined generic repository interfaces for tasks and actions, including reusable ListResourceInput, Filter, and SortParameter types.
    * `task.go`: interface for task repo
    * `common.go`, `filter.go`, `sort.go`, `request.go`: generic query utils interface
* `runs/repository/impl/`: Provided the GORM-based implementation for task operations and generic database utilities for filtering and pagination.
    * `task.go`: DB operation for task repo
    * `common.go`, `filter.go`, `sort.go`, `request.go`: generic query utils

* `runs/migrations/migrations.go`: Add auto migrate for tasks and task_specs tables.
* `runs/repository/mocks/`: Generated mockery-based mocks for the repository layer.
* `runs/tests/scripts/`: Added and updated shell scripts using `buf curl` for manual verification of task creation and listing.

## How was this patch tested?

- Unit tests (run `make test` under `./runs` and ensure all test passed)
- Start service locally and run scripts in `runs/tests/scripts/` for testing (see https://github.com/flyteorg/flyte/pull/6813#issuecomment-3686002341)

### Labels

Please add one or more of the following labels to categorize your PR:
- **added**: For new features.
- **changed**: For changes in existing functionality.
- **deprecated**: For soon-to-be-removed features.
- **removed**: For features being removed.
- **fixed**: For any bug fixed.
- **security**: In case of vulnerabilities

This is important to improve the readability of release notes.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

## Docs link

Update `runs/README.md`

PR #6825, opened 2025-12-29
Fix: Add timestamps to task creation

yuhuan130:
## Tracking issue

Related to #6813 (Initial task service implementation)

Timestamps (`created_at`, `updated_at`) are not being populated in the `tasks` table, which causes: 
1. `ListTasks` sorting by `created_at` to be ineffective (line 158 in `impl/task.go`)
2. API responses to show incorrect `deployedAt: "0001-01-01T00:00:00Z"` instead of actual deployment time
3. No audit trail for when tasks were created or last updated

Added timestamp handling to the `CreateTask` function in `runs/repository/impl/task.go`:

1. **Added columns to INSERT statement**: `created_at` and `updated_at`
2. **Set values**:  Both timestamps use `time.Now()` when creating new tasks
3. **ON CONFLICT behavior**: When a task is re-deployed (same version), only `updated_at` is refreshed while `created_at` preserves the original creation time
4. **Import added**: Added `time` package import

**Files changed:**
- `runs/repository/impl/task.go` (lines ~3, 35, 47, 64-65)

### Manual testing with local SQLite database: 

**Test 1: Create new task**
```bash
# Started runs service locally
make run

# Created a task using test script
./runs/tests/scripts/create_task.sh

# Verified in database
sqlite3 runs/runs.db "SELECT name, version, created_at, updated_at FROM tasks"
```

(base) yuhuan@yuhuan flyte runstestsscriptslist_tasks sh
Screenshot 2025-12-29 at 13 37 38

### DIFF
--- a/runs/repository/impl/task_test.go
+++ b/runs/repository/impl/task_test.go
@@ -73,7 +73,6 @@ func TestCreateTask(t *testing.T) {
 	assert.Equal(t, task.FunctionName, retrieved.FunctionName)
 	assert.WithinDuration(t, startTime, retrieved.CreatedAt, 1*time.Second, "created_at should be close to now")
 	assert.Equal(t, retrieved.CreatedAt, retrieved.UpdatedAt, "created_at and updated_at should be equal")
-
 }
 
 func TestGetTask_NotFound(t *testing.T) {


runs/repository/impl/task_test.go
```diff
@@ -69,6 +70,8 @@ func TestCreateTask(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, task.Environment, retrieved.Environment)
 	assert.Equal(t, task.FunctionName, retrieved.FunctionName)
+    assert.False(t, retrieved.CreatedAt.IsZero(), "created_at should not be zero")
+    assert.False(t, retrieved.UpdatedAt.IsZero(), "updated_at should not be zero")
```
machichima: nit: We can directly check if created at and updated at is set correctly, then we do not need `TestCreateTask_Timestamps`
```suggestion
	assert.WithinDuration(t, startTime, retrieved.CreatedAt, 2*time.Second, "created_at should be close to now")
	assert.WithinDuration(t, startTime, retrieved.UpdatedAt, 2*time.Second, "updated_at should be close to now")
```

note: also add `startTime := time.Now()` before `repo.CreateTask`

yuhuan130: That's actually right. @machichima I've merged it into TestCreateTask, would you mind taking a glance of it? Thank u.

PR #6826, opened 2025-12-29
[Executor] Update taskaction cr

machichima:

Based on the k8s api convention ([here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)), it mentioned:

> The pattern of using phase is deprecated. Newer API types should use conditions instead

It's better for us to remove Phase and Message and just use Conditions in our executor to follow the best practice.

Controller changes:

- Use `Conditions` for the reconcile logic, following best practice mentioned in [api-conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)
    - Use common top-level condition which summarizes more detailed conditions (`Succeed`, `Failed`, `Progressing`). Use `Reason` for status detail
- Follow how k8s job controller did, check the terminate condition first, then going through others ([link](https://github.com/kubernetes/kubernetes/blob/1d16b7f3a880144ff3447dbd93ba200381e9c414/pkg/controller/job/job_controller.go#L859-L859))
- Add fields in `Spec`, and update sample YAML
- Add more `kubebuilder:printcolumn` for observability
- Add `workflowconnect` mock and update taskaction controller test

Development Changes:

- Update `go.mod`, `Dockerfile`, and `Makefile` to build `executor` successfully
- Add `DEVELOPMENT.md` to guide through setting up local development environment and apply examples
- Update README to provide more descriptions

## How was this patch tested?

- task action controller unit test

```sh
❯ make test
/Users/nary/workData/open-source/flyte-v2/executor/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/Users/nary/workData/open-source/flyte-v2/executor/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
Setting up envtest binaries for Kubernetes version 1.34...
/Users/nary/workData/open-source/flyte-v2/executor/bin/k8s/1.34.1-darwin-arm64KUBEBUILDER_ASSETS="/Users/nary/workData/open-source/flyte-v2/executor/bin/k8s/1.34.1-darwin-arm64" go test $(go list ./... | grep -v /e2e) -coverprofile cover.out
        github.com/flyteorg/flyte/v2/executor/api/v1            coverage: 0.0% of statements
        github.com/flyteorg/flyte/v2/executor/cmd               coverage: 0.0% of statements
ok      github.com/flyteorg/flyte/v2/executor/pkg/controller    8.052s  coverage: 39.5% of statements
        github.com/flyteorg/flyte/v2/executor/test/utils                coverage: 0.0% of statements
```

- Manual test, following `DEVELOPMENT.md` for setup and apply example with `kubectl apply -k config/samples/`. Ensure the conditions change correctly

```sh
❯ k describe taskactions taskaction-sample
Name:         taskaction-sample
Namespace:    default
Labels:       app.kubernetes.io/managed-by=kustomize
              app.kubernetes.io/name=executor
Annotations:  <none>
API Version:  flyte.org/v1
Kind:         TaskAction
Metadata:
  Creation Timestamp:  2025-12-30T02:22:35Z
  Generation:          1
  Resource Version:    991
  UID:                 12fa6800-91de-4b11-9501-66b4ebd43df6
Spec:
  Action Name:      sample-task
  Domain:           dev
  Input Uri:        /tmp/input
  Org:              demo
  Project:          default
  Run Name:         sample-run
  Run Output Base:  /tmp/output
Status:
  Conditions:
    Last Transition Time:  2025-12-30T02:26:47Z
    Message:               TaskAction has completed
    Reason:                Completed
    Status:                False
    Type:                  Progressing
    Last Transition Time:  2025-12-30T02:26:47Z
    Message:               TaskAction completed successfully
    Reason:                Completed
    Status:                True
    Type:                  Succeeded
  State Json:              {"actionId":"sample-run/sample-task","phase":"Succeeded","timestamp":"2025-12-30T02:26:47Z"}
Events:                    <none>

```

### Labels


executor/Makefile
```diff
@@ -117,7 +117,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
 docker-build: ## Build docker image with the manager.
-	$(CONTAINER_TOOL) build -t ${IMG} .
+	$(CONTAINER_TOOL) build -t ${IMG} -f Dockerfile ..
```
machichima, line 120: Set working dir to root so that we can copy all required things in `Dockerfile`

gen/go/flyteidl2/workflow/workflowconnect/mocks/mock_StateServiceClient.go
machichima, line 1: This is used in `executor/pkg/controller/taskaction_controller_test.go`, which is the main reason we are generating mocks for `workflowconnect`

go.mod
machichima, line 1: Modify this as we got error when doing `make build` in `executor/`. The packages version is incompatible with Kubernetes v0.28.x packages

image

EngHabu (approved):

executor/config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: flyteorg/executor
+  newTag: nightly
```
popojk: Why do we change the tag to nightly?

executor/DEVELOPMENT.md
```diff
@@ -0,0 +1,154 @@
...
+
+To keep consistent code generation and testing result, it is recommended to remove outdated binaries installed by the Makefile.
+
+```sh
+make clean
+```
+
+## Local development on kind cluster
+
+Following steps requires you to switch your workign directory to the `executor/`.
```
popojk: nit: 
```suggestion
Following steps requires you to switch your working directory to the `executor/`.
```

executor/DEVELOPMENT.md
```diff
@@ -0,0 +1,154 @@
...
+You can apply the samples from the config/sample using the following command:
+
+```sh
+kubectl apply -k config/samples/
+```
+
+You can see the `TaskAction` CRD created:
+
+```sh
+❯ k get taskactions
```
popojk: nit: 
```suggestion
❯ kubectl get taskactions
```

executor/DEVELOPMENT.md
```diff
@@ -0,0 +1,154 @@
...
+NAME                RUN          ACTION        STATUS      AGE
+taskaction-sample   sample-run   sample-task   Completed   59m
+```
+
+Or use `-o wide` to view more details:
+
+```sh
+❯ k get taskactions -o wide
```
popojk: ditto

executor/DEVELOPMENT.md
```diff
@@ -0,0 +1,154 @@
...
+IMG=flyteorg/executor:nightly make docker-build
+```
+
+3. Load image into kind cluster
+
+```sh
+kind load docker-image flyteorg/executor:nightly --name flytev2
+```
+
+4. Install CRD in to the cluster
```
popojk: nit: 
```suggestion
4. Install CRD into the cluster
```

executor/DEVELOPMENT.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Run executor outside the cluster 
+
+1. Create a kind cluster
+
+```sh
+kind create cluster --image=kindest/node:v1.26.0 --name flytev2
+```
+
+2. Install CRD in to the cluster
```
popojk:
```suggestion
2. Install CRD into the cluster
```

executor/DEVELOPMENT.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Local development on kind cluster
+
+Following steps requires you to switch your workign directory to the `executor/`.
+
+```sh
+cd executor
+```
+
+### Run executor inside the cluster
```
popojk, line 41: Well, it's good for now, but I think maybe we can wrap all the steps into one makefile script so that it saves more time for OSS developer to run the service locally, WDYT?

executor/pkg/controller/taskaction_controller.go
```diff
@@ -93,98 +94,120 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-	case PhaseRunning:
-		// Running → Succeeded (simulated execution)
-		nextPhase = PhaseSucceeded
-		logger.Info("Transitioning from Running to Succeeded",
-			"name", taskAction.Name, "action", actionSpec.ActionId.Name)
+	// Sequential condition evaluation for Progressing
+	progressingCond := findConditionByType(taskAction.Status.Conditions, flyteorgv1.ConditionTypeProgressing)
+	if progressingCond != nil && progressingCond.Status == metav1.ConditionTrue {
+		// Check the Reason to determine sub-state
+		if progressingCond.Reason == string(flyteorgv1.ConditionReasonExecuting) {
```
popojk, line 130: nit: Right now we handle all condition cases in various if blocks, do you think it is more readable to handle them with case blocks? Like:
```
switch progressingCond.Reason {
case string(flyteorgv1.ConditionReasonExecuting):
   ...
```

executor/config/manager/kustomization.yaml
machichima: I think `make deploy` changes this automatically. I will change it back

executor/DEVELOPMENT.md
machichima: Yes, this would be great! I think we can consider this together with starting all services in `manager/`.

executor/pkg/controller/taskaction_controller.go
machichima: Great point! 
Conditions are different from Phases, they are more like status observations. A key distinction is that Conditions allow multiple states to be present at the same time (e.g., a resource can be both `Ready` and `Progressing`). Therefore, using a state-machine-like method (like `switch`) here violates the nature of Conditions.

Also, one of the main reasons Kubernetes deprecated Phase is that Enums are not extensible. If we wanted to add a new phase, it would be a breaking change for old clients (see discussion [[here](https://github.com/kubernetes/kubernetes/issues/7856)](https://github.com/kubernetes/kubernetes/issues/7856)). We shouldn't use a strict state-machine format with Conditions. Instead, we should use the "filter pattern" like we did here (e.g., if condition X is true -> do something), which allows us to handle priorities correctly.

machichima: Actually I found that my approach does not follow the best practice. We shouldn't write logic based on `Reason`, as this is intended to be used in concise output only.

We can keep this now as we do not actually create resources. Once we perform true operation and create resources, we should track the resource directly rather than depends on the `Reason` field

popojk (approved):

PR #6852, opened 2026-01-09
Feat: Make RayCluster head node ingress optional

fg91:

Currently, the flytepropeller ray plugin hardcodes the creation of an ingress for the ray cluster's head node, see [here](https://github.com/flyteorg/flyte/blob/8877200be9cbf98093e22026de51c17c9d0fdacd/flyteplugins/go/tasks/plugins/k8s/ray/ray.go#L133).

This is a potential security problem as the ray head node might be exposed to the public internet in clusters with a default ingress class without the user knowing.

* Add field `EnableIngress` to the ray plugin config
* Disable the creation of the ingress by default.

## How was this patch tested?

* Added a unit test.

## Related PRs

https://github.com/flyteorg/flytekit/pull/3365


flyteidl/protos/flyteidl/plugins/ray.proto
```diff
@@ -39,6 +39,9 @@ message HeadGroupSpec {
   map<string, string> ray_start_params = 1;
   // Pod Spec for the ray head pod
   core.K8sPod k8s_pod = 2;
+
+  // Optional. Whether to enable an ingress on the ray cluster head node
+  bool enable_ingress = 3;
```
pingsutw: Why don't we add it to the plugin config? https://github.com/flyteorg/flyte/blob/8877200be9cbf98093e22026de51c17c9d0fdacd/flyteplugins/go/tasks/plugins/k8s/ray/config.go#L61

fg91: I added it to the head node config because that's where this is configured in the RayCluster CRD and some of the other fields of the head node config can be configured from python too.
But I don't have a strong opinion, could change to the plugin config instead of the task_config in the task decorator if you prefer. 
My main concern is that it should be disabled by default and opt-in to not accidentally expose things :)

pingsutw: I think this config is more related to the cluster settings, so putting it in the plugin config map might be better.

pingsutw: We could add it to the Ray proto in the future if anyone wants to turn it off for certain tasks only, although i think it's not the common use case.

fg91: Agree. Can you please take another look?
https://github.com/flyteorg/flyte/pull/6852/commits/1f5adfdda11a75089b372f389c26f6f734c72f99

flyteplugins/go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -130,9 +130,10 @@ func (rayJobResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsC
 }
 
 func constructRayJob(taskCtx pluginsCore.TaskExecutionContext, rayJob *plugins.RayJob, objectMeta *metav1.ObjectMeta, taskPodSpec v1.PodSpec, headNodeRayStartParams map[string]string, primaryContainerIdx int, primaryContainer v1.Container) (*rayv1.RayJob, error) {
-	enableIngress := true
 	cfg := GetConfig()
 
+	enableIngress := cfg.EnableIngress
```
machichima, line 135: nit: I think we can put `EnableIngress:  & cfg.EnableIngress ` directly in `HeadGroupSpec`, as it's the only place that use this

pingsutw (approved):

flyteplugins/go/tasks/plugins/k8s/ray/ray.go
fg91: We could but the `ServiceType` - which is also part of the [RayCluster's `HeadGroupSpec`](https://github.com/ray-project/kuberay/blob/b8451500d4276e5c27a1025c732c562eaac04a37/ray-operator/apis/ray/v1/raycluster_types.go#L135) - is also in our top level plugin config. My preference would be to try to keep it consistent here at least?

This is not a strong opinion though.

machichima:
@fg91 I think the flyteplugins test CI failed. Could you please take a look? Thanks!

fg91:
https://github.com/flyteorg/flyte/pull/6905 and this PR tackled the same issue (ray clusters being hardcoded to create ingresses) in parallel with the difference that I argue for actually disabling the ingress by default, see PR description:

> This is a potential security problem as the ray head node might be exposed to the public internet in clusters with a default ingress class without the user knowing.

As a platform engineer, I'd admittedly be quite alarmed to discover that jobs launched by platform users have been creating potentially public ingresses without me being aware. This is why I don't like the default of enabling the ingress.

@EngHabu [here](https://github.com/flyteorg/flyte/pull/6905#discussion_r2780127649) you said

> I like booleans to default to false... So I think this should be DisableIngress

upon which you, @mickjermsurawong-openai, changed your initial `EnableIngress` (with default true) to `DisableIngress` with (default false).

Would you be ok with merging this PR too which goes back to `EnableIngress` but with default false?

This PR also adds a unit test for `BuildResource` + enabling/disabling the ingress.

mickjermsurawong-openai:
Fabio, curious without ingress, is Ray dashboard still available? My understanding is no (pls correct me i just started using the ray plugin). My thought is i'd setup KubeRay API for dashboard instead, although it's experimental feature, and we can set pod-template for ray job to add the right annotation.

mickjermsurawong-openai:
I'm fine changing the flag back to EnableIngress!

machichima:
> Fabio, curious without ingress, is Ray dashboard still available? My understanding is no (pls correct me i just started using the ray plugin). My thought is i'd setup KubeRay API for dashboard instead, although it's experimental feature, and we can set pod-template for ray job to add the right annotation.

Hi @mickjermsurawong-openai, 
Do you mean KubeRay "APIServer"? I think by using APIServer you can connect to Ray dashboard without needed to set the annotation

machichima:
> As a platform engineer, I'd admittedly be quite alarmed to discover that jobs launched by platform users have been creating potentially public ingresses without me being aware. This is why I don't like the default of enabling the ingress.

I think it's safer to keep the default as enable that aligns with the current behavior, to prevent users's confusion. For example when they upgrade the newer flyte version, and Ray Dashboard cannot be accessed, they might not know what's happening

EngHabu:
@fg91 can you confirm whether this is needed for accessing the ray dashboard through proxy? is this flag ONLY about creating the Ingress object?

fg91:
> @fg91 can you confirm whether this is needed for accessing the ray dashboard through proxy? is this flag ONLY about creating the Ingress object?

I just tested it again and can confirm that with `enableIngress: false` there is still a `NodePort` (default) service that can be port-forwarded to access the ray dashboard.
This flag only controls the creation of an ingress - which might accidentally make the dashboard public in some cases :/

EngHabu (approved):

Issue #6824, opened 2025-12-29
[Feat] Refactor and finalize Run service


machichima:
## Description

There are few tasks left in run service:

1. Some function pass protobuf request directly to the repository. This violates separation of concerns and creates a dependency from the repository layer to the API layer. We should refactor this to follow how we did in `task_service`

https://github.com/flyteorg/flyte/blob/5c20324bf44a22c12deee974fbaa0b984f4e0b51/runs/service/run_service.go#L53-L53

https://github.com/flyteorg/flyte/blob/5c20324bf44a22c12deee974fbaa0b984f4e0b51/runs/service/task_service.go#L66-L66

2. There's some TODOs left that we need to build the response from the run model

https://github.com/flyteorg/flyte/blob/5c20324bf44a22c12deee974fbaa0b984f4e0b51/runs/service/run_service.go#L163-L169

Can also follow what we have in task service

https://github.com/flyteorg/flyte/blob/5c20324bf44a22c12deee974fbaa0b984f4e0b51/runs/service/task_service.go#L82-L82

yuhuan130:
#take

PR #6844, opened 2026-01-05
[Fix] Ignore cargo lock

machichima:
## Tracking issue

Related to: #6845

## Why are the changes needed?

`Cargo.lock` in `gen/rust` is updated frequently even if there's no proto update. Every time we do `make gen`, we can see some version update in `gen/rust/Cargo.lock` (for example: [here](https://github.com/flyteorg/flyte/actions/runs/20701630862/job/59424791583)). It's is inconvenient to run `make gen` every time even if there's no proto changes.

## What changes were proposed in this pull request?

Add `Cargo.lock` to gitignore. It will not affect the usage when using `flyteidl2` as the package:
> "Cargo.lock does not affect the consumers of your package, only Cargo.toml does that"
mentioned in [the cargo book](https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control)

## How was this patch tested?

Ensure `make gen` will not make changes if there's no proto udpate

### Labels

Please add one or more of the following labels to categorize your PR:
- **added**: For new features.
- **changed**: For changes in existing functionality.
- **deprecated**: For soon-to-be-removed features.
- **removed**: For features being removed.
- **fixed**: For any bug fixed.
- **security**: In case of vulnerabilities

This is important to improve the readability of release notes.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [x] All commits are signed-off.

PR #6829, opened 2025-12-30
[v2] Refactor: Define all mockery in .mockery.yaml

majiayu000:
## Summary
This PR fixes #6815

Refactor to define all mockery configuration in `.mockery.yaml` instead of using `//go:generate mockery` comments in individual files.

This PR modify the way we generate the mocks following the best practice described in: https://vektra.github.io/mockery/v2.53/configuration

- Define packages for mocking in `.mockery.yaml`
- Use `EXPECT()` rather than `On()` (in docs: Use `with-expecter: True` to generate `EXPECT()` methods for your mocks. This is the preferred way to set up your mocks.) 

## Changes
- Updated `.mockery.yaml` with configurations for:
  - flyteplugins packages (pluginmachinery/core, workqueue, k8s, catalog, io)
  - flytestdlib packages (cache, utils, storage, fastcheck, random)
  - runs/repository/interfaces (already had config, removed duplicate directive)
- Removed all `//go:generate mockery` comments from interface files

This change consolidates all mock generation configuration into a single location, making it easier to maintain and ensuring consistency across the codebase.

**File changes**: 13 files changed, 64 insertions(+), 15 deletions(-)

Related to: https://github.com/flyteorg/flyte/pull/6813#discussion_r2649208841

PR #6853, opened 2026-01-10
[Refactor]: CreateRun add initial service and response structure

yuhuan130:
## Tracking issue

Closes part 1 of #6824

Currently, `CreateRun` passes protobuf requests directly to the repository layer, violating separation of concerns.  This creates tight coupling between the API and data layers, making testing harder and inconsistent with the pattern in `task_service.go` (PR #6813).

Refactored `CreateRun` to use the transformer pattern:

```
Before:  Service → Repository (protobuf) → Database
After:   Service → Transformer → Repository (domain model) → Database
```

**New Files:**

- `runs/repository/transformers/run.go` - Converts protobuf → domain model
  - `CreateRunRequestToModel`: Transforms incoming request to domain model
  - `RunModelToCreateRunResponse`: Builds complete API response with Action hierarchy (ID, Status, Metadata)

**Modified Files:**

- `runs/service/create_run.go` - Uses transformers instead of passing protobuf directly
- `runs/repository/interfaces/action.go` - Changed signature to accept `*models.Run`
- `runs/repository/impl/action_repo.go` - Accepts domain models, removed protobuf logic
- `runs/service/run_service_test.go` - Updated mocks and tests
- `runs/repository/mocks/mock_ActionRepo.go` - Regenerated mocks to match new interface signature

**Key improvements:**

- Repository layer no longer depends on protobuf definitions
- Clean separation between API and data layers
- Follows closely as pattern for `task_service.go`
- Completed `RunModelToCreateRunResponse` transformer to build full API response hierarchy (includes Action ID, Status with phase/timestamps, and Metadata with source)

## How was this patch tested?

- Added unit tests for transformer functions
- Updated service and repository tests to use domain models
- Verified response structure matches expected hierarchy
- All tests pass locally:  `go test ./runs/...  -v`

## Screenshots

Screenshot 2026-01-11 at 18 14 09
Screenshot 2026-01-12 at 18 13 30 Screenshot 2026-01-12 at 18 13 42

- `main` 
  - \#6583
    - **\[Refactor]: CreateRun add initial service and response structure** :point\_left:


runs/runs.
machichima, line 1: Should we remove this?

machichima:
Could you please add a script for testing under https://github.com/flyteorg/flyte/tree/v2/runs/test/scripts and test this API manually to ensure the runs are created successfully?

runs/runs.
yuhuan130: Oops, added accidentally

machichima:
Hi @yuhuan130 ,
We update the mockery format in https://github.com/flyteorg/flyte/pull/6829. Could you please merge `v2` and generate the mockery again?

runs/test/scripts/create_run.sh
machichima, line 1: Could we keep one buf request only and commented out others, so that when we execute this script, there's only one run being created.

We can do things like below. I keep creating run with run_id as default as this is something that we will use the most

```sh
# Create run by run ID
# buf curl --schema . "$ENDPOINT/flyteidl2.workflow.RunService/CreateRun" --data @- <<EOF
# {
#   "run_id": {
#     "org": "$ORG",
#     "project": "$PROJECT",
#     "domain": "$DOMAIN",
#     "name": "my-custom-run-123"
#   },
#   "task_id": {
#     "org": "$ORG",
#     "project": "$PROJECT",
#     "domain": "$DOMAIN",
#     "name": "$TASK_NAME",
#     "version": "$VERSION"
#   },
#   "source": "RUN_SOURCE_WEB"
# }
# EOF

# Create run by project ID
buf curl --schema . "$ENDPOINT/flyteidl2.workflow.RunService/CreateRun" --data @- <<EOF
{
  "project_id": {
    "organization": "$ORG",
    "name": "$PROJECT",
    "domain": "$DOMAIN"
  },
  "task_spec": {
    "short_name": "test"
  },
  "source": "RUN_SOURCE_CLI"
}
EOF
```

runs/repository/transformers/run.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/common"
+	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/core"
+	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/workflow"
+	"github.com/flyteorg/flyte/v2/runs/repository/models"
+)
+
+const InitialPhase = "PHASE_QUEUED"
+
+// CreateRunRequestToModel converts CreateRunRequest protobuf to Run domain model
+func CreateRunRequestToModel(ctx context.Context, req *workflow.CreateRunRequest) (*models.Run, error) {
```
machichima, line 22: I would prefer isolate service layer and repository layer to prevent the transformer function depends on service request. That is, let's not pass the `CreateRunRequest` and transform here. Instead, do things in service layer, and just pass `runID` and `actionSpec` into here. Like what we did in:

https://github.com/flyteorg/flyte/blob/442bcc12bb47ba67144b9ec8081be08eb2685d0d/runs/service/task_service.go#L53-L54

In this case, we can reuse the run proto to run model transform function in other places rather than only for create run function

machichima:
For the check generate CI failure, we removed `Cargo.lock` in `gen` in https://github.com/flyteorg/flyte/pull/6844. Please merge `v2` and run `git rm --cached gen/rust/Cargo.lock` to prevent git tracking the `Cargo.lock`.
Thank you!!

runs/repository/transformers/run.go
yuhuan130: Ahh~ I see. No problem

runs/test/scripts/create_run.sh
yuhuan130: Ok, super quick confirm, you mentioned run_id as the default since it's most commonly used. But I think your example showed it the other way around, so I wanted to double-check before making the change. Thanks!

Sovietaced:
Please reopen this against the new `main` branch.

Issue #4466, opened 2023-11-21
[UI Feature] Propagate dynamic workflow error messages from flytepropeller to UI


fg91:
### Motivation: Why do you think this is important?

Our platform users regularly ping us with executions where a dynamic workflow failed to correctly register.

Flyte's UI doesn't provide helpful information to understand that the registration of the dynamic sub workflow has failed and how to fix it.

In this example, the dynamic task pod itself succeeded but the node failed:

![image](https://github.com/flyteorg/flyte/assets/36511035/648f29d5-75e0-4f45-a73e-d369c7906a02)

We in the MLOps team then typically look in the flytepropeller logs to try to understand how the dynamic subworkflow needs to be fixed.

In this case, the propeller logs showed:

<img width="999" alt="issue" src="https://github.com/flyteorg/flyte/assets/36511035/3a259e95-11c4-47f8-a895-870229203f54">

### Goal: What should the final outcome look like, ideally?

It would be very helpful if FlyteConsole would show the dynamic workflow related errors occurring in flytepropeller so that users can fix them themselves.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

hamersaw:
I suspect this is related to https://github.com/flyteorg/flyte/issues/4483. These use a flyteidl error that is wrapped to support the `DeepCopy` function ([here](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/error.go#L12-L14) and [here](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/branch.go#L12-L14)), My concern is that the latest [update to k8s dependencies](https://github.com/flyteorg/flyte/pull/4199) changed how the k8s client unmarshals CRDs into json structs.

hamersaw:
@fg91 do you have a minimal repro? Think I found the issues and fixed similarly for [branch node error reporting](https://github.com/flyteorg/flyte/pull/4528), but want to double check here.

fg91:
> @fg91 do you have a minimal repro?

Sorry for the late reply. This is a minimal repro:

```py
from flytekit import task, dynamic, workflow


@task
def foo(a = 1):  # Notice the missing type hint
    print(a)


@dynamic
def sub_wf():
    foo(a=1)


@workflow
def wf():
    sub_wf()

```

[...]

PR #6854, opened 2026-01-12
Fix: Correctly handle malformed dynamic workflows to avoid 'failed + succeeded + running' Schroedinger state

fg91:
## Tracking issue

Fixes https://github.com/flyteorg/flyte/issues/4466

This is truly my arch nemesis in terms of Flyte issues. I reported this problem more than 2 years ago and it was discussed multiple times in the contributors' sync. Our platform users have run into this problem dozens of times, always leaving them profoundly confused and frustrated.

When a dynamic workflow is malformed in a way that is not detected on the flytekit level but only when compiling the workflow in flytepropeller, such as ...

```py
@dataclass
class Foo:
    a: int = 1

@task
def foo(a: str = 1): 
    print(a)

@task
def bar() -> Foo: 
    return Foo()


@dynamic
def sub_wf():
    foo(a=bar()) # Pass instance of Foo instead of string
```

... the user is presented with this confusing view in flyteconsole, showing a succeeded attempt, failed node, and running workflow:

Screenshot 2026-01-12 at 17 56 29

Only ~1.5h later, the workflow fails with

```console
RuntimeExecutionError: max number of system retry attempts [51/50] exhausted. Last known status message: 0: [User] malformed dynamic workflow ...
```

The discussions in the issue and in the contributors' sync always revolved around changing flyteconsole to show the underlying error to the user in the UI, see e.g. [here](https://github.com/flyteorg/flyte/issues/4466#issuecomment-2292164862).

I realized that this is actually not a UI issue but rather a bug in the dynamic handler logic:

* The dynamic workflow cannot be built due to a user error so the node transitions to the failing phase.
* The `Abort` and `Finalize` methods are called but both try to build the dynamic workflow another time - which again fails.
* Not being able to build the dynamic workflow makes `Abort` and `Finalize` return an error.
* This error is categorized as a transient system error so that the "system error retry logic" kicks in.
* This prevents the workflow from fully failing until the system retry budget is exhausted.
* Therefore, the workflow fails only ~1.5h later with `max number of system retry attempts [51/50] exhausted ... malformed dynamic workflow ...`

This PR modifies `Abort` and `Finalize` to not return an error when the dynamic workflow is malformed. This way, the workflow can be aborted and finalized immediately and the user is presented with there underlying user error right away.

## How was this patch tested?

* Ran propeller with the proposed changes and the example above. The workflow immediately fails and shows the underlying error to the user:

    Screenshot 2026-01-12 at 18 10 43

* Added unit tests for the fixed behaviour.


flytepropeller/pkg/controller/nodes/dynamic/handler_test.go
```diff
@@ -822,6 +822,119 @@ func createDynamicJobSpecWithLaunchPlans() *core.DynamicJobSpec {
 	}
 }
 
+func createNodeContext(t assert.TestingT, ttype string, finalOutput storage.DataReference) *nodeMocks.NodeExecutionContext {
```
fg91, line 825: This helper is not new, I just moved it out of `TestDynamicNodeTaskNodeHandler_Finalize` to be able to use it in the new `TestDynamicNodeTaskNodeHandler_Abort`. For that I added the argument `t assert.TestingT` which was captured from the outer scope before.

EngHabu (approved):
Oh.. nice find!

PR #6859, opened 2026-01-16
apply mockery in queue_service_test

### LINKED ISSUE #6855: [v2] Use EXPECT instead of On in `queue`
The only place with `On()` left is in `queue/service/queue_service_test.go`. Will need a few refactor there. I think we can do it in the follow-up PR

```sh
❯ find ./ -name "*_test.go" -type f -exec grep -l "\.On(" {} \;
./queue/service/queue_service_test.go
```

1. Current `queue_service_test.go` define the mock implementation manually. We should use mockery instead, following how we generate mocks in https://github.com/flyteorg/flyte/pull/6829: https://github.com/flyteorg/flyte/blob/ada62dc81f72a7d635afcdea74d731426ec88fc6/queue/service/queue_service_test.go#L18-L20
2. Use EXPECT instead of On following the mockery best practice 

_Originally posted by @machichima in https://github.com/flyteorg/flyte/pull/6829#pullrequestreview-3629782627_

WangWang0226:
## Tracking issue
Closes #6855

The repository has been migrating to a standardized mocking approach using `mockery` with `EXPECT()` semantics (see PR #6829).
However, `queue_service_test.go` was still relying on a manually defined mock implementation and the legacy `On()` API from `testify/mock`.

Aligning `queue_service_test.go` with the existing `mockery` configuration ensures consistency, improves readability, and follows the recommended mocking best practices adopted in the codebase.

This PR refactors the queue service tests to fully adopt `mockery`-generated mocks and `EXPECT()`-based expectations.

Specifically:
- Added `github.com/flyteorg/flyte/v2/queue/service` to `.mockery.yaml` so that mocks for `QueueClientInterface` are generated automatically.
- Removed the manually implemented `MockQueueClient` from `queue_service_test.go`.
- Replaced manual mock construction with `mockery`-generated mocks (`mocks.NewQueueClientInterface(t)`).
- Migrated all mock expectations from `On(...)` to `EXPECT().Method(...)`, aligning with the preferred mocking style used across the repository.
- Kept the existing test coverage and behavior unchanged, focusing purely on refactoring and consistency.

No production code behavior was modified as part of this change.

## How was this patch tested?

- Existing unit tests in `queue_service_test.go` were updated to use `mockery` mocks and continue to pass.
- `queue_service_test.go` were run locally using:
  ```
  go test ./queue/service
  ```

* No additional test cases were required, as this PR is a refactor of test infrastructure only and does not change runtime behavior.

### Labels

* **changed**

### Setup process

No special setup is required beyond the existing development environment.

### Screenshots

N/A

* [ ] I updated the documentation accordingly.
* [x] All new and existing tests passed.

## Related PRs

* #6829 – Refactor: Define all mockery configuration in `.mockery.yaml`

## Docs link

N/A


WangWang0226:
Hi @machichima, could you please take a look when you have a chance? Thanks!

.mockery.yaml
```diff
@@ -59,6 +59,11 @@ packages:
       include-auto-generated: true
       all: true
       dir: 'runs/repository/mocks'
+  github.com/flyteorg/flyte/v2/queue/service:
+    config:
+      include-auto-generated: true
+      all: true
+      dir: 'queue/service/mocks'
```
popojk, line 66: I Think we can add the settings to remove the "Interface" string in mock name. WDYT @machichima ?
```
mockname: "Mock{{ .InterfaceName | trimSuffix \"Interface\" }}"
```

machichima: I would prefer not using `trimSuffix` here, which can generate the mock with different name as the original one. For example, if the interface named `UserInterface`, using this will generate `MockUser`, which may cause confusion, and if we try to search `UserInterface`, we cannot find this mock interface

machichima: I agree that having `Interface` suffix is not the best Golang interface naming convention, to remove this I think we should directly modify the interface name rather than trim here. WDYT?

PR #6860, opened 2026-01-16
fix: propagate error in ArrayNode output gathering

Sanchit2662:
## Summary

This PR fixes a **silent error propagation bug** in the ArrayNode worker output-gathering path (`flytepropeller/pkg/controller/nodes/array/worker.go`).
Due to a hardcoded `nil` being returned on the response channel, **errors encountered while reading sub-node outputs were silently discarded**, leading to workflows completing successfully with **missing or corrupted outputs**.

This is a **one-character fix** that restores correct error propagation and aligns the behavior with the existing node execution handling logic.

---

### Fix

Return the actual captured error instead of hardcoding `nil`:

```go
// After
gatherOutputsRequest.responseChannel <- struct {
    literalMap map[string]*idlcore.Literal
    error
}{literalMap, err}
```

This mirrors the already-correct behavior used for `nodeExecutionRequest` earlier in the same worker loop and ensures consistent error handling.

---

### Before

* Output read failures silently ignored
* Workflows succeed with incorrect or missing data
* Downstream task behavior becomes nondeterministic
* Production issues are extremely difficult to diagnose

### After

* Output gathering errors are properly propagated
* Workflows fail explicitly and predictably
* Retry / failure semantics behave as expected
* No silent data corruption

This affects **all Flyte deployments using ArrayNodes**, which is a core execution primitive.

---

## How was this patch tested?

* Manually verified error propagation path by inducing output read failures
* Confirmed consistency with `nodeExecutionRequest` error handling logic
* Ensured `workerErrorCollector` correctly captures and reports the error after the fix

No new tests were added due to the minimal and localized nature of the change, but the behavior now matches the established and already-tested execution path.

---

### Labels

* **fixed**

---

## Checklist

* [ ] I updated the documentation accordingly.
* [x] All new and existing tests passed.

### DIFF
--- a/flytepropeller/pkg/controller/nodes/dynamic/handler.go
+++ b/flytepropeller/pkg/controller/nodes/dynamic/handler.go
@@ -243,6 +243,12 @@ func (d dynamicNodeTaskNodeHandler) Abort(ctx context.Context, nCtx interfaces.N
 		if err != nil {
 			if stdErrors.IsCausedBy(err, utils.ErrorCodeUser) {
 				logger.Errorf(ctx, "failed to build dynamic workflow, user error: %s", err)
+				// If the dynamic workflow cannot be built because of a user error, i.e. malformed dynamic workflow,
+				// the error is deterministic and there are no nodes that need to be cleaned up as the dynamic worklow
+				// was never able to launch sub nodes.
+				// Returning an error on Abort would be treated as a transient system error that is retried as often as the
+				// system retry budget allows, surfacing the underlying user error only later aftr the system retry budget is exhausted.
+				return nil
 			}
 			return err
 		}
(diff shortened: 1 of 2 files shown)


Sanchit2662:
Hi @machichima ,
This pr fixes a bug where ArrayNode output-gathering errors were silently swallowed, causing workflows to succeed with missing outputs.
Whenever you have time, I’d really appreciate it if you could take a look and review the PR. 
Thank you!

machichima:
Hi @Sanchit266,
Thank you for the fix! Do you mind providing result screenshot showing the error is raised correctly?

Sanchit2662:
Hi @machichima,

I verified this locally by simulating the case where a subnode is marked Succeeded
but its output file is missing.

Running:
go clean -testcache && go test -v -run "TestHandleArrayNodePhaseSucceeding/Success$" .

Now results in:
worker error(s) encountered: [0]: failed to read data from
s3://bucket/output/0/0/outputs.pb: file does not exist

This confirms the output read error is correctly propagated instead of being
silently ignored. The test failure is expected in this setup since outputs are missing.

errrr

machichima:
Thanks @Sanchit2662!
Could you also add a test here (like what you modified in TestHandleArrayNodePhaseSucceeding) to make sure the error is raised?
Thank you!

PR #6865, opened 2026-01-20
Upgrade spark-operator to v2.1.1

kevinliao852:
## Tracking issue

Related #6686, #6348

- These changes are required to support newer Spark Operator features

### Labels


flyteplugins/go/tasks/plugins/k8s/spark/spark_test.go
machichima, line 1: Found some commented out assertions, should we uncomment them?

flyteplugins/go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -95,7 +95,7 @@ func getSparkConfig(taskCtx pluginsCore.TaskExecutionContext, sparkJob *plugins.
 		sparkConfig[k] = v
 	}
 
-	if sparkJob.GetExecutorPath() != "" {
+	if sparkJob.GetExecutorPath() != "" && sparkConfig["spark.pyspark.python"] == "" && sparkConfig["spark.pyspark.driver.python"] == "" {
```
machichima: Could you add the changes here to the PR description?

flyteplugins/go.mod
```diff
@@ -17,8 +16,9 @@ require (
 	github.com/golang/protobuf v1.5.4
 	github.com/hashicorp/golang-lru v0.5.4
 	github.com/imdario/mergo v0.3.13
+	github.com/kubeflow/spark-operator v0.0.0-20250325114751-1905be6e1dbd
```
machichima, line 20: Just curious, why the version here is not showing `v2.1.1`

machichima (commented):
Overall LGTM, just a few comments

flyteplugins/go.mod
kevinliao852: The tagged release pulls in transitive deps that don’t align with our current module graph, and it doesn’t build.
So for now, we’re pinning to a specific `commit hash` that compiles cleanly with our existing Flyte deps.
If there’s a better way to resolve the dependency mismatch without pinning, happy to switch.

flyteplugins/go/tasks/plugins/k8s/spark/spark_test.go
kevinliao852: Good catch. I missed updating those after the operator change. Thanks for flagging it. I’ll adjust them in this PR.

kevinliao852:
As discussed with @pingsutw , we should run the tests against the v1 CRD to match the currently compatible operator version.

flyteplugins/go/tasks/plugins/k8s/spark/spark.go
kevinliao852: I rechecked this logic and realized the current conditional handling could be problematic. I’d prefer to revert to the previous behavior for now. Let me know if you see any concerns.

flyteplugins/go.mod
pingsutw: Could we upgrade it 2026

kevinliao852: We can proceed with the current hash, which corresponds to Spark Operator v2.1.1, as it is stable for our current use case.

For future upgrades to newer versions, we will first need to resolve the compatibility issues tracked in Flyte PR #7013. In particular, the dependency conflicts and config decoding differences need to be addressed before safely moving beyond v2.1.1.

As discussed, we will:

Use v2.1.1 as the default for now
Defer upgrading to newer versions until the above issue is resolved

pingsutw (dismissed):

pingsutw (dismissed):

pingsutw (dismissed):

pingsutw (approved):

PR #6872, opened 2026-01-26
fixed: Remove print that breaks json output

honnix:

When `-o json` is specified, the output must be a valid json, however this print breaks it.

Removing the print out.

Alternatively, we could check output format first and then decide on whether to print this or not.

### Labels

## Related PRs

#6538


honnix:
There is a check failure which I'm not sure is because of this change. @pingsutw Could you please help when you get time? Also will merging this create a new release automatically?

PR #6875, opened 2026-01-28
Add visited nodes tracking and parallelism bypass for subworkflow nodes

### LINKED ISSUE #6873: [BUG] Flyte subworkflow nodes can be stuck even after referenced launchplans executions are complete
### Flyte & Flytekit version

**flyte: v1.16.3
flytekit: 1.16.12**

### Describe the bug

**Summary**
Flyte subworkflow nodes with reference launchplans can stay in the `Running` state long after the target execution has completed. 

**Behavior**
The screenshots below once such example which can consistently reproduce the issue. 
- Parent workflow triggers a reference launch plan execution, node `n7 ` - `another_sleepy_lp`
- The child reference launchplan executions finishes in under 3.5 minutes. 
- Corresponding node `n7` in the parent executions takes over 1 hour for completion.

Note that the problematic behavior is not seen for all workflows and is highly dependent on the dag-structure and the run-times of individual nodes in the dag. 

### Expected behavior

Parent execution's reference launchplan node `n7` should get updated within about 30 seconds  for child's completion (workflow evaluation re-enqueue frequency)

The wokflows defined here can consistenly reproduce the issue: https://github.com/sshardool/flyte/blob/4feb41fd0036dde78009bc9962ef269546e54f08/workflows/max-parallel/max_parallelism_workflows.py

Note that structure of the dags is carefully crafted to trigger the issue with nodes appearing in a very specific order. 

### Screenshots

Image

Image

Image

- [x] Yes

- [x] Yes

sshardool:
## Tracking issue
Fixes #6873

**Problem:** In deeply nested subworkflow executions, workflow nodes can get stuck or experience significant slowdowns due to:

**Redundant evaluations:** The same workflow node being evaluated multiple times within a single execution round when it's already running
**Parallelism bottlenecks:** Already-running workflow nodes (subworkflows/launchplans) being blocked by max parallelism limits, preventing progress on child nodes even though the parent workflow has already consumed its parallelism slot.

**Impact:**

- Prevents workflows from getting stuck in nested subworkflow and launchplans scenarios
- Improves execution throughput for workflows with many nested reference launchplans

**1. Visited Nodes Tracking**:
- Added `VisitedNodes` interface to `ControlFlow` in `ExecutionContext`
- Tracks workflow nodes using fully-qualified unique IDs (generated via `common.GenerateUniqueID`)
- Shared across entire execution tree (all nested subworkflows share the same `ControlFlow`)
- Only workflow nodes (NodeKindWorkflow) are tracked; task nodes are excluded
- Nodes are added to visited set after successful handling
- Already-visited nodes return `NodeStatusRunning` without re-evaluation

**2. Configurable Parallelism Bypass**:
- New config: `node-config.bypass-parallelism-check-for-workflow-nodes` (default: `false`)
- When enabled and max parallelism is reached:
  - Workflow nodes in `Running`, `Queued`, `Succeeding`, or `Failing` phases bypass the check
  - Nodes in `NotYetStarted` phase are still blocked (must wait for parallelism slot)
  - Task and array nodes remain subject to parallelism limits
- Config is logged once at executor initialization for visibility

**3. Additional Fix**:
- Fixed error swallowing in `RecursiveNodeHandler` where `return status, nil` should have been `return status, err`

## How was this patch tested?

Multiple reproduction scenarios are included in the draft PR, all of which resulted in stuck workflow node. These test workflows also expose the issues with the max-parallelism handling overall.

Stuck workflows recovered immediately after deployment and newly launched flows are not stuck any more since the fix has been deployed. 
Several unit tests were added as well.

### Labels

- **added**: Flag for allowing already running subworkflow nodes to be evaluated even after max-parallel nodes have been evaluated in the current cycle - `bypass-parallelism-check-for-workflow-nodes`
- **changed**: Explicit tracking added for nodes with have been visited in the current cycle.
- **fixed**: Fixes the stuck subworkflow behavior in #6873, can be generalized to all node types where this issue is possible. 

### Setup process
Example workflows that be be registered and executed with pyflyte:
https://github.com/sshardool/flyte/blob/4feb41fd0036dde78009bc9962ef269546e54f08/workflows/max-parallel/max_parallelism_workflows.py

### Screenshots
Screenshots included in #6873


machichima:
Hi @sshardool ,
Could you add DCO and fix CI error?
Thank you!

sshardool:
> Hi @sshardool , Could you add DCO and fix CI error? Thank you!

Updated and the CI error is fixed. PTAL - @machichima 
Note that `workflows` directory contains flows that reproduce the issue and included here for reference. Will remove it before merge.

PR #6878, opened 2026-01-28
Add endpoint to translate raw json to inputs

ursucarina:
### Labels


flyteidl2/workflow/translator_service.proto
```diff
@@ -53,3 +56,15 @@ message TaskSpecToLaunchFormJsonResponse {
   // The JSON for the variables.
   google.protobuf.Struct json = 1;
 }
+
+message JsonValuesToLiteralsRequest {
+  // The type definitions (VariableMap) describing the expected structure.
+  flyteidl2.core.VariableMap variables = 1;
+  // The raw JSON values to convert to literals.
+  google.protobuf.Struct values = 2;
```
EngHabu, line 64: Ketan was saying we should also allow this as the input for a CreateRun request.. which isn't a bad idea. You may still need this API for FE validation on the fly.. but the one in CreateRun request may simplify clients even more...
We can do it as a follow up.

PR #6896, opened 2026-02-05
fix: Add deterministic error clearing for FAIL_AFTER_EXECUTABLE_NODES_COMPLETE policy

andreahlert:
## Summary

This PR addresses the determinism concern raised by @hamersaw in #4624 regarding error clearing behavior when multiple nodes fail with `FAIL_AFTER_EXECUTABLE_NODES_COMPLETE` policy.

### The Problem

When multiple nodes fail, we want to preserve only one error to reduce etcd state size. The original implementation in #4624 would clear errors during iteration, but didn't account for the fact that Go map iteration order is non-deterministic. This could cause ALL errors to be cleared if iteration order varied between reconciliation rounds:

1. Round 1: Iterate [node-a, node-b] - clear node-a's error, keep node-b's
2. Round 2: Iterate [node-b, node-a] - node-b has error, clear it, then node-a has no error anymore

### The Solution

Add a nil check before considering a node as having a "new" failure:

```go
if nodeStatus.GetExecutionError() != nil {
    if nodeStatusWithError != nil && !c.enableCRDebugMetadata {
        nodeStatusWithError.ClearExecutionError()
    }
    nodeStatusWithError = nodeStatus
}
```

Once an error is cleared, that node won't be re-processed in subsequent rounds, ensuring determinism regardless of iteration order.

### Changes

- Add `ClearExecutionError()` method to `MutableNodeStatus` interface
- Implement `ClearExecutionError()` in `NodeStatus` struct  
- Add nil check in `handleDownstream` before error clearing
- Add comprehensive unit tests validating the deterministic behavior

## Test Plan

- [x] Unit tests for `ClearExecutionError()` method
- [x] Test that only the last error is preserved when multiple failures occur
- [x] Test that clearing already-nil errors is a no-op
- [x] All existing node controller tests pass

## Related Issues

- Fixes part of #4569 (Reduce etcd state for large workflows)
- Addresses determinism concern in #4624

### DIFF
--- a/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/iface.go
+++ b/flytepropeller/pkg/apis/flyteworkflow/v1alpha1/iface.go
@@ -346,6 +346,10 @@ type MutableNodeStatus interface {
 	GetArrayNodeStatus() MutableArrayNodeStatus
 	GetOrCreateArrayNodeStatus() MutableArrayNodeStatus
 	ClearArrayNodeStatus()
+
+	// ClearExecutionError clears the execution error stored in the node status.
+	// This is used to reduce etcd state size by removing old errors when new ones occur.
+	ClearExecutionError()
 }
 
 type ExecutionTimeInfo interface {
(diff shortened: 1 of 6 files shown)


andreahlert:
Hi @hamersaw, friendly ping! This addresses the determinism concern you raised in #4624 regarding error clearing with non-deterministic map iteration. CI is green - would love your review when you get a chance. Thanks!

PR #6897, opened 2026-02-05
Fix RayJob submitter pod affinity to use default config

pingsutw:
# TL;DR
Fix the RayJob submitter pod template to use the default affinity from K8s plugin config, and update the corresponding test to verify this behavior correctly.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added
 - [x] Any pending items have an associated Issue

## Complete description
The submitter pod template for RayJob was already using `config.GetK8sPluginConfig().DefaultAffinity`, but the test was incorrectly verifying that the submitter pod affinity matched the head pod's affinity instead of the default config affinity. This PR:

- Adds a clarifying comment explaining why the submitter pod uses default affinity/tolerations (so it lands on the default node)
- Updates the test to set a `DefaultAffinity` in the K8s plugin config and verify the submitter pod template uses it, rather than setting affinity on the head pod and expecting it to be copied

## Tracking Issue
NA

## Follow-up issue
NA

### DIFF
--- a/gen/rust/Cargo.lock
+++ b/gen/rust/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.100"
+version = "1.0.101"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
+checksum = "5f0e0fee31ef5ed1ba1316088939cea399010ed7731dba877ed44aeb407a75ea"
 
 [[package]]
 name = "async-stream"


flyteplugins/go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -427,7 +427,9 @@ func buildSubmitterPodTemplate(rayClusterSpec *rayv1.RayClusterSpec) v1.PodTempl
 			},
 			RestartPolicy:      v1.RestartPolicyNever,
 			EnableServiceLinks: &enableServiceLinks,
-			Tolerations:        tolerations,
+			// Run the RayJob submitter pod with the default affinity/tolerations so it lands on the default node.
+			Tolerations: tolerations,
+			Affinity:    config.GetK8sPluginConfig().DefaultAffinity,
```
jeevb, line 432: Does this not have to be overridden from the pod spec/template?

jeevb (approved):

flyteplugins/go/tasks/plugins/k8s/ray/ray.go
pingsutw: it’s intentional to use the default since head node could use gpu, we don’t want to add gpu affinity/toleration to the submitter pod

PR #6902, opened 2026-02-08
Implement state service on top of etcd.

EngHabu:

### Labels

## Docs link

- `main` 
  - \#6583
    - \#6901
      - \#6902 :point\_left:
        - \#6904


state/service/state_service.go
```diff
@@ -199,37 +196,33 @@ func (s *StateService) Watch(ctx context.Context, req *connect.Request[workflow.
 
 	logger.Infof(ctx, "Sent initial state (%d actions) and sentinel for parent action: %s", len(childActions), parentActionID.Name)
 
-	// Watch for state updates
-	updates := make(chan *common.ActionIdentifier, 100)
-	errs := make(chan error, 1)
-
-	go s.repo.ActionRepo().WatchStateUpdates(ctx, updates, errs)
+	// Subscribe to updates
+	updateCh := s.k8sClient.Subscribe()
```
popojk: Should we subscribe first then call list API for the first batch of data? Otherwise some data might be missing between the time window?

state/service/state_service.go
```diff
@@ -79,12 +82,6 @@ func (s *StateService) Put(ctx context.Context, req *connect.Request[workflow.Pu
 		}), nil
 	}
 
-	// Send notification
-	if err := s.repo.ActionRepo().NotifyStateUpdate(ctx, msg.ActionId); err != nil {
-		logger.Warnf(ctx, "Failed to send state update notification: %v", err)
-		// Continue anyway - the update was saved
-	}
```
popojk: I think here instead we should send a notification to run service so that the run service can update DB and show latest action status in the UI.

state/service/state_service.go
```diff
@@ -240,73 +233,74 @@ func (s *StateService) Watch(ctx context.Context, req *connect.Request[workflow.
...
 		actionID.Run.Domain != parentActionID.Run.Domain ||
 		actionID.Run.Name != parentActionID.Run.Name {
 		return false
 	}
 
-	// For now, we'll include all actions in the run
-	// In production, you'd check the parent relationship
+	// For now, include all actions in the same run
+	// A more sophisticated implementation would check the parent-child relationship
 	return true
```
popojk, line 278: nit:
```suggestion
	// Must be same run
	// For now, include all actions in the same run
	// A more sophisticated implementation would check the parent-child relationship
    return actionID.Run.Org == parentActionID.Run.Org &&
    actionID.Run.Project == parentActionID.Run.Project &&
    actionID.Run.Domain == parentActionID.Run.Domain &&
    actionID.Run.Name == parentActionID.Run.Name
```

state/service/state_service.go
```diff
@@ -67,8 +70,8 @@ func (s *StateService) Put(ctx context.Context, req *connect.Request[workflow.Pu
 		}), nil
 	}
 
-	// Update action state in database
-	if err := s.repo.ActionRepo().UpdateActionState(ctx, msg.ActionId, msg.State); err != nil {
+	// Update TaskAction state in Kubernetes
+	if err := s.k8sClient.PutState(ctx, msg.ActionId, msg.State); err != nil {
```
popojk, line 75: nit: An idea that we can consider to implement in another PR. I suggest that we store the updated status in a local cache after k8s CR is updated. The Get/Watch request can retrieve data from cache first to prevent a possible bottleneck in k8s API.

state/k8s/client.go
```diff
@@ -0,0 +1,316 @@
...
+	taskAction := &executorv1.TaskAction{}
+	if err := c.k8sClient.Get(ctx, client.ObjectKey{
+		Name:      taskActionName,
+		Namespace: c.namespace,
+	}, taskAction); err != nil {
+		return fmt.Errorf("failed to get TaskAction %s: %w", taskActionName, err)
+	}
+
+	// Update state JSON
+	taskAction.Status.StateJSON = stateJSON
```
popojk, line 93: Should we check if current state equals to previous state, and skip update if true?

state/k8s/client.go
```diff
@@ -0,0 +1,316 @@
...
+	return taskAction, nil
+}
+
+// Subscribe creates a new subscription channel for action updates
+func (c *StateClient) Subscribe() chan *ActionUpdate {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	ch := make(chan *ActionUpdate, c.bufferSize)
+	c.subscribers[ch] = struct{}{}
```
popojk: The watch action API request was scoped by parent action ID. Should we maintain a {parent action ID: subscribers} map here here? The update should only be notified to subscribers listening on updated action parent ID

manager/cmd/main.go
```diff
@@ -128,21 +130,33 @@ func serve(ctx context.Context) error {
...
 	if err != nil {
 		return fmt.Errorf("failed to initialize Kubernetes client: %w", err)
 	}
 
-	// Initialize the executor API scheme
-	if err := queuek8s.InitScheme(); err != nil {
-		return fmt.Errorf("failed to initialize scheme: %w", err)
-	}
+	// Create a client.Client from the WithWatch client for services that don't need watch
+	var regularK8sClient client.Client = k8sClient
```
machichima:
```suggestion
	var k8sClientWithoutWatch client.Client = k8sClient
```
nit: make it clear that this k8s client do not have watch support

state/k8s/client.go
machichima: Make sense to me!

state/k8s/client.go
```diff
@@ -0,0 +1,316 @@
...
+// notifySubscribers sends an update to all subscribers
+func (c *StateClient) notifySubscribers(update *ActionUpdate) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	for ch := range c.subscribers {
+		select {
+		case ch <- update:
+		default:
+			// Channel full, skip (non-blocking)
```
machichima: If the channel full, the update can lost. Based on the proto, `Watch` should guarantee at-least-once 

https://github.com/flyteorg/flyte/blob/78afd14cbefc372e36b97bb1efba7f336ba91e1b/flyteidl2/workflow/state_service.proto#L21-L22

I think we can just leave a error log for now, add TODO, and can handle this in the future if needed

manager/cmd/main.go
```diff
@@ -128,21 +130,33 @@ func serve(ctx context.Context) error {
...
 
 	// Create queue client (for Kubernetes operations)
-	queueK8sClient := queuek8s.NewQueueClient(k8sClient, cfg.Kubernetes.Namespace)
+	queueK8sClient := queuek8s.NewQueueClient(regularK8sClient, cfg.Kubernetes.Namespace)
 	logger.Infof(ctx, "Kubernetes client initialized for namespace: %s", cfg.Kubernetes.Namespace)
 
+	// Create state client (K8s-based, for watching TaskAction CRs)
+	stateK8sClient := statek8s.NewStateClient(k8sClient, cfg.Kubernetes.Namespace, 100)
```
machichima: Related to https://github.com/flyteorg/flyte/pull/6902/changes#r2845032983

What might be the best default value for buffer size? and should we make this configurable?

state/service/state_service.go
machichima: I think we can directly get from informer cache and do not need to implement this on our own. This can be improved in the future PR

state/service/state_service.go
machichima: It would be better to create a new goroutine to subscribe to the state client (like what we did in the state service `Watch`). Every time there's an update, we can then run `UpdateActionState` and `NotifyStateUpdate`

manager/cmd/main.go
machichima: Discussed offline, 100 would be fine now. We added log when the updates are dropped. If it happens in the future, we can come back and update this size

state/service/state_service.go
machichima: updated in https://github.com/flyteorg/flyte/pull/6902/commits/7b3505d245e68efcb444b2140c79c0da87d97168

state/service/state_service.go
machichima: We decide to put it in `Watch` in state service, when the `Watch` is being called, we will also update the DB

Updated in https://github.com/flyteorg/flyte/pull/6902/commits/23b5e51e7d2b0ad5e4f53582565d3a885e46482e

state/k8s/client.go
machichima: Updated in https://github.com/flyteorg/flyte/pull/6902/commits/b9187ada6f139df95c85e5582de88d7ccd255097

state/k8s/client.go
machichima: Updated in https://github.com/flyteorg/flyte/pull/6902/commits/c3e616d19ee473e3a834ae184f0b1377f1daf50b

manager/cmd/main.go
machichima: updated in https://github.com/flyteorg/flyte/pull/6902/commits/7fdc8f8d4e4bb228049e0c1a1223a3ca22e7a147

state/k8s/client.go
machichima: Updated in https://github.com/flyteorg/flyte/pull/6902/commits/01f09d542aeadbdeb794baeed456c8e306c81676

state/service/state_service.go
machichima: This function is removed in https://github.com/flyteorg/flyte/pull/6902/commits/740de544de0e41943ecae629780024218eca178b based on the update here: https://github.com/flyteorg/flyte/pull/6902#discussion_r2850620889

state/k8s/client.go
```diff
@@ -0,0 +1,331 @@
...
+	watching    bool
+}
+
+// NewStateClient creates a new Kubernetes-based state client
+func NewStateClient(k8sClient client.WithWatch, namespace string, bufferSize int) *StateClient {
+	return &StateClient{
+		k8sClient:   k8sClient,
+		namespace:   namespace,
+		bufferSize:  bufferSize,
+		subscribers: make(map[string]chan *ActionUpdate),
```
popojk: I think the map value should be a list of channels because we may have more than 1 subscribers for the a given action ID?

state/k8s/client.go
```diff
@@ -0,0 +1,331 @@
...
+			},
+			Name: taskAction.Spec.ActionName,
+		},
+		StateJSON:        taskAction.Status.StateJSON,
+		Phase:            getPhaseFromConditions(taskAction),
+		IsDeleted:        event.Type == watch.Deleted,
+		ParentActionName: parentActionName,
+	}
+
+	c.notifySubscribers(ctx, update)
```
popojk, line 306: I think besides notify subscribers, we should also notify run service to store the event

state/k8s/client.go
machichima: Fixed in https://github.com/flyteorg/flyte/pull/6902/commits/df0808897e421a05e40a5f5e31161f908b33ff50

state/k8s/client.go
machichima: Discussed offline, we will just call run service to update the action state to DB when we get update in `k8s/client.go`. A future follow-up will be adding a background goroutine to query the run service for persisting the update, which we can handle retry there

machichima:
Closed by https://github.com/flyteorg/flyte/pull/6948

PR #6903, opened 2026-02-08
Initial executor <> Plugins integration

EngHabu:

### Labels

## Docs link

- `main` 
  - \#6583
    - \#6901
      - \#6902
        - \#6903 :point\_left:
          - \#6904

### DIFF
--- a/manager/cmd/main.go
+++ b/manager/cmd/main.go
@@ -147,10 +147,10 @@ func serve(ctx context.Context) error {
 	}
 
 	// Create a client.Client from the WithWatch client for services that don't need watch
-	var regularK8sClient client.Client = k8sClient
+	var k8sClientWithoutWatch client.Client = k8sClient
 
 	// Create queue client (for Kubernetes operations)
-	queueK8sClient := queuek8s.NewQueueClient(regularK8sClient, cfg.Kubernetes.Namespace)
+	queueK8sClient := queuek8s.NewQueueClient(k8sClientWithoutWatch, cfg.Kubernetes.Namespace)
 	logger.Infof(ctx, "Kubernetes client initialized for namespace: %s", cfg.Kubernetes.Namespace)
 
 	// Create state client (K8s-based, for watching TaskAction CRs)
--- a/state/k8s/client.go
+++ b/state/k8s/client.go
@@ -17,10 +17,11 @@ import (
 
 // ActionUpdate represents an update to a TaskAction
 type ActionUpdate struct {
-	ActionID  *common.ActionIdentifier
-	StateJSON string
-	Phase     string
-	IsDeleted bool
+	ActionID         *common.ActionIdentifier
+	StateJSON        string
+	Phase            string
+	IsDeleted        bool
+	ParentActionName string
 }
 
 // StateClient implements state operations using Kubernetes TaskAction CRs
@@ -30,8 +31,11 @@ type StateClient struct {
 	bufferSize int
 
 	// Watch management
-	mu          sync.RWMutex
-	subscribers map[chan *ActionUpdate]struct{}
+	mu sync.RWMutex
+	// Map parent action name to subscriber channels.
+	// Multiple callers may watch the same parent action concurrently.
+	// TODO: add a prometheus counter for dropped updates when metrics are wired up for the state service
+	subscribers map[string]map[chan *ActionUpdate]struct{}
 	stopCh      chan struct{}
 	watching    bool
 }
@@ -42,7 +46,7 @@ func NewStateClient(k8sClient client.WithWatch, namespace string, bufferSize int
 		k8sClient:   k8sClient,
 		namespace:   namespace,
 		bufferSize:  bufferSize,
-		subscribers: make(map[chan *ActionUpdate]struct{}),
+		subscribers: make(map[string]map[chan *ActionUpdate]struct{}),
 	}
 }
 
@@ -72,6 +76,11 @@ func (c *StateClient) PutState(ctx context.Context, actionID *common.ActionIdent
 		Namespace: c.namespace,
 	}, taskAction); err != nil {
 		return fmt.Errorf("failed to get TaskAction %s: %w", taskActionName, err)
+	}
+
+	// Skip update if the stateJSON does not change
+	if taskAction.Status.StateJSON == stateJSON {
+		return nil
 	}
 
 	// Update state JSON
@@ -137,24 +146,30 @@ func (c *StateClient) GetTaskAction(ctx context.Context, actionID *common.Action
 	return taskAction, nil
 }
 
-// Subscribe creates a new subscription channel for action updates
-func (c *StateClient) Subscribe() chan *ActionUpdate {
+// Subscribe creates a new subscription channel for action updates for specified parent action name
+func (c *StateClient) Subscribe(parentActionName string) chan *ActionUpdate {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
 	ch := make(chan *ActionUpdate, c.bufferSize)
-	c.subscribers[ch] = struct{}{}
+	if c.subscribers[parentActionName] == nil {
+		c.subscribers[parentActionName] = make(map[chan *ActionUpdate]struct{})
+	}
+	c.subscribers[parentActionName][ch] = struct{}{}
 	return ch
 }
 
-// Unsubscribe removes a subscription channel
-func (c *StateClient) Unsubscribe(ch chan *ActionUpdate) {
+// Unsubscribe removes the given channel from the subscription list for the parent action name
+func (c *StateClient) Unsubscribe(parentActionName string, ch chan *ActionUpdate) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	if _, ok := c.subscribers[ch]; ok {
-		delete(c.subscribers, ch)
+	if channels, ok := c.subscribers[parentActionName]; ok {
+		delete(channels, ch)
 		close(ch)
+		if len(channels) == 0 {
+			delete(c.subscribers, parentActionName)
+		}
 	}
 }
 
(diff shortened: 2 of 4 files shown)


executor/pkg/controller/taskaction_controller.go
```diff
@@ -72,106 +96,209 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-	if failedCond != nil && failedCond.Status == metav1.ConditionTrue {
-		logger.Info("TaskAction already failed",
-			"name", taskAction.Name, "action", actionSpec.ActionId.Name)
+	// Ensure finalizer is present
+	if !controllerutil.ContainsFinalizer(taskAction, taskActionFinalizer) {
+		controllerutil.AddFinalizer(taskAction, taskActionFinalizer)
+		if err := r.Update(ctx, taskAction); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
```
machichima, line 123:
```suggestion
		if err := r.Update(ctx, taskAction); err != nil {
		    logger.Error(err, "Failed to update TaskAction with finalizer")
			return ctrl.Result{}, err
		}
	}
```

nit: add log for debugging

executor/pkg/controller/taskaction_controller.go
```diff
@@ -72,106 +96,209 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	p, err := r.PluginRegistry.ResolvePlugin(taskAction.Spec.TaskType)
+	if err != nil {
+		logger.Error(err, "failed to resolve plugin", "taskType", taskAction.Spec.TaskType)
+		r.Recorder.Eventf(taskAction, corev1.EventTypeWarning, string(FailedPluginResolve),
+			"No plugin found for task type %q: %v", taskAction.Spec.TaskType, err)
+		setCondition(taskAction, flyteorgv1.ConditionTypeFailed, metav1.ConditionTrue,
+			flyteorgv1.ConditionReasonPluginNotFound, err.Error())
+		setCondition(taskAction, flyteorgv1.ConditionTypeProgressing, metav1.ConditionFalse,
+			flyteorgv1.ConditionReasonPluginNotFound, err.Error())
+		_ = r.Status().Update(ctx, taskAction)
```
machichima: Should we also deal with the status update error?

executor/pkg/controller/taskaction_controller.go
```diff
@@ -72,106 +96,209 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	if !controllerutil.ContainsFinalizer(taskAction, taskActionFinalizer) {
+		controllerutil.AddFinalizer(taskAction, taskActionFinalizer)
+		if err := r.Update(ctx, taskAction); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	// Resolve plugin from registry
+	p, err := r.PluginRegistry.ResolvePlugin(taskAction.Spec.TaskType)
+	if err != nil {
```
machichima, line 127: I'm thinking of moving this to a validation function, as well as validating other fields in `taskAction`, and put this validation function before we add the finalizer. In this case, if the validation failed, we can just not requeue it, and do not need to remove the finalizer.

We can do this in follow-up!

executor/pkg/controller/taskaction_controller.go
```diff
@@ -72,106 +96,209 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 
-			if err := r.Status().Update(ctx, taskAction); err != nil {
-				return ctrl.Result{}, err
-			}
-			return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
-		}
+	// Determine requeue behavior
+	if phaseInfo.Phase().IsTerminal() {
+		return ctrl.Result{}, nil
+	}
```
machichima: Should we just handle the terminal state above with `if isTerminal()` and remove this here?

executor/pkg/controller/taskaction_controller.go
```diff
@@ -72,106 +96,209 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 
-			setCondition(taskAction, flyteorgv1.ConditionTypeProgressing, metav1.ConditionTrue,
-				flyteorgv1.ConditionReasonExecuting, "TaskAction is executing")
+	taskAction.Status.PluginPhase = phaseInfo.Phase().String()
+	taskAction.Status.PluginPhaseVersion = phaseInfo.Version()
 
-			taskAction.Status.StateJSON = createStateJSON(actionSpec, "Running")
+	// Update status subresource
+	if err := r.Status().Update(ctx, taskAction); err != nil {
```
popojk: Should we only update when current phase not equal to previous one?

executor/pkg/controller/taskaction_controller.go
```diff
@@ -72,106 +96,209 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		setCondition(ta, flyteorgv1.ConditionTypeProgressing, metav1.ConditionTrue,
+			flyteorgv1.ConditionReasonQueued, info.Reason())
+
+	case pluginsCore.PhaseInitializing:
+		setCondition(ta, flyteorgv1.ConditionTypeProgressing, metav1.ConditionTrue,
+			flyteorgv1.ConditionReasonInitializing, info.Reason())
+
+	case pluginsCore.PhaseRunning:
+		setCondition(ta, flyteorgv1.ConditionTypeProgressing, metav1.ConditionTrue,
+			flyteorgv1.ConditionReasonExecuting, info.Reason())
```
machichima, line 295: For future PR: we should probably add max retry and fail this if exceed

executor/pkg/controller/taskaction_controller.go
```diff
@@ -72,106 +96,209 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	tCtx, err := plugin.NewTaskExecutionContext(
+		taskAction,
+		r.DataStore,
+		stateMgr,
+		r.SecretManager,
+		r.ResourceManager,
+		r.CatalogClient,
+	)
+	if err != nil {
+		logger.Error(err, "failed to build task execution context")
```
machichima, line 158: Can we also add an event here showing that the Abort/Finalize are skipped:

```go
r.Recorder.Eventf(taskAction, corev1.EventTypeWarning, "FinalizationSkipped",
          "Could not build task execution context; skipping Abort/Finalize. Underlying resources may need manual cleanup: %v", err)
```

executor/pkg/controller/taskaction_controller.go
```diff
@@ -72,106 +96,209 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-	failedCond := findConditionByType(taskAction.Status.Conditions, flyteorgv1.ConditionTypeFailed)
-	if failedCond != nil && failedCond.Status == metav1.ConditionTrue {
-		logger.Info("TaskAction already failed",
-			"name", taskAction.Name, "action", actionSpec.ActionId.Name)
+	// Ensure finalizer is present
+	if !controllerutil.ContainsFinalizer(taskAction, taskActionFinalizer) {
+		controllerutil.AddFinalizer(taskAction, taskActionFinalizer)
+		if err := r.Update(ctx, taskAction); err != nil {
+			return ctrl.Result{}, err
+		}
```
machichima, line 121: I think we should requeue again after the update, as the update action will update the resource version, and this will block the update statue below (the resource version mismatch)

docs: https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes

executor/pkg/controller/taskaction_controller.go
popojk: I think if the CR status is in terminal state in current reconcile round, we should stop requeue instead of do 1 more requeue?

machichima: My concern is:
1. We should based on the `TaskAction` state, rather than based on both plugin phase and `TaskAction` phase
2. Would be better to check the terminal state in one place rather than two places

I think it's fine for us to requeue again to make the code cleaner

popojk: Make sense, go go

executor/pkg/controller/taskaction_controller.go
machichima: Added in https://github.com/flyteorg/flyte/pull/6903/commits/42a904e4cd073bdf508098f5aa2d118affc058d1

executor/pkg/controller/taskaction_controller.go
machichima: Fixed in https://github.com/flyteorg/flyte/pull/6903/commits/aa1c6555026e6de7f3f84302a45c361fd1604927

executor/pkg/controller/taskaction_controller.go
machichima: Updated in https://github.com/flyteorg/flyte/pull/6903/commits/3705ed5e05a1625720464b5b05e76e1d53b1ef51

executor/pkg/controller/taskaction_controller.go
machichima: Added in https://github.com/flyteorg/flyte/pull/6903/commits/c7f20e5b913fc455e7f5c7d6596464f36641649b

executor/pkg/controller/taskaction_controller.go
machichima: Updated in https://github.com/flyteorg/flyte/pull/6903/commits/2c7d0bc39dd103c6b0351b84f343891faf4975d9

executor/pkg/controller/taskaction_controller.go
machichima: Added in https://github.com/flyteorg/flyte/pull/6903/commits/e3429fefae0d14dbd722971549f091a3dca92214

gen/rust/Cargo.lock
machichima, line 1: It was included in the`.gitignore`, I think some PR accidentally adding it back

https://github.com/flyteorg/flyte/blob/b142044027d7dec2f5d47a62818c954c10ceb545/.gitignore#L44

PR #6904, opened 2026-02-08
Switch to pflags

EngHabu:

### Labels

## Docs link

- `main` 
  - \#6583
    - \#6901
      - \#6902
        - \#6904 :point\_left:


manager/cmd/main.go
machichima, line 233: Rather than hard coded config here, I think we should use `managerconfig.GetConfig()` instead?

manager/config/config.go
machichima, line 92: I think we should follow how we did in other `config.go` file, for example:

```go
var configSection = config.MustRegisterSection(configSectionKey, defaultConfig)

// GetConfig returns the parsed data proxy configuration
func GetConfig() *DataProxyConfig {
	return configSection.GetConfig().(*DataProxyConfig)
}


```

executor/cmd/main.go
```diff
@@ -47,52 +52,77 @@ import (
...
+// nolint:gocyclo
+func run() error {
+	cfg := config.GetConfig()
+
 	opts := zap.Options{
 		Development: true,
 	}
-	opts.BindFlags(flag.CommandLine)
-	flag.Parse()
-
```
machichima, line 93: Seems like the zap flag is being ignored after the change. I draft a patch to add it back: https://github.com/machichima/flyte/commit/3a6037510fac654aa59580f2a80d10a64c9b6e1d

executor/cmd/main.go
machichima, line 222: There's multiple places under `run()` that use `os.Exit()` rather than return error. It would be better to return error in case we want to add defer block/tests in the future (`os.Exit()` will directly kill the process)

machichima:
Closed by https://github.com/flyteorg/flyte/pull/6943

PR #6905, opened 2026-02-09
Support flag in ray plugin config to optionally disable ingress for Ray cluster

mickjermsurawong-openai:
## Tracking issue

- Allow disabling ingress for KubeRay according to [spec](https://ray-project.github.io/kuberay/reference/api/#headgroupspec). 
- Default of this flag changes no behavior of existing Ray cluster creation

- Nginx ingress is now [official retired ](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/), and some organizations now banned its usage
- Disabling this mean that the dashboard won't work, but running of Rayjob itself isn't affected

### Labels

- Added unit test
- Built locally and tested that rayplugin still launch job, created ephemeral cluster, head and workers correctly.

### DIFF
--- a/flyteplugins/go/tasks/plugins/k8s/ray/config.go
+++ b/flyteplugins/go/tasks/plugins/k8s/ray/config.go
@@ -20,7 +20,7 @@ var (
 		TTLSecondsAfterFinished:  3600,
 		ServiceType:              "NodePort",
 		IncludeDashboard:         true,
-		EnableIngress:            true,
+		DisableIngress:           false,
 		DashboardHost:            "0.0.0.0",
 		EnableUsageStats:         false,
 		ServiceAccount:           "",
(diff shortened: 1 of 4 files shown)


flyteplugins/go/tasks/plugins/k8s/ray/config.go
```diff
@@ -20,6 +20,7 @@ var (
 		TTLSecondsAfterFinished:  3600,
 		ServiceType:              "NodePort",
 		IncludeDashboard:         true,
+		EnableIngress:            true,
```
EngHabu: I like booleans to default to false... So I think this should be `DisableIngress`

mickjermsurawong-openai: Addressed thank you for fast response!

EngHabu (approved):

Issue #6911, opened 2026-02-11
[Core feature] Add Databricks Serverless Compute Support to Databricks Connector


rohitrsh:
### Motivation: Why do you think this is important?

Add comprehensive support for **Databricks Serverless Compute** in the flytekit databricks connector, enabling Flyte tasks to run on auto-scaling, on-demand compute without cluster management.

## Motivation

Classic Databricks compute requires **10-15 minutes for cluster startup**, which significantly impacts:
- Developer productivity and iteration speed
- Time-to-production for ML experiments
- Resource costs (paying for idle cluster time)

Databricks Serverless provides:
- **Sub-minute job startup times**
- Automatic scaling based on workload
- Zero cluster management overhead
- Better cost efficiency (pay only for the compute used)

### 1. Serverless Compute Detection and Job Format

**Auto-detect** serverless vs classic compute based on configuration:

| Configuration | Compute Mode |
|--------------|--------------|
| `existing_cluster_id` | Classic |
| `new_cluster` | Classic |
| `environment_key` (no cluster) | **Serverless** |
| `environments` array (no cluster) | **Serverless** |

Generate correct Databricks Jobs API 2.1 format for serverless (multi-task format with `tasks` array).

### 2. SparkSession Preservation

Databricks Serverless provides a pre-configured SparkSession via Spark Connect. However, Flytekit's `fast_execute_task_cmd` clears Python module state, losing the SparkSession.

**Solution**:
- Custom serverless-compatible entrypoint (`entrypoint_serverless.py`)
- Direct task execution to preserve module state
- SparkSession stored in `sys.modules` and `builtins` for reliable retrieval
- New `get_spark()` helper function for simplified access

### 3. S3 Access via Databricks Service Credentials

Serverless doesn't have AWS instance metadata. Solution:
- Use `dbutils.credentials.getServiceCredentialsProvider()` to obtain credentials
- New `databricks_service_credential_provider` configuration option
- Automatic credential injection as environment variables

### 4. Notebook Task Support

Enable scheduling Databricks notebooks from Flyte workflows:
- New `notebook_path` and `notebook_base_parameters` configuration options
- Generate `notebook_task` instead of `spark_python_task` when notebook_path is set
- Support both workspace and git-sourced notebooks

### Serverless Spark Task

```python
from flytekitplugins.spark import DatabricksV2, get_spark

@task(
    task_config=DatabricksV2(
        databricks_conf={
            "run_name": "my-serverless-job",
[...]

PR #6919, opened 2026-02-17
Add RBAC support for cross-namespace secret reading

rohitrsh:
### Tracking issue

Related to https://github.com/flyteorg/flyte/issues/6911

### Why are the changes needed?

The Databricks Spark connector now supports [Add multi-tenant Databricks token support via cross-namespace K8S secrets](https://github.com/flyteorg/flytekit/pull/3394), allowing the connector to read Databricks tokens from Kubernetes secrets in workflow namespaces. This enables multi-tenant Databricks access, allowing each Flyte project to use its own Databricks workspace/token.

For this feature to work, the connector's ServiceAccount needs `get` permission on `secrets` across namespaces. Currently, the `flyteconnector` Helm chart creates a ServiceAccount but no RBAC resources (ClusterRole / ClusterRoleBinding), so the connector cannot read secrets from workflow namespaces.

#### New file: `templates/connector/rbac.yaml`

Adds an optional `ClusterRole` and `ClusterRoleBinding` for the flyteconnector ServiceAccount:

- **ClusterRole** Configurable rules via `values.yaml`, defaulting to `get` on `secrets`
- **ClusterRoleBinding**  Binds the ClusterRole to the flyteconnector ServiceAccount
- **Guarded**  Only created when `rbac.enabled: true` (default: `false` for backward compatibility)
- **Follows existing patterns**  Uses the same naming and label conventions as the rest of the chart

```yaml
# templates/connector/rbac.yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: {{ template "flyte.namespace" . -}}-{{- template "flyteconnector.name" . }}
  labels: {{ include "flyteconnector.labels" . | nindent 4 }}
rules:
  {{- toYaml .Values.rbac.rules | nindent 2 }}
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: {{ template "flyte.namespace" . -}}-{{- template "flyteconnector.name" . }}
  labels: {{ include "flyteconnector.labels" . | nindent 4 }}
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: {{ template "flyte.namespace" . -}}-{{- template "flyteconnector.name" . }}
subjects:
- kind: ServiceAccount
  name: {{ template "flyteconnector.name" . }}
  namespace: {{ template "flyte.namespace" . }}
```

#### Updated: `values.yaml`

Added `rbac` configuration block:

```yaml
# -- RBAC configuration for flyteconnector
rbac:
  # -- Should RBAC resources (ClusterRole/ClusterRoleBinding) be created
  enabled: false
  # -- Rules for the ClusterRole
  rules:
    # Allow reading secrets across namespaces (for per-project Databricks token resolution)
    - apiGroups:
        - ""
      resources:
        - secrets
      verbs:
        - get
```

Key design decisions:
- **`enabled: false` by default**  Opt-in to avoid breaking existing deployments that don't need cross-namespace secret access
- **Only `get` verb**  Least privilege; no `list` or `watch`
- **Rules are configurable**  Operators can customise via `values.yaml` overrides
- **No `resourceNames` restriction**  Supports both the default `databricks-token` and custom secret names

### How was this patch tested?

1. **Template rendering verified** with `helm template`:

```bash
helm template ml-flyte charts/flyteconnector/ \
  --set rbac.enabled=true \
  --show-only templates/connector/rbac.yaml
```

2. **Verified RBAC disabled by default**  no ClusterRole/ClusterRoleBinding rendered when `rbac.enabled` is omitted

3. **Verified cross-namespace secret access** after applying:

```bash
kubectl auth can-i get secrets \
  --as=system:serviceaccount:flyte:flyteconnector \
  -n <workflow-namespace>
```

4. **End-to-end tested** with the Databricks per-project token feature  connector successfully reads namespace-specific tokens

### Setup process

To enable Databricks per-project token support:

```yaml
# values.yaml override
rbac:
  enabled: true
```

Then create secrets in workflow namespaces:

```bash
kubectl create secret generic databricks-token \
  --from-literal=token='dapi_your_token' \
  --namespace=<workflow-namespace>
```

### Screenshots

N/A (infrastructure-only change)

### Related PRs

- [flytekit PR: Multi-tenant Databricks token support](https://github.com/flyteorg/flytekit/pull/XXXX)

### Docs link

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,13 @@
+> [!IMPORTANT]
+> ## Looking for Flyte 2?
+>
+> - **Use Flyte locally?** Head to **[flyte-sdk](https://github.com/flyteorg/flyte-sdk)** — the new Python SDK for Flyte 2.
+> - **Want to contribute to the distributed backend?** See the **[`v2` branch](https://github.com/flyteorg/flyte/tree/v2)** of this repo.
+>
+> **The README below is for Flyte 1.x.**
+
+---
+
 <p align="center">
   <img src="https://raw.githubusercontent.com/flyteorg/static-resources/main/flyte/readme/flyte_and_lf.png" alt="Flyte and LF AI & Data Logo" width="250">
 </p>
--- a/flyteplugins/go/tasks/plugins/k8s/ray/config.go
+++ b/flyteplugins/go/tasks/plugins/k8s/ray/config.go
@@ -18,9 +18,9 @@ var (
 	defaultConfig = Config{
 		ShutdownAfterJobFinishes: true,
 		TTLSecondsAfterFinished:  3600,
+		EnableIngress:            false,
 		ServiceType:              "NodePort",
 		IncludeDashboard:         true,
-		DisableIngress:           false,
 		DashboardHost:            "0.0.0.0",
 		EnableUsageStats:         false,
 		ServiceAccount:           "",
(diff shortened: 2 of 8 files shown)


pingsutw (dismissed):

charts/flyteconnector/templates/connector/rbac.yaml
```diff
@@ -0,0 +1,27 @@
+{{- if .Values.rbac.enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
```
machichima, line 3: Should we also use `{{- if $.Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1" }}` here?

rohitrsh: Good catch, updated the ClusterRole to use `{{- if $.Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1" }}` as well, matching the pattern in `flyte-core/propeller/rbac.yaml`. Both `ClusterRole` and `ClusterRoleBinding` now gracefully fall back to `v1beta1` on older clusters.

machichima:
I think the failing test is unrelated to this PR, it also failed on other PRs

PR #6946, opened 2026-02-26
Notifications IDL

iaroslav-ciupin:

The goal of this PR is to address following use cases:

### Flyte OSS

1. I am new to notifications and I want to test my delivery configuration for the first time before creating any runs. I should use flyte-sdk or UI to call `DeliveryConfigService.SendTestNotification` in which I should provide my configuration and required template data. Endpoint will synchronously compile notification and send a corresponding  notification(webhook, email, Slack).
2. Once I am happy with my delivery configuration, I can create a one-off run with it. I should use flyte-sdk or UI to call `RunService.CreateRun` and put my previously tested delivery configuration into `RunSpec.notification_settings.InlineRule` field. This will store all necessary notification settings in run's RunSpec. When run completes, ActionService will send a ActionStatus with the same RunSpec to `action_statuses` Redis Stream. ActionStatus consumer will forward this message to `notifications` stream. Our notification consumer will pick this message and will compile & send a notification using same code as in step 1.
3. Once I see that my runs are triggering notifications as expected, I may want to update my deployed triggers with my tested notification settings. For this, I need to adjust existing triggers' RunSpec and add notification settings to it as in step 2 and re-deploy my tasks.

### Union-only

4. If I want to send notifications for all runs(triggered and manual) in a certain domain, project, for certain tasks, I should call `DeliveryConfigService.SaveDeliveryConfig` and `RuleService.SaveRule` endpoints via flyte-sdk or UI to create a persisted rule that will automatically inject notification settings for all runs that match those rules. This means everytime `RunService.CreateRun` endpoint is called, it will inject notification settings into its RunSpec if we have a rule that matches it.
   We can have all notification rules cached in memory(we don't expect huge amount of rules created by our users, I would say hundreds) so that in the average case it should not affect `RunService.CreateRun` latency.
   For Flyte OSS, "persisted settings" APIs will return 5xx unimplemented and `RunService.CreateRun` with notification\_settings non-empty rule\_id will also be rejected.

### Labels

## Docs link

- `main` 
  - \#6583
    - \#6946 :point\_left:

### DIFF
--- a/flyteidl2/gen_utils/rust/src/lib.rs
+++ b/flyteidl2/gen_utils/rust/src/lib.rs
@@ -49,6 +49,10 @@ pub mod flyteidl {
         include!("flyteidl2.core.rs");
     }
 
+    pub mod notification {
+        include!("flyteidl2.notification.rs");
+    }
+
     pub mod task {
         include!("flyteidl2.task.rs");
     }
--- a/gen/rust/src/lib.rs
+++ b/gen/rust/src/lib.rs
@@ -49,6 +49,10 @@ pub mod flyteidl {
         include!("flyteidl2.core.rs");
     }
 
+    pub mod notification {
+        include!("flyteidl2.notification.rs");
+    }
+
     pub mod task {
         include!("flyteidl2.task.rs");
     }


kumare3:
I am a little confused, but do we need so many different services? Should it not be just the Notification proto and it can be part of RunSpec and TriggerSpec? Is there anything else? If so why? Is the goal to allow users to create "DeliveryConfigs" and then associate named entities with Trigger and Run? From the DevX perspective this is much simpler IMO. But, if the goal is to precreate some delivery config and then support named configs, this is different and might need an approach that you specify. But we could add that later?

cc @EngHabu fyi?

flyteidl2/notification/definition.proto
```diff
@@ -0,0 +1,131 @@
...
+
+  repeated string to = 2;
+
+  repeated string cc = 3;
+
+  repeated string bcc = 4;
+
+  string html = 5 [(buf.validate.field).string.min_len = 1]; // this is a template that may contain placeholders
+
+  string text = 6; // this is a template that may contain placeholders
```
EngHabu: Are these oneof?

flyteidl2/task/run.proto
```diff
@@ -93,4 +95,26 @@ message RunSpec {
 
   // CacheConfig contains configurations that influence the behavior of cache reads and writes
   CacheConfig cache_config = 9;
+
+  oneof notification_settings {
```
kumare3, line 100: ohh i see it now

kumare3:
> I am a little confused, but do we need so many different services? Should it not be just the Notification proto and it can be part of RunSpec and TriggerSpec? Is there anything else? If so why? Is the goal to allow users to create "DeliveryConfigs" and then associate named entities with Trigger and Run? From the DevX perspective this is much simpler IMO. But, if the goal is to precreate some delivery config and then support named configs, this is different and might need an approach that you specify. But we could add that later?
> 
> cc @EngHabu fyi?

Ok I see the runSpec addition. I think this is the most important thing. Everything else can be fast follow

flyteidl2/notification/definition.proto
iaroslav-ciupin: I've just put it here temporarily based on AWS SES docs https://docs.aws.amazon.com/ses/latest/APIReference/API_Body.html , but I will have to do a more detailed investigation on what email provider are we going to use(AWS SES or Sendgrid) and I will adjust these payloads accordingly.

EngHabu (approved):

PR #6949, opened 2026-02-27
[V2] Migrate action service

machichima:

We introduce new "Action Service" to replace queue service and state service. This PR add the action service and remove the usage of queue and state service

- Add actions service and remove queue and state services
- Move state client (interact with k8s client) to action client
- Update all usage for state/queue component to use action service / client
- Add `_U_USE_ACTIONS: "1"` to env var to let flyte-sdk use action service

## How was this patch tested?

- Run locally based on docs in `manager/DEVELOPMENT.md`, and ensure we can submit tasks through `flyte run` from flyte-sdk
- Add simple unit test for action service in `actions_service_test.go`

### Labels

## Docs link

- `main` 
  - \#6583
    - \#6949 :point\_left:

### DIFF
--- a/manager/cmd/main.go
+++ b/manager/cmd/main.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 
+	"github.com/flyteorg/flyte/v2/actions"
 	"github.com/flyteorg/flyte/v2/app"
 	"github.com/flyteorg/flyte/v2/dataproxy"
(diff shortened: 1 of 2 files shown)


queue/config/default/kustomization.yaml
machichima, line 1: I think we can add those YAML for actions later in the follow-up PR

PR #6951, opened 2026-02-27
feat(flyteidl2): add image_build_url field to TaskMetadata proto 

AdilFayyaz:

When a task's container image is built by a remote image build system, there is currently no way to surface a link back to that build run in the task metadata. Consumers of the API (e.g. the UI or CLI) have no way to navigate from a task to the image build that produced it, making debugging and traceability harder.

Added a new image\_build\_url field (field number 19, type string) to the TaskMetadata proto message in flyteidl2/core/tasks.proto.

This field stores the URL of the remote image build run that produced the container image for the task, enabling downstream systems to link a task back to its image build.

Generated code was updated across all supported language targets:

- Go: gen/go/flyteidl2/core/tasks.pb.go and tasks.pb.validate.go
- Python: gen/python/flyteidl2/core/tasks\_pb2.py and tasks\_pb2.pyi
- Rust: gen/rust/src/flyteidl2.core.rs
- TypeScript: gen/ts/flyteidl2/core/tasks\_pb.ts
- Swagger: gen/go/flyteidl2/connector/service.swagger.json

## How was this patch tested?

The change is a proto field addition with generated code only — no business logic was introduced. The generated files were verified to compile correctly. No unit tests were added as this is a pure IDL/schema change; testing of the field's population and consumption will occur in the service implementations that set/read this field.

### Labels

- added: New image\_build\_url field in TaskMetadata proto.

## Docs link

- `main` 
  - \#6583
    - \#6951 :point\_left:


flyteidl2/core/tasks.proto
```diff
@@ -192,6 +192,9 @@ message TaskMetadata {
 
   // Log links associated with this task.
   repeated core.TaskLog log_links = 18;
+
+  // URL of the remote image build run that produced the container image for this task.
+  string image_build_url = 19;
```
EngHabu: nit on naming:

```suggestion
  string image_build_run_url = 19;
```

But more fundamentally, should it be a RunIdentifier instead of a run url? @pingsutw

flyteidl2/core/tasks.proto
```diff
@@ -192,6 +193,9 @@ message TaskMetadata {
 
   // Log links associated with this task.
   repeated core.TaskLog log_links = 18;
+
+  // URL of the remote image build run that produced the container image for this task.
```
pingsutw:
```suggestion
  // RunIdentifier of the remote image build run that produced the container image for this task.
```

PR #6956, opened 2026-03-02
[V2] Add Internal run service

machichima:

Run service should only watch/update the DB, rather than using the action service client. This PR make run service fully depends on the DB, and add the internal run service for other service (action service)  to call for persisting data to DB.

- Add `runs/service/internal_run_service.go`, which contains the internal run service implementations
- Update `runs/service/run_service.go` to only update / watch the DB, and remove the usage of action client
- `actions/k8s/client.go` will call internal run service method to record action/action event
- Add `ActionEvent` table (`runs/repository/models/action_event.go`)

Fixes

- Update the `Phase` in the `Action` table from string to int, being able to parse from `ActionPhase` enum rather than hard coding `PHASE_....`

## How was this patch tested?

- Follow the development guide in `manager/DEVELOPMENT.md` to start single binary locally, and run example through flyte-sdk

- flyte-sdk `.wait` can get the phase update correctly image

- `TaskAction` CR completed

image

- Data persisted to DB correctly

Actions: image

ActionEvents: image

### Labels

## Docs link

- `main` 
  - \#6583
    - \#6956 :point\_left:

### DIFF
--- a/runs/service/run_service.go
+++ b/runs/service/run_service.go
@@ -635,21 +635,29 @@ func (s *RunService) WatchActions(
 	errsCh := make(chan error)
 	go s.repo.ActionRepo().WatchActionUpdates(ctx, runID, updatesCh, errsCh)
 
-	// Send existing actions from DB
-	existingActions, _, err := s.repo.ActionRepo().ListActions(ctx, runID, 1000, "")
-	if err != nil {
-		logger.Errorf(ctx, "Failed to list actions: %v", err)
-		// Continue — still watch for new updates
-	} else if len(existingActions) > 0 {
-		enriched := make([]*workflow.EnrichedAction, 0, len(existingActions))
-		for _, a := range existingActions {
-			enriched = append(enriched, s.convertActionToEnrichedProto(a))
+	// Send existing actions from DB (paginate through all pages)
+	token := ""
+	for {
+		batch, nextToken, err := s.repo.ActionRepo().ListActions(ctx, runID, 100, token)
+		if err != nil {
+			logger.Errorf(ctx, "Failed to list actions: %v", err)
+			break
 		}
-		if err := stream.Send(&workflow.WatchActionsResponse{
-			EnrichedActions: enriched,
-		}); err != nil {
-			return err
+		if len(batch) > 0 {
+			enriched := make([]*workflow.EnrichedAction, 0, len(batch))
+			for _, a := range batch {
+				enriched = append(enriched, s.convertActionToEnrichedProto(a))
+			}
+			if err := stream.Send(&workflow.WatchActionsResponse{
+				EnrichedActions: enriched,
+			}); err != nil {
+				return err
+			}
 		}
+		if nextToken == "" {
+			break
+		}
+		token = nextToken
 	}
 
 	for {


actions/k8s/client.go
```diff
@@ -415,6 +421,74 @@ func (c *ActionsClient) notifySubscribers(ctx context.Context, update *ActionUpd
...
+		Attempt:      1, // TODO(nary): hardcoded until retry support is added
+		Phase:        update.Phase,
+		Version:      taskAction.Status.PluginPhaseVersion,
+		UpdatedTime:  updatedTime,
+		ReportedTime: timestamppb.Now(),
+	}
+	recordReq := &workflow.RecordActionEventsRequest{
+		Events: []*workflow.ActionEvent{event},
+	}
+	if _, err := c.runClient.RecordActionEvents(ctx, connect.NewRequest(recordReq)); err != nil {
```
pingsutw, line 487: For the follow-up, we should also check whether the event has been sent

pingsutw: qq: Why don't we use RecordActionEventStream here?

actions/k8s/client.go
```diff
@@ -415,6 +421,74 @@ func (c *ActionsClient) notifySubscribers(ctx context.Context, update *ActionUpd
...
+			},
+		}
+		if _, err := c.runClient.UpdateActionStatus(ctx, connect.NewRequest(statusReq)); err != nil {
+			logger.Warnf(ctx, "Failed to update action status in run service for %s: %v", update.ActionID.Name, err)
+		}
+	}
+
+	// TODO(nary): some ActionEvent fields not populated here. Need further discussion on where and how we want 
+	// to handle the action event persistence
+	// - ErrorInfo:     not on the CR; the executor does not write structured error info to TaskAction status.
```
EngHabu, line 469: It should, right?

EngHabu: I think all of these should be pushed into the CR

flyteidl2/workflow/internal_run_service.proto
```diff
@@ -0,0 +1,160 @@
...
+import "google/rpc/status.proto";
+
+option go_package = "github.com/flyteorg/flyte/v2/gen/go/flyteidl2/workflow";
+
+service InternalRunService {
+  // Record a new action.
+  rpc RecordAction(RecordActionRequest) returns (RecordActionResponse) {}
+
+  // Record a new action (streaming implementation).
+  rpc RecordActionStream(stream RecordActionStreamRequest) returns (stream RecordActionStreamResponse) {}
```
EngHabu, line 17: I don't think we use/need this any more

flyteidl2/workflow/internal_run_service.proto
```diff
@@ -0,0 +1,160 @@
...
+
+  // Update the status of an action.
+  rpc UpdateActionStatus(UpdateActionStatusRequest) returns (UpdateActionStatusResponse) {}
+
+  // Update the status of an action (streaming implementation).
+  rpc UpdateActionStatusStream(stream UpdateActionStatusStreamRequest) returns (stream UpdateActionStatusStreamResponse) {}
```
EngHabu, line 23: ditto

flyteidl2/workflow/internal_run_service.proto
```diff
@@ -0,0 +1,160 @@
...
+
+  // Record a collection of action events.
+  rpc RecordActionEvents(RecordActionEventsRequest) returns (RecordActionEventsResponse) {}
+
+  // Record a new action event (streaming implementation).
+  rpc RecordActionEventStream(stream RecordActionEventStreamRequest) returns (stream RecordActionEventStreamResponse) {}
```
EngHabu, line 29: ditto

flyteidl2/workflow/internal_run_service.proto
```diff
@@ -0,0 +1,160 @@
...
+  }
+}
+
+// Response message for recording an action.
+message RecordActionResponse {
+  // The action id.
+  flyteidl2.common.ActionIdentifier action_id = 1 [(buf.validate.field).required = true];
+
+  // The result.
+  google.rpc.Status status = 2 [(buf.validate.field).required = true];
```
EngHabu, line 64: This is weird to me... not sure why are we returning the status in the response?

actions/k8s/client.go
machichima: I'm still not sure if we want to add the `[]ActinEvent` to CR, and keep adding action event to it when there's a new event. It's like using the CR as a queue, which may not be a good practice

I add this TODO here so that we can discuss this later and deal with it in the follow-up PR, otherwise this PR can be too big

actions/k8s/client.go
machichima: It's because this function will be triggered per watch call. To use `RecordActionEventStream` we need to start it in a background goroutine, and each watch update we will put element to the channel, then the background goroutine will take those elements out and send to `RecordActionEventStream`. 

This can be a follow-up if needed, but using unary is the simplest way here

actions/k8s/client.go
```diff
@@ -415,6 +421,74 @@ func (c *ActionsClient) notifySubscribers(ctx context.Context, update *ActionUpd
...
+			ActionId: update.ActionID,
+			Parent:   update.ParentActionName,
+			InputUri: taskAction.Spec.InputURI,
+		}
+		if taskAction.Spec.TaskType != "" {
+			recordReq.Spec = &workflow.RecordActionRequest_Task{
+				Task: &workflow.TaskAction{},
+			}
+		}
+		if _, err := c.runClient.RecordAction(ctx, connect.NewRequest(recordReq)); err != nil {
```
popojk, line 444: Curious, why not using the RecordActionStream API to reduce the connection overhead?

actions/k8s/client.go
```diff
@@ -415,6 +421,74 @@ func (c *ActionsClient) notifySubscribers(ctx context.Context, update *ActionUpd
...
+	}
+
+	if update.Phase != common.ActionPhase_ACTION_PHASE_UNSPECIFIED {
+		statusReq := &workflow.UpdateActionStatusRequest{
+			ActionId: update.ActionID,
+			Status: &workflow.ActionStatus{
+				Phase: update.Phase,
+			},
+		}
+		if _, err := c.runClient.UpdateActionStatus(ctx, connect.NewRequest(statusReq)); err != nil {
```
popojk, line 462: Why not using the UpdateActionStatusStream API?

flyteidl2/workflow/internal_run_service.proto
machichima: It's to align the response from the streaming version. If we decide to remove the streaming version, we can also remove this

runs/service/run_service.go
```diff
@@ -703,55 +647,31 @@ func (s *RunService) WatchActions(
 		}
 	}
 
+	// Step 2: Watch for updates from DB
+	updatesCh := make(chan *models.Action)
+	errsCh := make(chan error)
+	go s.repo.ActionRepo().WatchActionUpdates(ctx, runID, updatesCh, errsCh)
```
popojk, line 636: We should watch actions first then list or some events might be missing between the time window

runs/service/run_service.go
```diff
@@ -680,21 +630,15 @@ func (s *RunService) WatchActions(
...
-	// Step 1: Subscribe to K8s watch events, filter to this run
-	// Root-level actions have no parent action name, so subscribe with "" to receive their updates.
-	// TODO(nary): This is wrong here, we should watch the DB update, and the watch should depends on the filter
-	updatesCh := s.stateClient.Subscribe("")
-	defer s.stateClient.Unsubscribe("", updatesCh)
-
-	// Step 2: Send existing TaskAction CRs for this run
-	taskActions, err := s.stateClient.ListRunActions(ctx, runID)
+	// Step 1: Send existing actions from DB
+	existingActions, _, err := s.repo.ActionRepo().ListActions(ctx, runID, 1000, "")
```
popojk: Why hardcode the page size to 1000 and only stream 1 page of data to the client? I think we can send all pages of data to the stream when the client connected to this API so that we don't miss data

actions/k8s/client.go
```diff
@@ -415,6 +421,74 @@ func (c *ActionsClient) notifySubscribers(ctx context.Context, update *ActionUpd
...
+	event := &workflow.ActionEvent{
+		Id:           update.ActionID,
+		Attempt:      1, // TODO(nary): hardcoded until retry support is added
+		Phase:        update.Phase,
+		Version:      taskAction.Status.PluginPhaseVersion,
+		UpdatedTime:  updatedTime,
+		ReportedTime: timestamppb.Now(),
+	}
+	recordReq := &workflow.RecordActionEventsRequest{
+		Events: []*workflow.ActionEvent{event},
```
popojk, line 485: Curious, why do we need a events array here as we only insert 1 event here?

actions/k8s/client.go
machichima: Same as comment: https://github.com/flyteorg/flyte/pull/6956#discussion_r2875449715, this is for the simplification. Using `RecordActionStream` can add more complexity here. We can use it in the future if needed

actions/k8s/client.go
machichima: Same as https://github.com/flyteorg/flyte/pull/6956#discussion_r2875695526

actions/k8s/client.go
machichima: This is to align the `RecordActionEventsRequest` format, and we can leave it for the future if we want to implement batch recording

runs/service/run_service.go
machichima: Thanks for pointing this out! Updated in https://github.com/flyteorg/flyte/pull/6956/commits/533b85ed13eefd77396575d3de8d64a4707a8b78

runs/service/run_service.go
machichima: Make sense! Updated in https://github.com/flyteorg/flyte/pull/6956/commits/bc7c58985a396f2ca89f98936cebc85487635493

popojk (approved):

PR #6970, opened 2026-03-06
[V2][Feat] Add events service

machichima:

Previous we call internal run service to record action events (persist to DB) from action client (k8s watcher). However, we cannot get the full info we want from the CR (e.g. `LogInfo`, `ErrorInfo`, ...). In this PR, we add a new `EventService` running independently, receiving requests from executor and forward to internal run service through background workers.

### Alternatives

We've considered following alternatives:

#### 1. Store map of `ActionEvents` under `TaskAction` CR status

We can store a map of action events with ID under the `TaskAction` CR status, and add a `persisted_id` slice to record the persisted action events. Executor will put new action event to the CR, and action service will take the action event out -> check if it's persisted by checking `persisted_id`, and call request internal run service.

This is using `TaskAction` CR as a queue, which we need to:

1. take the action event info out
2. determine which action event is persisted
3. remove the persisted action event

This method has following disadvantages:

1. can be more complex, as we need to handle the queueing logic on our own
2. may cause performance issue in large scale
3. Using CR object as queue is not a good practice

#### 2. Store `ActionEvent` in a new CR

We can add a new `ActionEvent` CR to store the action events. Executor will create a new `ActionEvent` CR when there's status update, and action service can watch this CR. Whenever there's new `ActionEvent` CR create, it will parse info from the CR and send request to internal run service for persisting action event.

This method has following disadvantages:

1. If run service is down, we may create overwhelming CRs, causing pressure to etcd

- Add new `EventsService` proto and implementations
- Start a background goroutine `EventsWorker` to queue and send requests to internal run service (through unary)
- Clean-up `executor/cmd/main.go` and remove unused code

TODO next: Use streaming endpoint / implement batch request from `EventsService` to internal run service

## How was this patch tested?

1. Start Flyte locally following `manager/DEVELOPMENT.md`
2. Use this config in flyte-sdk

```yaml
admin:
  endpoint: dns:///localhost:8090
  insecure: True
image:
  builder: local
task:
  domain: development
  project: testproject
  org: testorg
```

6. Run an example (e.g. `python examples/basics/hello_v2.py`)
7. Look into the SQLite DB (`flyte.db`), check if the data exists

image

DB:

image

### Labels

## Docs link

- `main` 
  - \#6583
    - \#6970 :point\_left:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,7 @@ COPY flyteplugins flyteplugins
 COPY flytestdlib flytestdlib
 COPY gen gen
 COPY actions actions
+COPY events events
 COPY runs runs
 
 COPY go.mod go.sum ./
--- a/Makefile
+++ b/Makefile
@@ -48,29 +48,6 @@ sandbox-run: ## Start the flyte sandbox without rebuilding the image
 sandbox-stop: ## Stop the flyte sandbox
 	$(MAKE) -C docker/sandbox-bundled stop
 
-# =============================================================================
-# Local Cluster Commands
-# =============================================================================
-
-.PHONY: cluster-create
-cluster-create: ## Create k3d cluster with host gateway alias for pod-to-host connectivity
-	$(eval HOST_GATEWAY_IP := $(shell docker run --rm --add-host=probe:host-gateway busybox cat /etc/hosts 2>/dev/null | awk '$$1~/^[0-9]/&&$$2=="probe"{print $$1;exit}'))
-	@if [ -z "$(HOST_GATEWAY_IP)" ]; then \
-		echo "ERROR: Failed to detect HOST_GATEWAY_IP. Ensure Docker is running."; \
-		exit 1; \
-	fi
-	@echo "Host gateway IP: $(HOST_GATEWAY_IP)"
-	@if k3d cluster get $(CLUSTER_NAME) --no-headers >/dev/null 2>&1; then \
-		echo "Cluster $(CLUSTER_NAME) already exists, skipping creation"; \
-	else \
-		CLUSTER_NAME=$(CLUSTER_NAME) HOST_GATEWAY_IP=$(HOST_GATEWAY_IP) envsubst < config/k3d/cluster.yaml | k3d cluster create --config -; \
-	fi
-	@echo "Cluster $(CLUSTER_NAME) ready. Pods can reach host services via flyte-host:<port>"
-
-.PHONY: cluster-delete
-cluster-delete: ## Delete k3d cluster
-	@k3d cluster delete $(CLUSTER_NAME)
-
 .PHONY: help
 help: ## Show this help message
 	@echo '🆘  Showing help message'
--- a/manager/DEVELOPMENT.md
+++ b/manager/DEVELOPMENT.md
@@ -1,5 +1,7 @@
 # Development Guide
 
+> Note: this development guide is stale, should use `make sandbox-run` command
+
 Use following method to setup the whole system and run:
 
 1. Create k3d cluster with


executor/setup.go
```diff
@@ -105,8 +107,14 @@ func Setup(ctx context.Context, sc *app.SetupContext) error {
 			return fmt.Errorf("executor: failed to initialize plugin registry: %w", err)
 		}
 
+		eventsServiceURL := sc.BaseURL
+		if eventsServiceURL == "" {
+			eventsServiceURL = cfg.StateServiceURL
```
pingsutw: Should we rename config to EventsServiceURL

executor/pkg/controller/taskaction_controller.go
```diff
@@ -259,20 +272,172 @@ func (r *TaskActionReconciler) removeFinalizer(ctx context.Context, taskAction *
...
+	if r.eventsClient == nil {
+		return nil
+	}
+
+	actionEvent := r.buildActionEvent(newTaskAction, phaseInfo)
+	if _, err := r.eventsClient.Record(ctx, connect.NewRequest(&workflow.RecordRequest{
+		Events: []*workflow.ActionEvent{actionEvent},
+	})); err != nil {
+		// Best effort: status update succeeded, do not fail reconciliation on event publish errors.
+		logger.Error(err, "failed to publish action event", "action", actionEvent.GetId().GetName())
```
pingsutw: What If the events proxy is temporarily down? events are silently lost

executor/setup.go
machichima: Fixed in https://github.com/flyteorg/flyte/pull/6970/commits/b7cf247474510e4fe277da04c143d9d05f4c61a0

executor/pkg/controller/taskaction_controller.go
machichima: I raise an error here in: https://github.com/flyteorg/flyte/pull/6970/commits/e1a9f18edc217cd338186842c1040e2ed5671ff5

And we can rely on controller-runtime's reconcile error backoff logic

executor/pkg/controller/taskaction_controller.go
```diff
@@ -259,20 +272,180 @@ func (r *TaskActionReconciler) removeFinalizer(ctx context.Context, taskAction *
...
 	logger := log.FromContext(ctx)
 
 	if !taskActionStatusChanged(oldTaskAction.Status, newTaskAction.Status) {
 		return nil
 	}
 
 	logger.Info("updateTaskActionStatus", "name", oldTaskAction.Name, "old status", oldTaskAction.Status, "new status", newTaskAction.Status)
-	err := r.Status().Update(ctx, newTaskAction)
-	if err != nil {
+	if err := r.Status().Update(ctx, newTaskAction); err != nil {
```
pingsutw: We should only update the status after the event has been successfully sent to the event service. Otherwise, the event won’t be resent if it gets lost, since the old and new action statuses would be the same

executor/pkg/controller/taskaction_controller.go
```diff
@@ -259,20 +272,180 @@ func (r *TaskActionReconciler) removeFinalizer(ctx context.Context, taskAction *
...
+}
+
+func reportedTimestamp(info *pluginsCore.TaskInfo) *timestamppb.Timestamp {
+	if info != nil && info.ReportedAt != nil {
+		return timestamppb.New(*info.ReportedAt)
 	}
+	return timestamppb.Now()
+}
+
+func outputRefs(runOutputBase, actionName string) *task.OutputReferences {
```
popojk, line 372: Is the actionName always be different if we execute the same task multiple times? Just want to make sure we will not have conflict bucket name.

executor/pkg/controller/taskaction_controller.go
```diff
@@ -259,20 +272,180 @@ func (r *TaskActionReconciler) removeFinalizer(ctx context.Context, taskAction *
...
+	return out
+}
+
+func cacheStatusFromExternalResources(resources []*pluginsCore.ExternalResource) core.CatalogCacheStatus {
+	for _, resource := range resources {
+		if resource == nil {
+			continue
+		}
+		// Return the first explicit cache status signal.
+		if resource.CacheStatus != core.CatalogCacheStatus_CACHE_DISABLED || len(resources) == 1 {
```
popojk: nit: In `len(resources) == 1` the number "1" is kind of magical for me. Could we have a comment to explain why we return`resource.CacheStatus` when `len(resources) == 1`?

executor/pkg/controller/taskaction_controller.go
machichima: The root action is created in:
https://github.com/flyteorg/flyte/blob/e1a9f18edc217cd338186842c1040e2ed5671ff5/runs/service/run_service.go#L112-L121

And for sub-actions, they are created in:
https://github.com/flyteorg/flyte-sdk/blob/a87584a9bc7a2601a6754b4cc3dd2d6481221904/src/flyte/_internal/controllers/remote/_controller.py#L186-L188

I think for sub-actions, we will always have a unique name. For root action, unique name is guaranteed when request with `ProjectId`. If request with `RunId`, `CreateRun` will raise error.

Therefor I think it's safe here

executor/pkg/controller/taskaction_controller.go
machichima: I was thinking to return the cache status directly if there's only one resource. However, I think this is redundant. 

If the only resource is not `CatalogCacheStatus_CACHE_DISABLED`, it will get into `if resource.CacheStatus != core.CatalogCacheStatus_CACHE_DISABLED` block, otherwise it will continue to the end of the function and return `CatalogCacheStatus_CACHE_DISABLED`. 

I'll remove it here

machichima: Done in https://github.com/flyteorg/flyte/pull/6970/commits/98ad3ccd52cc53445b7a21a38e0cafbde1764952

machichima:
Got error when merging v2, fixing

Issue #6968, opened 2026-03-05
[BUG] failed to record action in the database


pingsutw:
### Flyte & Flytekit version

flyte2

### Describe the bug

It seems we’re attempting to create the same action in the database multiple times, when it should only be created once
```
2026/03/05 14:33:24 /Users/kevin/git/flyte/runs/repository/impl/action.go:253 UNIQUE constraint failed: actions.org, actions.project, actions.domain, actions.name
[0.900ms] [rows:0] INSERT INTO `actions` (`org`,`project`,`domain`,`name`,`parent_action_name`,`phase`,`action_spec`,`action_details`,`ended_at`) VALUES ("testorg","flytesnacks","development","2dly0jqbp82npd9fop3r9onon","run-1772749610",1,"{""action_id"":{""run"":{""org"":""testorg"",""project"":""flytesnacks"",""domain"":""development"",""name"":""run-1772749610""},""name"":""2dly0jqbp82npd9fop3r9onon""},""parent_action_name"":""run-1772749610"",""input_uri"":""s3://flyte-data/testorg/flytesnacks/development/run-1772749610/2dly0jqbp82npd9fop3r9onon/inputs.pb"",""Spec"":{""Task"":{}}}","{}",NULL) RETURNING `created_at`,`updated_at`,`id`
{"json":{"src":"internal_run_service.go:91"},"level":"warning","msg":"RecordAction: failed to create action 2fhazqnjb2uhh2ti5pc7kqfsd: failed to create action: UNIQUE constraint failed: actions.org, actions.project, actions.domain, actions.name","ts":"2026-03-05T14:33:24-08:00"}

2026/03/05 14:33:24 /Users/kevin/git/flyte/runs/repository/impl/action.go:253 UNIQUE constraint failed: actions.org, actions.project, actions.domain, actions.name
[...]

PR #6987, opened 2026-03-06
Add OppoBloomFilter to deduplicate RecordAction calls

pingsutw:
## Summary
closes https://github.com/flyteorg/flyte/issues/6968
- Add `OppoBloomFilter` to `ActionsClient` to deduplicate `RecordAction` gRPC calls caused by Kubernetes watch event replays on reconnect
- Add configurable `RecordFilterSize` (default 8M) to actions config
- Only add to filter after successful gRPC call so transient failures can be retried
- Graceful degradation: if filter creation fails, proceeds without dedup

## Test plan

- [x] `go build ./actions/...` passes
- [x] `go test ./actions/...` passes
- [x] Unit test: duplicate `watch.Added` events → `RecordAction` called only once
- [x] Unit test: failed `RecordAction` allows retry on next event
- [x] Unit test: nil filter (backward compat) calls `RecordAction` every time

* `main` 
  - \#6583
    - \#6987 :point\_left:

### DIFF
--- a/actions/config/config.go
+++ b/actions/config/config.go
@@ -18,7 +18,8 @@ var defaultConfig = &Config{
 	},
 	WatchBufferSize:  100,
 	RunServiceURL:    "http://localhost:8090",
-	RecordFilterSize: 1 << 23, // 8M
+	// 8M slots × 8 bytes/pointer = 64 MB; can track ~8M unique actions.
+	RecordFilterSize: 1 << 23,
 }
 
 var configSection = config.MustRegisterSection(configSectionKey, defaultConfig)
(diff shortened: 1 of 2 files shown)


runs/repository/impl/action.go
```diff
@@ -250,8 +250,20 @@ func (r *actionRepo) CreateAction(ctx context.Context, runID uint, actionSpec *w
 		ActionDetails:    datatypes.JSON([]byte("{}")), // Empty details initially
 	}
 
-	if err := r.db.WithContext(ctx).Create(action).Error; err != nil {
-		return nil, fmt.Errorf("failed to create action: %w", err)
+	result := r.db.WithContext(ctx).
+		Clauses(clause.OnConflict{DoNothing: true}).
```
machichima, line 254: Thank you for this!

Issue #6934, opened 2026-02-25
[V2] Move validation to a validation function


machichima:
I'm thinking of moving this to a validation function, as well as validating other fields in `taskAction`, and put this validation function before we add the finalizer. In this case, if the validation failed, we can just not requeue it, and do not need to remove the finalizer.

We can do this in follow-up!

_Originally posted by @machichima in https://github.com/flyteorg/flyte/pull/6903#discussion_r2839757969_

PR #7006, opened 2026-03-10
Add TaskAction validation before finalizer registration

AdilFayyaz:
## Tracking issue

Closes [#6934](https://github.com/flyteorg/flyte/issues/6934)

In PR #6903, it was noted that plugin resolution (the check for whether a registered plugin exists for a given taskType) happened after the finalizer was added to the resource. This meant that if a TaskAction had an unrecognized taskType, the controller would:

1. Add a finalizer on the first reconcile
2. Discover the plugin is missing on the second reconcile
3. Mark the resource terminal

At that point, deleting the resource requires the controller to run `handleAbortAndFinalize` to remove the finalizer even though no plugin ever ran and there is nothing to clean up. Additionally, `validateSpec` only checked 4 of the 9 required spec fields, leaving runName, org, project, domain, and actionName unvalidated at the controller level.

- Replaced the separate `validateSpec` function and inline `ResolvePlugin` block with a unified `validateTaskAction` function that validates all required spec fields and resolves the plugin in a single pass
- This validation now runs before the finalizer is added if validation fails, the resource is marked terminal (Failed condition), not requeued, and no finalizer is registered, making deletion trivial
- Introduced a `pluginResolver` interface so `validateTaskAction` is testable without a real plugin registry
- Added `taskaction_validation_test.go` with standard Go unit tests covering all 9 required field checks and the plugin-not-found case
- Fixed the existing integration test fixture which was missing required fields (taskType, taskTemplate)

## How was this patch tested?

- Unit tests: `go test ./pkg/controller/... -run TestValidateTaskAction -v`  covers all 9 missing-field cases and plugin-not-found
- Integration test: `go test ./pkg/controller/... -v` confirms reconcile completes without error and sets conditions

## Docs link

- `main` 
  - \#6583
    - \#7006 :point\_left:

### DIFF
--- a/executor/pkg/controller/taskaction_controller_test.go
+++ b/executor/pkg/controller/taskaction_controller_test.go
@@ -30,10 +30,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	flyteorgv1 "github.com/flyteorg/flyte/v2/executor/api/v1"
-	"github.com/flyteorg/flyte/v2/executor/pkg/plugin"
-	pluginsCore "github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/core"
-	k8sPlugin "github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/k8s"
-	"k8s.io/client-go/tools/record"
 	pluginsCore "github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/core"
 	k8sPlugin "github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/k8s"
 	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/core"
@@ -70,12 +66,6 @@ func buildTaskTemplateBytes(taskType, image string) []byte {
 	Expect(err).NotTo(HaveOccurred())
 	return data
 }
-
-// emptyPluginRegistry satisfies plugin.PluginRegistryIface with no registered plugins.
-type emptyPluginRegistry struct{}
-
-func (emptyPluginRegistry) GetCorePlugins() []pluginsCore.PluginEntry { return nil }
-func (emptyPluginRegistry) GetK8sPlugins() []k8sPlugin.PluginEntry   { return nil }
 
 // emptyPluginRegistry satisfies plugin.PluginRegistryIface with no registered plugins.
 type emptyPluginRegistry struct{}


executor/pkg/controller/taskaction_validation_test.go
```diff
@@ -0,0 +1,129 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
```
pingsutw: No need license here

executor/pkg/controller/taskaction_controller.go
```diff
@@ -50,9 +50,9 @@ const (
 type K8sEventType string
 
 const (
-	FailedUnmarshal     K8sEventType = "FailedUnmarshal"
-	FailedPluginResolve K8sEventType = "FailedPluginResolve"
-	FailedPluginHandle  K8sEventType = "FailedPluginHandle"
+	FailedUnmarshal    K8sEventType = "FailedUnmarshal"
+	FailedValidation   K8sEventType = "FailedValidation"
```
pingsutw: Could we keep the FailedPluginResolve type? 
Failed to find the plugin => FailedPluginResolve
missing project/domain => FailedValidation

executor/pkg/controller/taskaction_controller.go
```diff
@@ -112,20 +112,19 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-		logger.Error(err, "TaskAction spec is invalid")
-		r.Recorder.Eventf(taskAction, corev1.EventTypeWarning, string(flyteorgv1.ConditionReasonInvalidSpec),
-			"Invalid spec: %v", err)
-		setCondition(taskAction, flyteorgv1.ConditionTypeFailed, metav1.ConditionTrue,
-			flyteorgv1.ConditionReasonInvalidSpec, err.Error())
-		setCondition(taskAction, flyteorgv1.ConditionTypeProgressing, metav1.ConditionFalse,
-			flyteorgv1.ConditionReasonInvalidSpec, err.Error())
+	// Validate spec fields and resolve plugin before adding the finalizer
+	// If either fails, the resource is marked terminal and not requeued — no finalizer to clean up
+	p, reason, err := validateTaskAction(taskAction, r.PluginRegistry)
```
pingsutw, line 131: We can first check whether taskAction has a finalizer here. If it does, it means it has already been validated, so we can skip it.

AdilFayyaz: We can do that, however it would introduce branching complexity, the spec checks are just string comparisons so the cost is negligilble. Happy to make the change if you feel strongly.

PR #7009, opened 2026-03-10
Add file_extension field to BlobType

### LINKED ISSUE #7024: [BUG] [copilot] File extensions are missing when copilot downloads Blob/FlyteFile inputs
### Flyte & Flytekit version

Flyte 1.16

### Describe the bug

Copilot today has the following behavior:

```
# ContainerTask, as an example
def t1(file: FlyteFile["csv"]):
    ... # copilot downloads the file to e.g. /inputs/file
```

Sometimes, this is unintuitive (perhaps a bug? or an enhancement request? not clear...) and it may be expected that the file is downloaded to `/inputs/file.csv` instead.

### Expected behavior

The file is downloaded to `/inputs/file.csv` instead. However, the "bugfix" must be done carefully, as it is likely that existing workflows are already expecting the file to be downloaded to `/inputs/file`.

_No response_

### Screenshots

_No response_

- [x] Yes

- [x] Yes

ddl-rliu:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/7024 [BUG] [copilot] File extensions are missing when copilot downloads Blob/FlyteFile inputs

We'll assume the above behavior is not a bug (it is long-standing, and a "bugfix" will likely break existing workflows). Instead, this PR proposes enhancements to FlyteFile/Blob to support writing workflow inputs **with** the file extension. This PR includes several changes across flyteidl and flytecopilot.

The enhancements allow workflows to be flexible when writing blobs. Specifically, the existing behavior where flytecopilot writes blobs during the download phase without the file extension (e.g. "inputs/data") can now be enhanced so that the file extension is included (e.g. "inputs/data.csv").

edit: Just to add to the above - personally, I can use the `/inputs/data` path, and using user code the file can be renamed to `/inputs/data.csv` or something similar, if needed.

But the key issue(s)/use case(s) is this:
* We have a lot of users using Flyte, and bringing with them a great variety of user code. Highly technical users are comfortable changing the user code to rename file names for example, but the rest of the users get stuck and they find that their code fails when it runs in Flyte (namely, the user code expects `.csv` but flytecopilot writes the file without `.csv`)
* So Flyte uses blob storage. But in parallel, all the data that users work with is living in an NFS file system. Namely, all their files in the NFS are typical files (`data.csv`, etc. - files with extensions). Then, all of their user code expects paths more-or-less like `/inputs/data.csv`.
* Even though in Python, something like `pandas.read_csv(path)` is easily interchangeable with `path=/inputs/data vs. path=/inputs/data.csv`, many of our users write "SAS" code. SAS is a weird language where SAS datasets must have a particular name, like `/inputs/data.sas7bdat`, and it fails to process a file like `/inputs/data`. So unfortunately, a lot of the SAS scripts out-of-the-box will fail, but with a copilot file extensions fix, it does work!

[flyteidl]

Add a new `file_extension` string field to the BlobType protobuf message, allowing FlyteFile to optionally specify a file extension (e.g. "csv") that flytecopilot appends when writing blobs during the download phase (e.g. "data.csv"). When empty (the default), behavior is unchanged.

~Add a new `enable_legacy_filename` bool field to the BlobType protobuf message, allowing FlyteFile to optionally specify whether to preserve backward compatibility for tasks that read from the extensionless path.~

Regenerated all protobuf bindings (Go, Python, JS, Rust, ES, Swagger).

[flytecopilot]

The copilot download phase infers the desired download behavior(s) from the input interface. 

[flytekit] PR: https://github.com/flyteorg/flytekit/pull/3406

### Alternatives considered

The PR's approach configures the file download behaviors at the BlobType level (e.g. per FlyteFile). This has several pros (granularity, explicitness).

But, one con is that unlike `BlobType.format` (which can be inferred from the output filename e.g. "data.csv" -> `format: "csv"`), the new fields `BlobType.file_extension, BlobType.enable_legacy_filename` could not be inferred from the output filename (does "data.csv" match to `file_extension: "csv"`?). Ultimately, this seems like it is introducing a minor inconsistency, but acceptable given the benefits. (It also seems like this is all hypothetical - [copilot upload does not](https://github.com/flyteorg/flyte/pull/7009/changes#diff-a1630c62d571c8e788788eb138a75ab7c7b90d0e09cc2b3d991c95970ffe2d49L110) **actually** infer the format from the filename, so this concern may be somewhat moot?)

Here are the other approaches I considered:

**1. New flyte-copilot CLI flags `--file_extension_config ENUM=(disabled, enabled, legacy)`**

file_extension_config = disabled (by default) - Same behavior as today (`data: FlyteFile[csv]` is written to `outputs/data`)
file_extension_config = enabled - New behavior (`data: FlyteFile[csv]` is written to `outputs/data.csv`)
file_extension_config = legacy - New behavior but backwards compatible, (`data: FlyteFile[csv]` is written to `outputs/data` and `outputs/data.csv`)

Cons:
- The download behavior is granular at the file level, but the CLI flag is granular at the task level (more or less). This makes the configuration less flexible, and begins to somewhat overload the configurability of copilot at the CLI level.

**2. No changes, user code should be modified to read from the existing path e.g. `data` and add the extension itself**

Cons:
- Migrating large projects to Flyte becomes burdensome, when needing to add another extra step like this here - user code needs to be modified to handle adjusting the file extensions on inputs.

## How was this patch tested?

```
# Container task python code
def t1(datasetA: FlyteFile[TypeVar('csv')], datasetB=Annotated[FlyteFile[TypeVar('csv')], FileDownloadConfig(file_extension="csv", enable_legacy_filename=True)]): …

# flyte-copilot logs
{"json":{},"level":"debug","msg":"inputInterface: [variables:{key:\"datasetA\"  value:{type:{blob:{format:\"csv\"}}  description:\"datasetA\"}}  variables:{key:\"datasetB\"  value:{type:{blob:{format:\"csv\"  file_extension:\"csv\"  enable_legacy_filename:true}}  description:\"datasetB\"}}]","ts":"2026-03-10T23:20:55Z"}
{"json":{},"level":"debug","msg":"varFilenames: map[datasetA:[datasetA[] datasetB:[datasetB.csv datasetB]]","ts":"2026-03-10T23:20:55Z"}
...
{"json":{},"level":"info","msg":"Successfully copied [4119] bytes remote data from [s3://rliucp108882-flyte-data//a0c46ac7-eea3-427d-91c1-a3c6b4a27e83/datasetB] to local [/execution-vol/flows/workflow/inputs/datasetB]","ts":"2026-03-10T23:20:55Z"}
{"json":{},"level":"info","msg":"Successfully copied [4119] bytes remote data from [s3://rliucp108882-flyte-data//a0c46ac7-eea3-427d-91c1-a3c6b4a27e83/datasetB] to local [/execution-vol/flows/workflow/inputs/datasetB.csv]","ts":"2026-03-10T23:20:55Z"}
{"json":{},"level":"info","msg":"Successfully copied [3904] bytes remote data from [s3://rliucp108882-flyte-data//af859c71-811d-4c26-9845-b3fd8830ab6b/datasetA] to local [/execution-vol/flows/workflow/inputs/datasetA]","ts":"2026-03-10T23:20:55Z"}

```

### Labels


flytecopilot/data/download.go
```diff
@@ -486,14 +522,20 @@ func (d Downloader) RecursiveDownload(ctx context.Context, inputs *core.LiteralM
...
-		varPath := path.Join(dir, variable)
 		lit := literal
 		f[variable] = futures.NewAsyncFuture(childCtx, func(ctx2 context.Context) (interface{}, error) {
-			v, lit, err := d.handleLiteral(ctx2, lit, varPath, writePrimitiveToFile)
-			if err != nil {
-				return nil, err
+			var resultLit *core.Literal
+			var resultV interface{}
+			var err error
+			// TODO: Refactor handleLiteral to accept a list of file paths
```
ddl-rliu: Assuming this PR's general approach is acceptable, I can do this refactor in the same PR, or in a follow-up PR (Also, there can be an improvement here to use symlinks, if the output will be written 2+ times).

The refactor isn't actually needed for correctness, it's more for clarity. The returned resultV, resultLit is identical for calls to `handleLiteral` that only differ by the value of `varPath`.

flyteidl/protos/flyteidl/core/types.proto
```diff
@@ -80,6 +80,20 @@ message BlobType {
...
     string format = 1;
     BlobDimensionality dimensionality = 2;
+
+    // Optional file extension (e.g. "csv", "parquet") to use during copilot download.
+    // Default is "", which means no extension is appended.
+    // Differences from "format":
+    //   1. "format" is used for type validation in flytekit, "file_extension" is not.
+    //   2. "file_extension" controls the file extension of the blob when materializing
+    //    to local disk during copilot download, unlike "format".
+    string file_extension = 3;
```
ddl-rliu, line 87: I'm open to merging this concept with the existing `format` field, but I think we'd still need a new bool field like `enable_legacy_filename` for backwards compatibility.

The initial motivation was that in some case allowing `format != file_extension` would be more flexible; also, this way the Blob is very explicit about whether a file extension will be added to it (you can visually parse the BlobType values and immediately know what file extension, if any, will be added. otherwise, if the concepts were merged, you could not immediately assume that "format=csv" implies the file is written with the extension.)

machichima: We would like to prevent updating the proto as possible. Could we:
1.  Add the file extension based on the `format`. In flytekit we can add this format as file extension. Is there any case that you think using `format` here cannot fit in?
2. Rather than adding flag, could we directly fallback to find file without extension if the one with extension does not exists?

ddl-rliu: > Add the file extension based on the format

Totally open to this! Are you thinking that e.g. "format=csv" implies flytecopilot will always write the file with the extension? My initial suggestion was mostly from an explicitness/compatibility perspective.

Also, I wanted to understand more about avoiding updates to the proto. Was it a performance concern? I just did a quick check to 100% ensure the happy path proto messages are unaffected by this change.
```
from flyteidl.core.types_pb2 import BlobType
blob = BlobType(
    format="csv",
    dimensionality=BlobType.SINGLE,
)

# flyte@master (latest flyte head)
blob.SerializeToString().hex() # value is 0a03637376

# flyte@rliu.DOM-75010.file-ext-copilot (my branch)
blob.SerializeToString().hex() # value is 0a03637376
```
So the optional new fields are indeed being excluded from happy path existing proto messages, which is probably an important requirement.

machichima: > Also, I wanted to understand more about avoiding updates to the proto. Was it a performance concern?

It's not a performance concern. My main concern is that changing the proto means changing a public API, which also requires corresponding client-side updates.

I'm not opposed to modifying the proto here, especially since these are optional fields. But we still need to be careful about the cases where the fields are missing. In particular, we should make sure both of these version-skew scenarios behave correctly:
1. new flytekit with old backend
2. old flytekit with new backend

Therefor, I think the bigger concern is API compatibility and missing-field semantics. If we can avoid modifying the proto and still achieve the same behavior, I would prefer that.

ddl-rliu: Interesting, I'm thinking of ways to automate a version-skew test that gives reproducible results. I've created a proof of concept here: https://github.com/flyteorg/flyte/pull/7147/changes#r3030688248

I think the results are positive for the changes in this PR, meaning that version-skew testing is satisfactory for the new optional fields. Here are the results I achieved: https://github.com/flyteorg/flyte/pull/7147/changes#r3030772543
> Backend (flyte) and Client (flytekit)
>  - Old: master@558cab64ffb60044653f19f1ea952133e233208e
>  - New: rliu.DOM-75010.file-ext-copilot@9327eb02af6acac3d31c844413583cd9d13a74d0
> 
> Test 1: New client (flytekit) and old backend (flyte)
> 
> message = BlobType(
>   format="test",
>   dimensionality=BlobType.SINGLE,
>   file_extension="test",
>   enable_legacy_filename=True,
> )
> 
> New client (flytekit) serializes the message to: 0a04746573741a04746573742001
> Old backend (flyte) deserializes the message to: BlobType("test", BlobType.SINGLE)
> // [Author's note] the optional params are dropped, which is the desired behavior.
> 
> Test 2: Old client (flytekit) and new backend (flyte)
> 
> message = BlobType(
>   format="test",
>   dimensionality=BlobType.SINGLE,
> )
> 
> Old client (flytekit) serializes the message to: 0a0474657374
> New backend (flyte) deserializes the message to: BlobType("test", BlobType.SINGLE, "", False)
> // [Author's note] the optional params are populated with the default values "", False, which is the desired behavior.

If this kind of testing is directionally correct, that's awesome and I can extend it more in any way, if needed!

---

Prior to this PR, I considered two other approaches that do not involve modifying the proto: https://github.com/flyteorg/flyte/pull/7009#issue-4054799248

> **1. New flyte-copilot CLI flags `--file_extension_config ENUM=(disabled, enabled, legacy)`**
> 
> file_extension_config = disabled (by default) - Same behavior as today (`data: FlyteFile[csv]` is written to `outputs/data`)
> file_extension_config = enabled - New behavior (`data: FlyteFile[csv]` is written to `outputs/data.csv`)
> file_extension_config = legacy - New behavior but backwards compatible, (`data: FlyteFile[csv]` is written to `outputs/data` and `outputs/data.csv`)
> 
> Cons:
> - The download behavior is granular at the file level, but the CLI flag is granular at the task level (more or less). This makes the configuration less flexible, and begins to somewhat overload the configurability of copilot at the CLI level.
> 
> **2. No changes, user code should be modified to read from the existing path e.g. `data` and add the extension itself**
> 
> Cons:
> - Migrating large projects to Flyte becomes burdensome, when needing to add another extra step like this here - user code needs to be modified to handle adjusting the file extensions on inputs.

After my analysis, I concluded that extending the proto would have the most desirable result, and I'm hopeful the version-skew testing I included can help give an API compatibility/missing field handling guarantee. But I'm also happy to investigate any of the alternative approaches more, and give more confidence in the final decision.

flytecopilot/data/download.go
```diff
@@ -486,14 +519,26 @@ func (d Downloader) RecursiveDownload(ctx context.Context, inputs *core.LiteralM
...
+			var resultV interface{}
+			var err error
+			if len(varFilenames[variable]) == 0 {
+				filenames = []string{variable}
+			} else {
+				filenames = varFilenames[variable]
+			}
+			for _, filename := range filenames {
+				varPath := path.Join(dir, filename)
+				// TODO: Refactor handleLiteral to accept a list of file paths and return a list of downloaded results
```
ddl-ebrown: TODO here

flyteidl/protos/flyteidl/core/types.proto
machichima: Thank you for your detailed reply! I agree that we would need to modify the proto here. 

After re-thinking on using `format` as `file_extension`, this may not be a good way as users may already use `format=XXX` but does not expect having file extension. Changing the semantics of this field can cause breaking change. Therefor I agree that add `file_extension` here is better.

The question I got is whether we really need `enable_legacy_filename`. I imagine that if user want copilot to download with the extension, they will set `file_extension=XXX`, and copilot will download file as `inputs /data.XXX`. If they want to keep the original behavior, they can just define `FlyteFile` without `file_extension`, then copilot will download file as `inputs/data`, keeping the old behavior.

Could we only add `file_extension` field in proto here?

ddl-rliu: Yes, that makes sense. I changed it to only add `file_extension` to the proto! https://github.com/flyteorg/flyte/pull/7009/commits/32627a18c89c18482e5492168aaffc2fc773da9d

Re-ran locally to test the changes

```
# Container task python code
def t1(datasetA: Annotated[FlyteFile[TypeVar('csv')], FileDownloadConfig(file_extension="csv")], datasetB: FlyteFile[TypeVar('csv')]): …

# flyte-copilot logs
2026-04-08T23:48:42.480Z: {"json":{},"level":"info","msg":"Successfully copied [196608] bytes remote data from [s3://ls-engc-flyte-data//858fd656-11ec-41ec-b235-b3b2984f230e/ex] to local [/execution-vol/flows/workflow/inputs/datasetB]","ts":"2026-04-08T23:48:42Z"}
2026-04-08T23:48:42.564Z: {"json":{},"level":"info","msg":"Successfully copied [1703936] bytes remote data from [s3://ls-engc-flyte-data//25d4f8d8-7eb4-492b-97c4-ab3ce8bdcf74/ae] to local [/execution-vol/flows/workflow/inputs/datasetA.csv]","ts":"2026-04-08T23:48:42Z"}
```

Re-ran version skew test as well https://github.com/flyteorg/flyte/pull/7147/changes#r3054453586

flytecopilot/data/download.go
ddl-rliu: fixed! https://github.com/flyteorg/flyte/pull/7009/commits/32627a18c89c18482e5492168aaffc2fc773da9d

flyteidl/protos/flyteidl/core/types.proto
```diff
@@ -80,6 +80,14 @@ message BlobType {
     // csv, parquet etc
     string format = 1;
     BlobDimensionality dimensionality = 2;
+
+    // Optional file extension (e.g. "csv", "parquet") to use during copilot download.
+    // Default is "", which means no extension is appended.
+    // Differences from "format":
+    //   1. "format" is used for type validation in flytekit, "file_extension" is not.
+    //   2. "file_extension" controls the file extension of the blob when materializing
+    //    to local disk during e.g. flytecopilot download, unlike "format".
```
machichima:
```suggestion
    // Optional file extension (e.g. "csv", "parquet") appended to the blob filename during flytecopilot download.
    // Unlike "format", this field is not used for type validation in SDK.
    // It only controls the local filename when materializing to disk.
```

nit

machichima:
Thank you! The proto looks good to me. I will look into the implementation.
Could you please also update the flytekit PR so that we can test it locally?

ddl-rliu:
Done! https://github.com/flyteorg/flytekit/pull/3406

flytecopilot/data/download.go
```diff
@@ -26,13 +27,46 @@ import (
...
 
+// By default, blobs (FlyteFiles) were not and still are not written with a
+// file extension. For example, a data: FlyteFile["csv"] blob is written
+// to "inputs/data", even though Format="csv".
+//
+// When FileExtension="" (the default), this old behavior is preserved.
+//
+// However, an input blob
+// `data: Annotated[FlyteFile["csv"], FileDownloadConfig(file_extension="csv")]`
+// should be written to "inputs/data.csv" (when FileExtension="csv" - new behavior).
```
machichima, line 47:
```suggestion
// should be written to "inputs/data.csv" (when FileExtension="csv").
```
nit

flytecopilot/data/download.go
```diff
@@ -26,13 +27,46 @@ import (
...
+func resolveVarFilenames(ctx context.Context, vars *core.VariableMap) map[string]string {
+	varFilenames := make(map[string]string, len(vars.GetVariables()))
+	for varName, variable := range vars.GetVariables() {
+		varType := variable.GetType()
+		switch varType.GetType().(type) {
+		case *core.LiteralType_Blob:
+			ext := varType.GetBlob().GetFileExtension()
```
machichima, line 54: Could we check `varType.GetBlob().GetDimensionality() == core.BlobType_MULTIPART` to prevent adding file extension to directory name

flytecopilot/data/download.go
```diff
@@ -486,7 +520,11 @@ func (d Downloader) RecursiveDownload(ctx context.Context, inputs *core.LiteralM
 			}
 			logger.Infof(ctx, "read object at location [%s]", offloadedMetadataURI)
 		}
-		varPath := path.Join(dir, variable)
+		filename := variable
+		if varFilename, ok := varFilenames[variable]; ok {
+			filename = varFilename
+		}
+		varPath := path.Join(dir, filename)
```
machichima, line 527: Could we get file extension through
`literal.GetScalar().GetBlob().GetMetadata().GetType().GetFileExtension()` here instead? So that we  do not need to pass in extra `varFilenames`

flytecopilot/data/download.go
```diff
@@ -26,13 +27,46 @@ import (
 	"github.com/flyteorg/flyte/flytestdlib/storage"
 )
 
+var validFileExtensionRe = regexp.MustCompile(`^[a-zA-Z0-9]+(\.[a-zA-Z0-9]+)*$`)
```
machichima, line 30: nit: Could we add a comment saying that this matches single and multi-part file extension (e.g. `tar.gz`)?

Issue #5469, opened 2024-06-11
[BUG] New versions of viper breaks config loading


Sovietaced:
### Describe the bug

My custom Flyte plugin has a depenency that pulls in a more recent version of viper v1.11.0 -> v1.18.2 and this appears to break config loading since flyteplugins `config_load_test` fails. I realize that this is not a present bug but I figured I would open this up in case anyone has an issue and in the event that Flyte eventually updates viper to a more recent version. 

It appears that the latest entry in defined in a YAML map will be the only entry in the map that is parsed. For example:

```
=== RUN   TestLoadConfig/k8s-config-test
    config_load_test.go:30: 
        	Error Trace:	/home/jparraga/code/flyte/flyteplugins/go/tasks/config_load_test.go:30
        	Error:      	Not equal: 
        	            	expected: map[string]string{"annotationKey1":"annotationValue1", "annotationKey2":"annotationValue2"}
        	            	actual  : map[string]string{"annotationkey2":"annotationValue2"}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,4 +1,3 @@
        	            	-(map[string]string) (len=2) {
        	            	- (string) (len=14) "annotationKey1": (string) (len=16) "annotationValue1",
        	            	- (string) (len=14) "annotationKey2": (string) (len=16) "annotationValue2"
        	            	+(map[string]string) (len=1) {
        	            	+ (string) (len=14) "annotationkey2": (string) (len=16) "annotationValue2"
        	            	 }
        	Test:       	TestLoadConfig/k8s-config-test
```        	

It appears the error happens in [decoding the custom config into the root config](https://github.com/flyteorg/flyte/blob/master/flytestdlib/config/viper/viper.go#L303). From what I've seen there is a different in the typing of the configuration that is parsed by viper. Notably, map entries used to be parsed as `map[interface{}]interface{}` and are now parsed as `map[string]interface{}`. The code is bit hair but I figured I'd file a ticket before I dig deeper here. 

### Expected behavior

I would expect config loading to work correctly and unit tests to pass. 

### Additional context to reproduce

`go get github.com/spf13/viper@v1.18.2`

proceed to run unit tests in `flyteplugins`

### Screenshots

v1.11.0
<img width="508" alt="v1 11 0" src="https://github.com/flyteorg/flyte/assets/1274471/61feb4a4-d3ee-4297-8232-5e38a7250290">

v.1.18.2
[...]

PR #7013, opened 2026-03-11
Upgrade viper to v1.21.0 with case-sensitive key fix

honnix:
## Summary

Upgrade `github.com/spf13/viper` from v1.11.0 to v1.21.0 across all 7 Go modules.

This addresses the dependency incompatibility described in https://github.com/flyteorg/flyte/issues/5469 — where upgrading viper broke case-sensitive map keys and caused only the last map entry to be parsed.

### What changed

**Dependency updates:**
- `github.com/spf13/viper` v1.11.0 → v1.21.0
- `github.com/mitchellh/mapstructure` → `github.com/go-viper/mapstructure/v2` (viper's new fork)
- Transitive bumps: `fsnotify` v1.6.0 → v1.9.0, `spf13/afero`, `spf13/cast`, `spf13/pflag`, `pelletier/go-toml/v2`, `subosito/gotenv`

**Code fixes in `flytestdlib/config/viper/viper.go`:**

1. **`sliceToMapHook` update** — The hook that converts YAML arrays to maps (the workaround for viper's case-insensitive keys) now handles `map[string]interface{}` in addition to `map[interface{}]interface{}`. Newer YAML parsing in viper v1.21 returns the former type.

2. **Case-sensitive key restoration** — Viper v1.21 added `insensitiveArray()` which recursively lowercases all map keys inside arrays. This broke the array-based workaround for case-sensitive map keys that Flyte relies on.

   **Root cause:** In v1.11, `insensitiviseMap` only recursed into nested maps, leaving array contents with original case. In v1.21, the new `insensitiveArray` function recurses into arrays too, lowercasing everything.

   **Fix:** After calling `AllSettings()`, re-read config files directly with `gopkg.in/yaml.v3` (which preserves key case) and restore the original casing for values within arrays via `restoreCaseSensitiveArrayKeys()`. This selectively replaces lowercased array data from viper with case-preserved data from the raw YAML, while keeping viper's env var and pflag integration intact.

**Import migration (67+ files):**
- All `*_flags_test.go` files and the code generator template (`templates.go`) updated from `github.com/mitchellh/mapstructure` to `github.com/go-viper/mapstructure/v2`.

## Test plan

- [x] All 7 modules compile (`go build ./...`)
- [x] `flytestdlib/config/viper` tests pass
- [x] `flytestdlib/cli/pflags/api` tests pass (code generator)
- [x] `NestedMaps` test passes with case-sensitive keys (`itemA`, `itemAa`, etc.)
- [x] `Simple` map config test passes
- [x] Full `flytestdlib` test suite passes (except macOS-only symlink watcher, see below)
- [x] CI on Linux (symlink watcher test should pass with inotify)

### Known issue: macOS-only test failure

`TestAccessor_UpdateConfig/[Viper]_Change_handler_k8s_configmaps` fails on macOS due to `fsnotify` v1.9.0's kqueue backend not emitting events when a symlink within a watched directory is replaced via `ln -sfn`. This is a platform-specific limitation — on Linux with inotify, the watcher correctly detects symlink replacements. This test should pass in CI.


flytestdlib/config/viper/viper.go
```diff
@@ -6,16 +6,18 @@ import (
...
 	"fmt"
+	"os"
 	"reflect"
 	"strings"
 	"sync"
 
 	"github.com/fsnotify/fsnotify"
-	"github.com/mitchellh/mapstructure"
+	"github.com/go-viper/mapstructure/v2"
 	"github.com/pkg/errors"
```
honnix, line 16: Viper v1.21 switched from `mitchellh/mapstructure` to `go-viper/mapstructure/v2`. All 67 `*_flags_test.go` files across the repo have the same one-line import change.

flytestdlib/config/viper/viper.go
```diff
@@ -6,16 +6,18 @@ import (
...
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	viperLib "github.com/spf13/viper"
+	"gopkg.in/yaml.v3"
```
honnix, line 20: New import: used to re-read config files directly and restore case-sensitive keys that viper v1.21 lowercases (see `restoreCaseSensitiveArrayKeys` below).

flytestdlib/config/viper/viper.go
```diff
@@ -182,18 +184,22 @@ func sliceToMapHook(f reflect.Kind, t reflect.Kind, data interface{}) (interface
 	// Only handle slice -> map conversion
 	if f == reflect.Slice && t == reflect.Map {
 		// this will be the target result
-		res := map[interface{}]interface{}{}
+		res := map[string]interface{}{}
```
honnix, line 187: **`sliceToMapHook` fix**: Viper v1.21 uses a newer YAML parser that returns `map[string]interface{}` instead of `map[interface{}]interface{}`. Updated to handle both via type switch. Without this, array-to-map conversion silently returned raw data.

flytestdlib/config/viper/viper.go
```diff
@@ -261,7 +267,61 @@ func jsonUnmarshallerHook(_, to reflect.Type, data interface{}) (interface{}, er
 // Parses RootType config from parsed Viper settings. This should be called after viper has parsed config file/pflags...etc.
 func (v viperAccessor) parseViperConfig(root config.Section) error {
 	// We use AllSettings instead of AllKeys to get the root level keys folded.
-	return v.parseViperConfigRecursive(root, v.viper.AllSettings())
+	settings := v.viper.AllSettings()
```
honnix, line 270: **Core case-sensitivity fix**: Viper v1.11's `insensitiviseMap` did NOT recurse into arrays, so array contents preserved key case. Viper v1.21 added `insensitiveArray()` which lowercases everything — breaking the array-based workaround for case-sensitive map keys (ref: https://github.com/flyteorg/flyte/issues/5469).

Fix: after `AllSettings()`, re-read config files with `gopkg.in/yaml.v3` (which preserves case) and replace array values with the originals. Env var and pflag overrides from viper are unaffected since they apply to scalar values, not arrays.

flytestdlib/config/viper/viper.go
```diff
@@ -261,7 +267,61 @@ func jsonUnmarshallerHook(_, to reflect.Type, data interface{}) (interface{}, er
...
+		if err := yaml.Unmarshal(data, &rawSettings); err != nil {
+			continue
+		}
+
+		restoreCaseSensitiveArrayKeys(settings, rawSettings)
+	}
+
+	return v.parseViperConfigRecursive(root, settings)
+}
+
```
honnix, line 296: Walks viper's lowercased settings and the raw YAML in parallel. Only replaces `[]interface{}` (array) values — these are the ones that contain case-sensitive map keys encoded as YAML arrays. Nested maps are recursed into to find deeper arrays.

flytestdlib/cli/pflags/api/templates.go
```diff
@@ -98,7 +98,7 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/mitchellh/mapstructure"
+	"github.com/go-viper/mapstructure/v2"
```
honnix, line 101: This is the code generator template. This single import change is what produces the identical one-line diff in all 67 `*_flags_test.go` files across the repo.

flyteadmin/auth/cookie_manager_test.go
```diff
@@ -315,14 +315,11 @@ func TestExtractAccessTokenFromCookies(t *testing.T) {
 		t.Logf("Cookie 1 raw decoded (first 50 bytes): %s", string(decoded[:50]))
 	}
 	s := securecookie.New(hashKey, blockKey)
+	// Disable timestamp validation since this test uses hardcoded cookies
```
honnix: **Pre-existing bug** (not caused by viper upgrade): The hardcoded securecookie values have a creation timestamp of ~Jan 30 2026. With securecookie's default `MaxAge` of 86400s (1 day), they expired on Jan 31. `MaxAge(0)` disables timestamp validation — this test verifies token extraction, not cookie freshness.

datacatalog/pkg/config/config_flags_test.go
```diff
@@ -10,7 +10,7 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/mitchellh/mapstructure"
+	"github.com/go-viper/mapstructure/v2"
```
honnix, line 13: Representative of all 67 `*_flags_test.go` files: mechanical import change from `mitchellh/mapstructure` to `go-viper/mapstructure/v2`.

flytestdlib/config/viper/viper.go
```diff
@@ -261,7 +267,61 @@ func jsonUnmarshallerHook(_, to reflect.Type, data interface{}) (interface{}, er
...
+func restoreCaseSensitiveArrayKeys(viperData, rawData map[string]interface{}) {
+	for lowerKey, viperVal := range viperData {
+		// Find matching key in rawData (case-insensitive match)
+		var rawVal interface{}
+		for rawKey, rv := range rawData {
+			if strings.EqualFold(rawKey, lowerKey) {
+				rawVal = rv
+				break
+			}
+		}
```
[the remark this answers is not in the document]
honnix: I'm not sure whether the complexity of code pays off the performance gain, as the config file wouldn't be gigantic. That said, I'm open to adjust if needed.

flytestdlib/config/viper/viper.go
```diff
@@ -261,7 +267,61 @@ func jsonUnmarshallerHook(_, to reflect.Type, data interface{}) (interface{}, er
...
+
+		data, err := os.ReadFile(configFile)
+		if err != nil {
+			continue
+		}
+
+		var rawSettings map[string]interface{}
+		if err := yaml.Unmarshal(data, &rawSettings); err != nil {
+			continue
+		}
```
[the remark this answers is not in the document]
honnix: This makes sense.

Sovietaced (approved):

Sovietaced:
I reviewed the code. I'm not very familiar with the viper codebase but this seems to pass integration/functional tests that parse config files so I think it's worth landing since this has been a large source of frustration.

PR #7014, opened 2026-03-11
Fix Docker build for CGO-dependent sqlite3 package

pingsutw:
## Summary

- Remove cross-compilation (`--platform=${BUILDPLATFORM}`) from Dockerfile since `go-sqlite3` requires CGO which is incompatible with cross-compilation
- Enable `CGO_ENABLED=1` for native build
- Fix `.dockerignore` negation pattern by replacing unreliable `gen/` + `!gen/go/**` with explicit subdirectory exclusions (`gen/python/`, `gen/rust/`, `gen/ts/`)
- Copy only `gen/go` instead of full `gen` directory in Dockerfile

## Test plan

- [x] Verified `docker build -f Dockerfile .` completes successfully
- [x] Verified `make sandbox-build` completes successfully

- `main` 
  - \#6583
    - \#7014 :point\_left:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,11 @@
 # Todo(alex): We should add UI into the image when UI is done
 
-FROM golang:1.24-bookworm AS flytebuilder
+FROM --platform=${BUILDPLATFORM} golang:1.24-bookworm AS flytebuilder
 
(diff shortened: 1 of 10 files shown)


popojk (approved):
LGTM, thanks!

PR #7018, opened 2026-03-12
[Sandbox][RunService]Switch database from sqlite to postgresql

### LINKED ISSUE #7002: [V2][sandbox] Set default db to PostgreSQL in sandbox

We should set the default db of sandbox to be PostgreSQL 

Run service should call PostgreSQL db.

n/a

_No response_

- [x] Yes

- [x] Yes

popojk:
## Tracking issue

Closes #7002

We should switch the database from sqlite to PostgreSQL so that we can use pg notify in RunService

- Update sandbox Helm chart
- Update DB layer settings

## How was this patch tested?

1. Buid with `make sandbox_build` and run sandbox with `make sandbox_run`
2. Run an example in flyte-sdk `python examples/basics/devbox_one.py`
3. Link to local sandbox PostgreSQL with the config:

```
host: localhost
port: 30001
username: postgres
password: postgres
```

4. The actions data successfully written into db Screenshot 2026-03-12 at 2 10 24 PM

### Labels

## Docs link

- `main` 
  - \#6583
    - \#7018 :point\_left:

### DIFF
--- a/flytestdlib/database/postgres.go
+++ b/flytestdlib/database/postgres.go
@@ -42,13 +42,8 @@ func resolvePassword(ctx context.Context, passwordVal, passwordPath string) stri
 	return password
 }
 
-// GetPostgresDSN produces the DSN (data source name) for opening a postgres db connection.
-func GetPostgresDSN(ctx context.Context, pgConfig PostgresConfig) string {
-	return getPostgresDsn(ctx, pgConfig)
-}
-
 // Produces the DSN (data source name) for opening a postgres db connection.
-func getPostgresDsn(ctx context.Context, pgConfig PostgresConfig) string {
+func GetPostgresDsn(ctx context.Context, pgConfig PostgresConfig) string {
 	password := resolvePassword(ctx, pgConfig.Password, pgConfig.PasswordPath)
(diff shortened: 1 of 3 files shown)


flytestdlib/database/postgres.go
```diff
@@ -42,6 +42,11 @@ func resolvePassword(ctx context.Context, passwordVal, passwordPath string) stri
 	return password
 }
 
+// GetPostgresDSN produces the DSN (data source name) for opening a postgres db connection.
+func GetPostgresDSN(ctx context.Context, pgConfig PostgresConfig) string {
+	return getPostgresDsn(ctx, pgConfig)
+}
```
popojk: Make this function public so that the Run service repo layer will not need to double implement this string construction logic.

pingsutw: Could we make the function at line 51 public instead?

popojk: Fixed in [86bb050](https://github.com/flyteorg/flyte/pull/7018/commits/86bb050f4c73f6a8c72f4d9786e65cf4b57af743)

PR #7023, opened 2026-03-12
proto definitions for settings service

afbrock:
### Labels

## Docs link

- `main` 
  - \#6583
    - **proto definitions for settings service** :point\_left:


flyteidl2/settings/settings_definition.proto
```diff
@@ -0,0 +1,117 @@
...
+// StorageSettings controls data storage configuration.
+message StorageSettings {
+  // URI prefix for raw output data.
+  StringSetting raw_data_path = 1;
+}
+
+// TaskResourceDefaults defines default and limit resource requests for tasks.
+message TaskResourceDefaults {
+  StringSetting min_cpu = 1;
+  StringSetting min_gpu = 2;
```
wild-endeavor: did we say we were getting rid of GPU?

flyteidl2/settings/settings_definition.proto
```diff
@@ -0,0 +1,117 @@
...
+  StringSetting min_memory = 3;
+  StringSetting min_storage = 4;
+  StringSetting limits_cpu = 5;
```
wild-endeavor: limits are the max that can be specified right?  can we call it a cap or a maximum? i'm worried the word limit is confusing in the context of k8s.

Also does it make sense to support GPU here @EngHabu ?

flyteidl2/settings/settings_definition.proto
```diff
@@ -0,0 +1,117 @@
...
+// Settings holds all configurable settings for a scope.
+// Each field uses a typed setting wrapper supporting inherit/unset/value states.
+message Settings {
+  RunSettings run = 1;
+  SecuritySettings security = 2;
+  StorageSettings storage = 3;
+  TaskResourceSettings task_resource = 4;
+  StringListSetting labels = 5;
+  StringMapSetting annotations = 6;
+  StringMapSetting environment_variables = 7;
```
wild-endeavor: I think we need a property bag here as well.  i feel like that one is a generic struct @EngHabu?  Or should it be property bags?  dictionary of structs keyed off a string?

flyteidl2/settings/settings_definition.proto
```diff
@@ -0,0 +1,117 @@
...
+enum SettingScopeLevel {
+  SETTING_SCOPE_LEVEL_ORG = 0;
+  SETTING_SCOPE_LEVEL_DOMAIN = 1;
+  SETTING_SCOPE_LEVEL_PROJECT = 2;
+}
+
+// StringSetting is a string-valued setting with inheritance control.
+message StringSetting {
+  SettingState state = 1;
+  string value = 2;
```
wild-endeavor: weren't these originally going to be a oneof?  Is there a benefit to making these separate objects (StringSetting, IntSetting, etc.).  I understand you did this so that you can do
```
StringSetting raw_data_path = 1;
```
and then if someone receives a `StorageSettings` object, they'll know that it contains a string type... just wondering if this is a useful feature - in what scenario will this be useful to know above and beyond what the original oneof block would give you?

(see the SettingValue object in [the md file](https://github.com/unionai/cloud/blob/settings-protos/organizations/prompts/SETTINGS_SPEC_DETAILED.md))

flyteidl2/settings/settings_definition.proto
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// IntSetting is an integer-valued setting with inheritance control.
+message IntSetting {
+  SettingState state = 1;
```
wild-endeavor: we should add a description to each setting also.

flyteidl2/settings/settings_service.proto
```diff
@@ -0,0 +1,164 @@
...
+  // The key identifying this settings record.
+  SettingsKey key = 1;
+
+  // Resolved effective settings at the requested scope.
+  Settings settings = 2;
+
+  // Per-setting source level. Keys are dot-separated field paths
+  // (e.g. "run.default_queue", "task_resource.defaults.min_cpu").
+  // Only present for settings with state = SETTING_STATE_VALUE.
+  map<string, SettingScopeLevel> sources = 3;
```
wild-endeavor: what is this for again?  shouldn't it already be in each setting object?

flyteidl2/settings/settings_service.proto
```diff
@@ -0,0 +1,164 @@
...
+// UpdateSettingsRawRequest upserts unstructured settings at the scope implied
+// by the key. Intended for clients that are not up to date with the current
+// settings schema. Uses optimistic locking: version must match the current
+// stored version. Use version = 0 when creating a settings record for the
+// first time.
+message UpdateSettingsRawRequest {
+  SettingsKey key = 1 [(buf.validate.field).required = true];
+
+  // Serialized settings blob.
+  string data = 2 [(buf.validate.field).string.min_len = 1];
```
wild-endeavor: i don't think we should use string data here.  This is the endpoint that the SDK CLI will probably use right?  The sole purpose of this endpoint is so that changes to the settings object can be handled without forcing the user to upgrade.  But I think a plain string is too loose right?  I think we can assume that the SDK will always have access to the base setting value objects (be it `SettingValue` or the collection of `StringSetting`, `IntSetting`, etc.)... given that, shouldn't this be something like map<string, oneof stringsetting...> or map<string, SettingValue>?

Also, what was the plan for map/list types?  the original plan was to have them be additive. We don't need to do that - if we just force the user to copy/paste any values from higher scopes that actually makes the editing experience simpler.  (search for "Map-type settings should show each scope's contribution separately" in the md file to see what i mean)  Is that the plan? just make the user do it?

I know i'm referencing the original doc... don't worry about the yaml _formatting_ part in that doc please (we'll take care of the actual yaml parts), but the information does need to be in the api responses.

wild-endeavor (commented):
..

flyteidl2/settings/settings_definition.proto
EngHabu: sure..

flyteidl2/settings/settings_definition.proto
EngHabu: for "everything else" ?

flyteidl2/settings/settings_service.proto
afbrock: the reason for it to be string data is for the comments.  that may be unnecessary if we code into the sdk the comments about the level and description and which can be edited and which not, but if we wanted to add any other descriptive elements or anything like that this would allow it to be totally controlled by the back end and not the sdk.  
That said, we can change it (and update) so that there is less parsing work involved by the backend.

flyteidl2/settings/settings_service.proto
afbrock: oh, yeah, let me change it so that it is in the settings wrapper

flyteidl2/settings/settings_definition.proto
afbrock: I think there is only one way that really makes sense - and that is if we are making these consumable by customers or 3rd parties where we do not use or do not have control or simply cannot implement them in our control plane.  Adding an 'everything else' bucket basically means that the very next change someone makes to add a setting they need to answer the question 'does this get proper proto or does it go in the bag'.  and then every consumer has to ask 'did that new setting get proper proto definition, or is it in the bag, and if it is in the bag where is its name defined'.  and that mental work I really want to avoid.

flyteidl2/settings/settings_definition.proto
afbrock: so i think k8s has request/limit, I think if we want to make it match we should use that, but to me it is more intuitive to say min/max... except if we have scale down as a feature we might also start in the middle somewhere rather than at the minimum.  my gut says we should say min/max because that is pretty clear, but that also isnt coming from a k8s background

flyteidl2/settings/settings_definition.proto
afbrock: i am not that familiar with the oneofs - if someone makes a create request with a setting that is a oneof, can they put whatever type they want in there?

flyteidl2/settings/settings_definition.proto
wild-endeavor: chatted offline... let's ignore this for now and come to it when we need to work on cluster resource controller

flyteidl2/org/settings_definition.proto
```diff
@@ -0,0 +1,134 @@
...
+// The oneof encodes both control state and the typed value:
+//   - state field set to INHERIT (default) or UNSET: no associated data
+//   - string_value / int_value / list_value / map_value set: implies VALUE state
+message SettingValue {
+  oneof value {
+    // INHERIT (default) or UNSET — no data value associated.
+    SettingState state = 1;
+
+    // A plain string value.
+    string string_value = 2;
```
wild-endeavor, line 49: i think we also need boolean

flyteidl2/org/settings_service.proto
```diff
@@ -0,0 +1,176 @@
...
+//   org + domain + project → project scope
+// Settings are inherited from parent scopes where not explicitly overridden.
+message GetSettingsRequest {
+  SettingsKey key = 1 [(buf.validate.field).required = true];
+}
+
+// GetSettingsResponse contains resolved effective settings.
+// Each setting's scope_level field indicates which scope provided its value.
+message GetSettingsResponse {
+  SettingsKey key      = 1;
```
wild-endeavor, line 53: should we not just use SettingsRecord here?

flyteidl2/org/settings_service.proto
```diff
@@ -0,0 +1,176 @@
...
+// CreateSettingsRequest creates a new settings record at the scope implied
+// by the key. Fails if a record already exists at that scope.
+message CreateSettingsRequest {
+  SettingsKey key      = 1 [(buf.validate.field).required = true];
+  Settings    settings = 2 [(buf.validate.field).required = true];
+}
+
+// CreateSettingsResponse contains the created settings and initial version.
+message CreateSettingsResponse {
+  SettingsKey key      = 1;
```
wild-endeavor, line 100: Can this also be SettingsRecord

flyteidl2/org/settings_service.proto
```diff
@@ -0,0 +1,176 @@
...
+  SettingsKey key      = 1 [(buf.validate.field).required = true];
+  Settings    settings = 2 [(buf.validate.field).required = true];
+
+  // Current version of the settings record. Must match the server version.
+  uint64 version = 3;
+}
+
+// UpdateSettingsResponse contains the updated settings and new version.
+message UpdateSettingsResponse {
+  SettingsKey key      = 1;
```
wild-endeavor, line 117: ditto

flyteidl2/org/settings_service.proto
```diff
@@ -0,0 +1,176 @@
...
+  // Dot-notation field paths mapped to Setting values.
+  map<string, SettingValue> data = 2;
+
+  // Current version of the settings record. Must match the server version.
+  uint64 version = 3;
+}
+
+// UpdateSettingsRawResponse contains the stored raw settings and new version.
+message UpdateSettingsRawResponse {
+  SettingsKey key = 1;
```
wild-endeavor, line 137: i think message can just be empty.  we can attach error details

```golang
import (
    "context"

    "google.golang.org/grpc/codes"
    "google.golang.org/grpc/status"

    "google.golang.org/genproto/googleapis/rpc/errdetails"
)

func (s *Server) CreateUser(ctx context.Context, req *pb.CreateUserRequest) (*pb.CreateUserResponse, error) {
    var violations []*errdetails.BadRequest_FieldViolation

    if req.Email == "" {
        violations = append(violations, &errdetails.BadRequest_FieldViolation{
            Field:       "email",
            Description: "email is required",
        })
    }

    if req.Age < 18 {
        violations = append(violations, &errdetails.BadRequest_FieldViolation{
            Field:       "age",
            Description: "must be at least 18",
        })
    }

    if len(violations) > 0 {
        st := status.New(codes.InvalidArgument, "validation failed")

        br := &errdetails.BadRequest{
            FieldViolations: violations,
        }

        stWithDetails, err := st.WithDetails(br)
        if err != nil {
            return nil, st.Err() // fallback
        }

        return nil, stWithDetails.Err()
    }

    return &pb.CreateUserResponse{UserId: "123"}, nil
}
```

upon getting a validation error, the sdk/yaml client should call the get for edit endpoint again, and show the validation error message.

Maybe upon success we return the new version, but i think that is even not needed.

PR #7025, opened 2026-03-13
Adjust notifications IDL to support notifications per phases

iaroslav-ciupin:

1. Support different notifications per different phases. E.g. success email for SUCCESS phase, FAILED/TIMED\_OUT/ABORTED email for failed phases respectively.
2. Support 2 notifications type per same set of phases. E.g. email and Slack for any phase in FAILED/TIMED\_OUT/ABORTED

### Labels

## Docs link

- `main` 
  - \#6583
    - \#7025 :point\_left:


flyteidl2/notification/definition.proto
```diff
@@ -35,39 +35,29 @@ message RuleId {
...
 }
 
 message DeliveryConfigTemplate {
-  oneof config {
-    option (buf.validate.oneof).required = true;
+  option (buf.validate.message).cel = {
+    id: "at_least_one_required"
+    message: "at least one of the delivery options must be set"
+    expression: "has(this.webhook) || has(this.email)"
+  };
```
EngHabu, line 16: Why this vs oneof?

flyteidl2/task/run.proto
```diff
@@ -105,24 +106,31 @@ message RunSpec {
...
 message InlineRule {
-  repeated DeliveryOption delivery_options = 1 [(buf.validate.field).repeated.min_items = 1];
-
-  InlineRuleChecks checks = 2;
+  repeated InlineRuleMatcher matchers = 1 [(buf.validate.field).repeated.min_items = 1];
 }
 
-message DeliveryOption {
-  oneof option {
+message InlineRuleMatcher {
```
EngHabu: I would call this `InlineRule` and the other `InlineRuleList`

kumare3: +1

flyteidl2/notification/definition.proto
```diff
@@ -35,39 +35,29 @@ message RuleId {
 }
 
 message DeliveryConfigId {
-  // Org this rule belongs to.
+  // Org this config belongs to.
   string org = 1 [
```
kumare3: why keep org? We should just make all of this at the rule level no?

flyteidl2/notification/definition.proto
iaroslav-ciupin: Because in oneof you must have at most one. Here you can have as many notifications as you want. for same phase match, we may want both email & webhook & Slack for example.

flyteidl2/notification/definition.proto
iaroslav-ciupin: I don't follow, can you elaborate?
this is a separate entity that lives in its own table, in my understanding org is needed:
- for authz, to ensure you are referring to delivery config set up by your org, and not someone else's
- to avoid name collision in DB, so that you don't accidentally update/delete someone else's config.

kumare3: When you register a notification the runspec or the trigger already has the org. Why would you ever lookup some other org. We should drop it and infer from the context

iaroslav-ciupin: @kumare3  you are mixing one-off settings and persisted settings. 
One-off settings are stored in runspec via CreateRun or DeployTask(with triggers) endpoints.
Persisted settings are stored independently of any tasks, runs, triggers, etc.  Its a separate set of endpoints to create a Rule and a DeliveryConfig(see commit where I removed this API https://github.com/flyteorg/flyte/pull/6946/changes/75c97debe4f059401ff3a9d1aa02b53cdfa651cc )

EngHabu: I think what Ketan is probably saying is that in the RunSpec, instead of embedding this id, you just need the "string name" of the DeliveryConfig...

iaroslav-ciupin: ah, I thought he was referring to removing org from DeliveryConfigId object. 
Yeah, we can infer org from a run/trigger when creating a one-off notification setting. 👍

iaroslav-ciupin:
implemented this IDL on backend https://github.com/unionai/cloud/pull/14878

flyteidl2/task/run.proto
```diff
@@ -98,31 +99,39 @@ message RunSpec {
 
   // Optionally, you can send a notification when your run completes.
   oneof notification_settings {
-    flyteidl2.notification.RuleId rule_id = 10; // either refer to a previously stored notification rule
+    string notification_rule_name = 10; // either refer to a previously stored notification rule
```
kumare3, line 102: Nice

PR #7031, opened 2026-03-14
feat(runs): wire AbortRun and AbortAction through to ActionsService 

AdilFayyaz:

AbortRun and AbortAction previously only updated the run phase in the database, leaving the TaskAction CR and its pod still running in Kubernetes. This connects the abort path to ActionsService.Abort so execution is actually stopped.

- AbortRun now calls actionsClient.Abort() with the root action ID after the DB update, triggering K8s-side CR deletion and cascade to all descendants
- AbortAction similarly calls actionsClient.Abort() for the specific action after the DB update

## How was this patch tested?

- Start the manager (make -C manager run) against a local sandbox
- Submit a run via uv run flyte run ... and confirm pods appear in \[namespace]
- Run kubectl -n \[namespace] get taskactions to confirm the CR exists
- Abort with uv run flyte abort run <run-name>
- Verify TaskAction CR is deleted: kubectl -n \[namespace] get taskactions
- Verify run phase is ABORTED: uv run flyte get run
- Verify manager logs show "Aborted run..." and "Deleted TaskAction ... (descendants will be cascade deleted by K8s)"
- Added tests to run\_service\_test.go

## Docs link

- `main` 
  - \#6583
    - \#7031 :point\_left:

### DIFF
--- a/runs/service/run_service_test.go
+++ b/runs/service/run_service_test.go
@@ -92,7 +92,7 @@ func TestAbortRun(t *testing.T) {
 		actionRepo.On("AbortRun", mock.Anything, runID, "User requested abort", (*common.EnrichedIdentity)(nil)).Return(nil)
 		actionsClient.On("Abort", mock.Anything, mock.MatchedBy(func(req *connect.Request[actions.AbortRequest]) bool {
 			return req.Msg.ActionId.Run.Name == runID.Name &&
-				req.Msg.ActionId.Name == runID.Name &&
+				req.Msg.ActionId.Name == "a0" &&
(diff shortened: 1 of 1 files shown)


runs/service/run_service.go
```diff
@@ -246,6 +246,19 @@ func (s *RunService) AbortRun(
 		return nil, connect.NewError(connect.CodeInternal, err)
 	}
 
+	// Abort via actions service
+	rootActionID := &common.ActionIdentifier{
+		Run:  req.Msg.RunId,
+		Name: req.Msg.RunId.Name,
```
EngHabu: Isn't this 'a0'?

pingsutw (approved):

PR #7034, opened 2026-03-16
[V2]Sandbox webhook

### LINKED ISSUE #7003: [V2][sandbox] Sandbox support webhook

Sandbox should support webhook

n/a

n/a

_No response_

- [x] Yes

- [x] Yes

popojk:
## Tracking issue

Closes #7003

To handle secret injection from k8s cluster secrets to flyte task pods, we implement a webhook feature in this PR. Whenever a pod create request sent to the k8s api interface,  k8s will redirect the request to our webhook server. The webhook server will mutate the pod spec by adding k8s secrets to it, and the webhook will finally create pods with those secret ENVVAR. Please see bellow diagram for better understanding.

```
┌─────────────────────────────────────────────────────────────────────────┐
  │  Client (flytekit SDK)                                                  │
  │                                                                         │
  │  @env.task(secrets=Secret(group="my-secret", key="api-key"))            │
  │       │                                                                 │
  │       │  TaskTemplate.SecurityContext.Secrets = [{group, key}]          │
  └───────┼─────────────────────────────────────────────────────────────────┘
          │ flyte.run()
          ▼
  ┌───────────────────┐
  │   Flyte API       │  creates
  │   Server          │ ──────────────► TaskAction CRD
  └───────────────────┘                 (Spec.TaskTemplate = protobuf bytes)
                                                │
                                                │ controller watches
                                                ▼
  ┌─────────────────────────────────────────────────────────────────────────┐
  │  Executor (controller-runtime)                                          │
  │                                                                         │
  │  TaskActionReconciler.Reconcile()                                       │
  │       │                                                                 │
  │       ├─ NewTaskExecutionMetadata(ta)                                   │
  │       │    └─ extractSecurityContextFromTaskTemplate()                  │
  │       │         └─ SecurityContext.Secrets = [{group, key}]             │
  │       │                                                                 │
  │       └─ plugin.BuildResource(taskCtx)                  [pod plugin]   │
  │            ├─ flytek8s.BuildRawPod()                                    │
  │            └─ if len(SecurityContext.Secrets) > 0:                      │
  │                 pod.Labels["inject-flyte-secrets"] = "true"             │
  │                 pod.Annotations["flyte.secrets/s0"] = <encoded proto>   │
  └─────────────────────────────────┬───────────────────────────────────────┘
                                    │ client.Create(pod)
                                    ▼
  ┌─────────────────────────────────────────────────────────────────────────┐
  │  Kubernetes API Server                                                  │
  │                                                                         │
  │  MutatingWebhookConfiguration                                           │
  │    objectSelector: inject-flyte-secrets=true   ◄── only tagged pods    │
  │    clientConfig.service: flyte-pod-webhook                              │
  └─────────────────────────────────┬───────────────────────────────────────┘
                                    │ forwards pod create request
                                    ▼
  ┌─────────────────────────────────────────────────────────────────────────┐
  │  Webhook Server (same process as Executor, port 9443)                  │
  │                                                                         │
  │  PodMutator.Handle()                                                    │
  │       │                                                                 │
  │       └─ SecretsPodMutator.Mutate()                                     │
  │            ├─ UnmarshalStringMapToSecrets(pod.Annotations)              │
  │            │    └─ [{group:"my-secret", key:"api-key", ENV_VAR}]        │
  │            │                                                            │
  │            └─ K8sSecretInjector.Inject()                               │
  │                 └─ add env var to pod spec:                             │
  │                      name:  _UNION_MY-SECRET_API-KEY                   │
  │                      valueFrom.secretKeyRef:                            │
  │                        name: my-secret   (k8s Secret)                  │
  │                        key:  api-key                                    │
  └─────────────────────────────────┬───────────────────────────────────────┘
                                    │ mutated pod
                                    ▼
  ┌─────────────────────────────────────────────────────────────────────────┐
  │  Task Pod                                                               │
  │                                                                         │
  │  env:                                                                   │
  │    _UNION_MY-SECRET_API-KEY=<value from k8s Secret>                    │
  │    FLYTE_SECRETS_ENV_PREFIX=_UNION_                                     │
  └─────────────────────────────────────────────────────────────────────────┘
```

- Add a webhook server in controller runtime to handle pod mutation
- Add a service in helm chart in sandbox to expose webhook server
- Add logic in taskExecutionMetadata to consume and handle security context in task template

## Follow up

We should add an api endpoint in the backend for the user to add k8s secret with flyte cli i/o calling kubectl command

## How was this patch tested?

After sandbox up and running, we set a test secret in k8s cluster first:

```
kubectl create secret generic webhooktest \
--from-literal=apikey=hello-from-webhook \
-n flyte
```

Run bellow flyte workflow which has. secret config in it's task ENV:

```
import asyncio
import logging
import os
from typing import List

import flyte

env = flyte.TaskEnvironment(
    name="hello_world",
    secrets=flyte.Secret(group="webhooktest", key="apikey"),
)


@env.task
async def say_hello(data: str, lt: List[int]) -> str:
    print(f"Hello, world! - {flyte.ctx().action}")
    secret_val = os.environ.get("_UNION_WEBHOOKTEST_APIKEY", "NOT_FOUND")
    print(f"[webhook test] _UNION_WEBHOOKTEST_APIKEY={secret_val}")
    return f"Hello {data} {lt} secret={secret_val}"


@env.task
async def square(i: int = 3) -> int:
    print(flyte.ctx().action)
    return i * i


@env.task
async def say_hello_nested(data: str = "default string", n: int = 3) -> str:
    print(f"Hello, nested! - {flyte.ctx().action}")
    coros = []
    for i in range(n):
        coros.append(square(i=i))

    vals = await asyncio.gather(*coros)
    return await say_hello(data=data, lt=vals)


if __name__ == "__main__":
    flyte.init_from_config()
    run = flyte.with_runcontext(
        log_level=logging.DEBUG,
        env_vars={"KEY": "V"},
        labels={"Label1": "V1"},
        annotations={"Ann": "ann"},
        overwrite_cache=True,
        interruptible=False,
    ).run(say_hello_nested, data="hello world", n=10)
    print(run.name)
    print(run.url)
    run.wait()
    # print(run.outputs())
```

We will see secret ENV injected by webhook in task pod: Screenshot 2026-03-16 at 10 36 39 AM

### Labels

## Docs link

- `main` 
  - \#6583
    - **\[V2]Sandbox webhook** :point\_left:

### DIFF
--- a/executor/pkg/webhook/init_cert.go
+++ b/executor/pkg/webhook/init_cert.go
@@ -68,10 +68,8 @@ func createWebhookSecret(ctx context.Context, namespace string, cfg *webhookConf
 	// In full deployment the webhook should be running in a single pod and an init container will generate and inject the secret data
 	if cfg.LocalCert {
 		certPath := cfg.ExpandCertDir()
-		if _, err := os.Stat(certPath); os.IsNotExist(err) {
-			if err := os.Mkdir(certPath, folderPerm); err != nil {
-				return err
-			}
+		if err := os.MkdirAll(certPath, folderPerm); err != nil {
+			return err
 		}
 
 		if err := os.WriteFile(path.Join(certPath, CaCertKey), certs.CaPEM.Bytes(), permission); err != nil {
@@ -102,37 +100,14 @@ func createWebhookSecret(ctx context.Context, namespace string, cfg *webhookConf
 	}
 
 	if kubeErrors.IsAlreadyExists(err) {
-		logger.Infof(ctx, "A secret already exists with the same name. Validating.")
-		s, err := secretsClient.Get(ctx, cfg.SecretName, metav1.GetOptions{})
-		if err != nil {
+		logger.Infof(ctx, "Secret [%v] already exists, recreating with new certs.", cfg.SecretName)
+		if err := secretsClient.Delete(ctx, cfg.SecretName, metav1.DeleteOptions{}); err != nil {
 			return err
 		}
-
-		requiresUpdate := false
-		for key := range secretData {
-			if key == CaCertKey {
-				continue
-			}
-			if _, exists := s.Data[key]; !exists {
-				requiresUpdate = true
-				break
-			}
-		}
-
-		if requiresUpdate {
-			logger.Infof(ctx, "The existing secret is missing one or more keys.")
-			secret.Annotations = map[string]string{
-				"flyteLastUpdate": "system-updated",
-				"flyteUpdatedAt":  time.Now().String(),
-			}
-			_, err = secretsClient.Update(ctx, secret, metav1.UpdateOptions{})
-			if err != nil && kubeErrors.IsConflict(err) {
-				logger.Infof(ctx, "Another instance updated the same secret. Ignoring.")
-				err = nil
-			}
+		if _, err := secretsClient.Create(ctx, secret, metav1.CreateOptions{}); err != nil {
 			return err
 		}
-
+		logger.Infof(ctx, "Recreated secret [%v]", cfg.SecretName)
 		return nil
 	}
 
--- a/docker/sandbox-bundled/manifests/complete.yaml
+++ b/docker/sandbox-bundled/manifests/complete.yaml
@@ -1031,7 +1031,7 @@ type: Opaque
 ---
 apiVersion: v1
 data:
-  haSharedSecret: VWs4TWZpNFZiVHdGU0hSSA==
+  haSharedSecret: Rkk2Y1ZETzk4SnV2V3JZUA==
   proxyPassword: ""
   proxyUsername: ""
 kind: Secret
@@ -1557,7 +1557,7 @@ spec:
     metadata:
       annotations:
         checksum/config: 8f50e768255a87f078ba8b9879a0c174c3e045ffb46ac8723d2eedbe293c8d81
-        checksum/secret: 1b0294724750b2aad0728683df4cf90644c040758b3d48200366fead95ff5271
+        checksum/secret: 89994f7cdb81c298a0f3073a45e5fd8afd4cf154b99d3983ed597946794e2d2a
       labels:
         app: docker-registry
         release: flyte-sandbox
--- a/docker/sandbox-bundled/manifests/dev.yaml
+++ b/docker/sandbox-bundled/manifests/dev.yaml
@@ -753,7 +753,7 @@ metadata:
 ---
 apiVersion: v1
 data:
-  haSharedSecret: eDY0aTZaZzRXMW1wQ2E0aw==
+  haSharedSecret: U0VJZnFNeGFTWE9peDFlMw==
   proxyPassword: ""
   proxyUsername: ""
 kind: Secret
@@ -1176,7 +1176,7 @@ spec:
     metadata:
       annotations:
         checksum/config: 8f50e768255a87f078ba8b9879a0c174c3e045ffb46ac8723d2eedbe293c8d81
-        checksum/secret: 01c0a9c532acf51c227b5a8dc364b56b4464126872515c19a63678ff6406aa2a
+        checksum/secret: 770936f6bf0e6bc5cd48fc176eee8831648fb3c4204aa706e08f88ad9b96441b
       labels:
         app: docker-registry
         release: flyte-sandbox


flyteplugins/go/tasks/plugins/k8s/pod/plugin.go
```diff
@@ -168,15 +169,28 @@ func (p plugin) BuildResource(ctx context.Context, taskCtx pluginsCore.TaskExecu
...
+			pod.Annotations = make(map[string]string)
+		}
+		for k, v := range secretAnnotations {
+			pod.Annotations[k] = v
+		}
+		if pod.Labels == nil {
+			pod.Labels = make(map[string]string)
+		}
+		pod.Labels[secretUtils.PodLabel] = secretUtils.PodLabelValue
+	}
```
popojk: Right now our k8s plugin did not implement secret context handling logic, so this piece of code did it.

executor/pkg/webhook/pod.go
```diff
@@ -0,0 +1,153 @@
...
+	mutateConfig := &admissionregistrationv1.MutatingWebhookConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      pm.cfg.ServiceName,
+			Namespace: namespace,
+		},
+		Webhooks: []admissionregistrationv1.MutatingWebhook{
+			{
+				Name: webhookName,
+				ClientConfig: admissionregistrationv1.WebhookClientConfig{
+					CABundle: caBytes,
```
popojk, line 107: We need to register the webhook configuration with our self signed CA so that the k8s API can communicate with our webhook under https protocol

popojk:
Fix secret annotation implementation in [26fdf4b](https://github.com/flyteorg/flyte/pull/7034/commits/26fdf4bbc0daf2e63d2485c65a81ddf3a8e75578)

executor/pkg/webhook/init_cert.go
```diff
@@ -0,0 +1,210 @@
...
+
+func createWebhookSecret(ctx context.Context, namespace string, cfg *webhookConfig.Config, certs webhookCerts, secretsClient v1.SecretInterface) error {
+	isImmutable := true
+	secretData := map[string][]byte{
+		CaCertKey:            certs.CaPEM.Bytes(),
+		ServerCertKey:        certs.ServerPEM.Bytes(),
+		ServerCertPrivateKey: certs.PrivateKeyPEM.Bytes(),
+	}
+
+	if cfg.LocalCert {
```
machichima, line 69: If `LocalCert` is `false`, I think the webhook cannot get any certificate files? In this case places use `os.ReadFile()` will fail

machichima:
Is it possible for us to create the Secret before the pod starts, and mount the secret to pod volume? Maybe we can use cert-manager, or other ways to achieve this?

executor/pkg/webhook/init_cert.go
```diff
@@ -0,0 +1,210 @@
...
+		Immutable: &isImmutable,
+	}
+
+	_, err := secretsClient.Create(ctx, secret, metav1.CreateOptions{})
+	if err == nil {
+		logger.Infof(ctx, "Created secret [%v]", cfg.SecretName)
+		return nil
+	}
+
+	if kubeErrors.IsAlreadyExists(err) {
```
machichima, line 102: Could you please explain the secret update logic here? Based on the logic now, we do:
1. Get the old secret back
2. Check if `tls.crt` and `tls.key` exists in the old secret
3. if any of them do not exists, update the secret

This can result in stale secret (certificate files get updated locally, but Secret object have the old ones)

I got questions below:
1. Why do we skip the `ca.crt` check?
2. Should we also check if the old and new secret file are the same? Or can we directly update to new secret?

popojk: 1. ca.crt will be hold by client side(k8s cluster), whereas the server side(webhook) will only need tls.crt and tls.key.
2. The update will fail if we have conflicting secret. Please see [here](https://github.com/flyteorg/flyte/blob/3bc13ad7d44375e6f5b7c766cc312e275acd686a/executor/pkg/webhook/init_cert.go#L127)

popojk:
> Is it possible for us to create the Secret before the pod starts, and mount the secret to pod volume? Maybe we can use cert-manager, or other ways to achieve this?

@machichima Add a todo in [d5e66ee](https://github.com/flyteorg/flyte/pull/7034/commits/d5e66eeb4077981a86362c68375d7c38d60ab6fd). In future flyte full deployment, we should have a webhook pod with a init pod to generate and inject certs.

executor/pkg/webhook/init_cert.go
machichima: 1. Even if the webhook server itself only needs `tls.crt` and `tls.key` to serve TLS, `ca.crt` is still part of the generated cert set and is later used to populate `MutatingWebhookConfiguration.clientConfig.caBundle` for k8s api server to use. Skipping `ca.crt` here makes the Secret more likely to differ from the locally generated cert state. Was that intentional?

2. `Conflict` here refers to an optimistic concurrency failure on `Update`, i.e. the Secret was modified after our `Get` and before our `Update`. That is not directly related to whether the Secret already contains certificate files.
https://kubernetes.io/docs/reference/using-api/api-concepts/#patch-and-apply

popojk: @machichima On second thought, i think it is more reasonable to update full set of secret every time webhook start and running. Fixed in [6faacf3](https://github.com/flyteorg/flyte/pull/7034/commits/6faacf3dbe25c1dad8c27549842fe6e98dac6cd8) cc @pingsutw

PR #7035, opened 2026-03-16
[V2] Add Cache service

machichima:

Add cache service

Add cache service and reservation to:

- Directly fetch output from object store URI if cache enabled
- Add reservation to prevent two requests trying to write to cache at the same time

## How was this patch tested?

- Run `flyte-sdk` script below twice to trigger cache

```py
import random

import flyte
from flyte import Cache

env = flyte.TaskEnvironment(name="use_cache_simple")


@env.task(cache=Cache(behavior="override", version_override="v1"))
def main(x_list: list[int]) -> str:
    token = random.randint(1, 1_000_000)
    print(f"cache miss: generated token={token}")
    return f"sum={sum(x_list)}, token={token}"


if __name__ == "__main__":
    flyte.init_from_config()  # establish remote connection from within your script.
    run = flyte.run(main, x_list=list(range(10)))  # run remotely inline and pass data.

    print(run.name)
    print(run.url)

    run.wait()
```

```sh
❯ python examples/use_cache_simple.py

  > Building 1 image...
    > Building image flyte for environment use_cache_simple
    i Image ghcr.io/flyteorg/flyte:16b2b307f1864d4a76e0224b38da6444 already exists, skipping build
  ✓ Built image for environment use_cache_simple: ghcr.io/flyteorg/flyte:16b2b307f1864d4a76e0224b38da6444
  > Bundling code...
  ✓ Code bundle: 1 files, 0.009765625 MB (compressed 0.0006875991821289062 MB)
  > Uploading code bundle...
rgk8t2gb7ksw6cpq8llp
http://localhost:8080/v2/domain/development/project/testproject/runs/rgk8t2gb7ksw6cpq8llp
Run 'rgk8t2gb7ksw6cpq8llp' completed successfully.

~/workData/open-source/flyte-sdk main *22 !1 ?4                                                       13s  flyte-sdk
❯ python examples/use_cache_simple.py

  > Building 1 image...
    > Building image flyte for environment use_cache_simple
    i Image ghcr.io/flyteorg/flyte:16b2b307f1864d4a76e0224b38da6444 already exists, skipping build
  ✓ Built image for environment use_cache_simple: ghcr.io/flyteorg/flyte:16b2b307f1864d4a76e0224b38da6444
  > Bundling code...
  ✓ Code bundle: 1 files, 0.009765625 MB (compressed 0.0006875991821289062 MB)
  > Uploading code bundle...
rm42shl58npfxzccmjts
http://localhost:8080/v2/domain/development/project/testproject/runs/rm42shl58npfxzccmjts
Run 'rm42shl58npfxzccmjts' completed successfully.
```

There's only one pod created for the first execution `rgk8t2gb7ksw6cpq8llp`:

```sh
❯ k get pods -n testproject-development
NAME                        READY   STATUS      RESTARTS   AGE
rgk8t2gb7ksw6cpq8llp-a0-0   0/1     Completed   0          108s
```

### Labels

## Docs link

- `main` 
  - \#6583
    - **\[V2] Add Cache service** :point\_left:


cache_service/cmd/main.go
```diff
@@ -0,0 +1,47 @@
...
+			sc.Port = cfg.Server.Port
+
+			db, err := app.InitDB(ctx, database.GetConfig())
+			if err != nil {
+				return fmt.Errorf("failed to initialize database: %w", err)
+			}
+			sc.DB = db
+
+			labeled.SetMetricKeys(contextutils.ProjectKey, contextutils.DomainKey, contextutils.WorkflowIDKey, contextutils.TaskIDKey)
+			dataStore, err := storage.NewDataStore(storage.GetConfig(), promutils.NewTestScope())
```
pingsutw: Should use `promutils.NewScope("cache-service")`

actions/k8s/client.go
```diff
@@ -580,6 +582,23 @@ func buildActionSpec(action *actions.Action, runSpec *task.RunSpec) *workflow.Ac
 	return actionSpec
 }
 
+func extractTaskCacheKey(action *actions.Action) string {
+	taskSpec, ok := action.Spec.(*actions.Action_Task)
+	if !ok || taskSpec.Task == nil {
+		return ""
+	}
+
+	return getStringValue(taskSpec.Task.CacheKey)
```
pingsutw: Nit: instead calling `getStringValue`, could we do nil check in this function?

cache_service/cmd/main.go
pingsutw: We probably need `RunIDKey` and `ActionIDKey` instead. No need to do it now

executor/pkg/controller/taskaction_cache.go
```diff
@@ -0,0 +1,206 @@
...
+		info := cacheTaskInfo(corepb.CatalogCacheStatus_CACHE_HIT, "cache hit")
+		return pluginsCore.DoTransition(pluginsCore.PhaseInfoSuccess(info)), true, nil
+	}
+
+	if !catalog.IsNotFound(err) {
+		log.FromContext(ctx).Error(err, "cache lookup failed, continuing with task execution")
+		return pluginsCore.UnknownTransition, false, nil
+	}
+
+	if !cacheCfg.serializable {
```
pingsutw: If cacheCfg.serializable is False, we directly get the output from the cache
If cacheCfg.serializable is True, we should do GetOrExtendReservation

executor/pkg/controller/taskaction_cache.go
```diff
@@ -0,0 +1,206 @@
...
+	tCtx pluginsCore.TaskExecutionContext,
+	transition pluginsCore.Transition,
+	handledByCache bool,
+) (pluginsCore.Transition, error) {
+	cacheCfg, ok, err := buildTaskCacheConfig(ctx, taskAction, tCtx)
+	if err != nil || !ok || r.Catalog == nil {
+		return transition, err
+	}
+
+	if handledByCache {
```
pingsutw: I think we don't need handledByCache. If cache is disabled, buildTaskCacheConfig will always return not ok

machichima: `buildTaskCacheConfig` only tells us whether caching is configured for the task. `handledByCache` is still needed to distinguish if the output is:
1. get from the cache (`handledByCache = True`)
2. get from execution

If is case 1, we do not need to write result to cache

cache_service/cmd/main.go
machichima: Done in https://github.com/flyteorg/flyte/pull/7035/commits/20844863afdffe6a2dc6f40a858449add91fa4a4

actions/k8s/client.go
machichima: Done in https://github.com/flyteorg/flyte/pull/7035/commits/2dbf6cfc70b2fb75bda95c666bfb4cae989c621d

executor/pkg/controller/taskaction_cache.go
machichima: Fixed in https://github.com/flyteorg/flyte/pull/7035/commits/372b340abb072261c7e0229ed51905b0cbf21bbc

cache_service/manager/manager.go
```diff
@@ -0,0 +1,287 @@
...
+	Metadata  *cacheservicepb.Metadata
+}
+
+func New(cfg *cacheconfig.Config, outputs interfaces.CachedOutputRepo, reservations interfaces.ReservationRepo) *Manager {
+	maxHeartbeat := cfg.MaxReservationHeartbeat.Duration
+	if maxHeartbeat <= 0 {
+		maxHeartbeat = 10 * time.Second
+	}
+
+	graceMultiplier := time.Duration(cfg.HeartbeatGracePeriodMultiplier)
```
popojk: Although the business logic is correct, I'm wondering why do we set the type time.Duration for graceMultiplier? time.Duration by definition represents nanosecond, and I think it makes more sense to use int as type directly?

cache_service/manager/manager.go
```diff
@@ -0,0 +1,287 @@
...
+		reservations:                    reservations,
+		heartbeatGracePeriodMultiplier:  graceMultiplier,
+		maxReservationHeartbeatInterval: maxHeartbeat,
+	}
+}
+
+// Get looks up an already-materialized cache entry.
+//
+// This answers the question: "Does a usable cached output already exist for
+// this key?" If yes, the caller can short-circuit execution and reuse it.
```
popojk: Should we keep this in comment?

cache_service/manager/manager.go
```diff
@@ -0,0 +1,287 @@
...
+func mergeMetadata(existing *models.CachedOutput, request *cacheservicepb.Metadata, now time.Time) *cacheservicepb.Metadata {
+	var metadata *cacheservicepb.Metadata
+	if request != nil {
+		metadata = proto.Clone(request).(*cacheservicepb.Metadata)
+	} else {
+		metadata = &cacheservicepb.Metadata{}
+	}
+
+	if existing != nil && len(existing.Metadata) > 0 {
+		if existingMetadata, err := unmarshalMetadata(existing.Metadata); err == nil && metadata.GetCreatedAt() == nil {
```
popojk, line 252: If err occurred here due to unmarshall fail, the metadata will be tagged with `metadata.CreatedAt = timestamppb.New(now)` and persisted into db. However, should we actually return an error back to client side as the metadata provided is not valid?

cache_service/manager/manager.go
```diff
@@ -0,0 +1,287 @@
...
+		if repository.IsAlreadyExists(err) {
+			// Another caller created the reservation after our initial read.
+			current, getErr := m.reservations.Get(ctx, reservationKey)
+			if getErr != nil {
+				return nil, connect.NewError(connect.CodeInternal, getErr)
+			}
+			return reservationFromModel(current), nil
+		}
+		return nil, connect.NewError(connect.CodeInternal, err)
+	}
```
popojk: nit: There is a potential race condition here. When an existing reservation is expired and two callers race to claim it, the loser receives a stale pre-read snapshot instead of the current DB state. We can fix it in future PR by wrapping the db operations into a transaction, your call ;)

cache_service/service/service.go
```diff
@@ -0,0 +1,169 @@
...
+	if base == nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("base_request is required"))
+	}
+	if err := validateBaseKey(base.GetKey()); err != nil {
+		return nil, err
+	}
+	if err := validateOwnerID(base.GetOwnerId()); err != nil {
+		return nil, err
+	}
+	copy := proto.Clone(base).(*cacheservicepb.GetOrExtendReservationRequest)
```
popojk: Curious why do we need to deep copy the request here? The request will not be used anywhere else, do we need to worry about the in memory object being over written?

popojk (commented):
Overall looks good, left a few comments.

cache_service/manager/manager.go
machichima: I add this here to pair with the docstring for `GetOrExtendReservation` showing their difference and why we need two of them. Let me refactor it to make it more clear

https://github.com/flyteorg/flyte/blob/27bf2b50554d1e4a3e136ee79b5926e2c4decbc9/cache_service/manager/manager.go#L156-L167

cache_service/manager/manager.go
machichima: Ah yes, int if more appropriate here. I did this because we can directly do `now.Add(heartbeat * m.heartbeatGracePeriodMultiplier)` below.

Let me fix this

cache_service/manager/manager.go
machichima: The unmarshal here is against `existing.Metadata`, i.e. previously persisted data from storage, not the current request payload. If this fails, it points to corrupted/incompatible stored metadata rather than a bad client request.

Here I skip the `err != nil` case so that we can directly overwrite with fresh metadata.

WDYT?

cache_service/service/service.go
machichima: We will modify the `Key` in the request below:

```go
copy.Key = scopedKey(base.GetKey(), req.Msg.GetIdentifier())
```

I wanted to avoid mutating the incoming transport request in place, so the original request remains unchanged within the handler.

But I agree that `proto.Clone` may be too heavy here, we can construct a new request and modify the key for the new struct

cache_service/manager/manager.go
machichima: Updated in https://github.com/flyteorg/flyte/pull/7035/commits/59d5a6800f7c08d753882aac1bcc935febecd00d

cache_service/manager/manager.go
machichima: Updated in https://github.com/flyteorg/flyte/pull/7035/commits/bfc8720951ea2efaf25f4985a1442ec1548d9ef7

cache_service/manager/manager.go
machichima: I solve it by get again before returning if we didn't get the reservation. Updated in https://github.com/flyteorg/flyte/pull/7035/commits/18a4f29b2b1c766a6c712503fc5e00971bc68fca

cache_service/service/service.go
machichima: Updated in https://github.com/flyteorg/flyte/pull/7035/commits/14e1059c209125aa1e2c23a12ea0680ab179db57

cache_service/manager/manager.go
popojk: ok make sense

popojk (approved):
LGTM, thanks!

PR #7041, opened 2026-03-18
fix(propeller): treat K8s BadRequest/Invalid as permanent failure instead of retrying

### LINKED ISSUE #6531: [BUG] Flyte Propeller treats a BadRequest code from K8s when launching a resource as a retryable failure
### Flyte & Flytekit version

1.15

### Describe the bug

Currently, the plugin manager code has a line commented:
https://github.com/flyteorg/flyte/blob/61b5b4aec1b03c9c9604de8e78c4166ee415ccdc/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go#L257-L259

which results in any BadRequest status codes from K8s to be treated as a retryable failure, since that is the base case: 
https://github.com/flyteorg/flyte/blob/61b5b4aec1b03c9c9604de8e78c4166ee415ccdc/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go#L264-L266

This is particularly problematic for our use-case where we have a validating admission webhook to reject any invalid pods. By default webhook rejections are BadRequests, which do make sense for validations, but this should not be treated as retryable since there is no chance for success, and we don't want the workflow to end up in an aborted state. 

Our mitigation has been to make the validating webhook send an EntityTooLarge code, which seems to be the only case treated as non retryable:
https://github.com/flyteorg/flyte/blob/61b5b4aec1b03c9c9604de8e78c4166ee415ccdc/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go#L260-L262

I think just uncommenting L259 would fix this bug, but since the line exists and has been explicitly commented out, I'm opening this ticket to get some insight about the motivation, and if it's safe to uncomment.

### Expected behavior

Ideally, BadRequest codes should be treated as non retryable.

_No response_

### Screenshots

_No response_

- [x] Yes

- [x] Yes

themavik:
## Summary

- Uncomment the `PhaseInfoFailure` return for `BadRequest`/`Invalid` K8s errors in `plugin_manager.go` so they are treated as permanent failures instead of being retried indefinitely

## Problem

When a Kubernetes API server returns a `BadRequest` (400) or `Invalid` (422) error during resource creation (e.g., an admission webhook rejection), FlytePropeller logs the error but falls through to the generic system error handler. This causes the task to be retried indefinitely, wasting resources and delaying user feedback.

The return statement on line 264 was commented out:
```go
} else if k8serrors.IsBadRequest(err) || k8serrors.IsInvalid(err) {
    logger.Errorf(ctx, "Badly formatted resource for plugin [%s], err %s", e.id, err)
    // return pluginsCore.DoTransition(pluginsCore.PhaseInfoFailure("BadTaskFormat", err.Error(), nil)), nil
}
```

Without the return, execution falls through to line 271 which wraps the error as a generic system error and returns `UnknownTransition`, causing infinite retries.

## Root Cause

The sibling code in `flyteplugins/go/tasks/plugins/array/k8s/subtask.go` handles the same case correctly (uncommented), confirming this was an oversight.

## Fix

Uncomment the return statement so `BadRequest`/`Invalid` errors immediately transition to `PhasePermanentFailure` with a clear `"BadTaskFormat"` reason. Added a unit test (`jobBadRequest`) that verifies the fix.

## Testing

- Added `TestK8sTaskExecutor_Handle_LaunchResource/jobBadRequest` test that creates a fake client returning `k8serrors.NewBadRequest(...)` and verifies the transition phase is `PhasePermanentFailure`
- Existing tests remain unchanged

Closes #6531

### DIFF
--- a/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go
+++ b/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go
@@ -261,7 +261,7 @@ func (e *PluginManager) launchResource(ctx context.Context, tCtx pluginsCore.Tas
 			return pluginsCore.DoTransition(pluginsCore.PhaseInfoRetryableFailure("RuntimeFailure", err.Error(), nil)), nil
 		} else if k8serrors.IsBadRequest(err) || k8serrors.IsInvalid(err) {
 			logger.Errorf(ctx, "Badly formatted resource for plugin [%s], err %s", e.id, err)
-			// return pluginsCore.DoTransition(pluginsCore.PhaseInfoFailure("BadTaskFormat", err.Error(), nil)), nil
+			return pluginsCore.DoTransition(pluginsCore.PhaseInfoFailure("BadTaskFormat", err.Error(), nil)), nil
 		} else if k8serrors.IsRequestEntityTooLargeError(err) {
 			logger.Errorf(ctx, "Badly formatted resource for plugin [%s], err %s", e.id, err)
 			return pluginsCore.DoTransition(pluginsCore.PhaseInfoFailure("EntityTooLarge", err.Error(), nil)), nil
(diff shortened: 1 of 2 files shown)


themavik:
Any updates on this?

PR #7044, opened 2026-03-18
fix(plugins): classify CreateContainerError as system-retryable instead of user error

### LINKED ISSUE #6786: [BUG] CreateContainerError being misclassified as UserError and not honoring retries
### Flyte & Flytekit version

**flyte version:** v1.15.3
**flytekit version:** v1.16.5

### Describe the bug

When a pod fails to start due to a `ContainersNotReady|CreateContainerError` (particularly "failed to reserve container name"), and the error message includes "Grace period exceeded", the error is incorrectly classified as kind: USER instead of kind: SYSTEM.

This causes a gap where neither retry mechanism triggers:

- System retries don't apply because the error is classified as USER
- User retries don't apply because the container never started and the task code never ran

The task fails immediately with no retry attempts, even though retries=5 is configured on the task.

### Expected behavior

`ContainersNotReady|CreateContainerError`/grace period exceeded should be classified as a SYSTEM error, which would allow max-node-retries-system-failures to automatically retry the task.
Alternatively, user retries should be triggered for pre-execution failures when the error is classified as USER.

Example error message:
`Grace period [3m0s] exceeded|containers with unready status: [primary]|failed to reserve container name "primary_xxxxx_namespace_uuid_0": name "primary_xxxxx_namespace_uuid_0" is reserved for "containerid123..."`

Example CRD information:
```
nodeStatus:
  n0:
    error:
      code: ContainersNotReady|CreateContainerError
      kind: USER
      message: 'Grace period [3m0s] exceeded|containers with unready status: [primary]|failed
        to reserve container name...'
    phase: 6
```

### Screenshots

_No response_

- [x] Yes

- [x] Yes

themavik:
## Summary

When a pod fails with `CreateContainerError` or `CreateContainerConfigError` and the configured grace period is exceeded, the error is classified as a USER permanent failure (`PhaseInfoFailureWithCleanup`). This prevents system retries from triggering — the task fails immediately even when `retries` is configured.

These errors are typically transient infrastructure issues (e.g. "failed to reserve container name", "container name already in use") that should be retried by the system.

## Root Cause

In `pod_helper.go`, both `CreateContainerError` and `CreateContainerConfigError` use `PhaseInfoFailureWithCleanup` after the grace period, which creates:
- Phase: `PhasePermanentFailure`
- Error kind: `ExecutionError_USER`

USER errors do not trigger system retries (`max-node-retries-system-failures`), so the task fails immediately.

## Fix

Changed both error paths to use `PhaseInfoSystemRetryableFailureWithCleanup`, which creates:
- Phase: `PhaseRetryableFailure`
- Error kind: `ExecutionError_SYSTEM`

This allows the system retry mechanism to handle transient container creation failures.

## Testing

- Updated existing `CreateContainerErrorOutsideGracePeriod` test to assert `PhaseRetryableFailure` instead of `PhasePermanentFailure`
- Updated existing `CreateContainerConfigErrorOutsideGracePeriod` test similarly
- Both tests verify `CleanupOnFailure()` remains `true`
- Within-grace-period tests remain unchanged (still `PhaseInitializing`)

Closes #6786

### DIFF
--- a/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -1030,7 +1030,8 @@ func demystifyPendingHelper(status v1.PodStatus, info pluginsCore.TaskInfo) (plu
 								// allow a grace period for kubelet to resolve
 								// transient issues with the container runtime. If the
 								// error persists beyond this time, the corresponding
-								// task is marked as failed.
+								// task is marked as a system-retryable failure so that
+								// system retries can handle transient infrastructure errors.
 								// NOTE: The current implementation checks for a timeout
 								// by comparing the condition's LastTransitionTime
 								// based on the corresponding kubelet's clock with the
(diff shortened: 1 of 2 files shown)


kumare3:
Cc @machichima good one to carry over

PR #7045, opened 2026-03-18
fix(flytectl): fall back to Docker Desktop socket paths when default is unavailable

### LINKED ISSUE #6138: [BUG] flytectl demo start fails on Mac when using Docker Desktop
### Describe the bug

It seems flyectl demo fails to use default Docker configuration.

```
brew install flyteorg/homebrew-tap/flytectl (worked)
% flytectl demo start
INFO[0000] [0] Couldn't find a config file []. Relying on env vars and pflags. 
🧑‍🏭 Bootstrapping a brand new Flyte cluster... 🔨 🔧
Error: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
{"json":{},"level":"error","msg":"Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?","ts":"2025-01-03T23:09:33-08:00"}
```

```
% docker ps | head -2
CONTAINER ID   IMAGE                       COMMAND                  CREATED          STATUS          PORTS     NAMES
ccd36f830cc9   5ddc6458eb95                "/metrics-server --c…"   42 minutes ago   Up 42 minutes             k8s_metrics-server_metrics-server-86776f5749-qmp8c_kube-system_e9f65542-d8fc-43b7-9956-d22a6cf6ef98_6
```

### Expected behavior

I was expecting the demo to start. It seems straightforward (the issue), but I tried to look for relating settings and so far have not succeeded. I thought I would find something to change the Docker sock address.

```
% docker info | grep address
WARNING: daemon is not using the default seccomp profile
  com.docker.desktop.address=unix:///Users/me/Library/Containers/com.docker.docker/Data/docker-cli.sock
```

```
brew install flyteorg/homebrew-tap/flytectl
flytectl demo start
```

### Screenshots

_No response_

- [X] Yes

- [X] Yes

themavik:
## Summary

On macOS with Docker Desktop, the Docker socket may not be at the default `/var/run/docker.sock`. Docker Desktop places it at `~/.docker/run/docker.sock` or `~/Library/Containers/com.docker.docker/Data/docker-cli.sock` instead. This causes `flytectl demo start` to fail with:

```
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

## Root Cause

`GetDockerClient()` in `docker_util.go` uses `client.NewClientWithOpts(client.FromEnv)`, which defaults to `/var/run/docker.sock` when `DOCKER_HOST` is not set. Docker Desktop for Mac does not always create a symlink at that path.

## Fix

After the initial client creation, if a ping fails and `DOCKER_HOST` is not explicitly set, try well-known Docker Desktop socket paths:
1. `~/.docker/run/docker.sock` (Docker Desktop default on macOS and Linux)
2. `~/Library/Containers/com.docker.docker/Data/docker-cli.sock` (macOS-specific)

For each candidate, check if the socket file exists, create a client, and verify with a ping. Use the first working socket. If none work, return the original client (preserving the existing error behavior).

## Testing

- Existing behavior unchanged when `/var/run/docker.sock` works or `DOCKER_HOST` is set
- On macOS Docker Desktop without the default symlink, the fallback finds the correct socket
- Socket paths are only tried when the initial connection fails (no performance impact on working setups)

Closes #6138

### DIFF
--- a/flytectl/pkg/docker/docker_util.go
+++ b/flytectl/pkg/docker/docker_util.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/docker/docker/api/types"
@@ -60,6 +62,24 @@ var (
 	ExtraHosts         = []string{"host.docker.internal:host-gateway"}
 )
 
+// dockerDesktopSocketPaths returns well-known Docker Desktop socket paths
+// for macOS and Linux when the default /var/run/docker.sock is absent.
+func dockerDesktopSocketPaths() []string {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return nil
+	}
+	paths := []string{
+		filepath.Join(home, ".docker", "run", "docker.sock"),
+	}
+	if runtime.GOOS == "darwin" {
+		paths = append(paths,
+			filepath.Join(home, "Library", "Containers", "com.docker.docker", "Data", "docker-cli.sock"),
+		)
+	}
+	return paths
+}
+
 // GetDockerClient will returns the docker client
 func GetDockerClient() (Docker, error) {
 	if Client == nil {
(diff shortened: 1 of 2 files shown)


themavik:
Any updates on this?

PR #7046, opened 2026-03-18
fix: close HTTP response bodies and file handles to prevent resource leaks

themavik:
## Summary

Several resource leaks found across the codebase where HTTP response bodies and file handles are not properly closed. These can cause connection pool exhaustion, file descriptor leaks, and memory leaks in long-running processes.

### 1. `flyteadmin/auth/authzserver/metadata_provider.go`
- Added `defer response.Body.Close()` after successful HTTP response in `GetOAuth2Metadata`
- Added `response.Body.Close()` before retry when receiving 4xx/5xx status codes in `sendAndRetryHTTPRequest` — without this, each retry iteration leaks the previous response body

### 2. `flytectl/cmd/register/register_util.go`
- Added `defer res.Body.Close()` in `DirectUpload` — the response body was never closed on either success or error paths

### 3. `flytectl/pkg/filesystemutils/file_system_utils.go`
- Added `outFile.Close()` before returning on non-EOF `io.CopyN` error during tar extraction — the file handle was leaked when extraction failed mid-stream

### 4. `flytestdlib/config/viper/collection.go`
- Added `reader.Close()` after `MergeConfig` in `GetConfig` — the config file reader was opened but never closed

## Testing

- All changes are mechanical resource cleanup additions
- No behavioral changes — only ensures proper cleanup of OS resources
- Existing tests continue to pass as the changes only affect cleanup paths

### DIFF
--- a/flyteadmin/auth/authzserver/metadata_provider.go
+++ b/flyteadmin/auth/authzserver/metadata_provider.go
@@ -84,6 +84,7 @@ func (s OAuth2MetadataProvider) GetOAuth2Metadata(ctx context.Context, r *servic
 		if err != nil {
 			return nil, err
 		}
+		defer response.Body.Close()
 
 		raw, err := ioutil.ReadAll(response.Body)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


kumare3:
Cc @EngHabu

PR #7047, opened 2026-03-18
fix: prevent os.RemoveAll on empty path and add missing transaction rollbacks

themavik:
## Summary

Three safety fixes addressing potential data loss and transaction leaks.

### 1. `flytectl/cmd/compile/compile.go` — Critical: `os.RemoveAll("")` can delete current directory

When `GetSerializeOutputFiles` fails, `tmpDir` is `""`. The deferred `os.RemoveAll(tmpDir)` then calls `os.RemoveAll("")`, which resolves to the current working directory and **deletes it recursively**.

**Fix:** Guard the defer with `if tmpDir != ""`.

### 2. `flyteadmin/pkg/repositories/gormimpl/launch_plan_repo.go` — Transaction leak on Commit failure

`SetActive` uses `tx.Begin()` with manual `tx.Rollback()` calls on error paths, but if `tx.Commit()` fails (line 104), the transaction is never rolled back. Depending on the database driver, this can leave the transaction open and leak a connection.

**Fix:** Add `defer tx.Rollback()` immediately after `Begin()`. Rollback after a successful commit is a no-op in GORM, making this safe and idiomatic. Removed the now-redundant manual `tx.Rollback()` calls.

### 3. `flyteadmin/pkg/repositories/config/seed_data.go` — Same transaction leak pattern

Same issue as above in `SeedProjects`.

**Fix:** Add `defer tx.Rollback()` after `Begin()`.

## Testing

- `os.RemoveAll("")` fix is a guard condition — no behavioral change on the success path
- Transaction rollback changes are idiomatic Go patterns — GORM's `Rollback()` is a no-op after `Commit()`
- Existing tests continue to pass

### DIFF
--- a/flyteadmin/pkg/repositories/config/seed_data.go
+++ b/flyteadmin/pkg/repositories/config/seed_data.go
@@ -58,6 +58,7 @@ func uniqueProjects(seedProjects []string, seedProjectsWithDetails []SeedProject
 // Returns a function to seed the database with default values.
 func SeedProjects(db *gorm.DB, projects []SeedProject) error {
 	tx := db.Begin()
+	defer tx.Rollback()
 	for _, project := range projects {
 		projectModel := models.Project{
 			Identifier:  project.Name,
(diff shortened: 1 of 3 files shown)


themavik:
Any updates on this?

PR #7048, opened 2026-03-18
fix(flytectl): handle NewRequest error and close response body in SendRequest

themavik:
## Summary

Two bugs in `SendRequest` in `flytectl/pkg/util/util.go`:

### 1. Ignored `http.NewRequest` error causes nil pointer panic

```go
req, _ := http.NewRequest(method, url, option)
response, err := client.Do(req)  // panics if req is nil
```

If the method or URL is invalid, `http.NewRequest` returns an error and a nil `*Request`. The error was silently discarded with `_`, and `client.Do(nil)` panics.

**Fix:** Check the error from `http.NewRequest` and return it.

### 2. HTTP response body leaked on non-200 status

When the response has a non-200 status code, the function returns an error but never closes `response.Body`, leaking the underlying TCP connection.

**Fix:** Close `response.Body` before returning the error.

## Testing

- Existing test `TestSendRequest` with `"htp://github.com"` now returns a proper error from `NewRequest` instead of relying on `client.Do` to fail
- No behavioral change for callers on the success path

### DIFF
--- a/flytectl/pkg/util/util.go
+++ b/flytectl/pkg/util/util.go
@@ -128,12 +128,16 @@ func PrintSandboxTeardownMessage(flyteConsolePort int, kubeconfigLocation string
 // SendRequest will create request and return the response
 func SendRequest(method, url string, option io.Reader) (*http.Response, error) {
 	client := &http.Client{}
-	req, _ := http.NewRequest(method, url, option)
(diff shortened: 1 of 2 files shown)


themavik:
Any updates on this?

PR #7049, opened 2026-03-18
settings service proto definitions

afbrock:

This PR introduces the new api for settings.

## How was this patch tested?

Tests run in devbox

### Labels

## Docs link

- `main` 
  - \#6583
    - **settings service proto definitions** :point\_left:


buf.yaml
```diff
@@ -31,6 +31,7 @@ lint:
       - flyteidl2/project/project_service.proto
       - flyteidl2/project/common.proto
     ENUM_ZERO_VALUE_SUFFIX:
+      - flyteidl2/org/settings_definition.proto
```
EngHabu: Can we address lint instead of adding more exceptions here? Kind of keeping it for backward compatibility

flyteidl2/org/settings_definition.proto
```diff
@@ -0,0 +1,137 @@
...
+import "buf/validate/validate.proto";
+
+option go_package = "github.com/flyteorg/flyte/v2/gen/go/flyteidl2/org";
+
+// SettingState controls how a setting at a given scope behaves when no value
+// is set. The VALUE state is implicit: it is inferred from which value field
+// inside the SettingValue oneof is populated.
+enum SettingState {
+  // Delegate to the parent scope. This is the default.
+  INHERIT = 0;
```
EngHabu: Can you check the best practices on naming?

flyteidl2/org/settings_definition.proto
```diff
@@ -0,0 +1,137 @@
...
+
+  // Explicitly no value at this scope. Stops the inheritance chain.
+  UNSET = 1;
+}
+
+// SettingScopeLevel identifies which scope level a resolved setting came from.
+enum ScopeLevel {
+  ORG = 0;
```
EngHabu: Ditto

flyteidl2/org/settings_definition.proto
afbrock: yes, updated.

flyteidl2/org/settings_definition.proto
afbrock: updated

buf.yaml
afbrock: i just removed this line, lint does ok without it.

Issue #449, opened 2020-08-05
[Feature]Remove namespace from all non kustomize configuration


kumare3:
**Motivation: Why do you think this is important?**
Deployments of Flyte should not be bound to flyte namespace and should be kustomizable

**Goal: What should the final outcome look like, ideally?**
Kustomization.yaml allows namespace as a primitive

**Describe alternatives you've considered**
to change namespace we have to recreate the yamls

**Flyte component**
- [ ] Overall
- [x] Flyte Setup and Installation scripts
- [ ] Flyte Documentation
- [ ] Flyte communication (slack/email etc)
- [ ] FlytePropeller
- [ ] FlyteIDL (Flyte specification language)
- [ ] Flytekit (Python SDK)
- [ ] FlyteAdmin (Control Plane service)
- [ ] FlytePlugins
- [ ] DataCatalog 
- [ ] FlyteStdlib (common libraries)
- [ ] FlyteConsole (UI)
- [ ] Other

kumare3:
A complete kustomize overhaul has been merged into master

PR #7056, opened 2026-03-19
Add ARCHIVED state to LaunchPlanState enum

blaketastic2:
## Tracking issue

N/A — this addresses a common operational pain point with no existing tracking issue.

Every `pyflyte register` creates new launch plan versions, and nothing cleans up old ones. Over time this leads to heavier list queries, more data to page through, and increased memory pressure on Admin.

Today the only cleanup mechanism is `UpdateLaunchPlan` to set state `INACTIVE`, but `INACTIVE` has meaning in the active/inactive schedule lifecycle — conflating "old cruft" with "intentionally deactivated" makes filtering unreliable. Users who discover downstream launch plans via `ListLaunchPlans` (e.g. filtering on `workflow.short_description`) get back all versions across all deploys when they only care about the latest.

Adding `ARCHIVED` (value `2`) gives operators a distinct state they can set on old versions and filter out (`ne(state,2)`), enabling a scheduled cleanup pattern: keep the most recent N versions per named entity, archive the rest, never touch the active one.

**flyteidl (proto + generated code):**
- Add `ARCHIVED = 2` to the `LaunchPlanState` enum in `launch_plan.proto`
- Regenerate bindings for Go, Python, Rust, TypeScript, and JavaScript
- Update both swagger JSON files (gateway + client assets)

**flyteadmin:**
- Add `archiveLaunchPlan` method to `LaunchPlanManager` — same behavior as `disableLaunchPlan` (disables any active schedule, updates the DB) but sets state to `ARCHIVED` instead of `INACTIVE`
- Add `case admin.LaunchPlanState_ARCHIVED` to the `UpdateLaunchPlan` switch

**flytectl:**
- Repurpose the `--archive` flag to set `LaunchPlanState_ARCHIVED` instead of mapping to `INACTIVE`. This flag was previously deprecated in #449 as an alias for `--deactivate` because "archive" was misleading when no `ARCHIVED` state existed. Now that the state exists, `--archive` does what the name implies.
- Update mutual exclusion logic to handle three flags (`--activate`, `--deactivate`, `--archive`)
- Remove the now-unused `deprecatedCommandWarning` function
- Update help text and flag descriptions

**Note on `--archive` behavioral change:** Prior to this PR, `--archive` was a deprecated alias for `--deactivate` (set state to `INACTIVE`/`0`). After this PR, `--archive` sets state to `ARCHIVED`/`2`. Anyone relying on the deprecated `--archive` flag to deactivate launch plans should switch to `--deactivate`.

## How was this patch tested?

- Updated `TestLaunchPlanCanBeArchived` in flytectl to verify `--archive` now sets `ARCHIVED` state
- All existing flyteadmin manager impl tests pass (launch plan manager, transformers, utils, validation)
- All existing flytectl update tests pass
- Both `flyteadmin` and `flytectl` compile cleanly

### Labels

- **added**: New `ARCHIVED` launch plan state

### DIFF
--- a/flyteadmin/pkg/manager/impl/launch_plan_manager_test.go
+++ b/flyteadmin/pkg/manager/impl/launch_plan_manager_test.go
@@ -33,6 +33,7 @@ import (
 
 var active = int32(admin.LaunchPlanState_ACTIVE)
 var inactive = int32(admin.LaunchPlanState_INACTIVE)
+var archived = int32(admin.LaunchPlanState_ARCHIVED)
 var mockScheduler = &mocks.EventScheduler{}
 var launchPlanIdentifier = &core.Identifier{
 	ResourceType: core.ResourceType_LAUNCH_PLAN,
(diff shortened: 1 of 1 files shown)


flytectl/cmd/update/launch_plan.go
```diff
@@ -47,15 +53,19 @@ func updateLPFunc(ctx context.Context, args []string, cmdCtx cmdCore.CommandCont
...
+	}
 	if archive {
-		deprecatedCommandWarning(ctx, "archive", "deactivate")
-		deactivate = true
-	} else {
-		deactivate = launchplan.UConfig.Deactivate
+		setCount++
 	}
-	if activate == deactivate && deactivate {
+	if setCount > 1 {
```
pingsutw, line 68: What if we do deactivate and archive? It shouldn't raise an error

blaketastic2: These are three distinct states in the enum (`ACTIVE`, `INACTIVE`, `ARCHIVED`), and each flag maps to exactly one `LaunchPlanState` value. If we allowed `--deactivate --archive` together it would be ambiguous — we'd have to silently pick one, which is more confusing than just asking the user to specify a single flag.

Archive already implies the launch plan won't be scheduled (the server-side behavior disables any active schedule), so there's no need to also pass `--deactivate`. If the intent is to archive, `--archive` alone is sufficient.

PR #7057, opened 2026-03-19
Fix manager hanging on shutdown

pingsutw:
## Summary

- Fix the manager process hanging indefinitely on Ctrl+C by adding `wg.Wait()` with a 30-second timeout for worker goroutines
- Support force exit on a second SIGINT/SIGTERM signal, matching the common Go pattern used by controller-runtime and Kubernetes components
- Make HTTP server shutdown errors non-fatal so the rest of the shutdown sequence (worker cleanup) still runs

## Test plan

- `main` 
  - \#6583
    - **Fix manager hanging on shutdown** :point\_left:

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -167,11 +167,13 @@ func (a *App) serve(ctx context.Context) error {
 	}()
 
 	workerCancel()
+	var shutdownErr error
 	if server != nil {
 		shutdownCtx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
 		defer cancel()
 		if err := server.Shutdown(shutdownCtx); err != nil {
 			logger.Errorf(ctx, "Server shutdown error: %v", err)
+			shutdownErr = err
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


app/app.go
```diff
@@ -155,17 +155,38 @@ func (a *App) serve(ctx context.Context) error {
...
+	}()
+
 	workerCancel()
 	if server != nil {
-		shutdownCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
 		defer cancel()
 		if err := server.Shutdown(shutdownCtx); err != nil {
-			return fmt.Errorf("server shutdown failed: %w", err)
+			logger.Errorf(ctx, "Server shutdown error: %v", err)
```
machichima, line 175: Can we return this error rather than only log here? e.g. we can store the error here and return the error  below when done or timeout

pingsutw: fixed

machichima (approved):

PR #6991, opened 2026-03-08
Fix: [Run Service] Ensure GetActionData works expected

SharpKoi:
## Tracking issue

Closes #6974

## Why are the changes needed?

The `GetActionData` endpoint in Run Service currently returns inputs missing the literals and context fields, which is unexpected. This PR fixes the issue #6974 .

## What changes were proposed in this pull request?

**Key Changes:**

- Fix the read path in `GetActionData` to match the write path used during storage.
- In `CreateRun`, store the full `task.Inputs` (which preserves both literals and context) instead of only `task.Inputs.Literals`.

## How was this patch tested?

This PR changes how `task.Inputs` is written and read: previously, `task.Inputs` was stored and read as a `core.LiteralMap`, preserving only literals. Now, the full `task.Inputs` is stored and read directly, preserving both literals and context.
To ensure backward compatibility between the old and new formats, the following wire-compatibility tests were added in `run_service_test.go`:

- `task.Inputs` written → read as `task.Inputs`
- `task.Inputs` written → read as `core.LiteralMap`
- `core.LiteralMap` written → read as `task.Inputs`

### Labels

- **changed**: For changes in existing functionality.
- **fixed**: For any bug fixed.

This is important to improve the readability of release notes.

## Check all the applicable boxes 

- [x] I updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] All commits are signed-off.

## Docs link

- `main` 
  - \#6583
    - **Fix: \[Run Service] Ensure GetActionData works expected** :point\_left:

PR #7064, opened 2026-03-20
Fix CreateRun response, output URI, and GetActionData

pingsutw:
## Tracking issue

Related to #6991

1. **CreateRunResponse was incomplete** — only returned the action ID without metadata or status from the DB model
2. **Output URI was missing `/outputs.pb` suffix** — the executor's `outputRefs` stored the prefix path (e.g., `s3://flyte-data/<id>`) instead of the full file path (`s3://flyte-data/<id>/outputs.pb`), causing `GetActionData` to fail reading outputs
3. **GetActionData was fragile** — constructed output paths manually from `runOutputBase` instead of reading the URI stored in action attempts

- **`runs/service/run_service.go`**: Use `convertRunToProto(run)` for `CreateRunResponse`; refactor `GetActionData` to read outputs from attempt URIs with concurrent `errgroup`; set root action name to `runId.Name`
- **`executor/pkg/controller/taskaction_controller.go`**: Append `/outputs.pb` to output URI in `outputRefs`
- **`runs/repository/impl/action.go`**: Append `/inputs.pb` to inputUri when creating run; set initial `Attempts` to 1; remove debug logging
- **`runs/repository/models/action.go`**: Add `Attempts` field to Action model
- **`runs/repository/interfaces/action.go`**: Rename `inputUri` param to `inputPrefix` for clarity

## How was this patch tested?

Unit tests added:

- `TestFillDefaultInputs` — default input filling with various edge cases
- `TestTaskIdFromTaskSpec` — task ID extraction from spec
- `TestGenerateCacheKeyForTask` — cache key generation determinism and input filtering
- `TestCreateRun_ResponseUsesRunModel` — verifies response populated from DB model
- `TestCreateRun_ActionIDUsesRunName` — verifies action ID uses run name
- `TestGetActionData_ReadsOutputFromAttempts` — verifies output read from attempt URI
- `TestGetActionData_NonSucceededSkipsOutputs` — verifies output skipped for non-terminal actions
- `TestOutputRefs` — verifies executor output URI includes `/outputs.pb`

### Labels

- **fixed**: Bug fix for output URI and CreateRun response
- **changed**: GetActionData refactored to use attempt URIs

- `main` 
  - \#6583
    - **Fix CreateRun response, output URI, and GetActionData** :point\_left:

### DIFF
--- a/runs/service/run_service.go
+++ b/runs/service/run_service.go
@@ -127,7 +127,6 @@ func (s *RunService) CreateRun(
 	// the run name here and normalise the request to a RunId so that the repo
 	// receives a single, pre-formed identifier — avoiding a second independent
 	// name generation inside the repo layer.
-	var org, project, domain, name string
 	var runId *common.RunIdentifier
 	switch id := request.Id.(type) {
 	case *workflow.CreateRunRequest_RunId:
@@ -177,13 +176,13 @@ func (s *RunService) CreateRun(
 
 	inputs = fillDefaultInputs(inputs, taskSpec.GetDefaultInputs())
 	// Compute storage URIs before DB insert so they're persisted in the ActionSpec
-	inputPrefix := buildInputPrefix(s.storagePrefix, org, project, domain, name)
-	runOutputBase := buildRunOutputBase(s.storagePrefix, org, project, domain, name)
+	inputPrefix := buildInputPrefix(s.storagePrefix, runId.GetOrg(), runId.GetProject(), runId.GetDomain(), runId.GetName())
+	runOutputBase := buildRunOutputBase(s.storagePrefix, runId.GetOrg(), runId.GetProject(), runId.GetDomain(), runId.GetName())
 	runSpec.RawDataStorage = &task.RawDataStorage{RawDataPrefix: s.storagePrefix}
 
 	// Persist inputs to storage. The task runtime always tries to load inputs.pb,
 	// even when the task has no user inputs, so we must write an empty LiteralMap.
-	literalMap := inputsToLiteralMap(req.Msg.Inputs)
+	literalMap := inputsToLiteralMap(inputs)
 	inputRef := storage.DataReference(inputPrefix + "/inputs.pb")
 	if err := s.dataStore.WriteProtobuf(ctx, inputRef, storage.Options{}, literalMap); err != nil {
 		logger.Errorf(ctx, "Failed to write inputs to storage: %v", err)
@@ -221,7 +220,7 @@ func (s *RunService) CreateRun(
 				},
 			},
 		},
-		RunSpec: request.RunSpec,
+		RunSpec: runSpec,
 	}))
 	if err != nil {
 		logger.Errorf(ctx, "Failed to enqueue root action: %v", err)


runs/repository/impl/action_test.go
```diff
@@ -47,6 +47,7 @@ func setupActionDB(t *testing.T) *gorm.DB {
 		updated_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
 		ended_at DATETIME,
 		duration_ms INTEGER,
+		attempts INTEGER NOT NULL DEFAULT 0,
```
machichima: Should we start from 1 or 0?

pingsutw: 1

pingsutw: updated

runs/service/run_service.go
```diff
@@ -125,32 +128,58 @@ func (s *RunService) CreateRun(
...
-		org = id.RunId.Org
-		project = id.RunId.Project
-		domain = id.RunId.Domain
-		name = id.RunId.Name
+		runId = request.GetRunId()
 	case *workflow.CreateRunRequest_ProjectId:
-		org = id.ProjectId.Organization
-		project = id.ProjectId.Name
-		domain = id.ProjectId.Domain
-		name = generateRunName(time.Now().UnixNano())
```
machichima, line 138: Seems like we didn't set value for those variables, we will get empty values below

runs/service/run_service.go
machichima: We should use `inputs` here rather than `req.Msg.Inputs`?

runs/service/run_service.go
```diff
@@ -162,66 +191,47 @@ func (s *RunService) CreateRun(
...
+			RunOutputBase: runOutputBase,
+			Spec: &actions.Action_Task{
+				Task: &workflow.TaskAction{
+					Id:       taskID,
+					Spec:     taskSpec,
+					CacheKey: wrapperspb.String(cacheKey),
+				},
+			},
+		},
+		RunSpec: request.RunSpec,
```
machichima: I think we should use `runSpec` here rather than `request.RunSpec`?

runs/service/run_service.go
```diff
@@ -576,55 +587,100 @@ func (s *RunService) GetActionData(
...
+			}
+
+			if len(attempts) == 0 {
+				return app.NewServerError(codes.NotFound, "outputs not available, no attempts for action")
+			}
+
+			outputUri := attempts[len(attempts)-1].GetOutputs().GetOutputUri()
+			if outputUri == "" {
+				return app.NewServerError(codes.NotFound, "outputs not available")
+			}
```
machichima, line 662: Should we fallback to check if the output is in object store rather than directly fail here? Will there be a case that the output is already being written into the object store but DB hasn't updated

runs/service/run_service.go
pingsutw: good catch, thanks. I remove those variables and use runId.GetOrg(), runId.GetProject() instead.

runs/service/run_service.go
pingsutw: fixed

runs/service/run_service.go
pingsutw: No, executor always write the output first before sending the event (output uri is in the event).

runs/service/run_service.go
pingsutw: fixed

PR #7070, opened 2026-03-20
[V2][Bug] Fix WatchActionDetails to return full action details in streamed updates

pingsutw:

`WatchActionDetails` was using `actionModelToDetails()` for streamed updates, which only converts basic model fields. This meant streamed updates were **missing task spec and attempt details** — information that was included in the initial response via `getActionDetails()`. Clients receiving streamed updates would see incomplete action details compared to the initial snapshot.

- Extract `buildActionDetails(ctx, model, actionId)` from `getActionDetails` — takes a pre-fetched `*models.Action` and enriches it with task spec + attempts
- Make `getActionDetails` a thin wrapper: calls `GetAction` then delegates to `buildActionDetails`
- Replace `actionModelToDetails` with `buildActionDetails` in the `WatchActionDetails` streaming path so that streamed updates include full action details (task spec, attempts) matching the initial response
- Add comprehensive unit tests for `WatchActionDetails` covering:
  - Streamed updates include full details (task spec, attempts)
  - Error info propagation in streamed updates
  - Filtering of updates for other actions
  - Error handling when `buildActionDetails` fails

## How was this patch tested?

- `go test ./runs/service/ -run TestWatchActionDetails -v` — all 4 tests pass
- `go test ./runs/service/ -v` — no regressions
- `go build ./runs/...` — compiles cleanly

* `main` 
  - \#6583
    - **\[V2]\[Bug] Fix WatchActionDetails to return full action details in streamed updates** :point\_left:


runs/service/watch_action_details_test.go
machichima, line 1: I think this is an integration test rather than unit test? Should we move it to somewhere else and keep unit test here only?

PR #7072, opened 2026-03-21
[V2] Add RunLogsService for streaming pod logs

pingsutw:
# TL;DR

Add `RunLogsService` implementation to the OSS Flyte repo, enabling the UI to stream pod logs for action attempts directly from Kubernetes.

## Type

- [ ] Bug Fix
- [x] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [x] Unit tests added

## Complete description

Implements the `RunLogsService` with an extensible `LogStreamer` interface architecture:

- **`RunLogsService`** (orchestrator) — resolves `LogContext` from action events, delegates to a `LogStreamer`
- **`LogStreamer` interface** — abstracts log fetching so backends can be swapped (K8s direct, data proxy, etc.)
- **`K8sLogStreamer`** — reads pod logs directly via the K8s API, parses RFC3339Nano timestamps, and batches \~100 lines per response
- **`runs/setup.go`** — registers the service when `K8sConfig` is available
- **12 unit tests** covering orchestrator logic (happy path, not found, DB errors, streamer errors) and K8s streamer logic (timestamp parsing, pod/container resolution)

The design mirrors the cloud implementation's separation of concerns but uses direct K8s pod log streaming instead of a data proxy.

## Tracking Issue

NA

## Follow-up issue

NA

- `main` 
  - \#6583
    - **\[V2] Add RunLogsService for streaming pod logs** :point\_left:

### DIFF
--- a/runs/repository/impl/action.go
+++ b/runs/repository/impl/action.go
@@ -331,7 +331,7 @@ func (r *actionRepo) GetLatestEventByAttempt(ctx context.Context, actionID *comm
 	result := r.db.WithContext(ctx).
 		Where("org = ? AND project = ? AND domain = ? AND run_name = ? AND name = ? AND attempt = ?",
 			actionID.Run.Org, actionID.Run.Project, actionID.Run.Domain, actionID.Run.Name, actionID.Name, attempt).
-		Order("version DESC, phase DESC").
+		Order("phase DESC, version DESC").
 		First(&event)
 	if result.Error != nil {
 		if result.Error == gorm.ErrRecordNotFound {
--- a/docker/sandbox-bundled/manifests/complete.yaml
+++ b/docker/sandbox-bundled/manifests/complete.yaml
@@ -1094,7 +1094,7 @@ type: Opaque
 ---
 apiVersion: v1
 data:
-  haSharedSecret: Y25XVlo4aGdiV05WRFpXRQ==
+  haSharedSecret: ZEFIMEQzSW1OeE5HZ1Z1UQ==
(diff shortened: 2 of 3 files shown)


executor/pkg/controller/taskaction_controller.go
```diff
@@ -360,7 +360,6 @@ func (r *TaskActionReconciler) updateTaskActionStatus(
 		return err
 	}
 
-	logger.Info("updateTaskActionStatus", "name", oldTaskAction.Name, "old status", oldTaskAction.Status, "new status", newTaskAction.Status)
```
pingsutw, line 363: too noisy

runs/repository/impl/action.go
```diff
@@ -732,22 +789,47 @@ func (r *actionRepo) WatchActionUpdates(ctx context.Context, runID *common.RunId
...
+				// deduplicating by action name so we only query the DB once
+				// per action during a burst of updates.
+				pending := make(map[string]bool)
 				if len(notifPayload) > len(runPrefix) && notifPayload[:len(runPrefix)] == runPrefix {
-					// Extract action name from payload
-					actionName := notifPayload[len(runPrefix):]
+					pending[notifPayload[len(runPrefix):]] = true
+				}
+				// Drain buffered notifications without blocking.
+			drain:
```
pingsutw, line 800: if we don't drain here, action will fire 10 notifications in quick succession, we'd call GetAction() 10 times and send 10 nearly-identical updates

PR #7080, opened 2026-03-23
feat(runs): async abort reconciler with retry and crash recovery

AdilFayyaz:

AbortRun/AbortAction previously called actionsClient.Abort() synchronously after updating the DB. If the call failed, the DB showed ABORTED but the pod kept running with no retry path. Abort needed to be made reliable without blocking the user-facing response.

- Add three columns to actions (abort\_requested\_at, abort\_attempt\_count, abort\_reason); DB update is immediate so the UI sees ABORTED optimistically
- Introduce AbortReconciler — a background worker pool that drives pod termination to completion with exponential backoff and a configurable max-attempts cap; uses a deduplicated in-memory queue so concurrent pushes for the same action are no-ops
- AbortRun/AbortAction push directly to the reconciler queue after the DB update; startup scan re-enqueues any pending aborts left over from a process crash
- Scope all abort-related DB queries with run\_name to correctly target individual actions when node names (e.g. a0) repeat across runs

## How was this patch tested?

- go test ./runs/... passes
- TestAbortReconciler\_SuccessOnFirstAttempt — happy path: worker calls actionsClient.Abort, clears DB flag
- TestAbortReconciler\_RetriesOnFailure — fails N−1 times, succeeds on Nth attempt
- TestAbortReconciler\_GivesUpAtMaxAttempts — always fails; flag cleared after max attempts with no further calls
- TestAbortReconciler\_DeduplicatesQueue — same action pushed twice; actionsClient.Abort called exactly once
- TestAbortReconciler\_StartupScanPicksUpPending — pending aborts from before startup are all processed
- TestAbortReconciler\_NotFoundTreatedAsSuccess — NotFound from actions service clears the flag without retry
- Manually ran an example from the sdk and tested the abort functionality

## Docs link

- `main` 
  - \#6583
    - **feat(runs): async abort reconciler with retry and crash recovery** :point\_left:

### DIFF
--- a/runs/repository/impl/action.go
+++ b/runs/repository/impl/action.go
@@ -553,8 +553,10 @@ func (r *actionRepo) ListPendingAborts(ctx context.Context) ([]*models.Action, e
 // MarkAbortAttempt increments abort_attempt_count and returns the new value.
 // Called by the reconciler before each actionsClient.Abort call.
 func (r *actionRepo) MarkAbortAttempt(ctx context.Context, actionID *common.ActionIdentifier) (int, error) {
+	var action models.Action
 	result := r.db.WithContext(ctx).
-		Model(&models.Action{}).
+		Model(&action).
+		Clauses(clause.Returning{Columns: []clause.Column{{Name: "abort_attempt_count"}}}).
 		Where("org = ? AND project = ? AND domain = ? AND run_name = ? AND name = ?",
 			actionID.Run.Org, actionID.Run.Project, actionID.Run.Domain, actionID.Run.Name, actionID.Name).
 		Updates(map[string]interface{}{
@@ -563,16 +565,6 @@ func (r *actionRepo) MarkAbortAttempt(ctx context.Context, actionID *common.Acti
 		})
 	if result.Error != nil {
 		return 0, fmt.Errorf("failed to mark abort attempt: %w", result.Error)
-	}
-
-	// Re-fetch the updated count.
-	var action models.Action
-	if err := r.db.WithContext(ctx).
-		Select("abort_attempt_count").
-		Where("org = ? AND project = ? AND domain = ? AND run_name = ? AND name = ?",
-			actionID.Run.Org, actionID.Run.Project, actionID.Run.Domain, actionID.Run.Name, actionID.Name).
-		First(&action).Error; err != nil {
-		return 0, fmt.Errorf("failed to read abort attempt count: %w", err)
 	}
 	return action.AbortAttemptCount, nil
 }
--- a/runs/service/abort_reconciler.go
+++ b/runs/service/abort_reconciler.go
@@ -3,6 +3,7 @@ package service
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"sync"
 	"time"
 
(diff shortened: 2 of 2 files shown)


runs/service/abort_reconciler.go
```diff
@@ -0,0 +1,317 @@
...
+
+// Run starts the reconciler. It blocks until ctx is cancelled.
+func (r *AbortReconciler) Run(ctx context.Context) error {
+	logger.Infof(ctx, "AbortReconciler starting (%d workers, max %d attempts)", r.cfg.Workers, r.cfg.MaxAttempts)
+
+	// Startup scan: enqueue any actions that were left pending from before this process started.
+	if err := r.startupScan(ctx); err != nil {
+		logger.Errorf(ctx, "AbortReconciler startup scan failed: %v", err)
+		// Non-fatal — the NOTIFY watcher will still pick up new aborts.
+	}
```
pingsutw: We should start the worker first, and then push the abort jobs to the queue. Otherwise, if the number of jobs exceeds the queue size, the push in startupScan could block forever

runs/repository/impl/action.go
```diff
@@ -474,13 +484,143 @@ func (r *actionRepo) AbortAction(ctx context.Context, actionID *common.ActionIde
...
+	}
+	return actions, nil
+}
+
+// MarkAbortAttempt increments abort_attempt_count and returns the new value.
+// Called by the reconciler before each actionsClient.Abort call.
+func (r *actionRepo) MarkAbortAttempt(ctx context.Context, actionID *common.ActionIdentifier) (int, error) {
+	result := r.db.WithContext(ctx).
+		Model(&models.Action{}).
+		Where("org = ? AND project = ? AND domain = ? AND name = ?",
```
pingsutw: Should we add run name here as well?

AdilFayyaz: (org,project,domain,name) already uniquely identify exactly one row. Adding run name would be redundant in my opinion.

pingsutw: but you could have different runs, both of them have subtask. The name of the subtask could be the same.

pingsutw: also, isn't the root action name always `a0`?

AdilFayyaz: you're right, ill push a fix

runs/repository/impl/action.go
```diff
@@ -509,28 +512,89 @@ func (r *actionRepo) UpdateActionPhase(
...
+	// Re-fetch the updated count.
+	var action models.Action
+	if err := r.db.WithContext(ctx).
+		Select("abort_attempt_count").
+		Where("org = ? AND project = ? AND domain = ? AND run_name = ? AND name = ?",
+			actionID.Run.Org, actionID.Run.Project, actionID.Run.Domain, actionID.Run.Name, actionID.Name).
+		First(&action).Error; err != nil {
+		return 0, fmt.Errorf("failed to read abort attempt count: %w", err)
+	}
+	return action.AbortAttemptCount, nil
```
pingsutw: We can combine it into single query by using clause.Returning.  https://gorm.io/docs/update.html#Returning-Data-From-Modified-Rows

runs/service/abort_reconciler.go
```diff
@@ -0,0 +1,274 @@
...
+		r.queue.remove(task.key)
+		return
+	}
+
+	// Exponential backoff: initialDelay * 2^(attempt-1), capped at maxDelay.
+	delay := r.cfg.InitialDelay * (1 << (attemptCount - 1))
+	if delay > r.cfg.MaxDelay {
+		delay = r.cfg.MaxDelay
+	}
+	logger.Infof(ctx, "AbortReconciler: scheduling retry for %s in %s", task.key, delay)
```
pingsutw: it would be great to add jitter here as well. https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

PR #7088, opened 2026-03-25
[V2] Fix single binary and proxy pod slow start issue

popojk:
## Tracking issue

n/a

We originally give the single binary and proxy pod a 30s readinessProbe initialDelaySeconds, which caused a delay for the pods marked as ready eventhough they are ready already. This PR fixes the issues by decrease the initialDelaySeconds

### Labels

## Docs link

- `main` 
  - \#6583
    - **\[V2] Fix single binary and proxy pod slow start issue** :point\_left:

### DIFF
--- a/charts/flyte-sandbox/templates/proxy/deployment.yaml
+++ b/charts/flyte-sandbox/templates/proxy/deployment.yaml
@@ -28,10 +28,12 @@ spec:
           livenessProbe:
             tcpSocket:
               port: http
-            initialDelaySeconds: 30
+            initialDelaySeconds: 15
+            failureThreshold: 10
           readinessProbe:
             tcpSocket:
               port: http
+            initialDelaySeconds: 5
       volumes:
         - name: config
           configMap:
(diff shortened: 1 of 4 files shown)


charts/flyte-sandbox/templates/proxy/deployment.yaml
```diff
@@ -28,10 +28,12 @@ spec:
           livenessProbe:
             tcpSocket:
               port: http
-            initialDelaySeconds: 30
+            initialDelaySeconds: 15
+            failureThreshold: 10
```
popojk, line 32: Sometimes we see a restart count because the pod is not ready yet but livenessProbe check executed. According to k8s [doc](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#before-you-begin), a common practice is to give a higher failureThreshold to prevent the pod from being killed.

pingsutw (approved):

PR #7091, opened 2026-03-25
[v2] Event watcher

machichima:

Previously `WatchClusterEvents` incorrectly return action phase rather than cluster events.

- Add k8s event watcher to capture events for objects and store in `AdditionalReasons` in phase info
  - taskaction controller will take them out from `AdditionalReasons` and put into `ClusterEvents` in action event
- `WatchClusterEvent` get from `ClusterEvents`

## How was this patch tested?

- Run on sandbox and ensure we can see k8s events

image

### Labels

## Docs link

- `main` 
  - \#6583
    - **\[v2] Event watcher** :point\_left:

### DIFF
--- a/runs/service/run_service.go
+++ b/runs/service/run_service.go
@@ -296,7 +296,7 @@ func (s *RunService) GetRunDetails(
 	}
 
 	details := &workflow.RunDetails{
-		RunSpec: extractRunSpec(run.ActionSpec),
+		RunSpec: extractRunSpecFromActionModel(ctx, run),
 		Action:  actionDetails,
 	}
 
@@ -1282,10 +1282,10 @@ func extractActionSpec(specProto []byte) *workflow.ActionSpec {
 		return nil
 	}
 	var specMsg workflow.ActionSpec
-	if err := json.Unmarshal(specProto, &specMsg); err != nil {
-		return nil
+	if err := proto.Unmarshal(specProto, &specMsg); err == nil {
+		return &specMsg
 	}
-	return &specMsg
+	return nil
 }
 
 // extractRunSpec parses ActionSpec JSON to extract the RunSpec
@@ -1298,6 +1298,25 @@ func extractRunSpec(specProto []byte) *task.RunSpec {
 		return nil
 	}
 	return spec.RunSpec
+}
+
+func extractRunSpecFromActionModel(ctx context.Context, action *models.Action) *task.RunSpec {
+	if action == nil {
+		return nil
+	}
+
+	if len(action.RunSpec) > 0 {
+		spec := &task.RunSpec{}
+		if err := proto.Unmarshal(action.RunSpec, spec); err == nil {
+			return spec
+		} else {
+			logger.Warnf(ctx, "Failed to unmarshal action.RunSpec for %s/%s/%s/%s/%s, falling back to ActionSpec: %v",
+				action.Org, action.Project, action.Domain, action.RunName, action.Name, err)
+		}
+	}
+
+	// Fallback to extract from ActionSpec
+	return extractRunSpec(action.ActionSpec)
 }
 
 // Helper functions
--- a/actions/config/config.go
+++ b/actions/config/config.go
@@ -17,6 +17,7 @@ var defaultConfig = &Config{
 		Namespace: "flyte",
 	},
 	WatchBufferSize: 100,
+	WatchWorkers:    4,
 	RunServiceURL:   "http://localhost:8090",
 	// 8M slots × 8 bytes/pointer = 64 MB; can track ~8M unique actions.
 	RecordFilterSize: 1 << 23,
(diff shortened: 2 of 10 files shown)


runs/service/run_service.go
```diff
@@ -1105,41 +1105,58 @@ func (s *RunService) sendChangedActions(
...
-	}
-
-	// Step 2: Watch for updates from DB
+	// Watch for updates from DB
 	updatesCh := make(chan *models.Action, 50)
 	errsCh := make(chan error, 1)
 	go s.repo.ActionRepo().WatchActionUpdates(ctx, actionID.Run, updatesCh, errsCh)
 
-	lastPhase := action.Phase
+	const maxEvents = 500
```
machichima, line 1096: Just set a reasonable number here first, normally we will not get to this vaue. We can decide whether we want to increase/decrease it later one

executor/pkg/plugin/k8s/event_watcher.go
```diff
@@ -0,0 +1,151 @@
...
+	CreatedAt  time.Time
+	RecordedAt time.Time
+}
+
+type objectEventWatcher interface {
+	List(objectKey watchedObjectKey, createdAfter time.Time) []*eventInfo
+}
+
+type controllerRuntimeEventWatcher struct {
+	mu          sync.RWMutex
```
pingsutw: write will block other readers if we just use one lock. should we create one lock per eventInfo, like
```
type objectEvents struct {
	mu         sync.RWMutex
	eventInfos map[types.NamespacedName]*eventwatcheriface.EventInfo
}
```

machichima: Thanks for the advice! Updated in https://github.com/flyteorg/flyte/pull/7091/commits/2ada28c21c8e3c47e38f7f86ceefcfbfaa5ccf6b

runs/service/run_service.go
```diff
@@ -1105,41 +1105,60 @@ func (s *RunService) sendChangedActions(
...
 	ctx context.Context,
 	req *connect.Request[workflow.WatchClusterEventsRequest],
 	stream *connect.ServerStream[workflow.WatchClusterEventsResponse],
 ) error {
 	actionID := req.Msg.Id
+	attempt := req.Msg.GetAttempt()
 	logger.Infof(ctx, "Received WatchClusterEvents request for: %s/%s", actionID.Run.Name, actionID.Name)
 
-	// Step 1: Send existing events from current DB state
 	action, err := s.repo.ActionRepo().GetAction(ctx, actionID)
```
popojk, line 1091: Should we watch first then query action db? I t is likely that some updates might be lost between the time window

runs/service/run_service.go
```diff
@@ -1149,40 +1168,48 @@ func (s *RunService) WatchClusterEvents(
 			if !ok {
 				return nil
 			}
-			// Filter to this specific action
 			if updated.Name != actionID.Name {
 				continue
```
popojk: Looks like this continue here will makes the program go back to the top while loop and do the drain process again. Is there any reason for doing so?

machichima: Update: This is to prevent the event miss. I tried to skip when the update is not from target action, we will only get partial event (see following).
I think it's a bug now and my implementation is a workaround. Will try to deal with this in follow-ups

Image

runs/service/run_service.go
machichima: Fixed in https://github.com/flyteorg/flyte/pull/7091/commits/2868d5e8a2f693dacdaa00c8afc5b2a69ef92252

runs/service/run_service.go
machichima: add a todo comment here: https://github.com/flyteorg/flyte/pull/7091/commits/ee6efcb238fb1db2a5a69581e33e61efecc3995b

executor/pkg/plugin/k8s/event_watcher.go
pingsutw: We could use sync.Map since read is more than write here

executor/pkg/plugin/k8s/event_watcher.go
```diff
@@ -0,0 +1,190 @@
...
+	ctrlcache "sigs.k8s.io/controller-runtime/pkg/cache"
+)
+
+type watchedObjectKey struct {
+	Namespace string
+	Name      string
+	Kind      string
+}
+
+type eventInfo struct {
```
pingsutw, line 21: Should we add `reason` to eventInfo too?

executor/pkg/plugin/k8s/event_watcher.go
```diff
@@ -0,0 +1,190 @@
...
+		eventInfos.mu.Lock()
+
+		// Revalidate this bucket is still current before writing.
+		w.mu.RLock()
+		stillCurrent := w.objectCache[objectKey] == eventInfos
+		w.mu.RUnlock()
+		if !stillCurrent {
+			// eventInfos being deleted/changed, we should get the newest object again
+			eventInfos.mu.Unlock()
+			continue
```
pingsutw: do we still need revalidate if we already have a lock per event?

machichima: Fix by not deleting `eventInfo` when it's empty in https://github.com/flyteorg/flyte/pull/7091/commits/79769955964a7144b42093345bc3b8bbd4beeaa1
Also use `sync.Map` instead for `objectCache`

executor/pkg/plugin/k8s/event_watcher.go
machichima: Added in https://github.com/flyteorg/flyte/pull/7091/commits/24f4da62d9c6e2e1b93178efad22df5a23e155b8

PR #7093, opened 2026-03-25
feat(executor): in-place task retry with scoped output paths per attempt

AdilFayyaz:

When a task pod fails with a recoverable error, the executor relied on the runs service to create a new TaskAction CR for each retry. This external orchestration was broken, leaving retryable tasks stuck in a non-terminal RetryableFailure loop indefinitely.

- Intercepts PhaseRetryableFailure in TaskActionReconciler.Reconcile() before updating status
- If retries remain (currentAttempts < maxAttempts): calls p.Abort() to delete the failed pod, increments Status.Attempts, clears PluginState, and overrides the transition to PhaseQueued — causing the next reconcile to launch a fresh pod under the same TaskAction
- If retries are exhausted: converts PhaseRetryableFailure → PhasePermanentFailure, making the TaskAction terminal (Failed=True)
- Pod naming is naturally collision-free: buildGeneratedName encodes the attempt number (<name>-0, <name>-1, ...), so each restart creates a distinct pod

## How was this patch tested?

- Deploy a task with retries: 2 (maxAttempts=3) that always exits non-zero
- Verify pods `[name]-0, [name]-1, [name]-2` are each created and deleted in sequence:
  `kubectl get pods -n [namespace] | grep [taskaction-name]`
- Verify Status.Attempts increments on each failure:\
  `kubectl get taskaction [name] -n [namespace] -o jsonpath='{.status.attempts}'`
- After the third failure, verify the TaskAction is terminal:\
  `kubectl get taskaction [name] -n [namespace] -o jsonpath='{.status.conditions}'`

***

Failed=True, Reason=PermanentFailure

- Deploy a task with retries: 0 — confirm first failure immediately produces PermanentFailure

## Docs link

- `main` 
  - \#6583
    - **feat(executor): in-place task retry with scoped output paths per attempt** :point\_left:

### DIFF
--- a/actions/k8s/client.go
+++ b/actions/k8s/client.go
@@ -599,7 +599,7 @@ func buildOutputUri(ctx context.Context, ta *executorv1.TaskAction) string {
 		return ""
 	}
 	attempt := ta.Status.Attempts
-	if attempt == 0 {
+	if attempt == 0 { // if attempts is not set, default to 1
 		attempt = 1
 	}
 	prefix, err := plugin.ComputeActionOutputPath(ctx, ta.Namespace, ta.Name, ta.Spec.RunOutputBase, ta.Spec.ActionName, attempt)
--- a/executor/pkg/plugin/task_exec_context.go
+++ b/executor/pkg/plugin/task_exec_context.go
@@ -137,7 +137,7 @@ func NewTaskExecutionContext(
 	// Output writer — scope outputs per action and attempt so retries don't overwrite each other.
 	// Path: <RunOutputBase>/<shard>/<ActionName>/<attempt>/
 	attempt := taskAction.Status.Attempts
-	if attempt == 0 {
+	if attempt == 0 { // if attempts is not set, default to 1
 		attempt = 1
 	}
 	outputPrefix, err := ComputeActionOutputPath(ctx, taskAction.Namespace, taskAction.Name, taskAction.Spec.RunOutputBase, taskAction.Spec.ActionName, attempt)


executor/pkg/controller/taskaction_controller.go
```diff
@@ -215,6 +215,40 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	// Map transition phase to TaskAction conditions
 	phaseInfo := transition.Info()
+
+	// In-place pod restart: when a recoverable failure occurs, restart the pod within the
```
pingsutw: The plugin may restart pod, spark job, ray job, connector task. 
```suggestion
	// In-place task restart: when a recoverable failure occurs, restart the pod within the
```

executor/pkg/controller/taskaction_controller.go
```diff
@@ -215,6 +215,40 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			// Abort (delete) the current pod before incrementing attempts.
+			// tCtx was built with the current attempt number so Abort targets the right pod.
+			if abortErr := p.Abort(ctx, tCtx); abortErr != nil {
+				logger.Error(abortErr, "failed to abort pod during in-place restart")
+			}
+			// Track the new attempt count; applied to Status.Attempts after the stateMgr block.
+			restartAttempts = currentAttempts + 1
+			// Override the transition to Queued so the TaskAction stays non-terminal.
+			transition = pluginsCore.DoTransition(pluginsCore.PhaseInfoQueued(time.Now(), pluginsCore.DefaultPhaseVersion,
+				fmt.Sprintf("restarting pod (attempt %d/%d)", currentAttempts+1, maxAttempts)))
```
pingsutw:
```suggestion
				fmt.Sprintf("restarting task (attempt %d/%d)", currentAttempts+1, maxAttempts)))
```

executor/pkg/plugin/task_exec_context.go
```diff
@@ -95,12 +134,16 @@ func NewTaskExecutionContext(
...
-	// RunOutputBase is run-level (e.g. s3://bucket/org/proj/domain/run/),
-	// append the action name to make it action-specific.
-	outputPrefix := storage.DataReference(
-		strings.TrimRight(taskAction.Spec.RunOutputBase, "/") + "/" + taskAction.Spec.ActionName,
-	)
+	// Output writer — scope outputs per action and attempt so retries don't overwrite each other.
+	// Path: <RunOutputBase>/<shard>/<ActionName>/<attempt>/
+	attempt := taskAction.Status.Attempts
+	if attempt == 0 {
+		attempt = 1
```
pingsutw, line 141: should we use attempt++? what if `taskAction.Status.Attempts` is 1

actions/k8s/client.go
```diff
@@ -571,11 +572,21 @@ func buildNamespace(runID *common.RunIdentifier) string {
...
+// It uses the same path structure as the executor's ComputeActionOutputPath so that
+// the SDK can find outputs written by the executor.
+func buildOutputUri(ctx context.Context, ta *executorv1.TaskAction) string {
 	if ta.Spec.RunOutputBase == "" {
 		return ""
 	}
-	return strings.TrimRight(ta.Spec.RunOutputBase, "/") + "/" + ta.Spec.ActionName
+	attempt := ta.Status.Attempts
+	if attempt == 0 {
+		attempt = 1
```
pingsutw: should we use attempt++ here

executor/pkg/controller/taskaction_controller.go
```diff
@@ -215,6 +215,40 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			// Abort (delete) the current pod before incrementing attempts.
+			// tCtx was built with the current attempt number so Abort targets the right pod.
+			if abortErr := p.Abort(ctx, tCtx); abortErr != nil {
+				logger.Error(abortErr, "failed to abort pod during in-place restart")
+			}
+			// Track the new attempt count; applied to Status.Attempts after the stateMgr block.
+			restartAttempts = currentAttempts + 1
+			// Override the transition to Queued so the TaskAction stays non-terminal.
+			transition = pluginsCore.DoTransition(pluginsCore.PhaseInfoQueued(time.Now(), pluginsCore.DefaultPhaseVersion,
+				fmt.Sprintf("restarting task (attempt %d/%d)", currentAttempts+1, maxAttempts)))
```
pingsutw: use logger?

executor/pkg/plugin/task_exec_context.go
AdilFayyaz: attempt ++ would offset the values incorrectly.
Here's what's happening here - if `Status.Attempts` = 0, attempt = 1
if `Status.Attempts` = 1, attempt = 1, if `Status.Attempts` = 2, attempt = 2 ... and so on. 
When attempt = 0 it means it is not set yet.

`Status.Attempts` is incremented to 2 and written to k8s before the retry reconcile runs, so a retry always enters `NewTaskExecutionContext` with Status.Attempts >= 2. The value 1 is only ever seen during the first run, so no collision is possible.

actions/k8s/client.go
AdilFayyaz: Following the same convention from my comment above^^

executor/pkg/plugin/task_exec_context.go
pingsutw: got it, could you add a comment like `// if attempts is not set, default to 1`

actions/k8s/client.go
pingsutw: same, add a comment

executor/pkg/plugin/task_exec_context.go
pingsutw: thanks for the explanation

PR #7094, opened 2026-03-26
[V2] Pass run context to subaction + return in GetRunDetails

machichima:

- Subaction should also get the same run context from its parent.
- To update the Env vars when re-run a task in UI, we need to return env vars through `GetRunDetails`

- Pass run context down to child actions
- Return RunSpec in `GetRunDetails` API

## How was this patch tested?

- Run sandbox and ensure 1) run context appears in subactions; 2) We can see env var in the re-run page

- subaction TaskAction CR:

image

***

- subaction pod: image

***

- UI re-run page image

### Labels

## Docs link

- `main` 
  - \#6583
    - **\[V2] Pass run context to subaction + return in GetRunDetails** :point\_left:

### DIFF
--- a/actions/k8s/client.go
+++ b/actions/k8s/client.go
@@ -15,6 +15,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	executorv1 "github.com/flyteorg/flyte/v2/executor/api/v1"
+	"github.com/flyteorg/flyte/v2/executor/pkg/plugin"
 	"github.com/flyteorg/flyte/v2/flytestdlib/fastcheck"
 	k8sutil "github.com/flyteorg/flyte/v2/flytestdlib/k8s"
 	"github.com/flyteorg/flyte/v2/flytestdlib/logger"
@@ -429,7 +430,7 @@ func (c *ActionsClient) handleWatchEvent(ctx context.Context, event watch.Event)
 		ParentActionName: parentName,
 		StateJSON:        taskAction.Status.StateJSON,
 		Phase:            GetPhaseFromConditions(taskAction),
-		OutputUri:        buildOutputUri(taskAction),
+		OutputUri:        buildOutputUri(ctx, taskAction),
 		IsDeleted:        event.Type == watch.Deleted,
 		TaskType:         taskAction.Spec.TaskType,
 		ShortName:        shortName,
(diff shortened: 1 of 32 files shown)


actions/k8s/client.go
```diff
@@ -637,6 +643,50 @@ func applyRunSpecToTaskAction(taskAction *executorv1.TaskAction, runSpec *task.R
 	}
 }
 
+func inheritRunContextFromParentTaskAction(taskAction *executorv1.TaskAction, parentTaskAction *executorv1.TaskAction) {
+	if taskAction == nil || parentTaskAction == nil {
+		return
+	}
+	taskAction.Spec.EnvVars = cloneStringMap(parentTaskAction.Spec.EnvVars)
```
popojk, line 681: Should we append parent action's ENVVAR into child action's ENVVAR instead of overwrite? Child action probably has it's own ENVVAR

runs/service/run_service.go
```diff
@@ -1308,6 +1308,22 @@ func extractRunSpec(specProto []byte) *task.RunSpec {
...
 }
 
+func extractRunSpecFromActionModel(action *models.Action) *task.RunSpec {
+	if action == nil {
+		return nil
+	}
+
+	if len(action.RunSpec) > 0 {
+		spec := &task.RunSpec{}
+		if err := proto.Unmarshal(action.RunSpec, spec); err == nil {
```
popojk, line 1289: Curious why do we not handling the unmarshal error? Don't we need to log it or return it?

actions/k8s/client.go
machichima: In the current implementation, `TaskAction.Spec.EnvVars` is treated as run-scoped context projected from RunSpec, not per-subaction overrides.

In flyte-sdk, when enqueue a subaction, we do not pass any `RunSpec` that can override parent action's context setting

https://github.com/flyteorg/flyte-sdk/blob/ef9eda5b55102c850b82e14deb530d7a95dae6c1/src/flyte/_internal/controllers/remote/_core.py#L381-L392

runs/service/run_service.go
machichima: Yes, I think we should log here. Added in https://github.com/flyteorg/flyte/pull/7094/commits/75290ffd7402dd9e1fc0effb26a5014f0a7a2b19

PR #7096, opened 2026-03-26
Fix: Make flytectl hydrate pod template spec in task node overrides

fg91:

```py
@workflow
def wf():
    some_task().with_overrides(
        pod_template=PodTemplate(
            primary_container_name="primary",
            pod_spec=client.V1PodSpec(
                containers=[],
                node_selector={"foo", "bar"}
            ),
        )
    )
```

When running this workflow with `pyflyte run --remote`, it works. With `pyflyte package` and `flytectl register`, it fails because the code tarball bucket uri isn't templated in the pod container args:

```yaml
  - args:
    - pyflyte-fast-execute
    - --additional-distribution
    - '{{ .remote_package_path }}'
```

The templating happens during what `flytectl register` calls "hydration". It is done correctly for tasks specs but it is just omitted for task node overrides. This PR fixes this.

## How was this patch tested?

* Added unit test.
* Tested that the example workflow above succeeds in a cluster when registered with the fixed `flytectl`.

### DIFF
--- a/flytectl/cmd/register/register_util.go
+++ b/flytectl/cmd/register/register_util.go
@@ -35,6 +35,7 @@ import (
 	"github.com/flyteorg/flyte/flytestdlib/utils"
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
+	structpb "github.com/golang/protobuf/ptypes/struct"
 	"github.com/google/go-github/v42/github"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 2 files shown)


flytectl/cmd/register/register_util.go
```diff
@@ -301,6 +315,29 @@ func hydrateIdentifier(identifier *core.Identifier, version string, force bool)
 	}
 }
 
+// hydrateFastRegistrationArgs substitutes fast-registration template placeholders
```
fg91, line 318: Just moved here out of `hydrateTaskSpec`.

EngHabu (approved):

PR #7098, opened 2026-03-26
fix: convert CoPilot timeout to seconds for terminationGracePeriodSeconds

### LINKED ISSUE #7097: [BUG] CoPilot sets terminationGracePeriodSeconds in nanoseconds instead of seconds, causing pods to hang in Terminating state
## Describe the bug

In [`flyteplugins/go/tasks/pluginmachinery/flytek8s/copilot.go`](https://github.com/flyteorg/flyte/blob/master/flyteplugins/go/tasks/pluginmachinery/flytek8s/copilot.go), the CoPilot timeout is assigned directly to `TerminationGracePeriodSeconds` without converting from nanoseconds to seconds:

```go
coPilotPod.TerminationGracePeriodSeconds = (*int64)(&cfg.Timeout.Duration)
```

`cfg.Timeout.Duration` is a `time.Duration`, which stores nanoseconds as `int64`. Kubernetes expects `terminationGracePeriodSeconds` in seconds.

For example, a 1-hour copilot timeout produces `terminationGracePeriodSeconds: 3600000000000`.

## Expected behavior

The duration should be converted to seconds before assignment:

```go
seconds := int64(cfg.Timeout.Duration.Seconds())
coPilotPod.TerminationGracePeriodSeconds = &seconds
```

## How to reproduce

1. Run a Flyte task with CoPilot enabled (i.e., with output handling) which ignores SIGTERM
2. Cancel or timeout the execution
3. Observe the pod enters `Terminating` state and never completes termination until manual force-deletion
4. Inspect the pod: `kubectl get pod <name> -o jsonpath='{.spec.terminationGracePeriodSeconds}'` returns `3600000000000`
5. Inspect the deletion timestamp: it will be decades/centuries in the future

## Environment

- **FlytePropeller version:** v1.16.3 (also present on `master`/v2.0.9)

mitja-kleider:
## Summary

- Fix unit mismatch in `AddCoPilotToPod`: `cfg.Timeout.Duration` is a `time.Duration` (nanoseconds) but `TerminationGracePeriodSeconds` expects seconds
- Previously a 1h timeout produced `terminationGracePeriodSeconds: 3600000000000` (~114,000 years), causing pods to hang in Terminating state indefinitely

Fixes #7097

### DIFF
--- a/flyteplugins/go/tasks/pluginmachinery/flytek8s/copilot.go
+++ b/flyteplugins/go/tasks/pluginmachinery/flytek8s/copilot.go
@@ -274,7 +274,8 @@ func AddCoPilotToPod(ctx context.Context, cfg config.FlyteCoPilotConfig, coPilot
 			// Let the sidecar container start before the downloader; it will ensure the signal watcher is started before the main container finishes.
 			coPilotPod.InitContainers = append([]v1.Container{sidecar}, coPilotPod.InitContainers...)
 
-			coPilotPod.TerminationGracePeriodSeconds = (*int64)(&cfg.Timeout.Duration)
+			terminationGracePeriodSeconds := int64(cfg.Timeout.Duration.Seconds())
+		coPilotPod.TerminationGracePeriodSeconds = &terminationGracePeriodSeconds
 		}
 	}
 


Sovietaced:
Is it possible to add a unit test for this?

Sovietaced:
Fixed in https://github.com/flyteorg/flyte/pull/7183

PR #7107, opened 2026-03-27
Use dedicated connection for NOTIFY to avoid pool contention

pingsutw:
# TL;DR

Fix slow (\~1s) PostgreSQL NOTIFY calls by using a dedicated connection instead of competing for GORM's shared connection pool.

## Type

- [x] Bug Fix
- [ ] Feature
- [ ] Plugin

## Are all requirements met?

- [x] Code completed
- [x] Smoke tested
- [ ] Unit tests added

## Complete description

`NOTIFY action_updates` and `NOTIFY run_updates` were executed via `r.db.WithContext(ctx).Exec(...)`, which acquires a connection from GORM's shared pool on every call. Under concurrent load (many action updates happening simultaneously), this caused \~1s latency per NOTIFY due to pool contention — the NOTIFY itself is microseconds on PG, but waiting for a free connection dominated.

**Fix:**

- Added two buffered channels (`actionNotifyCh`, `runNotifyCh`, capacity 256) so callers (`notifyActionUpdate`, `notifyRunUpdate`) do a non-blocking channel send and return immediately.
- A single goroutine (`startNotifyLoop`) holds a dedicated `*sql.Conn` and drains both channels, completely bypassing the GORM pool.
- **Batching**: when multiple payloads queue up (common under load), they are drained and combined into a single `SELECT pg_notify(...), pg_notify(...)` round-trip, making it *more* efficient under high concurrency.
- Auto-reconnects on connection errors.

## Tracking Issue

*NA*

## Follow-up issue

*NA*

- `main` 
  - \#6583
    - **Use dedicated connection for NOTIFY to avoid pool contention** :point\_left:

### DIFF
--- a/runs/repository/impl/action.go
+++ b/runs/repository/impl/action.go
@@ -1201,7 +1201,7 @@ func (r *actionRepo) notifyActionUpdate(ctx context.Context, actionID *common.Ac
 	select {
 	case r.actionNotifyCh <- payload:
 	case <-ctx.Done():
-		logger.Warnf(ctx, "Action NOTIFY send cancelled for %s: %v", payload, ctx.Err())
+		logger.Errorf(ctx, "Action NOTIFY send cancelled for %s: %v", payload, ctx.Err())
 	}
 }
 


runs/repository/impl/action.go
```diff
@@ -1060,7 +1077,83 @@ func (r *actionRepo) ListRootActions(ctx context.Context, org, project, domain s
...
+// the error is logged and the notification is skipped.
+func (r *actionRepo) runNotifyLoop(sqlDB *sql.DB, conn *sql.Conn) {
+	defer conn.Close()
+
+	reconnect := func() {
+		conn.Close()
+		var err error
+		conn, err = sqlDB.Conn(context.Background())
+		if err != nil {
+			logger.Errorf(context.Background(), "Failed to re-acquire NOTIFY connection: %v", err)
```
AdilFayyaz, line 1139: Shouldn't we return here if the reconnect error is logged because the conn will be nil?

runs/repository/impl/action.go
```diff
@@ -1060,7 +1077,83 @@ func (r *actionRepo) ListRootActions(ctx context.Context, org, project, domain s
...
+		}
+	}
+
+	// drainAndExec collects all immediately available payloads from a channel
+	// and sends them along with the initial payload in a single batched SQL call.
+	drainAndExec := func(channel, firstPayload string, ch <-chan string) {
+		var b strings.Builder
+		b.WriteString(fmt.Sprintf("SELECT pg_notify('%s', '%s')", channel, firstPayload))
```
AdilFayyaz: Potential SQL Injection here. We can try to parameterize, and also across other calls as well

PR #7110, opened 2026-03-29
[v2][Service] Change the return of the app and trigger service endpoint

SZL741023:

The endpoint app and trigger service are not implemented

Change the return of create function in app and trigger endpoint service.

## How was this patch tested?

These two endpoint service don't have test function, just use command `go build ./runs/...` to confirm that no failure occured.

### Labels

## Docs link

- `main` 
  - \#6583
    - **\[v2]\[Service] Change the return of the app and trigger service endpoint** :point\_left:

### DIFF
--- a/runs/service/run_service.go
+++ b/runs/service/run_service.go
@@ -33,17 +33,17 @@ import (
 
 // RunService implements the RunServiceHandler interface
 type RunService struct {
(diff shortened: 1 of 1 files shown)


machichima:
Tested locally and can see not implemented errors:

image

image

PR #7119, opened 2026-03-30
perf: parallel worker pool for TaskAction watch event processing

AdilFayyaz:

Watch events for TaskActions were processed serially in the watch loop, causing status update delays under load. This adds a configurable worker pool with consistent sharding to parallelize processing while preserving per-resource ordering.

- Added WatchWorkers config field (default: 4) controlling the number of parallel event-processing goroutines.
- ActionsClient now owns a pool of worker channels; StartWatching spawns N workers on init.
- dispatchEvent routes events to workers via FNV-32a hash of TaskAction.Name, ensuring all events for the same resource go to the same worker (ordering preserved).
- Full channels drop events with a warning log rather than blocking the watch loop.

## How was this patch tested?

- Run new unit tests in actions/k8s/client\_test.go:
  - TestDispatchEvent\_ConsistentSharding — same name always routes to the same shard
  - TestDispatchEvent\_DifferentNamesCanLandOnDifferentShards — events spread across shards
  - TestDispatchEvent\_DropsErrorEvents / DropsNonTaskActionObjects — invalid events are discarded
  - TestDispatchEvent\_FullChannelDropsWithoutBlocking — full channel doesn't block
  - TestWorker\_ExitsOnStopCh / ExitsOnContextCancel — worker lifecycle

## Docs link

- `main` 
  - \#6583
    - **perf: parallel worker pool for TaskAction watch event processing** :point\_left:

### DIFF
--- a/actions/k8s/client_test.go
+++ b/actions/k8s/client_test.go
@@ -572,14 +572,28 @@ func TestDispatchEvent_DropsNonTaskActionObjects(t *testing.T) {
 	}
 }
 
-func TestDispatchEvent_FullChannelDropsWithoutBlocking(t *testing.T) {
+func TestDispatchEvent_FullChannelBlocks(t *testing.T) {
 	c := newWorkerTestClient(1, 1) // single worker, capacity 1
 	event := newTaskActionEvent("run1-action1", watch.Modified)
 
 	c.dispatchEvent(context.Background(), event) // fills the channel
-	c.dispatchEvent(context.Background(), event) // must drop, not block
 
-	assert.Equal(t, 1, len(c.workerChs[0]))
+	// Second dispatch must block because the channel is full.
+	// Run it in a goroutine and verify it unblocks once the channel is drained.
+	done := make(chan struct{})
+	go func() {
+		c.dispatchEvent(context.Background(), event)
+		close(done)
+	}()
+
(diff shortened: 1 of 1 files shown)


actions/k8s/client.go
```diff
@@ -386,11 +404,54 @@ func (c *ActionsClient) doWatch(ctx context.Context) error {
...
+		logger.Warnf(ctx, "received non-TaskAction object in watch event: %T", event.Object)
+		return
+	}
+	h := fnv.New32a()
+	_, _ = h.Write([]byte(ta.Name)) // FNV Write never returns an error
+	shard := h.Sum32() % uint32(c.numWorkers)
+	select {
+	case c.workerChs[shard] <- event:
+	default:
+		logger.Warnf(ctx, "worker %d channel full, dropping event for TaskAction %s", shard, ta.Name)
```
pingsutw: If we drop the event, we won't see the status update on UI. I think it's okay to remove select loop here and block the [loop](https://github.com/flyteorg/flyte/pull/7119/files#diff-6a117c32ff7d78b4638d24c944659b96756a92a3548d77eb13978da8a9ba93f0R358-R410) 
```suggestion
	c.workerChs[shard] <- event
```

PR #7120, opened 2026-03-30
fix(runs): allow phase update after terminal state for retry

pingsutw:
## Summary

- Remove the phase-forward-only guard (`phase <= ?`) from `UpdateActionPhase` so that actions in terminal states (e.g. FAILED) can transition back to QUEUED when retried.
- Update tests to validate the retry transition (FAILED → QUEUED).

## Test plan

- [x] Updated `TestUpdateActionPhase_AllowsRetryTransition` to verify FAILED → QUEUED transition succeeds
- [x] All existing action tests pass locally

* `main` 
  - \#6583
    - **fix(runs): allow phase update after terminal state for retry** :point\_left:

### DIFF
--- a/runs/repository/impl/action.go
+++ b/runs/repository/impl/action.go
@@ -512,10 +512,16 @@ func (r *actionRepo) UpdateActionPhase(
 		}
 	}
 
+	// Allow forward phase transitions (phase <= new) and retries from
+	// retryable terminal states (FAILED, TIMED_OUT) back to earlier phases.
+	retryablePhases := []int32{
+		int32(common.ActionPhase_ACTION_PHASE_FAILED),
+		int32(common.ActionPhase_ACTION_PHASE_TIMED_OUT),
+	}
 	result := r.db.WithContext(ctx).
 		Model(&models.Action{}).
-		Where("org = ? AND project = ? AND domain = ? AND run_name = ? AND name = ?",
-			actionID.Run.Org, actionID.Run.Project, actionID.Run.Domain, actionID.Run.Name, actionID.Name).
+		Where("org = ? AND project = ? AND domain = ? AND run_name = ? AND name = ? AND (phase <= ? OR phase IN ?)",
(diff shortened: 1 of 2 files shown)


runs/repository/impl/action.go
```diff
@@ -512,11 +512,10 @@ func (r *actionRepo) UpdateActionPhase(
 		}
 	}
 
-	// Only move the phase forward or re-apply the same phase — never downgrade.
 	result := r.db.WithContext(ctx).
 		Model(&models.Action{}).
-		Where("org = ? AND project = ? AND domain = ? AND run_name = ? AND name = ? AND phase <= ?",
-			actionID.Run.Org, actionID.Run.Project, actionID.Run.Domain, actionID.Run.Name, actionID.Name, phase).
+		Where("org = ? AND project = ? AND domain = ? AND run_name = ? AND name = ?",
+			actionID.Run.Org, actionID.Run.Project, actionID.Run.Domain, actionID.Run.Name, actionID.Name).
```
AdilFayyaz: Doesn't a complete removal of the phase filter open doors to unintended backward transitions. E.g., we could now move from Running to Queued or Succeeded to Queued. 
Maybe we should have a conditional phase query that explicitly allows phase update after terminal state.

runs/repository/impl/action_test.go
```diff
@@ -186,26 +186,28 @@ func TestUpdateActionPhase_PhaseGuard(t *testing.T) {
 	}, nil)
 	require.NoError(t, err)
 
-	// Move to RUNNING
```
AdilFayyaz, line 189: The existing test was checking if RUNNING -> QUEUED fails, which is no longer being tested with the updated test. Don't we still expect this to fail?

runs/repository/impl/action.go
pingsutw: > Maybe we should have a conditional phase query that explicitly allows phase update after terminal state

fixed. good call, thanks

PR #7122, opened 2026-03-31
fix(runs): fix B-tree index contention on actions table

AdilFayyaz:

`UpdateActionPhase` writes `updated_at/ended_at` as `NOW()` on every call, causing B-tree right-hand page contention on those indexes under concurrent load (200ms+ lock-wait). The SQLite polling path in `WatchActionUpdates` also lacked a `run_name filter`, returning updates across all runs in a project instead of just the target run.

- Remove `idx_actions_updated` and `idx_actions_ended` to eliminate right-side B-tree hotspot contention on monotonically-increasing timestamp columns
- Replace four separate single-column indexes `(idx_actions_org/project/domain/run_name)` with one composite `idx_actions_run_lookup(org, project, domain, run_name)` to serve run-scoped queries with a single selective index scan
- Fix `WatchActionUpdates` SQLite polling query to include `run_name` in the WHERE clause, scoping results to the specific run
- Add `migration 20260330_fix_action_indexes` to drop the 6 stale indexes and trigger `AutoMigrate` to create the composite replacement

## How was this patch tested?

- ran two psql queries to verify that the migration and the indexes were created correctly

1. `Select id from migrations`
   - Found the id: 20260330\_fix\_action\_indexes
2. `SELECT indexname FROM pg_indexes WHERE tablename = 'actions' ORDER BY indexname;`
   - idx\_actions\_run\_lookup is present (new composite index)
   - idx\_actions\_updated, idx\_actions\_ended, idx\_actions\_org, idx\_actions\_project, idx\_actions\_domain, idx\_actions\_run\_name are all gone
3. ran `hello.py` from the sdk on the sandbox as a smoke test

## Docs link

- `main` 
  - \#6583
    - **fix(runs): fix B-tree index contention on actions table** :point\_left:

### DIFF
--- a/runs/migrations/migrations.go
+++ b/runs/migrations/migrations.go
@@ -21,7 +21,6 @@ var AllModels = []interface{}{
 }
 
 const MigrationIDInitSchema = "20260327_runs_init_schema"
-const MigrationIDFixActionIndexes = "20260330_fix_action_indexes"
 
 var RunsMigrations = []*gormigrate.Migration{
 	{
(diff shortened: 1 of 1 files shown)


runs/migrations/migrations.go
```diff
@@ -33,6 +34,42 @@ var RunsMigrations = []*gormigrate.Migration{
...
+// migrateFixActionIndexes drops high-contention indexes and creates replacements.
+//
+// idx_actions_updated and idx_actions_ended caused B-tree right-hand page contention:
+// UpdateActionPhase writes updated_at/ended_at ≈ NOW() on every call, so all concurrent
+// writers compete for the same rightmost leaf page, producing 200ms+ lock-wait latency.
+//
+// idx_actions_org/project/domain/run_name are replaced by the composite
+// idx_actions_run_lookup(org, project, domain, run_name) which serves ListActions,
+// AbortAction, MarkAbortAttempt, and ClearAbortRequest with a single selective index.
+func migrateFixActionIndexes(db *gorm.DB) error {
```
pingsutw: It's okay to remove the code here (directly update the db schema) since no one is running flyte 2 in production for now.

PR #7126, opened 2026-03-31
[CON2-143] Add provider based email templates for notifications

pmahindrakar-oss:

The shared notification email contract currently supports only inline email content such as `subject`, `html_template`, and `text_template`.

Some notification producers need to use provider-managed email templates identified by a template ID and a set of dynamic template parameters.
The current contract does not model that use case cleanly.

This PR extends `flyteidl2.notification.EmailDeliveryTemplate` to support provider-backed email templates in a structured way.

Changes in this PR:

- Added a new `ProviderEmailTemplate` message with:
  - `template_id`
  - `template_data`
- Added `provider_template` to `EmailDeliveryTemplate`
- Kept the existing inline email fields intact for backward compatibility
- Added validation so an email template is either:
  - inline content (`subject` + `html_template`), or
  - a provider template (`provider_template`)
- Regenerated checked-in generated artifacts

This preserves existing inline-email behavior while allowing providers to use externally managed templates with dynamic data.

## How was this patch tested?

This change was tested by:

- running `buf lint --path flyteidl2/notification/definition.proto`
- regenerating the checked-in artifacts and verifying the generated outputs

### Labels

`changed`

### Setup process

No special setup required beyond the normal protobuf generation workflow.

### Screenshots

N/A

## Related PRs

N/A

## Docs link

N/A

- `main` 
  - \#6583
    - **\[CON2-143] Add provider based email templates for notifications** :point\_left:


flyteidl2/notification/definition.proto
```diff
@@ -81,16 +81,44 @@ message EmailRecipient {
...
+    max_pairs: 100
+  }];
+}
+
 message EmailDeliveryTemplate {
-  string subject = 1 [(buf.validate.field).string.min_len = 1]; // this is a Go template that may contain placeholders
+  option (buf.validate.message).cel = {
+    id: "inline_email_or_provider_template_required"
+    message: "either inline email subject/html_template or provider_template must be set"
+    expression: "(has(this.provider_template) && size(this.subject) == 0 && size(this.html_template) == 0 && size(this.text_template) == 0) || (!has(this.provider_template) && size(this.subject) > 0 && size(this.html_template) > 0)"
```
iaroslav-ciupin: we don't have notifications on prod yet, so its ok to do a breaking change and maybe introduce a `oneof` of InlineEmail and ProviderEmailTemplate to simplify this validation and make structure more self-explanatory

flyteidl2/notification/definition.proto
```diff
@@ -81,16 +81,44 @@ message EmailRecipient {
...
+    keys: {
+      string: {
+        min_len: 1
+        max_len: 100
+      }
+    }
+    values: {
+      string: {max_len: 4000}
+    }
+    max_pairs: 100
```
katrogan, line 116: out of curiosity, why this limitation?

flyteidl2/notification/definition.proto
```diff
@@ -81,16 +81,44 @@ message EmailRecipient {
   string address = 2;
 }
 
+message ProviderEmailTemplate {
+  string template_id = 1 [(buf.validate.field).string = {
+    min_len: 1
+    max_len: 200
+  }];
+
+  map<string, string> template_data = 2 [(buf.validate.field).map = {
```
katrogan, line 106: nit: can we add a comment for how this data is used?

flyteidl2/notification/definition.proto
pmahindrakar-oss: Done cleaned it up , so we dont have to support backward compatibility

flyteidl2/notification/definition.proto
pmahindrakar-oss: This is just a defensive upper bound. We expect `template_data` to be a relatively small set of dynamic template variables, so `100` felt permissive enough without leaving the payload unbounded. Happy to relax/remove it if that seems too arbitrary

flyteidl2/notification/definition.proto
pmahindrakar-oss: Added comments

flyteidl2/notification/definition.proto
```diff
@@ -81,16 +81,53 @@ message EmailRecipient {
...
 
-  repeated EmailRecipient cc = 3;
+  repeated EmailRecipient cc = 2;
 
-  repeated EmailRecipient bcc = 4;
+  repeated EmailRecipient bcc = 3;
 
-  string html_template = 5 [(buf.validate.field).string.min_len = 1]; // this is a Go template that may contain placeholders
+  // Content may be provided either inline or via an external provider template.
+  oneof content {
```
iaroslav-ciupin, line 128: can you add a buf validate annotation to make sure at least one is set?

PR #7151, opened 2026-04-03
[V2] Implement trace action

popojk:
## Tracking issue

n/a

Make trace action works

1. Handle trace action in ActionService Enqueue API
2. In InternalRunService RecordAction API, handle task and trace action separately
3. Refactor RunService action repository CreateAction implementation
4. Change the default sort field from id to created\_at in action repository ListAction

### Labels

## Docs link

- `main` 
  - \#6583
    - **\[V2] Implement trace action** :point\_left:


runs/repository/impl/action.go
```diff
@@ -478,14 +426,16 @@ func (r *actionRepo) ListActions(ctx context.Context, runID *common.RunIdentifie
...
-		query = query.Where("id > ?", token)
+		if t, err := time.Parse(time.RFC3339Nano, token); err == nil {
+			query = query.Where("created_at > ?", t)
+		}
 	}
 
 	var actions []*models.Action
 	result := query.
-		Order("id ASC").
+		Order("created_at ASC").
```
popojk, line 454: I change the default sorting field to created_at because trace action use the start time generated in sdk runtime as created_at time and stored in db, and this change makes sure that the ordering is correct.

runs/repository/impl/action.go
```diff
@@ -478,14 +426,16 @@ func (r *actionRepo) ListActions(ctx context.Context, runID *common.RunIdentifie
 		Where("org = ? AND project = ? AND domain = ? AND run_name = ?",
 			runID.Org, runID.Project, runID.Domain, runID.Name)
 
-	// Apply pagination token
+	// Apply pagination token (encoded as RFC3339Nano created_at cursor)
 	if token != "" {
-		query = query.Where("id > ?", token)
+		if t, err := time.Parse(time.RFC3339Nano, token); err == nil {
```
popojk, line 447: Use created_at field to do pagination as it is the default sorting field

runs/service/internal_run_service.go
```diff
@@ -72,21 +83,37 @@ func (s *RunService) recordSingleAction(ctx context.Context, req *workflow.Recor
...
+	// Build RunInfo for storage URIs and task spec digest.
 	info := &workflow.RunInfo{
 		InputsUri: req.GetInputUri(),
 	}
 
-	logger.Infof(ctx, "RecordAction: action=%s parent=%s taskId=%v taskSpec=%v",
-		actionID.GetName(), req.GetParent(), req.GetTask().GetId(), req.GetTask().GetSpec() != nil)
-
 	switch v := req.GetSpec().(type) {
 	case *workflow.RecordActionRequest_Task:
```
popojk, line 93: Refactor this piece of code to handle task action and trace action separately because the required field in db model is defferent

actions/k8s/client.go
```diff
@@ -90,77 +90,122 @@ func (c *ActionsClient) Enqueue(ctx context.Context, action *actions.Action, run
...
+		if _, err := c.runClient.UpdateActionStatus(ctx, connect.NewRequest(statusReq)); err != nil {
+			logger.Warnf(ctx, "Failed to mark trace action %s as succeeded: %v", actionID.Name, err)
+		}
+
+		// Notify SDK runtime informer
+		c.notifySubscribers(ctx, &ActionUpdate{
+			ActionID:         actionID,
+			ParentActionName: action.GetParentActionName(),
+			Phase:            common.ActionPhase_ACTION_PHASE_SUCCEEDED,
+			OutputUri:        action.GetTrace().GetOutputs().GetOutputUri(),
```
pingsutw, line 201: Do we need to notify the SDK? Will the task get stuck if we don’t notify it here

popojk: Yes the task will get stock because sdk runtime informer also watch for trace action, so we need to do it here

pingsutw:
### Code review

Found 4 issues:

1. **No unit tests for new trace action code path in `client.go`** (global CLAUDE.md says "Add unit tests based on the diff of this branch and master or main branch")

The new `case *actions.Action_Trace` branch in `Enqueue()` (~40 lines) calls `RecordAction`, `UpdateActionStatus`, and `notifySubscribers` but has no corresponding test coverage.

https://github.com/flyteorg/flyte/blob/36a5995e2f540e9d0ff9d84fe7449a4202d2e72c/actions/k8s/client.go#L166-L207

2. **`UpdateActionStatus` overwrites `Attempts` from 1 to 0 for trace actions** (bug: zero-valued proto field)

`recordSingleAction` sets `action.Attempts = 1` before insert. Then `Enqueue` calls `UpdateActionStatus` with an `ActionStatus` that only sets `Phase` -- `Attempts` defaults to `0`. `UpdateActionPhase` unconditionally writes `attempts` to the DB, overwriting the correct value.

https://github.com/flyteorg/flyte/blob/36a5995e2f540e9d0ff9d84fe7449a4202d2e72c/actions/k8s/client.go#L189-L197

3. **`notifySubscribers` called unconditionally even when `UpdateActionStatus` fails** (bug: DB/subscriber state inconsistency)

If `UpdateActionStatus` fails for the trace action, the error is logged as a warning but `notifySubscribers` still fires with `ACTION_PHASE_SUCCEEDED`. The parent unblocks thinking the trace succeeded, but the DB phase was never updated. The task action path avoids this because notifications go through the K8s watch loop, which only fires after the status is persisted.

https://github.com/flyteorg/flyte/blob/36a5995e2f540e9d0ff9d84fe7449a4202d2e72c/actions/k8s/client.go#L195-L207

4. **Missing `default` case in `Enqueue` action type switch** (bug: silent no-op for unknown action types)

The switch on `action.Spec.(type)` handles `Action_Task` and `Action_Trace` but has no `default` case. An unknown action type silently returns `nil` (success) without logging or returning an error.

https://github.com/flyteorg/flyte/blob/36a5995e2f540e9d0ff9d84fe7449a4202d2e72c/actions/k8s/client.go#L93-L209

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #7152, opened 2026-04-03
Implement CreateRun and UploadInputs APIs

katrogan:

Implements <https://github.com/flyteorg/flyte/pull/7139> in OSS

Decouples uploading inputs from calling CreateRun and switches to support CreateRun with inputs passed by URI

## How was this patch tested?

Tested locally by running devbox. The UI doesn't seem to like orgs even though they're required by the CRD so I verified the hello example locally like so

```
 Bash(curl -s -H "Content-Type: application/json" \
        -d '{"runId":{"org":"testorg","project":"flytesnacks","domain":"development","name":"rf89z2m7s7q47pfnj6gw"}}' \…)
  ⎿  phase: ACTION_PHASE_SUCCEEDED
     start: 2026-04-03T04:17:17.656266Z
     end: 2026-04-03T04:17:27.750213Z
```

### Labels

## Docs link

- `main` 
  - \#6583
    - **Implement CreateRun and UploadInputs APIs** :point\_left:


runs/service/run_service.go
```diff
@@ -177,14 +176,31 @@ func (s *RunService) CreateRun(
...
+		if err := s.dataStore.ReadProtobuf(ctx, offloadedRef, inputs); err != nil {
+			logger.Errorf(ctx, "Failed to read offloaded inputs from %s: %v", offloadedRef, err)
+			return nil, connect.NewError(connect.CodeNotFound, fmt.Errorf("failed to read offloaded inputs: %w", err))
+		}
+		logger.Infof(ctx, "Read offloaded inputs from %s", offloadedRef)
+	default:
+		inputs = request.GetInputs()
+	}
+
+	inputs = fillDefaultInputs(inputs, taskSpec.GetDefaultInputs())
```
katrogan: not sure how necessary this is for all cases, it's not ideal we read the offloaded inputs and then reupload them again below

pingsutw: We use the inputs to generate the cache key at line 216. is it possible that dataproxy generates the same hash here, so we don't need to read the offload inputs?

pingsutw: or I think we can only read the inputs only when cache is enabled at least.

pingsutw (approved):

runs/service/run_service.go
pingsutw: look at the sdk code, i think it's possible. we can pass a task template to UploadInputRequest if we want

pingsutw: task template is in the task spec. https://github.com/flyteorg/flyte-sdk/blob/aad83a1bb9aa4690fd09c65858b68417208ff697/src/flyte/_run.py#L401-L407

pingsutw: Never mind—ignore me. I think we just need to update generateCacheKeyForTask to make it the same as the closed-source version

pingsutw: generateCacheKeyForTask(taskTemplate, precompute-hash)

PR #7176, opened 2026-04-09
feat: implement AppService control plane proxy with TTL cache

AdilFayyaz:
Depends on: <https://github.com/flyteorg/flyte/pull/7175>

## Why are the changes needed?

The control plane needs a stable API surface for app management without making direct K8s calls. AppService proxies to InternalAppService (data plane) and adds a TTL cache to reduce cross-plane RPC calls on hot Get paths.

## What changes were proposed in this pull request?

- Added app/service/AppService — cache-aside proxy over InternalAppService; Get reads from cache, mutations invalidate, Watch is a streaming passthrough
- Added app/setup.go — mounts AppService on the control plane mux; in unified mode routes to /internal on the same mux with no network hop
- Restructured config: app/internal/config/InternalAppConfig (data plane) and app/config/AppConfig (control plane, holds CacheTTL + InternalAppServiceURL)
- Removed AppService stub and wiring from runs/ — no longer needed there

## How was this patch tested?

- go test ./app/service/... — 8 unit tests covering cache hit, miss, TTL expiry, invalidation on Create/Update/Delete, List always forwarding, and Watch stream proxying
- go test ./app/internal/service/... — covers InternalAppService RPCs including DesiredState branching, pagination, and Watch initial snapshot
- go test ./app/... — all packages pass

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

## Docs link

- `main` 
  - \#6583
    - **feat: implement AppService control plane proxy with TTL cache** :point\_left:
      - \#7188
        - \#7197

PR #7188, opened 2026-04-09
feat: resilient Watch RPC with reconnect and resourceVersion tracking

AdilFayyaz:
## Tracking issue

Depends on: <https://github.com/flyteorg/flyte/pull/7176>, <https://github.com/flyteorg/flyte/pull/7175>, <https://github.com/flyteorg/flyte/pull/7166>

K8s watches time out every \~5 minutes by default. The previous implementation closed the client's stream silently on disconnect with no reconnect, making Watch unreliable for long-lived connections.

- Replaced the single-shot watch goroutine in AppK8sClient.Watch() with a reconnect loop (watchLoop + drainWatcher) that transparently reopens the K8s watch on unexpected closes or Error events
- Added resourceVersion tracking — extracted from every Added/Modified/Deleted/Bookmark event and passed to the next watch call, ensuring no events are missed or replayed across reconnects
- Added exponential backoff (1s → 2s → 4s → 30s max) between reconnect attempts; backoff resets on any successful event or Bookmark
- K8s Error events are now logged with code/reason/message instead of being silently dropped

## How was this patch tested?

- go test ./app/internal/k8s/... -run TestWatch — 6 new tests covering:\
  channel close reconnect, Error event reconnect, Bookmark RV propagation to next watch call, exponential backoff timing, ctx cancel stops the goroutine,  initial watch error surfaces synchronously
- go test ./app/... — full suite passes with no regressions

## Docs link

- `main` 
  - \#6583
    - **feat: resilient Watch RPC with reconnect and resourceVersion tracking** :point\_left:
      - \#7197

### DIFF
--- a/app/internal/k8s/app_client.go
+++ b/app/internal/k8s/app_client.go
@@ -232,18 +232,17 @@ func (c *AppK8sClient) Watch(ctx context.Context, project, domain, appName strin
 }
 
 // openWatch starts a K8s watch from resourceVersion (empty = watch from now).
+// AllowWatchBookmarks is always set so K8s sends Bookmark events on every session,
+// keeping lastResourceVersion current even when no objects change.
 func (c *AppK8sClient) openWatch(ctx context.Context, ns string, labels map[string]string, resourceVersion string) (k8swatch.Interface, error) {
+	rawOpts := &metav1.ListOptions{AllowWatchBookmarks: true}
+	if resourceVersion != "" {
+		rawOpts.ResourceVersion = resourceVersion
+	}
 	opts := []client.ListOption{
 		client.InNamespace(ns),
 		client.MatchingLabels(labels),
-	}
-	if resourceVersion != "" {
-		opts = append(opts, &client.ListOptions{
-			Raw: &metav1.ListOptions{
-				ResourceVersion:     resourceVersion,
-				AllowWatchBookmarks: true,
-			},
-		})
+		&client.ListOptions{Raw: rawOpts},
 	}
 	watcher, err := c.k8sClient.Watch(ctx, &servingv1.ServiceList{}, opts...)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


app/internal/k8s/app_client.go
```diff
@@ -171,51 +173,185 @@ func (c *AppK8sClient) Delete(ctx context.Context, appID *flyteapp.Identifier) e
...
 		client.MatchingLabels(labels),
-	)
+	}
+	if resourceVersion != "" {
+		opts = append(opts, &client.ListOptions{
+			Raw: &metav1.ListOptions{
+				ResourceVersion:     resourceVersion,
+				AllowWatchBookmarks: true,
+			},
+		})
```
pingsutw, line 245: Don't we still need to bookmark at first call?

app/internal/k8s/app_client.go
```diff
@@ -171,51 +173,185 @@ func (c *AppK8sClient) Delete(ctx context.Context, appID *flyteapp.Identifier) e
...
+		watcher.Stop()
+		state.consecutiveErrors++
+		delay := state.nextBackoff()
+		logger.Warnf(ctx, "KService watch in namespace %s closed unexpectedly (attempt %d); reconnecting in %v",
+			ns, state.consecutiveErrors, delay)
+
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(delay):
```
pingsutw, line 283: Do we need backoff for Kubernetes watch timeouts? After the app service has been running for 1 hour, it always waits for 30 seconds (max(30, 2^60/5)) every 5 minutes. I think we only need backoff for other errors

app/internal/k8s/app_client.go
```diff
@@ -171,51 +173,185 @@ func (c *AppK8sClient) Delete(ctx context.Context, appID *flyteapp.Identifier) e
...
+	for {
+		select {
+		case <-ctx.Done():
+			return false
+		case event, ok := <-watcher.ResultChan():
+			if !ok {
+				return true
+			}
+
+			c.updateResourceVersion(event, state)
```
pingsutw, line 314: Should we update the resource version after we successfully send the response? so the `drainWatcher` will process the same item in the next loop if it failed to send the event in the previous loop.

app/internal/k8s/app_client.go
AdilFayyaz: Thanks for pointing this out, I didn't know this

app/internal/k8s/app_client.go
AdilFayyaz: We do, otherwise it might drop events, adding the fix

app/internal/k8s/app_client.go
AdilFayyaz: True, we might skip an event otherwise

pingsutw (approved):

PR #7192, opened 2026-04-10
[V2] List run accept filter

machichima:

`ListRuns` API currently does not accept filters, we should support it to list runs for tasks / triggers / ...

1. Remove repo impl `ListRuns` and only use `ListActions`, as run is just action with no parent
2. Enable filtering and sorting in `ListActions`
3. Add more filtering functions in `runs/repository/impl/filters.go` to be used in `ListRuns`

## How was this patch tested?

- Run sandbox locally and ensure task / trigger page can only see their runs

Total Runs:

image

Runs for "every\_minute" trigger:

image

Runs for "every\_minute" task:

image

### Labels

## Docs link

- `main` 
  - \#6583
    - **\[V2] List run accept filter** :point\_left:


runs/service/run_service.go
```diff
@@ -1159,27 +1188,20 @@ func (s *RunService) listAndSendAllActions(
...
-		if err != nil {
-			return err
-		}
-
-		if err := s.sendChangedActions(runID, updates, stream); err != nil {
-			return err
-		}
+	batch, err := s.repo.ActionRepo().ListActions(ctx, interfaces.ListResourceInput{
+		Filter: impl.NewRunActionsFilter(runID),
+		Limit:  10000,
```
pingsutw, line 1193: Is it intentional? it was 100

runs/service/run_service.go
```diff
@@ -871,37 +899,38 @@ func (s *RunService) ListActions(
...
 
-	// List actions from database
-	limit := 50
-	if req.Msg.Request != nil && req.Msg.Request.Limit > 0 {
-		limit = int(req.Msg.Request.Limit)
+	listInput, err := impl.NewListResourceInputFromProto(req.Msg.Request, models.ActionColumnsSet)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
 	}
+	listInput.Filter = impl.NewRunActionsFilter(req.Msg.RunId)
```
pingsutw, line 910: It may override the original filter in listInput. should we merge it instead?

runs/service/run_service.go
pingsutw: don't we need pagination

machichima: Fixed in https://github.com/flyteorg/flyte/pull/7192/commits/c6dae13ab338b972d9c83c47823126ee000400f3 to change limit back to 100 and add pagination

runs/service/run_service.go
machichima: Sure! Change to merge in https://github.com/flyteorg/flyte/pull/7192/commits/9c87857a97cbade52e0e1fac91e1eefe25f6a37a

pingsutw (approved):

runs/repository/impl/action.go
```diff
@@ -324,11 +339,12 @@ func (r *actionRepo) ListActions(ctx context.Context, input interfaces.ListResou
 			queryBuilder.WriteString(sp.GetOrderExpr())
 		}
 	} else {
-		queryBuilder.WriteString(" ORDER BY created_at ASC")
+		// Default sorting non-terminal runs at the top, then by most recent start time.
+		queryBuilder.WriteString(" ORDER BY phase ASC, created_at DESC")
```
pingsutw, line 343: Why do we need order by phase here? shouldn't we show the action in chronological order

machichima: I thought we should put non-terminated actions at the top. Sometimes an action can be created earlier but runs longer. In this case I think order by phase can be helpful

pingsutw (approved):

PR #7193, opened 2026-04-10
Fix: Allow toggling dynamic log links for Ray, Dask, Spark plugins

fg91:

Dynamic log links currently work for the pod plugin (normal `PythonFunctionTask`) and for the kubeflow plugin. Currently they cannot be toggled for Ray, Dask or Spark tasks. This PR fixes this.

The so-called `TaskTemplate` that contains the dynamic log link config needs to be included in the `tasklog.Input{...}` based on which the log links are generated.

## How was this patch tested?

Added unit tests.


pingsutw (approved):
Nice, Thank you

flyteplugins/go/tasks/plugins/k8s/dask/dask.go
```diff
@@ -297,13 +297,19 @@ func (p daskResourceHandler) GetTaskPhase(ctx context.Context, pluginContext k8s
 		OccurredAt: &occurredAt,
 	}
 
+	taskTemplate, err := pluginContext.TaskReader().Read(ctx)
```
EngHabu, line 300: This may be an s3 hit if I recall.. do we want to do that every time?

pingsutw: I think it's okay for now since we already do that for PythonFunctionTask. We can have a separate PR to do it.

fg91: True, [here](https://github.com/flyteorg/flyte/blob/334ace089c4d263c328a4680ffcc436fb3269fa0/flyteplugins/go/tasks/plugins/k8s/pod/plugin.go#L146) via [this](https://github.com/flyteorg/flyte/blob/334ace089c4d263c328a4680ffcc436fb3269fa0/flyteplugins/go/tasks/plugins/k8s/pod/plugin.go#L156) line 🤔 

But this is a good point, I will check audit logs next week how many of those requests propeller makes!

EngHabu: ah, good point...

EngHabu (approved):

PR #7197, opened 2026-04-10
feat: path based ingress routing for Flyte apps via Traefik + Knative

AdilFayyaz:

- Deploys each Flyte app with a deterministic public URL at `<base>/<project>/<domain>/<app>/localhost` by\
  creating Traefik IngressRoute + Middleware CRDs alongside each KService
- Adds InternalAppService to the unified manager binary, wiring the data plane app service into the existing HTTP\
  mux
- Injects cluster-internal env vars (minio endpoint, Flyte manager endpoint) into every KService pod so fserve can
  download code bundles and connect back to the manager
- Fixes sandbox-run to correctly forward FLYTE\_DEV to the sub-make

App service wiring (app/, manager/)

- app/config/config.go: Add InternalAppConfig (consolidated from app/internal/config) with ingress and default env
  var fields
- app/internal/config/config.go: Thin type alias to the public config to avoid circular imports
- app/setup.go: Add SetupInternal entry point for registering InternalAppService on the mux
- manager/config/config.go: Add Apps and InternalApps fields to manager Config
- manager/cmd/main.go: Register Knative scheme, wire up SetupInternal + Setup in the right order (internal before\
  external so the proxy can reach /internal/...)

Ingress routing (app/internal/k8s/app\_client.go)

- deployIngress: creates three Traefik resources per app using unstructured.Unstructured (no Traefik Go dependency):
  strip-<name> Middleware (strip path prefix), host-<name> Middleware (rewrite Host header to Knative hostname so\
  Kourier routes correctly), app-<name> IngressRoute
- deleteIngress: cleans up all three resources on app deletion
- publicIngress: returns path-based URL when ingress is enabled, falls back to Knative host-based URL
- buildPodSpec: adds Command, port mapping, resource requirements; injects DefaultEnvVars before user env vars
- kserviceToStatus: derives public URL from the app ID annotation rather than ksvc.Status.URL so the URL is\
  consistent before Knative route readiness

Public URL (app/internal/service/internal\_app\_service.go)

- publicIngress: returns {IngressBaseURL}/{project}/{domain}/{app} when IngressEnabled, falling back to Knative host
  pattern with configurable scheme

Sandbox (charts/, docker/sandbox-bundled/, Makefile)

- charts/flyte-sandbox/templates/proxy/traefik-config.yaml: Add allowExternalNameServices: true — required because\
  Knative creates ExternalName K8s services pointing to Kourier, which Traefik blocks by default
- charts/flyte-sandbox/values.yaml: Add manager.internalApps inline config including defaultEnvVars (minio endpoint + \_U\_EP\_OVERRIDE) so KService pods can reach minio and the Flyte manager
- docker/sandbox-bundled/Makefile: Add setup-knative target (installs Knative Serving + Kourier, configures domain\
  template and domain)
- Makefile: sandbox-run now forwards FLYTE\_DEV and calls setup-knative

Runtime (flytestdlib/app/app.go)

- Add requestGzipDecompressMiddleware: pre-decompresses Content-Encoding: gzip request bodies before the connect-rpc handler sees them, fixing a chunked-transfer + gzip interaction with the Python SDK

## How was this patch tested?

1. make sandbox-run FLYTE\_DEV=True — started the sandbox with Knative installed
2. make -C manager run — ran the manager locally (dev mode)
3. uv run python examples/apps/basic\_app.py — deployed the streamlit-hello-v2 app via the SDK
4. curl -sI <http://localhost:30080/flytesnacks/development/streamlit-hello-v2> — confirmed HTTP/1.1 200 OK
5. Tested via browser as well

## Docs link

- `main` 
  - \#6583
    - **feat: path based ingress routing for Flyte apps via Traefik + Knative** :point\_left:


app/config/config.go
```diff
@@ -20,3 +20,55 @@ func DefaultAppConfig() *AppConfig {
...
+
+	// IngressAppsPort is the port appended to the public app URL (e.g. 30081).
+	// Set to 0 to omit the port when behind a standard 80/443 proxy.
+	IngressAppsPort int `json:"ingressAppsPort" pflag:",Port for app subdomain URLs (0 = omit)"`
+
+	// DefaultEnvVars is a list of environment variables injected into every KService
+	// pod at deploy time, in addition to any env vars specified in the app spec.
+	// Use this to inject cluster-internal endpoints (e.g. _U_EP_OVERRIDE) that app
+	// processes need to connect back to the Flyte manager.
+	DefaultEnvVars []EnvVar `json:"defaultEnvVars" pflag:"-,Default env vars injected into every app pod"`
```
pingsutw: nit: Could we use map[string]string here. same as we do in k8s plugin

app/service/app_service.go
```diff
@@ -18,20 +18,15 @@ import (
 type AppService struct {
 	appconnect.UnimplementedAppServiceHandler
 	internalClient appconnect.AppServiceClient
-	// cache is nil when cacheTTL=0 (caching disabled).
-	cache *expirable.LRU[string, *flyteapp.App]
```
pingsutw, line 22: why do we revert this?

app/internal/k8s/app_client.go
```diff
@@ -451,6 +466,152 @@ func (c *AppK8sClient) List(ctx context.Context, project, domain, appName string
...
+//   - app-<name>    IngressRoute: matches Host header and applies the middleware
+//
+// No-ops when IngressEnabled is false.
+func (c *AppK8sClient) deployIngress(ctx context.Context, app *flyteapp.App) error {
+	if !c.cfg.IngressEnabled {
+		return nil
+	}
+	appID := app.GetMetadata().GetId()
+	ns := appNamespace(appID.GetProject(), appID.GetDomain())
+	ksvcName := kserviceName(appID)
```
pingsutw: Why do we need it? don't we always use k ingress?

app/service/app_service.go
AdilFayyaz: probably missed it in the merge conflict, thanks for pointing it out.

app/internal/k8s/app_client.go
AdilFayyaz: We technically don't, in the future if we're to move from Traefik the it makes sense to have this. But for now, I've removed the flag `IngressEnabled` (and its usages - simplified the logic) since it doesn't add any value.

pingsutw (approved):

PR #7200, opened 2026-04-14
fix: propagate abort phase to sub-actions and fix WatchActionDetails race

AdilFayyaz:

Aborting a parent action left sub-actions and group members stuck on their previous phase in the UI. Two independent gaps — missing DB cascade and a notification race — prevented the ABORTED state from reaching clients.

- AbortAction cascade: recursive CTE marks all non-terminal descendants ABORTED in a single round-trip, inserts abort events, and fires notifyActionUpdate for each
- UpdateActionPhase abort event: inserts into action\_events before notifying so WatchActionDetails phase transitions are always consistent with the action's phase
- K8s informer DeleteFunc: unwraps tombstones and sets ABORTED for cascade-deleted CRs that carry no terminal condition; skips markTerminalStatusRecorded on already-deleted objects
- WatchActionDetails subscription order: goroutine now subscribes before the initial DB fetch, closing the race window where cascade notifications fired before the handler was registered
- AbortRun: returns child identifiers so the reconciler can push all non-terminal children for pod termination, not just the root

## How was this patch tested?

- Abort a parent action with running K8s task sub-actions — confirm all sub-actions immediately show ABORTED in both the sidebar list and detail panel
- Abort a group action — confirm group and all trace children update without requiring a page refresh
- Abort with an active phase filter (e.g. ?status=RUNNING) — confirm sub-actions are updated before being removed from the filtered list (frontend fix in RunStore.ts)
- Abort from CLI against a parent action — confirm children are cascade-aborted in DB even though the reconciler only deletes the parent CR
- Run all unit tests

## Docs link

- `main` 
  - \#6583
    - **fix: propagate abort phase to sub-actions and fix WatchActionDetails race** :point\_left:

### DIFF
--- a/actions/k8s/client.go
+++ b/actions/k8s/client.go
@@ -422,6 +422,18 @@ func (c *ActionsClient) setupInformer(ctx context.Context) error {
 			}
 			c.dispatchEvent(taskAction, watch.Modified)
 		},
+		DeleteFunc: func(obj interface{}) {
+			// The informer may deliver a DeletedFinalStateUnknown tombstone
+			// when a delete event was missed; unwrap it first.
+			if tombstone, ok := obj.(toolscache.DeletedFinalStateUnknown); ok {
+				obj = tombstone.Obj
+			}
+			taskAction, ok := obj.(*executorv1.TaskAction)
+			if !ok {
+				return
+			}
+			c.dispatchEvent(taskAction, watch.Deleted)
+		},
 	})
 	if err != nil {
 		return fmt.Errorf("failed to add TaskAction informer handler: %w", err)
@@ -494,6 +506,14 @@ func buildActionUpdate(ctx context.Context, taskAction *executorv1.TaskAction, e
 		shortName = extractShortNameFromTemplate(taskAction.Spec.TaskTemplate)
 	}
 
+	phase := GetPhaseFromConditions(taskAction)
+	// A TaskAction deleted without a terminal condition was cascade-deleted by K8s
+	// (e.g. its parent was aborted). Treat it as aborted so the DB and UI reflect
+	// the real outcome instead of leaving the phase as UNSPECIFIED.
+	if eventType == watch.Deleted && phase == common.ActionPhase_ACTION_PHASE_UNSPECIFIED {
+		phase = common.ActionPhase_ACTION_PHASE_ABORTED
+	}
+
 	return &ActionUpdate{
 		ActionID: &common.ActionIdentifier{
 			Run: &common.RunIdentifier{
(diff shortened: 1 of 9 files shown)


tasks/lessons.md
```diff
@@ -0,0 +1,19 @@
+# Claude Lessons Learned
```
pingsutw, line 1: remove this file?

runs/repository/impl/action.go
```diff
@@ -136,25 +138,100 @@ func (r *actionRepo) ListRuns(ctx context.Context, req *workflow.ListRunsRequest
...
+	if errors.Is(err, sql.ErrNoRows) {
+		return nil, fmt.Errorf("run not found: %w", sql.ErrNoRows)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("failed to abort run: %w", err)
+	}
+
+	// Cascade abort to non-terminal child actions so the reconciler terminates their pods.
+	// Skip SUCCEEDED and ABORTED rows — they are already done.
+	rows, err := r.db.QueryxContext(ctx,
```
pingsutw, line 170: I think we can combine above query together

actions/k8s/client.go
```diff
@@ -494,6 +506,14 @@ func buildActionUpdate(ctx context.Context, taskAction *executorv1.TaskAction, e
...
 	}
 
+	phase := GetPhaseFromConditions(taskAction)
+	// A TaskAction deleted without a terminal condition was cascade-deleted by K8s
+	// (e.g. its parent was aborted). Treat it as aborted so the DB and UI reflect
+	// the real outcome instead of leaving the phase as UNSPECIFIED.
+	if eventType == watch.Deleted && phase == common.ActionPhase_ACTION_PHASE_UNSPECIFIED {
+		phase = common.ActionPhase_ACTION_PHASE_ABORTED
+	}
+
```
pingsutw, line 516:
```suggestion
	if eventType == watch.Deleted {
		phase = common.ActionPhase_ACTION_PHASE_ABORTED
	}

```

runs/repository/impl/action.go
```diff
@@ -437,27 +522,123 @@ func (r *actionRepo) UpdateActionPhase(
...
+	}
+
 	// Notify action subscribers.
 	r.notifyActionUpdate(ctx, actionID)
 
-	logger.Infof(ctx, "Aborted action: %s", actionID.Name)
+	// Cascade ABORTED to all descendants of this action (recursive: children,
+	// grandchildren, etc.) that are not yet in a terminal phase.
+	// We use a recursive CTE so all depths are handled in a single round-trip.
+	rows, err := r.db.QueryxContext(ctx,
```
pingsutw, line 562: merge above query if possible

actions/k8s/client.go
```diff
@@ -422,6 +422,18 @@ func (c *ActionsClient) setupInformer(ctx context.Context) error {
...
+			// The informer may deliver a DeletedFinalStateUnknown tombstone
+			// when a delete event was missed; unwrap it first.
+			if tombstone, ok := obj.(toolscache.DeletedFinalStateUnknown); ok {
+				obj = tombstone.Obj
+			}
+			taskAction, ok := obj.(*executorv1.TaskAction)
+			if !ok {
+				return
+			}
+			c.dispatchEvent(taskAction, watch.Deleted)
```
pingsutw, line 435: I'm thinking maybe we don't need it. we have inserted event in the run service, right?

AdilFayyaz:
Handled correctly in: https://github.com/flyteorg/flyte/pull/7225

PR #7214, opened 2026-04-15
[V2] Set ReportUri in ActionEvent output references

pingsutw:

The `OutputReferences` proto (`flyteidl2/task/common.proto`) defines both `output_uri` and `report_uri`, but the executor only populated `output_uri` when building `ActionEvent`s. Consumers that need the HTML report URI had no way to locate it.

In `executor/pkg/controller/taskaction_controller.go`, `outputRefs` now also sets `ReportUri` using the convention path `<outputPrefix>/report.html`, alongside the existing `<outputPrefix>/outputs.pb`.

## How was this patch tested?
image

Added a unit test in `executor/pkg/controller/output_refs_test.go` that verifies:

- The report URI ends with `/report.html`.
- The report URI and output URI share the same directory prefix.

```
go test ./executor/pkg/controller/ -run TestOutputRefs -count=1
ok  	github.com/flyteorg/flyte/v2/executor/pkg/controller	0.851s
```

### Labels

- changed

## Stack

- `main` 
  - \#6583
    - **\[V2] Set ReportUri in ActionEvent output references** :point\_left:

### DIFF
--- a/executor/pkg/controller/output_refs_test.go
+++ b/executor/pkg/controller/output_refs_test.go
@@ -70,6 +70,18 @@ func TestOutputRefs(t *testing.T) {
 		assert.Equal(t, "1", parts[len(parts)-1])
 	})
 
+	t.Run("report URI ends with report.html and shares prefix with output URI", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


AdilFayyaz (approved):
LGTM

PR #7194, opened 2026-04-10
Move TailLogs endpoint to dataproxy

katrogan:
And add a `GetActionLogContext` to fetch log context from RunsService

## Why are the changes needed?

Required to reduce dataflow in a secure environment.

## What changes were proposed in this pull request?

See PR title

## How was this patch tested?

Ran the demo cluster locally. Tested with [flyte-sdk](https://github.com/flyteorg/flyte-sdk/pull/962) and this script

```
@env.task
async def spew_logs(duration_sec: int = 300, interval_sec: int = 5) -> str:
    elapsed = 0
    iteration = 0
    while elapsed < duration_sec:
        iteration += 1
        print(f"[{elapsed:>4}s] iteration {iteration} - still running...", flush=True)
        await asyncio.sleep(interval_sec)
        elapsed += interval_sec
    print(f"Done after {elapsed}s ({iteration} iterations)", flush=True)
    return f"completed {iteration} iterations"


if __name__ == "__main__":
    import sys

    config_path = sys.argv[1] if len(sys.argv) > 1 else None
    if config_path:
        from flyte.config import Config

        flyte.init_from_config(Config.auto(config_path))
    else:
        flyte.init_from_config()
    run = flyte.run(spew_logs)
    print(f"Run: {run.name}")
    print(f"URL: {run.url}")
    print("Tailing logs...")
    run.action.show_logs(raw=True)
```

Verified output worked fine

```
 python examples/basics/long_running.py .flyte/config-oss-local.yaml
  > Bundling code...
  ✓ Code bundle: 1 files, 0.009765625 MB (compressed 0.0006895065307617188 MB)
  > Uploading code bundle...
Run: r8xqdw87c7l266b9j99x
URL: http://localhost:30080/v2/domain/development/project/flytesnacks/runs/r8xqdw87c7l266b9j99x
Tailing logs...
 [flyte] Flyte runtime started for action a0 with run name r8xqdw87c7l266b9j99x [   0s] iteration 1 - still running... [   5s] iteration 2 - still running... [  10s] iteration 3 - still running...
```

### Labels

Please add one or more of the following labels to categorize your PR:

- **added**: For new features.
- **changed**: For changes in existing functionality.
- **deprecated**: For soon-to-be-removed features.
- **removed**: For features being removed.
- **fixed**: For any bug fixed.
- **security**: In case of vulnerabilities

This is important to improve the readability of release notes.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
[...]

PR #7221, opened 2026-04-16
[V2] Add pod_name field to TailLogsRequest

pingsutw:

Currently, \`TailLogsRequest\` only accepts an action execution ID and attempt number, which means the service has to guess which pod to tail logs from. Adding a \`pod\_name\` field allows callers to explicitly specify the target pod, which is useful when an action has multiple pods.

Add a \`pod\_name\` field (field number 3) to \`TailLogsRequest\` in `run_logs_service.proto`. If not provided, the service will attempt to find the primary pod, else assume the first pod.

Regenerated all language bindings (Go, Python, Rust, TypeScript).

## How was this patch tested?

Proto compilation and code generation verified.

### Labels

- **added**

* `main` 
  - \#6583
    - **\[V2] Add pod\_name field to TailLogsRequest** :point\_left:


flyteidl2/workflow/run_logs_service.proto
```diff
@@ -24,6 +24,9 @@ message TailLogsRequest {
 
   // The attempt number.
   uint32 attempt = 2 [(buf.validate.field).uint32.gt = 0];
+
+  // The pod name to tail logs from. If not provided, attempt to find the primary pod, else assume the first pod.
+  string pod_name = 3 [(buf.validate.field).string.min_len = 1];
```
katrogan: this service is deprecated, I just moved this over to dataproxy :) can we update?
https://github.com/flyteorg/flyte/pull/7194

pingsutw: updated both

flyteidl2/workflow/run_logs_service.proto
```diff
@@ -24,6 +24,9 @@ message TailLogsRequest {
 
   // The attempt number.
   uint32 attempt = 2 [(buf.validate.field).uint32.gt = 0];
+
+  // The pod name to tail logs from. If not provided, attempt to find the primary pod, else assume the first pod.
+  string pod_name = 3;
```
katrogan, line 29: can we remove this? since run logs service is deprecated we shouldn't be updating it anymore

pingsutw: removed it

katrogan (approved):

PR #7225, opened 2026-04-16
fix: correct abort flow — phase transitions, reason propagation, and K8s cascade

AdilFayyaz:

The abort implementation on v2 had several bugs: the reconciler was pushing the wrong action ID, AbortInfo was never populated in action details, aborted actions were missing phase-transition events (so WatchActionDetails never showed ABORTED), and  WatchActionDetails had a race condition that could miss the abort notification entirely.

- Action service: add DeleteFunc handler to the K8s informer so CRD deletions (triggered  by abort) propagate ABORTED back to the run service; skip label-patching on deleted CRs which would always 404
- Run service: fix reconciler push to use "a0" (was using run name); populate AbortInfo.Reason in buildActionDetails (was a TODO); add 404 handling for unknown run/action in AbortRun/AbortAction; fix WatchActionDetails subscribe-before-read ordering to close the notification race
- Repo: insert an action\_events row with ABORTED phase on abort so WatchActionDetails returns a phaseTransitions entry; set ended\_at/duration\_ms on abort; simplify abort to only mark the root/targeted action — K8s OwnerReferences cascade CRD deletion to children, and the informer handles marking them ABORTED
- Reconciler: handle sql.ErrNoRows in MarkAbortAttempt gracefully; broaden isAlreadyTerminated to catch K8s "not found" errors wrapped as CodeInternal

## How was this patch tested?

- go test ./runs/... passes
- Call AbortRun on a running workflow; confirm root action is immediately ABORTED in DB with reason populated; confirm child actions transition to ABORTED after reconciler deletes the root CRD and the informer fires delete events
- Call WatchActionDetails on an aborted action; confirm the response includes a phaseTransitions entry with ABORTED and the AbortInfo.Reason field is set

## Docs link

- `main` 
  - \#6583
    - **fix: correct abort flow — phase transitions, reason propagation, and K8s cascade** :point\_left:


docker/demo-bundled/manifests/complete.yaml
```diff
@@ -301,6 +301,20 @@ metadata:
...
+automountServiceAccountToken: true
+kind: ServiceAccount
+metadata:
+  labels:
+    app.kubernetes.io/instance: flyte-demo
+    app.kubernetes.io/managed-by: Helm
+    app.kubernetes.io/name: minio
+    helm.sh/chart: minio-12.6.7
+  name: flyte-demo-minio
+  namespace: flyte
```
pingsutw: You may need to delete the files in charts/flyte-demo/charts/minio

runs/repository/impl/action.go
```diff
@@ -444,27 +464,58 @@ func (r *actionRepo) UpdateActionPhase(
...
+	logger.Infof(ctx, "AbortAction: aborted %s", actionID.Name)
 	return nil
 }
 
+// insertAbortEvent writes a single ABORTED phase-transition event into action_events so that
+// WatchActionDetails returns a phaseTransitions entry with phase = ABORTED for the action.
+// Failures are non-fatal — the caller should log and continue.
+func (r *actionRepo) insertAbortEvent(ctx context.Context, actionID *common.ActionIdentifier, attempts uint32, reason string, now time.Time) error {
+	// Build the minimal ActionEvent proto that mergeEvents will pick up.
+	event := &workflow.ActionEvent{
```
pingsutw: can delete it too

runs/repository/impl/action.go
```diff
@@ -431,6 +443,14 @@ func (r *actionRepo) UpdateActionPhase(
 		return err
 	}
 	if rowsAffected > 0 {
+		// Insert an abort event so WatchActionDetails phaseTransitions include ABORTED.
+		// This must happen before notifyActionUpdate so the event is visible when the
+		// subscriber re-fetches action events in response to the notification.
+		if phase == common.ActionPhase_ACTION_PHASE_ABORTED {
+			if err := r.insertAbortEvent(ctx, actionID, attempts, "", now); err != nil {
+				logger.Warnf(ctx, "UpdateActionPhase: failed to insert abort event for %s: %v", actionID.Name, err)
```
pingsutw: I think we can delete it too if close the connection unit we get the terminal action evnet.

AdilFayyaz: we still need this if WatchActionDetails waits for action_events to have a terminal entry.

runs/repository/impl/action.go
AdilFayyaz: This is called in UpdateActionPhase - insertAbortEvent is why action_events get an Aborted row.

pingsutw (approved):

PR #7233, opened 2026-04-18
use Helm to init rustfs in sandbox

### LINKED ISSUE #7209: [V2] use Helm to init rustfs in sandbox

According to rustfs [doc](https://github.com/rustfs/helm?tab=readme-ov-file#rustfs-helm-mode), we can run it with helm chart instead of scripting the yaml on our own. 
We run an init container to [create default bucket ](https://github.com/flyteorg/flyte/blob/v2/charts/flyte-demo/templates/storage/rustfs/deployment.yaml#L23-L32) now because rustfs do not support default bucket config now. After migrating to helmchart, we can run a k8s post-install Job to create default bucket.

1. Run make sandbox-build.
2. Run an example workflow and make sure the result is success.

n/a

_No response_

- [x] Yes

- [x] Yes

BarryWu0812:
## Tracking issue

Closes #7209

According to rustfs [doc](https://github.com/rustfs/helm?tab=readme-ov-file#rustfs-helm-mode), we can run it with helm chart instead of scripting the yaml on our own.
We run an init container to [create default bucket ](https://github.com/flyteorg/flyte/blob/v2/charts/flyte-demo/templates/storage/rustfs/deployment.yaml#L23-L32)now because rustfs do not support default bucket config now. After migrating to helmchart, we can run a k8s post-install Job to create default bucket.

Use Helm to init rustfs in sandbox and run k8s post-install Job to create default bucket.

## How was this patch tested?

Run make sandbox-build.
Run an example workflow and make sure the result is success.

### Labels

## Docs link

- `main` 
  - \#6583
    - **use Helm to init rustfs in sandbox** :point\_left:


charts/flyte-demo/values.yaml
```diff
@@ -27,7 +27,7 @@ flyte-binary:
         s3:
           disableSSL: true
           v2Signing: true
-          endpoint: http://rustfs.{{ .Release.Namespace }}:9000
```
popojk, line 30: Could we keep using `rustfs.{{ .Release.Namespace }}` as service name instead of hard coding?

charts/flyte-demo/templates/storage/rustfs-bucket-init-job.yaml
```diff
@@ -0,0 +1,39 @@
...
+metadata:
+  name: {{ printf "%s-rustfs-bucket-init" .Release.Name | trunc 63 | trimSuffix "-" }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- include "flyte-demo.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-weight": "10"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+spec:
```
popojk, line 15: I noticed that the job pod remain in cluster after job finished. could we add a `ttlSecondsAfterFinished` under spec section to delete it?

charts/flyte-demo/templates/storage/rustfs-bucket-init-job.yaml
```diff
@@ -0,0 +1,39 @@
...
+  backoffLimit: 6
+  template:
+    metadata:
+      labels:
+        {{- include "flyte-demo.selectorLabels" . | nindent 8 }}
+        app.kubernetes.io/component: rustfs-bucket-init
```
popojk, line 21: Curious why do we need this label here? I thought this is just an one-off k8s job and maybe we don't need label?

charts/flyte-demo/values.yaml
```diff
@@ -111,8 +111,35 @@ flyte-binary:
...
+    repository: rustfs/rustfs
+    tag: sandbox
+  service:
+    type: NodePort
+    endpoint:
+      nodePort: 30002
+  secret:
+    rustfs:
+      access_key: rustfs
+      secret_key: rustfsstorage
```
popojk, line 142: I noticed that there is a rustfs test connection pod remain in the cluster, which is not required because the sandbox should failed to startup if connection failed. We can do so by:
```
test:
      enabled: false
```

docker/demo-bundled/Makefile
```diff
@@ -51,7 +52,7 @@ manifests: dep_update
 		--load-restrictor=LoadRestrictionsNone \
 		kustomize/dev > manifests/dev.yaml
 
-CONSOLE_IMAGE := ghcr.io/flyteorg/flyte-client-v2
+CONSOLE_IMAGE := ghcr.io/unionai-oss/flyteconsole-v2
```
popojk, line 55: @pingsutw I'm not sure if we will pull console image from ghcr.io/unionai-oss in the future? Should we do this fix here?

popojk (commented):
Overall looks good, thanks! Left a few comments, PTAL

docker/demo-bundled/Makefile
pingsutw: I already changed the image name to ghcr.io/unionai-oss/flyteconsole-v2 in another PR

popojk:
Hey @BarryWu0812 .
1. I still saw a rustfs init pod created on routine basis, could you please check it?
2. Please rebase this PR after 1st item resolved

pingsutw (commented):
Could we add new path (/rustfs) to the ingress rules for the rustfs console and make sure it’s accessible from the browser?
https://github.com/flyteorg/flyte/blob/61a555c68d97eb921909f59c0bafabfa7905fbc1/charts/flyte-devbox/templates/proxy/ingress.yaml#L38-L55

charts/flyte-devbox/templates/storage/rustfs-bucket-init-job.yaml
```diff
@@ -0,0 +1,37 @@
...
+spec:
+  backoffLimit: 6
+  ttlSecondsAfterFinished: 60
+  template:
+    spec:
+      restartPolicy: OnFailure
+      containers:
+      - name: bucket-init
+        image: "{{ .Values.rustfs.bucketInit.image.repository }}:{{ .Values.rustfs.bucketInit.image.tag }}"
+        imagePullPolicy: {{ .Values.rustfs.bucketInit.image.pullPolicy }}
```
pingsutw: Could we add a patch to complete/dev [kustomization.yaml](https://github.com/flyteorg/flyte/blob/05d89b4e32a6972c39877238ddf55c702d10a27a/docker/devbox-bundled/kustomize/complete/kustomization.yaml#L104-L110) to add an init container to the rustfs deployment, so that we don't need to add `docker.io/minio/mc:latest` and another k8s job into the devbox
```yaml
- patch: |-
    apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: rustfs
      namespace: flyte
    spec:
      template:
        spec:
          initContainers:
          - name: create-default-buckets
            image: busybox:stable
            imagePullPolicy: IfNotPresent
            command:
            - sh
            - -c
            - mkdir -p /data/flyte-data
            securityContext:
              allowPrivilegeEscalation: false
              capabilities:
                drop:
                - ALL
              readOnlyRootFilesystem: true
              runAsNonRoot: true
            volumeMounts:
            - mountPath: /data
              name: data
```

BarryWu0812:
Hi @popojk @pingsutw , I have create new [PR](https://github.com/flyteorg/flyte/pull/7310) based on main branch. Please help me to review. Thanks.

PR #7234, opened 2026-04-18
fix(helm): set catalog-cache to noop when datacatalog is disabled in flyte-core

### LINKED ISSUE #7155: [BUG] Datacatalog enable and catalog-cache setting mismatch cause DNS
### Flyte & Flytekit version

flyte-core v1.16.4

### Describe the bug

By default, datacatalog is enabled ([here](https://github.com/flyteorg/flyte/blob/master/charts/flyte-core/values.yaml#L273)) which works with the default settings for catalog-cache ([here](https://github.com/flyteorg/flyte/blob/master/charts/flyte-core/values.yaml#L1055)).

However, this two settings are miles apart especially the catalog-cache setting is deeply embedded inside configmaps. So if datacatalog is disabled which is not uncommon to reduce operational complexity, the default settings on catalog-cache will cause DNS failure due to datacatalog is not there. This is a gotcha moment because users may not know there is a default setting on configmaps that explicitly sets the cache entrypoint to use datacatalog.

### Expected behavior

These two related configs should be linked together. If datacatalog is disabled, flyte should automatically disable the catalog-cache instead of using datacatalog service which doesn't exist.

Disable datacatalog without changing catalog-cache settings in configmaps setting.

### Screenshots

Workflow[xxxx] failed.
RuntimeExecutionError: max number of system retry attempts [31/30] exhausted.
Last known status message: Failed to check Catalog for previous results:
DataCatalog failed to get dataset for ID
  resource_type:TASK
  project:"xxxx"
  domain:"xxxx"
  name:"xxxx"
  version:"xxxx":
rpc error: code = Unavailable
  desc = connection error: desc = "transport: Error while dialing: dial tcp: lookup datacatalog on x.x.x.x:x no such host"

- [x] Yes

- [x] Yes

nuthalapativarun:
## Tracking issue
Closes #7155

When `datacatalog.enabled=false` in the flyte-core Helm chart, the `catalog-cache` configuration still defaults to `type: datacatalog` with `endpoint: datacatalog:89`. Since the datacatalog service is not deployed in this case, every task execution triggers a DNS lookup failure:

```
Failed to check Catalog for previous results:
rpc error: code = Unavailable
  desc = connection error: dial tcp: lookup datacatalog on x.x.x.x:x: no such host
```

These two related settings are far apart in `values.yaml` (datacatalog toggle at line 273, catalog-cache config buried inside configmaps at line 1055), making it easy to miss the dependency.

In `charts/flyte-core/templates/propeller/configmap.yaml`, the `catalog.yaml` configmap entry is now rendered conditionally:

- When `datacatalog.enabled=true` (default): renders the existing `configmap.catalog` value as before, preserving full backward compatibility.
- When `datacatalog.enabled=false`: renders `catalog-cache.type: noop` so propeller skips catalog lookups entirely, preventing DNS failures.

## How was this patch tested?
Verified with `helm template` that:
- `--set datacatalog.enabled=false` produces `catalog-cache:\n  type: noop`
- `--set datacatalog.enabled=true` (default) produces the original datacatalog endpoint config unchanged

### Labels
- **fixed**

### DIFF
--- a/charts/flyte-core/templates/propeller/configmap.yaml
+++ b/charts/flyte-core/templates/propeller/configmap.yaml
@@ -9,8 +9,14 @@ data:
 {{- with .Values.configmap.admin }}
   admin.yaml: | {{ tpl (toYaml .) $ | nindent 4 }}
 {{- end }}
+{{- if .Values.datacatalog.enabled }}
 {{- with .Values.configmap.catalog }}
   catalog.yaml: | {{ tpl (toYaml .) $ | nindent 4 }}
+{{- end }}
+{{- else }}
+  catalog.yaml: |
+    catalog-cache:
+      type: noop
 {{- end }}
 {{- with .Values.configmap.catalog_cache }}
   catalog_cache.yaml: | {{ tpl (toYaml .) $ | nindent 4 }}
--- a/deployment/eks/flyte_helm_dataplane_generated.yaml
+++ b/deployment/eks/flyte_helm_dataplane_generated.yaml
@@ -85,11 +85,9 @@ data:
       capacity: 1000
       rate: 500
       type: admin
-  catalog.yaml: | 
+  catalog.yaml: |
     catalog-cache:
-      endpoint: datacatalog:89
-      insecure: true
-      type: datacatalog
+      type: noop
   copilot.yaml: | 
     plugins:
       k8s:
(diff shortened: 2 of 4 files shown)


EngHabu (approved):

pingsutw:
@nuthalapativarun could you run `make helm` locally and push the changes

pingsutw (dismissed):

Sovietaced:
@nuthalapativarun Can you rebase and signoff your commit?

Sovietaced:
@nuthalapativarun please run `make helm` and push those changes

Sovietaced:
@nuthalapativarun Not sure why but you seem to generate different values than CI does. Wonder if you're using a different version of helm or something.

Sovietaced:
Closing in favor of https://github.com/flyteorg/flyte/pull/7433

PR #6294, opened 2025-03-02
Dep: Upgrade kubeflow training operator

fg91:
## Why are the changes needed?

Goal: Enable https://github.com/flyteorg/flyte/pull/6295 which prevents the flyteplugins kubeflow plugins from failing a task when a PytorchJob, TfJob, ... hasn't been updated by the kubeflow training operator *because the job was suspended* (e.g. by an external queueing system).

The run policy's "suspend" attribute is introduced only in a newer version of the kubeflow training operator.

## What changes were proposed in this pull request?

Upgrade the kubeflow training operator dependency.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [x] All commits are signed-off.

## Related PRs

https://github.com/flyteorg/flyte/pull/6295 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request upgrades the kubeflow training operator dependency and related k8s libraries to support new suspend functionality, preventing task failures when jobs remain suspended. It includes standardizing API usage by replacing legacy commonOp types with kubeflowv1 types in PyTorch and TensorFlow operators, critical fixes to error handling logic, missing return statements, and type corrections in protobuf definitions across multiple language bindings.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 5
</div>

PR #7241, opened 2026-04-20
Fix type error for suspended kubeflow jobs

strigazi:
When kubeflow jobs are suspended by an external system (eg kueue), the controller error's with the following [0].

https://github.com/flyteorg/flyte/pull/6295/changes relies only on app.Spec.RunPolicy.Suspend and does not take into account the phase.

[0]  `failed Execute for node. Error: failed at Node[dn1]. RuntimeExecutionError: failed during plugin execution, caused by: Invalid transition for plugin [pytorch]: transition doesn't have task info nor an execution error filled [TransitionTypeEphemeral,Phase<PhaseUndefined:0 <nil> Reason:>]`

related: #6294 #6295

Observed in versions:  v1.16.3, v1.16.4, v1.16.5

### DIFF
--- a/flyteplugins/go/tasks/plugins/k8s/kfoperators/mpi/mpi_test.go
+++ b/flyteplugins/go/tasks/plugins/k8s/kfoperators/mpi/mpi_test.go
@@ -255,6 +255,18 @@ func dummyMPIJobResource(mpiResourceHandler mpiOperatorResourceHandler,
 			Time: now.Add(3 * time.Minute),
 		},
 	}
+	jobSuspended := kubeflowv1.JobCondition{
+		Type:    kubeflowv1.JobSuspended,
+		Status:  corev1.ConditionTrue,
+		Reason:  "MPIJobSuspended",
+		Message: "MPIJob the-job is suspended.",
+		LastUpdateTime: v1.Time{
+			Time: now.Add(time.Minute),
+		},
+		LastTransitionTime: v1.Time{
+			Time: now.Add(time.Minute),
+		},
+	}
 
 	switch conditionType {
 	case kubeflowv1.JobCreated:
@@ -284,6 +296,11 @@ func dummyMPIJobResource(mpiResourceHandler mpiOperatorResourceHandler,
 			jobRunningInactive,
 			jobFailed,
 			jobRestarting,
+		}
+	case kubeflowv1.JobSuspended:
+		jobConditions = []kubeflowv1.JobCondition{
+			jobCreated,
+			jobSuspended,
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


strigazi:
Hello @fg91, following your PR #6295  to make flyte work with suspended kubeflow jobs, can you have a quick look in this PR? Does the direction of this patch look good to you?

pingsutw (dismissed):

fg91 (changes requested):
To me it doesn't make sense that a kubeflow job that is `.JobSuspended` (meaning not yet running) is treated as `pluginsCore.PhaseInfoRunning`. The ray plugin treats it as "phase info queued", see [here](https://github.com/flyteorg/flyte/blob/e938b03c9ca9b3593344da89a345d11716adc1fe/flyteplugins/go/tasks/plugins/k8s/ray/ray.go#L657) which makes more sense to me. Let's return `pluginsCore.PhaseInfoQueued(...` instead.

pingsutw:
but kubeflow may also suspend a running job too, (running -> suspended)

fg91:
> but kubeflow may also suspend a running job too, (running -> suspended)

Do you know in which situation kubeflow does this? Haven't seen this yet, only by external systems like kueue via an admission webhook right after creation of the job object.

strigazi:
I think that pluginsCore.PhaseInfoQueued makes more sense as well. Kueue suspends workloads to queue them for later. Either right after creation or after already running.

flyteplugins/go/tasks/plugins/k8s/kfoperators/mpi/mpi_test.go
```diff
@@ -556,7 +556,7 @@ func TestGetTaskPhase(t *testing.T) {
 	assert.Equal(t, pluginsCore.PhaseInfoUndefined, taskPhase)
 
 	// Training operator did not modify the job because it is suspended
-	mpiJobSuspended := dummyMPIJobResourceCreator(kubeflowv1.JobCreated)
+	mpiJobSuspended := dummyMPIJobResourceCreator(kubeflowv1.JobSuspended)
```
fg91: When the job was just created, we do treat this as "phase info queued" so the behavior that is tested here is correct, let's not remove coverage but add another text case for suspended please.

fg91: In the two files below as well please.

strigazi: done

fg91 (approved):
LGTM now, thanks!

Sovietaced (approved):

strigazi:
Are there any requirements to cut a [release](https://github.com/flyteorg/flyte/blob/master/RELEASE.md)? In this case 1.16.6 (with this patch)?

Sovietaced:
> Are there any requirements to cut a [release](https://github.com/flyteorg/flyte/blob/master/RELEASE.md)? In this case 1.16.6 (with this patch)?

Not really. I can push one out tomorrow if someone else doesn't.

PR #7265, opened 2026-04-23
feat: implement TailLogs for AppLogsService

pingsutw:

`AppLogsService.TailLogs` (`/flyteidl2.app.AppLogsService/TailLogs`) was previously unimplemented — calls to it returned `Unimplemented`. This blocks any UI or CLI that needs to stream logs from app replicas.

Implements `TailLogs` end-to-end with the same control-plane / data-plane split used by `AppService`:

**Data plane** (`app/internal/service/`)

- `InternalAppLogsService` resolves a request target to replica pods — `app_id` expands to all replicas via `AppK8sClient.GetReplicas`, `replica_id` maps to a single pod. It sends an initial `ReplicaIdentifierList` so the client knows which replicas will be streamed, then fans out per-pod log streams. `stream.Send` is guarded by a mutex because connect server streams are not safe for concurrent writes. Concurrency is capped via a semaphore.
- `K8sAppLogStreamer` tails pod logs via `kubernetes.Interface`. It picks the first non-`queue-proxy` container (Knative injects `queue-proxy` as a sidecar), batches 100 lines at a time, flushes on idle so clients aren't stuck on a partial batch, and detaches from the inbound gRPC deadline so long follows aren't cut off.

**Control plane** (`app/service/app_logs_service.go`)

- `AppLogsService` proxies the server-streaming RPC to the internal endpoint — same pattern as `AppService.Watch`.

**Wiring**

- `app/internal/setup.go` mounts `InternalAppLogsService` at `/internal/flyteidl2.app.AppLogsService/` (uses `sc.K8sConfig` for the streaming clientset).
- `app/setup.go` mounts the control-plane `AppLogsService` at `/flyteidl2.app.AppLogsService/`.

## How was this patch tested?

Screenshot 2026-04-23 at 1 48 45 PM

- `go build ./app/...`
- `go vet ./app/...`
- `go test ./app/...`

### Labels

added

* `main` 
  - \#6583
    - **feat: implement TailLogs for AppLogsService** :point\_left:


docker/devbox-bundled/manifests/complete.yaml
```diff
@@ -7970,7 +7969,7 @@ metadata:
 ---
 apiVersion: v1
 data:
-  domain-template: '{{.Name}}-{{.Namespace}}.{{.Domain}}'
+  domain-template: '{{.Name}}.{{.Domain}}'
```
pingsutw, line 7972: since we run all the apps in the flyte namespace

AdilFayyaz (approved):

PR #7271, opened 2026-04-24
fix(spark): populate timestamp fields in log template input

### LINKED ISSUE #6328: [Core feature] Add some start time template variable for Spark log links

When linking to some external logging service its very convenient if it automatically opens the correct time window. Currently this is not really possible for Spark tasks. 

Approximate parity with normal python tasks in terms of the template variables available for forming log links https://github.com/flyteorg/flyte/blob/f5a4ce0deabdddd16ea1c18fb9b44ac0be4f831f/flyteplugins/go/tasks/plugins/k8s/kfoperators/common/common_operator.go#L138-L146. 

Just provide log links with a really long time window. Unfortunately this can slow down the queries on the logging tool and occasionally leads to timeouts. 

For spark tasks the log links are created by https://github.com/flyteorg/flyte/blob/f5a4ce0deabdddd16ea1c18fb9b44ac0be4f831f/flyteplugins/go/tasks/plugins/k8s/spark/spark.go#L342-L347

- [x] Yes

- [x] Yes

cmditch:
## Summary

The Spark plugin's `getEventInfoForSpark` constructs `tasklog.Input` without populating any timestamp fields (`PodUnixStartTime`, `PodUnixFinishTime`, `PodRFC3339StartTime`, `PodRFC3339FinishTime`). This causes log template URIs referencing `{{ .podUnixStartTime }}` to render as `0`, making it impossible to construct time-bounded log links for external logging services like Grafana Loki or Kibana.

The standard K8s plugin (`logging_utils.go`) and kfoperators (`common_operator.go`) already populate these fields from the resource's `CreationTimestamp`. This change applies the same pattern to all four `tasklog.Input` structs in `getEventInfoForSpark`.

### Changes

- Compute `startTime` from `sj.CreationTimestamp.Unix()` (always available)
- Compute `finishTime` from `sj.Status.TerminationTime` only when the app is in a terminal state (`Completed` or `Failed`) and `TerminationTime` is non-zero — otherwise `finishTime` remains `0`
- Populate `PodUnixStartTime`, `PodUnixFinishTime`, `PodRFC3339StartTime`, and `PodRFC3339FinishTime` in all four `tasklog.Input` structs (Driver, User, System, and All-User logs)
- Added `TestGetEventInfoTimestamps` test with two subtests:
  - **Running app**: verifies start time is populated and finish time is `0`
  - **Completed app**: verifies both start and finish times are populated from `CreationTimestamp` and `TerminationTime` respectively

### Testing

- All existing Spark plugin tests pass
- New test verifies that a log template URI containing `{{ .podUnixStartTime }}` renders the actual creation timestamp rather than `0`
- New test verifies that finish time is only populated for terminal states using the CRD's `TerminationTime`
- Manually tested against a Flyte Binary deployment with Grafana Loki log links using `{{ .podUnixStartTime }}000` for epoch-ms time ranges

Fixes #6328

### DIFF
--- a/flyteplugins/go/tasks/plugins/k8s/spark/spark.go
+++ b/flyteplugins/go/tasks/plugins/k8s/spark/spark.go
@@ -394,6 +394,14 @@ func getEventInfoForSpark(pluginContext k8s.PluginContext, sj *sparkOp.SparkAppl
 	taskLogs := make([]*core.TaskLog, 0, 3)
 	taskExecID := pluginContext.TaskExecutionMetadata().GetTaskExecutionID()
 
+	startTime := sj.CreationTimestamp.Unix()
+	var finishTime int64
+	isTerminal := sj.Status.AppState.State == sparkOp.ApplicationStateCompleted ||
+		sj.Status.AppState.State == sparkOp.ApplicationStateFailed
+	if isTerminal && !sj.Status.TerminationTime.IsZero() {
+		finishTime = sj.Status.TerminationTime.Unix()
+	}
+
 	if sj.Status.DriverInfo.PodName != "" {
 		p, err := logs.InitializeLogPlugins(&sparkConfig.LogConfig.Mixed)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


flyteplugins/go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -391,6 +391,9 @@ func getEventInfoForSpark(pluginContext k8s.PluginContext, sj *sparkOp.SparkAppl
 	taskLogs := make([]*core.TaskLog, 0, 3)
 	taskExecID := pluginContext.TaskExecutionMetadata().GetTaskExecutionID()
 
+	startTime := sj.CreationTimestamp.Unix()
+	finishTime := time.Now().Unix()
```
pingsutw: Does spark job crd have finish time? Also, we should only set it when job is completed

cmditch: Good call — looks like the SparkApplication CRD does have Status.TerminationTime. Updated to use that, only set when the app is in a terminal state (Completed/Failed).

pingsutw (approved):

PR #7283, opened 2026-04-25
Fix: return error when update affects zero rows in execution repo

### LINKED ISSUE #7257: InternalRunService: silent data loss when RecordAction fails before UpdateActionStatus
## Summary

When the Actions Service's TaskAction CRD watcher (`actions/k8s/client.go:notifyRunService`) fails to call `InternalRunService.RecordAction`, subsequent `UpdateActionStatus` calls for the same action **silently no-op** — the gRPC response is success, but zero rows are updated and the action is effectively invisible in the Runs DB until a K8s watch re-list re-emits `ADDED`.

## Reproduction

1. Runs Service is briefly unavailable when a TaskAction's `ADDED` event is processed.
2. `RecordAction` fails. The Actions Service logs a `Warnf` and moves on (no retry). The bloom filter is **not** updated, but no further action is taken.
3. Runs Service recovers.
4. Subsequent `MODIFIED` events for that TaskAction call `UpdateActionStatus`. The SQL `UPDATE` matches zero rows. Runs Service returns `nil` error (see `runs/repository/impl/action.go:397-416`, `runs/service/internal_run_service.go:232`).
5. The action never appears to clients using `WatchRunDetails` until a watch re-list occurs (controller restart / watcher reconnect).

## Root cause

Two coupled issues in `InternalRunService`:

1. **`UpdateActionPhase` returns success on 0 rows affected.** The caller has no way to know the action is missing from the DB.
2. **`RecordAction` and `UpdateActionStatus` are split RPCs.** The client must call both in order; there is no single idempotent "report" operation.

## Proposed fix

**1. Make `UpdateActionStatus` return `NotFound` when no row is matched.**

```go
// runs/repository/impl/action.go
if rowsAffected == 0 {
    return ErrActionNotFound
}
```

…mapped to `connect.CodeNotFound` in the gRPC handler.

**2. Add `ActionStatus status = 6` to `RecordActionRequest`** so the initial insert carries the correct phase (no transient `PHASE_UNSPECIFIED` window).

**3. Self-heal in `actions/k8s/client.go:notifyRunService`:**

```go
_, err := c.runClient.UpdateActionStatus(ctx, statusReq)
if connect.CodeOf(err) == connect.CodeNotFound {
    // Row missing — RecordAction must have failed earlier. Rebuild from CRD
    // and re-record. Follow the existing pattern: Add to the bloom filter only
    // on success (the filter is add-only; never Remove).
    recordReq := buildRecordRequestFromCRD(taskAction)
    if _, recErr := c.runClient.RecordAction(ctx, recordReq); recErr == nil {
        if c.recordedFilter != nil {
            c.recordedFilter.Add(ctx, actionKey)
        }
        // Optional: retry UpdateActionStatus now that the row exists.
    }
}
```

The Actions Service still has the TaskAction CRD in its informer cache, so rebuilding the `RecordActionRequest` is free.

**4. On `ADDED` events, skip the separate `UpdateActionStatus`** — `RecordAction` now carries the status.

### Independent improvements

- **Bounded retry with backoff** for transient errors (`Unavailable`, `DeadlineExceeded`) — 3 attempts, exponential (100ms / 500ms / 2s). Don't block the sharded worker on long outages.
- **Periodic reconciliation sweep** — every ~5min list TaskAction CRDs without the `flyte.org/terminal-status-recorded` label and force a `notifyRunService` pass. Backstop for cases where the watcher doesn't reconnect but Runs Service has recovered.

## References

- `actions/k8s/client.go:554-646` — `notifyRunService`
- `runs/service/internal_run_service.go:191-255` — `UpdateActionStatus` handler
- `runs/repository/impl/action.go:361-417` — `UpdateActionPhase` (the silent-success site)

muskan-creates352:
Fixes #7257

This PR ensures that Update operations in ExecutionRepo return an error when no rows are affected.

Previously, updates silently succeeded even if no matching record existed, which could lead to hidden data issues.

This change adds a check on RowsAffected and returns a not found error when no rows are updated.

### DIFF
--- a/flyteadmin/pkg/repositories/gormimpl/execution_repo.go
+++ b/flyteadmin/pkg/repositories/gormimpl/execution_repo.go
@@ -70,9 +70,16 @@ func (r *ExecutionRepo) Update(ctx context.Context, execution models.Execution) 
 	timer := r.metrics.UpdateDuration.Start()
 	tx := r.db.WithContext(ctx).Model(&models.Execution{}).Where(getIDFilter(execution.ID)).Updates(execution)
 	timer.Stop()
+
 	if err := tx.Error; err != nil {
 		return r.errorTransformer.ToFlyteAdminError(err)
 	}
+
+	
+	if tx.RowsAffected == 0 {
+		return adminErrors.GetMissingEntityError("execution", nil)
+	}
+
 	return nil
 }
 
--- a/flyteadmin/pkg/repositories/gormimpl/node_execution_repo.go
+++ b/flyteadmin/pkg/repositories/gormimpl/node_execution_repo.go
@@ -100,8 +100,13 @@ func (r *NodeExecutionRepo) Update(ctx context.Context, nodeExecution *models.No
 	tx := r.db.WithContext(ctx).Model(&models.NodeExecution{}).Where(getIDFilter(nodeExecution.ID)).Updates(nodeExecution)
 	timer.Stop()
 	if err := tx.Error; err != nil {
-		return r.errorTransformer.ToFlyteAdminError(err)
-	}
+    return r.errorTransformer.ToFlyteAdminError(err)
+}
+
+if tx.RowsAffected == 0 {
+    return adminErrors.GetMissingEntityError("node execution", nil)
+}
+
 	return nil
 }
 
--- a/flyteadmin/pkg/repositories/gormimpl/task_execution_repo.go
+++ b/flyteadmin/pkg/repositories/gormimpl/task_execution_repo.go
@@ -88,6 +88,13 @@ func (r *TaskExecutionRepo) Update(ctx context.Context, execution models.TaskExe
 	if err := tx.Error; err != nil {
 		return r.errorTransformer.ToFlyteAdminError(err)
 	}
+
+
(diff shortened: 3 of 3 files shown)


muskan-creates352:
Hi @maintainers,

I've now extended the fix to NodeExecutionRepo as well to ensure consistent handling of zero rows affected across repositories.

Please let me know if any further adjustments are needed.

muskan-creates352:
Hi @maintainers,

The fix has now been applied consistently across ExecutionRepo, TaskExecutionRepo, and NodeExecutionRepo.

Looking forward to your feedback. Thanks!

muskan-creates352:
Hi ,

This PR ensures that update operations return an error when no rows are affected, preventing silent failures.

The fix has been applied consistently across ExecutionRepo, TaskExecutionRepo, and NodeExecutionRepo.

Please let me know if any further changes are needed. Thanks!

muskan-creates352:
Hi,

Could you please approve the workflows so that the checks can run?

Thanks!

Sovietaced:
The issue you linked to references code in the v2 branch but you've fixed some unrelated code in the `master` branch.

PR #7284, opened 2026-04-26
Delegate RunService.GetActionData to DataProxyService 

### LINKED ISSUE #7253: Delegate RunService.GetActionData to DataProxyService to remove duplicate implementation
## What

`RunService.GetActionData` (`runs/service/run_service.go:741`) and `dataproxy.Service.GetActionData` (`dataproxy/service/dataproxy_service.go:444`) both do the same job: resolve the input/output URIs for an action and read the protobuf payloads back from blob storage.

The canonical home for blob-backed data access is now the DataProxy service. The `RunService.GetActionData` RPC is only kept for backward compatibility with existing clients.

## Why

Today both services:

- Hold a `storage.DataStore` and implement parallel reads of `inputs.pb` / outputs via `errgroup`.
- Resolve storage URIs from the action record (DataProxy goes through `RunService.GetActionDataURIs`; RunService reads them directly from the DB).
- Must stay in lock-step whenever we evolve the storage layout, artifact types, or error semantics (not found, partial results, trace vs task actions, etc.).

This is duplicate surface area with no good reason now that DataProxy owns data access. It also makes the storage-layout contract implicit in two places.

## Proposed change

Refactor `RunService.GetActionData` into a thin shim that forwards to `DataProxyServiceClient.GetActionData`:

- Inject a `dataproxyconnect.DataProxyServiceClient` into `RunService` (likely the in-process one for the all-in-one binary).
- In `GetActionData`:
  1. Validate the request as today.
  2. Call `DataProxyService.GetActionData` with the same `action_id`.
  3. Convert the `dataproxy.GetActionDataResponse` to `workflow.GetActionDataResponse` (schemas need to be checked for 1:1 compatibility — they currently both use `task.Inputs` / `task.Outputs` so this should be trivial).
- Remove the direct `dataStore` / `repo` reads from this RPC once delegation is in place. The URI resolution logic already lives behind `GetActionDataURIs`, which DataProxy calls.
- Keep the `runs` RPC registered so existing clients keep working.

## Open questions

- Response compatibility: confirm `dataproxy.GetActionDataResponse` ⇄ `workflow.GetActionDataResponse` field compatibility (both appear to use `task.Inputs` / `task.Outputs`; double-check trace-action output handling, which is special-cased in the run-service implementation around lines 794-818).
- Whether trace-action output handling (reading from `RunInfo.OutputsUri`) is already covered on the DataProxy side — if not, that logic needs to move to DataProxy as part of this work, not be left in the run-service shim.

## References

- `runs/service/run_service.go:741` — current `RunService.GetActionData`
- `dataproxy/service/dataproxy_service.go:444` — DataProxy `GetActionData` (target delegate)
- #7252 — sibling issue doing the same delegation for `TailLogs`

BarryWu0812:
Delegate RunService.GetActionData to DataProxyService to remove duplicate implementation

Closes #7253

Today both services:

- Hold a `storage.DataStore` and implement parallel reads of `inputs.pb` / outputs via `errgroup`.

- Resolve storage URIs from the action record (DataProxy goes through RunService.GetActionDataURIs; RunService reads them directly from the DB).

- Must stay in lock-step whenever we evolve the storage layout, artifact types, or error semantics (not found, partial results, trace vs task actions, etc.).

This is duplicate surface area with no good reason now that DataProxy owns data access. It also makes the storage-layout contract implicit in two places.

Refactor `RunService.GetActionData` into a thin shim that forwards to `DataProxyServiceClient.GetActionData`:

- Inject a `dataproxyconnect.DataProxyServiceClient` into `RunService` (likely the in-process one for the all-in-one binary).
- In `GetActionData`:

  1. Validate the request as today.
  2. Call `DataProxyService.GetActionData` with the same `action_id`.
  3. Convert the `dataproxy.GetActionDataResponse` to `workflow.GetActionDataResponse` (schemas need to be checked for 1:1 compatibility — they currently both use `task.Inputs` / `task.Outputs` so this should be trivial).
- Remove the direct `dataStore` / `repo` reads from this RPC once delegation is in place. The URI resolution logic already lives behind `GetActionDataURIs`, which DataProxy calls.
- Keep the `runs` RPC registered so existing clients keep working.

### Labels

## Docs link

- `main` 
  - \#6583
    - \*\*Delegate RunService.GetActionData to DataProxyService \*\* :point\_left:

### DIFF
--- a/dataproxy/service/dataproxy_service.go
+++ b/dataproxy/service/dataproxy_service.go
@@ -19,11 +19,11 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/flyteorg/flyte/v2/dataproxy/config"
+	"github.com/flyteorg/flyte/v2/dataproxy/logs"
 	"github.com/flyteorg/flyte/v2/flytestdlib/logger"
 	"github.com/flyteorg/flyte/v2/flytestdlib/storage"
 	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/common"
 	flyteIdlCore "github.com/flyteorg/flyte/v2/gen/go/flyteidl2/core"
-	"github.com/flyteorg/flyte/v2/dataproxy/logs"
 	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/dataproxy"
 	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/dataproxy/dataproxyconnect"
 	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/project"
@@ -470,6 +470,10 @@ func (s *Service) GetActionData(
 	ctx context.Context,
 	req *connect.Request[dataproxy.GetActionDataRequest],
 ) (*connect.Response[dataproxy.GetActionDataResponse], error) {
+	if err := req.Msg.Validate(); err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
+	}
+
 	actionId := req.Msg.GetActionId()
 
 	urisResp, err := s.runClient.GetActionDataURIs(ctx, connect.NewRequest(&workflow.GetActionDataURIsRequest{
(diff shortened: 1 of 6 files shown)


dataproxy/service/dataproxy_service.go
```diff
@@ -495,11 +499,17 @@ func (s *Service) GetActionData(
 			}
 			logger.Infof(groupCtx, "GetActionData: reading inputs from %s", inputRef)
 			if err := s.dataStore.ReadProtobuf(groupCtx, inputRef, resp.Inputs); err != nil {
-				logger.Errorf(groupCtx, "GetActionData: failed to read inputs from %s: %v", inputRef, err)
-				return connect.NewError(connect.CodeInternal, fmt.Errorf("failed to read inputs from %s: %w", inputRef, err))
+				if !storage.IsNotFound(err) {
```
pingsutw, line 502: Should we return an error if the input is not found?

dataproxy/service/dataproxy_service.go
```diff
@@ -508,11 +518,16 @@ func (s *Service) GetActionData(
 			logger.Infof(groupCtx, "GetActionData: reading outputs from %s", outputRef)
 			var inputsOrOutputs task.Inputs
 			if err := s.dataStore.ReadProtobuf(groupCtx, outputRef, &inputsOrOutputs); err != nil {
-				logger.Errorf(groupCtx, "GetActionData: failed to read outputs from %s: %v", outputRef, err)
-				return connect.NewError(connect.CodeInternal, fmt.Errorf("failed to read outputs from %s: %w", outputRef, err))
-			}
-			resp.Outputs = &task.Outputs{
-				Literals: inputsOrOutputs.GetLiterals(),
+				if !storage.IsNotFound(err) {
```
pingsutw, line 521: ditto

dataproxy/service/dataproxy_service.go
BarryWu0812: https://github.com/flyteorg/flyte/pull/7284/changes#diff-5698a432711a8b677ea5e6d843f3a66a0c8cff4ebfaa6d262b844456364b46fbL11-L790

I saw the original logic is to skip the error that input is not found, so I kept the same logic.

pingsutw: ah, i see

pingsutw (approved):

PR #7289, opened 2026-04-27
fix(executor): bound consecutive Plugin.Handle errors with a system-failure counter

pingsutw:

When the k8s API rejects pod creation via an admission webhook (e.g. `flyte-pod-webhook` denies the request because a referenced secret cannot be found in any configured secret manager), `launchResource` in `executor/pkg/plugin/k8s/plugin_manager.go` returned the error as a raw Go error with `pluginsCore.UnknownTransition` and no phase transition.

In `taskaction_controller.go`, when `Plugin.Handle` returns a Go `error`, the controller only logs `"Plugin Handle failed"`, emits a `FailedPluginHandle` event, and requeues with the default duration. It does **not** record a phase, increment any system-failure counter, or convert to `PermanentFailure`. As a result, the TaskAction stays in `Queued` forever — the webhook will keep denying the same request, and there is no max-system-failure handling on this path.

Repro:

```python
@env.task
def fn(x: int) -> int:
    print("my secret is: ", os.getenv("TEST1"))   # secret "test1" does not exist
    ...
```

Before the fix, the TaskAction loops forever with the webhook denial message in the controller log:

```
admission webhook "flyte-pod-webhook.flyte.org" denied the request:
  none of the secret managers injected secret [key:"test1" ...]
```

In `launchResource`, before the generic system-error fall-through:

1. Detect admission-webhook denials via a small helper `isAdmissionWebhookDenial(err)` that matches the canonical error string (`"admission webhook"` + `"denied the request"`).
2. Also detect `k8serrors.IsInvalid(err)` (invalid pod specs are equally non-transient).
3. Return `pluginsCore.PhaseInfoFailure("AdmissionDenied", err.Error(), nil)` — a permanent failure phase — so the TaskAction terminates immediately with the underlying webhook message preserved for the user.

## How was this patch tested?

Screenshot 2026-04-27 at 11 56 45 AM

Manual end-to-end test using the bundled devbox:

1. Rebuilt `flyte-devbox:latest` with this change (`make -C docker/devbox-bundled build`) and restarted the devbox.
2. Ran a task that references a non-existent secret (`secrets=["test1"]`).
3. Confirmed the TaskAction transitions to `PermanentFailure` in \~8s (single attempt) instead of getting stuck:

```
NAMESPACE                 NAME                      RUN                    TASKTYPE   STATUS             AGE
flytesnacks-development   rf5v6h9d79n2ndnnqkpr-a0   rf5v6h9d79n2ndnnqkpr   python     PermanentFailure   8s
```

`Error State`:

```
Code:    AdmissionDenied
Kind:    USER
Message: admission webhook "flyte-pod-webhook.flyte.org" denied the request: ...
```

### Labels

- **fixed**

* `main` 
  - \#6583
    - **fix(executor): bound consecutive Plugin.Handle errors with a system-failure counter** :point\_left:

### DIFF
--- a/executor/pkg/controller/taskaction_controller.go
+++ b/executor/pkg/controller/taskaction_controller.go
@@ -328,6 +328,11 @@ func (r *TaskActionReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return r.recordSystemError(ctx, taskAction, originalTaskActionInstance, p.GetID(), systemErrorFromPhaseInfo(phaseInfo))
 	}
 
+	// Reset the consecutive system-failure counter on any non-system-error
+	// transition so transient blips earlier in the lifecycle don't accumulate
+	// toward the permanent-failure threshold.
+	taskAction.Status.SystemFailures = 0
+
 	// In-place task restart on USER-kind retryable failure: bump Status.Attempts
 	// and relaunch the pod under the same TaskAction.
 	var restartAttempts uint32
--- a/executor/pkg/controller/taskaction_controller.go
+++ b/executor/pkg/controller/taskaction_controller.go
@@ -55,6 +55,14 @@ const (
 	TaskActionDefaultRequeueDuration = 10 * time.Second
 	taskActionFinalizer              = "flyte.org/plugin-finalizer"
 
+	// DefaultMaxSystemFailures bounds consecutive system errors before the
+	// TaskAction is forced to PermanentFailure.
+	DefaultMaxSystemFailures uint32 = 3
+
+	// MaxSystemFailuresExceededCode is the ExecutionError.Code stamped on
+	// TaskActions that hit the system-failure ceiling.
+	MaxSystemFailuresExceededCode = "MaxSystemFailuresExceeded"
+
 	// LabelTerminationStatus marks a TaskAction as terminated for GC discovery.
 	LabelTerminationStatus = "flyte.org/termination-status"
 	// LabelCompletedTime records the UTC time (minute precision) when the TaskAction became terminal.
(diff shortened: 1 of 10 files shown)


executor/pkg/controller/taskaction_controller.go
```diff
@@ -77,16 +85,116 @@ const (
...
+) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	logger.Error(handleErr, "system error from plugin", "plugin", pluginID)
+	if r.Recorder != nil {
+		r.Recorder.Eventf(taskAction, corev1.EventTypeWarning, string(FailedPluginHandle),
+			"Plugin %q system error: %v", pluginID, handleErr)
+	}
+
+	taskAction.Status.SystemFailures++
+	if taskAction.Status.SystemFailures > r.maxSystemFailures() {
```
AdilFayyaz, line 160: shouldn't this be >=

executor/pkg/controller/taskaction_controller.go
```diff
@@ -77,16 +85,116 @@ const (
...
+	handleErr error,
+) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	logger.Error(handleErr, "system error from plugin", "plugin", pluginID)
+	if r.Recorder != nil {
+		r.Recorder.Eventf(taskAction, corev1.EventTypeWarning, string(FailedPluginHandle),
+			"Plugin %q system error: %v", pluginID, handleErr)
+	}
+
+	taskAction.Status.SystemFailures++
```
AdilFayyaz, line 159: are we resetting this back to 0? Suppose we hit 2 system errors and then a success, shouldn't this be reset?

executor/pkg/controller/taskaction_controller.go
pingsutw: It’s intentional. If maxSystemFailures is 1, we want the task to retry once. If we use `>=` here, the phase will be converted to a permanent error and won’t retry

executor/pkg/controller/taskaction_controller.go
pingsutw: I'm not sure actually. we don't reset it in v1.

AdilFayyaz: wouldnt it be an issue for long running tasks that may have transient failures followed by successes in between. Task could silently die after accumulating nonconsecutive `maxSystemFailures`

pingsutw: that makes sense, let me update it

pingsutw: updated it

AdilFayyaz (approved):

PR #7154, opened 2026-04-03
[V2] App Service design doc

pingsutw:
Design doc and spec for implementing the App Service in OSS Flyte v2, enabling deployment of long-running applications (serving endpoints, LLM inference, APIs) alongside workflow executions. Covers data model, service layer, K8s integration via ActionsService pattern (KService CRs), and phased implementation plan.

### Labels

Please add one or more of the following labels to categorize your PR:

- **added**: For new features.
- **changed**: For changes in existing functionality.
- **deprecated**: For soon-to-be-removed features.
- **removed**: For features being removed.
- **fixed**: For any bug fixed.
- **security**: In case of vulnerabilities

This is important to improve the readability of release notes.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

## Docs link

- `main` 
  - \#6583
    - **\[V2] App Service design doc** :point\_left:

PR #7296, opened 2026-04-28
[V2] Add App events

machichima:

Show APP events

- Add `app_conditions` table in DB to store conditions (events) for app
- Append conditions to DB when informer cache watch event triggered

## How was this patch tested?

- Serve app through devbox and check if we can see events

image

### Labels

## Docs link

- `main` 
  - \#6583
    - **\[V2] Add App events** :point\_left:


app/internal/migrations/migrations.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/flyteorg/flyte/v2/flytestdlib/database"
+)
+
+//go:embed sql/*.sql
+var migrationFS embed.FS
+
+// RunMigrations applies all pending app migrations.
+func RunMigrations(ctx context.Context, db *sqlx.DB) error {
+	return database.Migrate(ctx, db, "app", migrationFS)
```
pingsutw, line 17: We intentionally don’t use the database in the app service for simplicity. Could we directly get the events from the CRD when watching the events

pingsutw: https://github.com/flyteorg/flyte/pull/7154/files#diff-a74f1bc2eb071647a301a4510f31826e6bb1dead03f72dfdf1b8310974f306feR68-R104

machichima: In this case we will miss historical conditions as k8s conditions will only store the latest state

```sh
  Conditions:
    Last Transition Time:        2026-04-28T07:41:44Z
    Status:                      True
    Type:                        ConfigurationsReady
    Last Transition Time:        2026-04-28T07:41:44Z
    Status:                      True
    Type:                        Ready
    Last Transition Time:        2026-04-28T07:41:44Z
    Status:                      True
    Type:                        RoutesReady
```

pingsutw: > we will miss historical conditions

that's okay

Issue #7307, opened 2026-04-28
Support for exp backoff in retries and various timeout knobs


kumare3:
Today's controls collapse three different failure questions into one knob:

- **"How long can this attempt run?"** — answered by `max_runtime`.
- **"How long am I willing to wait for this to start?"** — unanswerable. An action stuck in `QUEUED` or `WAITING_FOR_RESOURCES` waits forever (or until a parent timeout fires elsewhere).
- **"What is the total wall-clock budget for this work, including retries?"** — unanswerable. With `retries=5` and `max_runtime=1h`, an action can legally consume 6 hours plus queue time before it gives up.

Retries have the same flattening problem. `retries=N` means "try up to N+1 times back-to-back," with no way to say "wait 30s, then 1m, then 2m before retrying" — which is exactly the right policy for a flaky external API. Authors today either embed manual backoff inside the task body (which loses observability) or accept that retries will hammer a recovering downstream into the ground.

We want to support
```python
@env.task(
    retries=flyte.RetryStrategy(
        count=5,
        backoff=flyte.Backoff(base=timedelta(seconds=30), factor=2.0, cap=timedelta(minutes=5)),
    ),
    timeout=flyte.Timeout(
        max_runtime=timedelta(minutes=10),
        deadline=timedelta(hours=1),
    ),
)
async def bounded_retries():
    ...
```

PR #7308, opened 2026-04-28
RetryStrategy and TimeoutStrategy

kumare3:
This PR adds and updates the way Retries and Timeouts are handled in the system. This is just the definition, full implementation will follow

## Tracking issue
Part of #7307

## Follow up
Backend update and flyte-sdk


EngHabu (dismissed):

wild-endeavor (dismissed):
I didn't know we were bringing back the other two messages... will let Haytham give the final stamp since I think he has to anyways.

EngHabu (approved):

PR #7323, opened 2026-04-30
Inject INTERNAL_APP_ENDPOINT_PATTERN into KService pods for in-cluster app discovery

AdilFayyaz:
When multiple Flyte apps run in the same cluster (e.g. a Gradio app calling a vLLM model-serving app), apps need to call each other via the internal Kubernetes service URL (*.flyte.svc.cluster.local). Previouly the `INTERNAL_APP_ENDPOINT_PATTERN` env var — which the Flyte SDK's `AppEndpoint` uses to resolve internal URLs — was never injected into KService pods. This meant `AppEndpoint(app_name="...")` had no pattern to resolve against, and developers had to manually construct and hardcode the full internal cluster URL.

 - Inject `INTERNAL_APP_ENDPOINT_PATTERN` into every KService container at deploy time  in buildKService. The value is a URL template with an {app_fqdn} placeholder that the SDK substitutes at runtime with the target app's name:                         
                                                                  
  `http://{app_fqdn}-{project}-{domain}.flyte.svc.cluster.local`
                                                                                     
 - The project and domain are baked in as a suffix to match Flyte's KService naming convention ({name}-{project}-{domain}). When the SDK resolves AppEndpoint(app_name="gemma4-26b") in project flytesnacks / domain development, it produces:                                                       
                                                          
  `http://gemma4-26b-flytesnacks-development.flyte.svc.cluster.local`

 - No auth is required on this path (bypasses ingress), and traffic stays in-cluster

### Code Changes                                                                                         
  - `app/config/config.go` — adds `NamespacedNameSuffixTemplate` field to `InternalAppConfig` 
  - `app/internal/k8s/app_client.go` — injects `INTERNAL_APP_ENDPOINT_PATTERN` in `buildKService`   

## How was this patch tested?

 Verified using the `app_calling_app` SDK example (`examples/apps/app_calling_app/app.py`). After this change, hitting `app2-calls-app1.../greeting/world` correctly proxies through to app1 via the internal cluster URL

## Stack

### DIFF
--- a/app/config/config.go
+++ b/app/config/config.go
@@ -65,6 +65,11 @@ type InternalAppConfig struct {
 	// processes need to connect back to the Flyte manager.
 	DefaultEnvVars map[string]string `json:"defaultEnvVars" pflag:"-,Default env vars injected into every app pod"`
 
+	// NamespacedNamePrefixTemplate is the template for generating the INTERNAL_APP_ENDPOINT_PATTERN env var.
+	// Supported variables: {{ project }}, {{ domain }}, {{ org }}.
+	// Example: "{{ project }}-{{ domain }}-"
+	NamespacedNamePrefixTemplate string `json:"namespacedNamePrefixTemplate" pflag:",Template for internal app endpoint pattern (e.g., {{ project }}-{{ domain }}-)"`
+
 	// WatchBufferSize is the buffer size for each subscriber's event channel.
 	// A larger value reduces the chance of dropped events under burst load.
 	WatchBufferSize int `json:"watchBufferSize" pflag:",Buffer size for watch subscriber channels"`
--- a/app/config/config.go
+++ b/app/config/config.go
@@ -65,10 +65,10 @@ type InternalAppConfig struct {
 	// processes need to connect back to the Flyte manager.
 	DefaultEnvVars map[string]string `json:"defaultEnvVars" pflag:"-,Default env vars injected into every app pod"`
 
-	// NamespacedNamePrefixTemplate is the template for generating the INTERNAL_APP_ENDPOINT_PATTERN env var.
-	// Supported variables: {{ project }}, {{ domain }}, {{ org }}.
-	// Example: "{{ project }}-{{ domain }}-"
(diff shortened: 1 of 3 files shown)


app/config/config.go
```diff
@@ -65,6 +65,11 @@ type InternalAppConfig struct {
 	// processes need to connect back to the Flyte manager.
 	DefaultEnvVars map[string]string `json:"defaultEnvVars" pflag:"-,Default env vars injected into every app pod"`
 
+	// NamespacedNamePrefixTemplate is the template for generating the INTERNAL_APP_ENDPOINT_PATTERN env var.
+	// Supported variables: {{ project }}, {{ domain }}, {{ org }}.
+	// Example: "{{ project }}-{{ domain }}-"
+	NamespacedNamePrefixTemplate string `json:"namespacedNamePrefixTemplate" pflag:",Template for internal app endpoint pattern (e.g., {{ project }}-{{ domain }}-)"`
```
pingsutw: It's for dev mode (running manager locally) right? Could we add a default here https://github.com/flyteorg/flyte/pull/7323/files#diff-2a2301646ea40f7638775b9a0cfbd0cb8c9f081b5f8732a73f44767aeae95e9dL23-L26

AdilFayyaz: this isnt only for the dev mode, it should work in both

pingsutw: Got it. Is it supposed to be used in app_client.go? I didn’t see any files using it

AdilFayyaz: It's coming from the sdk

app/internal/k8s/app_client.go
```diff
@@ -536,6 +545,18 @@ func (c *AppK8sClient) buildKService(app *flyteapp.App) (*servingv1.Service, err
 		podSpec.Containers[0].Env = append(defaults, podSpec.Containers[0].Env...)
 	}
 
+	// Inject INTERNAL_APP_ENDPOINT_PATTERN so app code can construct internal cluster URLs
+	// for other apps by substituting {app_fqdn} with the target app name.
+	// The suffix is rendered from NamespacedNamePrefixTemplate to match the KService name format
```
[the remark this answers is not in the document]
pingsutw: should we update the comment?

pingsutw (approved):

PR #7341, opened 2026-05-05
fix: make pod_name and container_name optional in LoggingContext; add ContainerIdentifier to TailLogsResponse

pingsutw:
## Why

`LoggingContext.kubernetes_pod_name` and `kubernetes_container_name` had a `min_len = 1` buf validation constraint, which rejected label-based log requests where the caller provides pod labels instead of a specific pod name. Union Cloud's DP dataproxy uses label-based queries for multi-pod jobs (e.g. Ray) where the specific pod name is not known upfront.

`TailTaskExecutionLogsResponse.Logs` was also missing a `container` field needed by the client to attribute log lines to specific pod/container pairs in the UI dropdown.

## What

- Remove `min_len` from `LoggingContext.kubernetes_pod_name` and `kubernetes_container_name` fields
- Add `ContainerIdentifier container = 3` to `TailTaskExecutionLogsResponse.Logs`
- Regenerate all gen files

## Testing

Verified via Union Cloud DP dataproxy integration — label-based log queries now pass proto validation and the container identifier is correctly surfaced to the client.


flyteidl2/dataproxy/dataproxy_service.proto
```diff
@@ -203,8 +203,15 @@ message TailLogsRequest {
   // The attempt number.
   uint32 attempt = 2 [(buf.validate.field).uint32.gt = 0];
 
-  // The pod name to tail logs from. If not provided, attempt to find the primary pod, else assume the first pod.
-  string pod_name = 3;
+  // Selects which pod(s) to tail. If unset, defaults to primary_pod.
+  oneof pod_selector {
```
katrogan, line 207: nit: maybe we can call this pod_filter and then we can drop the all_pods option altogether since we default to it when no filter is set

flyteidl2/dataproxy/dataproxy_service.proto
```diff
@@ -203,8 +203,15 @@ message TailLogsRequest {
...
+    // Tail logs from the primary pod only.
+    bool primary_pod = 6;
+    // Tail logs from all pods. This is the default when pod_selector is unset.
```
katrogan: this is a departure from the currently checked in behavior - will it cause any issues for older UI clients?

katrogan (dismissed):

flyteidl2/dataproxy/dataproxy_service.proto
pingsutw: sorry for the confusion. The default is primary pod (same as before). I just updated the comment

flyteidl2/dataproxy/dataproxy_service.proto
pingsutw: updated it

flyteidl2/dataproxy/dataproxy_service.proto
```diff
@@ -204,10 +204,10 @@ message TailLogsRequest {
   uint32 attempt = 2 [(buf.validate.field).uint32.gt = 0];
 
   // Selects which pod(s) to tail. If unset, defaults to primary_pod.
-  oneof pod_selector {
-    // Tail logs from the primary pod only.
+  oneof pod_filter {
+    // Tail logs from the primary pod only. This is the default when pod_filter is unset.
```
katrogan: oh I see, if this is the default totally fine keeping the oneof field name pod_selector then

katrogan (dismissed):

flyteidl2/dataproxy/dataproxy_service.proto
pingsutw: updated it

EngHabu (dismissed):

flyteidl2/core/execution.proto
```diff
@@ -151,6 +151,7 @@ message LogContext {
   repeated PodLogContext pods = 1;
   string primary_pod_name = 2;
   ConnectorLogContext connector = 3;
+  string pod_name_prefix = 4;
```
katrogan, line 155: nit: can we add the same comment from logs/dataplane/payload.proto here?

katrogan (dismissed):

flyteidl2/dataproxy/dataproxy_service.proto
```diff
@@ -203,8 +203,15 @@ message TailLogsRequest {
...
-  // The pod name to tail logs from. If not provided, attempt to find the primary pod, else assume the first pod.
-  string pod_name = 3;
+  // Selects which pod(s) to tail. If unset, defaults to primary_pod.
+  oneof pod_selector {
+    // Tail logs from the primary pod only. This is the default when pod_selector is unset.
+    bool primary_pod = 6;
+    // Tail logs from all pods.
+    bool all_pods = 5;
+    // Tail logs from the named pod only.
+    string pod_name = 3;
```
katrogan, line 213: would we want to support pod_name_prefix here too?

katrogan: what about other filters (repeated pod_names for example)

pingsutw: Not necessary now. If pod selector is equal all_pods, the log service will use the podNamePrefix in the log context to fetch the logs from all the workers

pingsutw: We can add them in the future when needed

flyteidl2/core/execution.proto
pingsutw: added

katrogan (approved):

PR #7343, opened 2026-05-05
fix: stop Knative apps privately + keep UI state accurate after stop/start

AdilFayyaz:

Users expect “Stop App” to (1) scale the app down to zero and (2) make the public ingress inaccessible. The prior approach relied on `autoscaling.knative.dev/max-scale: "0"` as a stop signal, but in Knative that value represents “unlimited” upper bound, so it didn’t reliably enforce the intended stop semantics. Separately, the UI polls/Get calls can briefly show stale “stopped” state during stop→start transitions because the control plane can cache transitional responses while the data plane is still converging.

 - Replaced the incorrect max-scale=0 stop semantics with Knative-native behavior:
 - Stop() labels the Knative Service `networking.knative.dev/visibility=cluster-local` so it is not published to the external gateway (public ingress becomes inaccessible).
 - Stop() marks the Service with `flyte.org/app-stopped=true` as the control-plane source of truth for STOPPED state.
 - Stop() sets `autoscaling.knative.dev/min-scale="0"` and `autoscaling.knative.dev/initial-scale="0"` to converge cleanly to zero pods.
 - Stop() deletes the latest ready Revision so existing pods terminate immediately (no waiting for the stable window).
 - Deploy() (“Start”) clears the stopped/private labels so the service becomes externally routable again.
 - Status mapping now treats STOPPED based on `flyte.org/app-stopped` (not max-scale).
 
### Devbox / cluster configuration
 - Enabled `config-autoscaler.allow-zero-initial-scale="true"` in devbox-bundled kustomize overlays and rendered manifests.
 - Updated `docker/devbox-bundled/Makefile` setup-knative to also patch config-autoscaler so the “manual install” path matches the bundled manifests behavior.

### UI correctness / control-plane behavior
 - Added a small TTL cache in the control-plane AppService for Get responses to reduce cross-plane chatter, while explicitly avoiding caching “transitional” stop/start windows (desired state running, but status still stopped) so the UI doesn’t get stuck showing STOPPED after clicking Start.

## How was this patch tested?
 - Unit tests:
   - go test ./app/internal/k8s
   - go test ./app/service
   - Manual verification (devbox / cluster):
       - Create app → confirm public URL reachable.
       - Stop app → confirm pods scale to 0 and public URL is not reachable (service is cluster-local).
       - Start app → confirm visibility label cleared, service becomes publicly reachable again, and UI
      transitions to ACTIVE without being stuck in STOPPED due to caching.
  

## Stack

### DIFF
--- a/app/internal/k8s/app_client.go
+++ b/app/internal/k8s/app_client.go
@@ -52,7 +52,11 @@ const (
 	annotationAppOrg  = "flyte.org/app-org"
 	annotationSpec    = "flyte.org/spec"
 
-	maxScaleZero = "0"
+	labelAppStopped        = "flyte.org/app-stopped"
+	labelKnativeVisibility = "networking.knative.dev/visibility"
+	visibilityClusterLocal = "cluster-local"
+
+	scaleZero = "0"
 
 	// maxKServiceNameLen is the Kubernetes DNS label limit.
 	maxKServiceNameLen = 63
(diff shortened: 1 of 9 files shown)


pingsutw (commented):
Looks good, one nit, should we override the desired state (if labelAppStopped is set to true) here before returning the spec?
https://github.com/flyteorg/flyte/blob/6b423494858c478c5be306ee63bb097e40e39a9a/app/internal/k8s/app_client.go#L449

AdilFayyaz:
@pingsutw yes we should override the desired state. Although, the UI did seem to toggle the button correctly. It should be reading from the spec but probably its reading the status from the watch events.

pingsutw (approved):

PR #7350, opened 2026-05-07
Measure plugin initializing latency and fix queue latency

fg91:

* Flytepropeller currently emits a metric measuring how much time is spent by a plugin in the queued phase, i.e. a Pod waiting for being scheduled on a node.
  * This metric currently undercounts the queued time because the time stamp when we transitioned into this phase is reset too often. This PR fixes this.
* I now also add a metric measuring how much time is spent  by a plugin in the initializing phase, i.e. how long it takes for a Pod to transition from `ContainerCreating` to `Running`.

* Fix measurement of phase transition time stamps
* Add a new metric that measures duration in initializing phase

## How was this patch tested?

* Added unit test
* Ran propeller with this change in a staging cluster and compared metrics to measured transitioning times


flytepropeller/pkg/controller/nodes/task/handler.go
```diff
@@ -553,6 +554,14 @@ func (t Handler) invokePlugin(ctx context.Context, p pluginCore.Plugin, tCtx *ta
 		}
 	}
 
+	// Emit the initializing latency if the task has just transitioned from Initializing to Running.
```
fg91, line 557: This metric is new. (I want to measure progress regarding startup times, i.e. image pulls.)

flytepropeller/pkg/controller/nodes/task/handler.go
```diff
@@ -800,12 +809,20 @@ func (t Handler) Handle(ctx context.Context, nCtx interfaces.NodeExecutionContex
...
+	lastPhaseUpdatedAt := ts.LastPhaseUpdatedAt
+	if ts.PluginPhase != pluginTrns.pInfo.Phase() {
+		lastPhaseUpdatedAt = time.Now()
+	}
 	err = nCtx.NodeStateWriter().PutTaskNodeState(handler.TaskNodeState{
 		PluginState:                        pluginTrns.pluginState,
 		PluginStateVersion:                 pluginTrns.pluginStateVersion,
 		PluginPhase:                        pluginTrns.pInfo.Phase(),
 		PluginPhaseVersion:                 pluginTrns.pInfo.Version(),
-		LastPhaseUpdatedAt:                 time.Now(),
```
fg91, line 808: We called this not only when transitioning from one phase to another but also within a single phase when the phase version was increased. This lead to wrong queue latency measurements.

flytepropeller/pkg/controller/nodes/task/handler_test.go
```diff
@@ -1313,3 +1314,264 @@ func Test_task_Handle_ValidateOutputErr(t *testing.T) {
...
+// actually changes. Intra-phase version bumps (which MaybeUpdatePhaseVersion
+// produces when a plugin's reason string changes — e.g. as a Pending pod
+// progresses through scheduler reasons during autoscaling) must preserve
+// the timestamp.
+//
+// The plugin_queue_latency and plugin_initialize_latency metrics anchor
+// off this timestamp; if version bumps reset it, both metrics undercount
+// and only capture the final segment before exiting the phase.
+func Test_task_Handle_LastPhaseUpdatedAt(t *testing.T) {
+	inputs := &core.LiteralMap{
```
fg91, line 1329: The boilerplate is mostly taken from existing tests above.

go.mod
fg91, line 1: Otherwise CI failed: https://github.com/flyteorg/flyte/actions/runs/25498439222/job/74824407222?pr=7350

pingsutw (approved):

Sovietaced:
Not necessary but I wonder if there is a way to assert the metrics are also called/updated. I am familiar with how to do this using open telemetry libraries but not as sure with prometheus.

Sovietaced (approved):

Sovietaced (approved):

PR #7363, opened 2026-05-11
Fix incorrect "not found" error check.

NullHypothesis:

So far, the following command would fail:
```
$ flytectl get workflow-execution-config -p PROJECT -d DOMAIN --attrFile FILE --gen
Error: attribute not found
```

The attribute file is not written.

The above command failed because the command line parsing code would check the error returned by `FetchAndUnDecorateMatchableAttr` for a gRPC "not found" error but `AdminFetcherExtClient` maps this gRPC code to an internal `NotFoundError` type: It is this error that we need to check for.

This PR adapts the error check, which unbreaks attribute file generation. (While at it, I also added a missing '\n' to a log message.)

## How was this patch tested?

I ran the above command again. It now works.


Sovietaced:
I checked the code and this seems sound to me.

Sovietaced (approved):

PR #7366, opened 2026-05-12
fix(propeller/k8s): treat BadRequest / Invalid from apiserver as permanent failure

### LINKED ISSUE #6531: [BUG] Flyte Propeller treats a BadRequest code from K8s when launching a resource as a retryable failure

1fanwang:
## Tracking issue

Closes #6531

## Why

When propeller creates a task pod and the apiserver rejects it with HTTP 400 (`BadRequest`) or HTTP 422 (`Invalid`) — typically a validating admission webhook saying no, or an invalid field caught by the apiserver — the failure path falls through to the default branch and surfaces as a retryable system error. The workflow then burns through its retry budget making the exact same request, with the exact same outcome.

These errors are intrinsic to the request payload. Retrying with the same input produces the same rejection. Neighboring `EntityTooLarge` is already handled this way.

The right transition was already written but commented out since #1 (2019) with no documented rationale. Uncommenting and adding a regression test.

## What changed

- `flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go`: uncomment the `PhaseInfoFailure("BadTaskFormat", ...)` transition so BadRequest / Invalid produce a permanent failure instead of falling through to the system-error path. Added a short comment so this doesn't get re-commented in the future.

## How was this patch tested?

Added two sub-tests to `TestK8sTaskExecutor_Handle_LaunchResource` — `jobBadRequest` (using `k8serrors.NewBadRequest`) and `jobInvalid` (using `k8serrors.NewInvalid`). Both assert `pluginsCore.PhasePermanentFailure` with error code `BadTaskFormat`. Both fail on master and pass with this PR; the full `plugin_manager_test.go` suite stays green.

```
ok  	github.com/flyteorg/flyte/flytepropeller/pkg/controller/nodes/task/k8s	1.724s
```

### DIFF
--- a/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go
+++ b/flytepropeller/pkg/controller/nodes/task/k8s/plugin_manager.go
@@ -262,8 +262,14 @@ func (e *PluginManager) launchResource(ctx context.Context, tCtx pluginsCore.Tas
 		} else if k8serrors.IsForbidden(err) {
 			return pluginsCore.DoTransition(pluginsCore.PhaseInfoRetryableFailure("RuntimeFailure", err.Error(), nil)), nil
 		} else if k8serrors.IsBadRequest(err) || k8serrors.IsInvalid(err) {
+			// BadRequest (HTTP 400) and Invalid (HTTP 422) errors are intrinsic
+			// to the request payload and not transient. The most common source
+			// is a validating admission webhook rejecting the pod spec; retrying
+			// with the same input will produce the same rejection. Treat as a
+			// permanent failure so the validation error surfaces to the user
+			// instead of exhausting the workflow's retry budget.
 			logger.Errorf(ctx, "Badly formatted resource for plugin [%s], err %s", e.id, err)
-			// return pluginsCore.DoTransition(pluginsCore.PhaseInfoFailure("BadTaskFormat", err.Error(), nil)), nil
+			return pluginsCore.DoTransition(pluginsCore.PhaseInfoFailure("BadTaskFormat", err.Error(), nil)), nil
 		} else if k8serrors.IsRequestEntityTooLargeError(err) {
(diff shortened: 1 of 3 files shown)


Sovietaced (approved):

Sovietaced:
Strange that this code has been commented out for 7 years but it sure seems like uncommenting it is the right move.

PR #7371, opened 2026-05-13
fix(executor): wire previous checkpoint path for in-place retries

AdilFayyaz:

On retries (`Status.Attempts >= 2`), `NewTaskExecutionContext` passed "" as the previous checkpoint path to `NewCheckpointRemoteFilePaths`. As a result, `{{.PrevCheckpointPrefix}}` substituted to empty in pod args, `checkpoint.load_sync()` always returned `None`, and tasks restarted from scratch on every retry.

Mirrors the worker-side fix from unionai/cloud#15143.

When `Status.Attempts >= 2`, derive the previous checkpoint path via `ComputeActionOutputPath(..., attempt-1) + ConstructCheckpointPath` and pass it through.

## How was this patch tested?
  - `go test ./executor/pkg/plugin/... ./executor/pkg/controller/...`
  - End-to-end sdk example: `examples/checkpoint/generic_data_checkpoint.py` succeeds on attempt 4

## Stack

### DIFF
--- a/executor/pkg/plugin/task_exec_context.go
+++ b/executor/pkg/plugin/task_exec_context.go
@@ -145,7 +145,17 @@ func NewTaskExecutionContext(
 		return nil, err
 	}
 	rawOutputPaths := ioutils.NewRawOutputPaths(ctx, outputPrefix)
-	outputFilePaths := ioutils.NewCheckpointRemoteFilePaths(ctx, dataStore, outputPrefix, rawOutputPaths, "")
+
+	var prevCheckpointPath storage.DataReference
+	if taskAction.Status.Attempts >= 2 {
+		prevOutputPrefix, err := ComputeActionOutputPath(ctx, taskAction.Namespace, taskAction.Name, taskAction.Spec.RunOutputBase, taskAction.Spec.ActionName, attempt-1)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


executor/pkg/plugin/task_exec_context.go
```diff
@@ -145,7 +145,17 @@ func NewTaskExecutionContext(
 		return nil, err
 	}
 	rawOutputPaths := ioutils.NewRawOutputPaths(ctx, outputPrefix)
-	outputFilePaths := ioutils.NewCheckpointRemoteFilePaths(ctx, dataStore, outputPrefix, rawOutputPaths, "")
+
+	var prevCheckpointPath storage.DataReference
+	if taskAction.Status.Attempts >= 2 {
```
pingsutw, line 152: nit: could we add a small comment to explain why we use 2 here? like first attempt start from 1 and there is no checkpoint for the first attempt or something like that.

pingsutw (dismissed):

executor/pkg/plugin/task_exec_context.go
AdilFayyaz: yeah ofc

pingsutw (approved):

PR #7425, opened 2026-05-26
feat(workflow): add RunService.SignalEvent + EventPayload

SVilgelm:
SignalEvent gives UI/CLI a run-scoped endpoint to resolve a paused condition action. It carries the user-facing typed EventPayload (bool/int/float/string oneof) and an ActionIdentifier; the run service converts the payload to a core.Literal and delegates to ActionsService.Signal.


flyteidl2/workflow/run_service.proto
```diff
@@ -365,6 +371,36 @@ message AbortActionRequest {
...
+}
+
+// Request message for signalling a condition action within a run.
+message SignalEventRequest {
+  // The condition action to signal.
+  common.ActionIdentifier action_id = 1 [(buf.validate.field).required = true];
+
+  // The parent action that owns the condition. Used to route the request to
+  // the correct actions-service shard (partitioned by parent action name).
+  string parent_action_name = 2;
```
pingsutw, line 395: Is it possible to get the parent action name in the run service somehow? so we don't need to pass it at the client side

flyteidl2/workflow/run_service.proto
```diff
@@ -365,6 +371,36 @@ message AbortActionRequest {
...
+// user-facing simplified payload surface — the run service converts it to a
+// core.Literal before forwarding to ActionsService.Signal. Exactly one field
+// must be set, matching the condition's declared type.
+message EventPayload {
+  oneof value {
+    option (buf.validate.oneof).required = true;
+    bool bool_value = 1;
+    int64 int_value = 2;
+    double float_value = 3;
+    string string_value = 4;
```
pingsutw, line 384: Do we only want to support these types? or should we use literal?

flyteidl2/workflow/run_service.proto
SVilgelm: we will need to read from the database in the Run Service. Is it hard to pass from the client side?

flyteidl2/workflow/run_service.proto
SVilgelm: We use literal in the Actions Services and the Run Service does the conversion. @kumare3 explicitly mentioned only these types: https://www.notion.so/Conditions-3448cc06513d803eb4afd879b9cef1e9

kumare3 (approved):

PR #7430, opened 2026-05-27
Add labels to create run /list runs api.

afbrock:
## Tracking issue
https://linear.app/unionai/issue/PRD-422/be-listruns-should-include-run-labels

These changes add labels to the list runs api.  This is done by adding effectively a duplication of the labels that are currently in runspec into the top level run object, as list runs only returns a repeated run object instead of the full
run details object (which is the right thing to do.)   The other change is that I added operators specific to (and important for) filtering key value pairs into the common list filters enum.  A user can add flyte specific key value pair labels in order to organize and filter their runs by adding filters where the field is the concatenation of 'labels." and the key name.

## How was this patch tested?

Along with 
https://github.com/flyteorg/flyte-sdk/pull/1140
and https://github.com/unionai/cloud/pull/16161
running in devbox, runs can be created and filtered based on label data that is persisted along with run information.


flyteidl2/workflow/run_service.proto
```diff
@@ -130,7 +128,13 @@ message CreateRunRequest {
...
   RunSource source = 8;
 
-  // Next field number: 10
+  // User-defined labels for filtering and organizing this run. These are the same
+  // labels as RunSpec.labels (the k8s resource labels applied to pods/jobs at
+  // execution time, also configurable via settings). Labels provided here are
+  // merged with settings/RunSpec labels and take precedence over them.
+  map<string, string> labels = 10;
+
+  // Next field number: 11
```
EngHabu: Did you find a thing that does this??? We had that internally at Google and it was pretty helpful specially with huge Protos and wanting to add fields closer to other fields where it makes sense while numbers have to keep going monotonically up... just asking

flyteidl2/workflow/run_service.proto
```diff
@@ -130,7 +128,13 @@ message CreateRunRequest {
   // Indicates client that created this run.
   RunSource source = 8;
 
-  // Next field number: 10
+  // User-defined labels for filtering and organizing this run. These are the same
+  // labels as RunSpec.labels (the k8s resource labels applied to pods/jobs at
+  // execution time, also configurable via settings). Labels provided here are
+  // merged with settings/RunSpec labels and take precedence over them.
+  map<string, string> labels = 10;
```
EngHabu: These are already inside RunSpec, right? Maybe that's what you were stating yesterday that went over my head... why do we need it in both places within the proto?

flyteidl2/workflow/run_definition.proto
```diff
@@ -30,6 +37,13 @@ message RunDetails {
 
   // Detailed information about the root action.
   ActionDetails action = 2;
+
+  // User-defined labels attached to this run at creation time. These are the same
+  // labels as RunSpec.labels (the k8s resource labels applied to pods/jobs at
+  // execution time, also configurable via settings). On create, labels supplied on
+  // the request are merged with settings/RunSpec labels, with request labels taking
+  // precedence, and the merged set is stored here and on the RunSpec.
+  map<string, string> labels = 3;
```
EngHabu: It's already within RunSpec, right? ditto

flyteidl2/workflow/run_definition.proto
```diff
@@ -22,6 +22,13 @@ option go_package = "github.com/flyteorg/flyte/v2/gen/go/flyteidl2/workflow";
 message Run {
   // Lightweight information about the root action.
   Action action = 1;
+
+  // User-defined labels attached to this run at creation time. These are the same
+  // labels as RunSpec.labels (the k8s resource labels applied to pods/jobs at
+  // execution time, also configurable via settings). On create, labels supplied on
+  // the request are merged with settings/RunSpec labels, with request labels taking
+  // precedence, and the merged set is stored here and on the RunSpec.
+  map<string, string> labels = 2;
```
EngHabu, line 32: Why do we need it here? RunDetails message has the RunSpec which has Labels

flyteidl2/workflow/run_definition.proto
afbrock: right, i will get rid of this one.

flyteidl2/workflow/run_definition.proto
afbrock: It needs to be in this one because a list request returns Run not RunDetail.

flyteidl2/workflow/run_service.proto
afbrock: No tool, just did it manually (since there was a comment already present it was easy enough to maintain.)

flyteidl2/workflow/run_definition.proto
```diff
@@ -499,3 +507,25 @@ message TaskGroup {
...
+  LABEL_SELECTOR_OPERATOR_NOT_IN = 2;
+  LABEL_SELECTOR_OPERATOR_EXISTS = 3;
+  LABEL_SELECTOR_OPERATOR_DOES_NOT_EXIST = 4;
+}
+
+message LabelSelectorRequirement {
+  string key = 1;
+  LabelSelectorOperator operator = 2;
+  repeated string values = 3;
+}
```
pmahindrakar-oss: Can we  not use filter in common.ListRequest 
https://github.com/unionai/cloud/blob/8aab6e4ffbd883231790bb22a8f658e0564906c4/gen/pb-go/common/list.pb.go#L308-L318

pmahindrakar-oss (commented):
Can we use the existing ListRequest filter instead of introducing a parallel label selector filtering and the comparision parameter around those

flyteidl2/workflow/run_definition.proto
afbrock: no, we can't because we are matching key value pairs, and so the selector can express something like the key 'team' exists, or the key 'team' has values in 'infra' 'frontend' or 'ml'
the operators seem similar, but the requirements are really different than common.list.filters.

pmahindrakar-oss: I get that label selector semantics are different because they operate on the labels map, not directly on a top-level field.

But since Flyte runs are already filtered through ListRequest.filters, I’m wondering if labels should still be modeled inside the same filtering abstraction.

Example query:

status = SUCCEEDED
created_by = praful
label team in (infra, frontend, ml)
label env exists

With separate LabelSelectorRequirement, clients now need to split this into two APIs:

filters[]:
  status = SUCCEEDED
  created_by = praful

label_selectors[]:
  team in (infra, frontend, ml)
  env exists

Could we instead model this as one filter system, e.g.

filters[]:
  status EQUAL SUCCEEDED
  created_by EQUAL praful
  labels.team VALUE_IN [infra, frontend, ml]
  labels.env EXISTS

or add a filter target/scope like FIELD vs LABEL?

That keeps ListRequest as the single query abstraction while still preserving label-specific semantics.

flyteidl2/workflow/run_definition.proto
```diff
@@ -22,6 +22,14 @@ option go_package = "github.com/flyteorg/flyte/v2/gen/go/flyteidl2/workflow";
...
   // Lightweight information about the root action.
   Action action = 1;
+
+  // User-defined labels attached to this run at creation time. These are the same
+  // labels as RunSpec.labels (the k8s resource labels applied to pods/jobs at
+  // execution time, also configurable via settings). On create, labels supplied on
+  // the request are merged with settings/RunSpec labels, with request labels taking
+  // precedence, and the merged set is stored here and on the RunSpec.
+  // It is part of Run so that it gets returned in a list request.
+  map<string, string> labels = 2;
```
machichima, line 32: Could we use `Label` as type here?

https://github.com/flyteorg/flyte/blob/c206519975d1a9adc1515c3dc803aa1945827cd1/flyteidl2/task/run.proto#L72-L72

flyteidl2/workflow/run_definition.proto
```diff
@@ -499,3 +507,25 @@ message TaskGroup {
   // Average time from run creation to entering RUNNING phase.
   google.protobuf.Duration average_time_to_running = 14;
 }
+
+message LabelSelector {
+  // ALL of these key=value pairs must be present (AND semantics).
+  map<string, string> match_labels = 1;
+
+  // Set-based requirements — reserved for future use.
+  repeated LabelSelectorRequirement match_expressions = 2;
```
machichima: If we decided to add this, should we add `buf.validate` here to limit the max length? We can follow the limit set in Kubernetes

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

flyteidl2/workflow/run_definition.proto
afbrock: Ok, yeah I like that - let me double check that it works impl wise.  I do like the idea of just using common list but i couldnt see how it could work.  the dot separated labels key makes sense.

pmahindrakar-oss (dismissed):

flyteidl2/workflow/run_definition.proto
EngHabu: ah I see... is there a `Label` type to use here 👇🏽 ?

pmahindrakar-oss (approved):

PR #7433, opened 2026-05-28
Set catalog cache to noop when data catalog is disabled

### LINKED ISSUE #7155: [BUG] Datacatalog enable and catalog-cache setting mismatch cause DNS

Sovietaced:
## Tracking issue
Closes #7155

Following up on #7234

When `datacatalog.enabled=false` in the flyte-core Helm chart, the `catalog-cache` configuration still defaults to `type: datacatalog` with `endpoint: datacatalog:89`. Since the datacatalog service is not deployed in this case, every task execution triggers a DNS lookup failure:

```
Failed to check Catalog for previous results:
rpc error: code = Unavailable
  desc = connection error: dial tcp: lookup datacatalog on x.x.x.x:x: no such host
```

These two related settings are far apart in `values.yaml` (datacatalog toggle at line 273, catalog-cache config buried inside configmaps at line 1055), making it easy to miss the dependency.

In `charts/flyte-core/templates/propeller/configmap.yaml`, the `catalog.yaml` configmap entry is now rendered conditionally:

- When `datacatalog.enabled=true` (default): renders the existing `configmap.catalog` value as before, preserving full backward compatibility.
- When `datacatalog.enabled=false`: renders `catalog-cache.type: noop` so propeller skips catalog lookups entirely, preventing DNS failures.

## How was this patch tested?
Verified with `helm template` that:
- `--set datacatalog.enabled=false` produces `catalog-cache:\n  type: noop`
- `--set datacatalog.enabled=true` (default) produces the original datacatalog endpoint config unchanged

### Labels
- **fixed**

### DIFF
--- a/charts/flyte-core/templates/propeller/configmap.yaml
+++ b/charts/flyte-core/templates/propeller/configmap.yaml
@@ -9,11 +9,14 @@ data:
 {{- with .Values.configmap.admin }}
   admin.yaml: | {{ tpl (toYaml .) $ | nindent 4 }}
 {{- end }}
+{{- if .Values.datacatalog.enabled }}
 {{- with .Values.configmap.catalog }}
   catalog.yaml: | {{ tpl (toYaml .) $ | nindent 4 }}
 {{- end }}
-{{- with .Values.configmap.catalog_cache }}
-  catalog_cache.yaml: | {{ tpl (toYaml .) $ | nindent 4 }}
+{{- else }}
+  catalog.yaml: |
+    catalog-cache:
+      type: noop
 {{- end }}
 {{- with .Values.configmap.copilot }}
   copilot.yaml: | {{ tpl (toYaml .) $ | nindent 4 }}
@@ -32,9 +35,6 @@ data:
 {{- end }}
 {{- with .Values.configmap.otel }}
   otel.yaml: | {{ tpl (toYaml .) $ | nindent 4 }}
-{{- end }}
-{{- with .Values.configmap.qubole }}
-  qubole.yaml: | {{ tpl (toYaml .) $ | nindent 4 }}
 {{- end }}
 {{- with .Values.configmap.resource_manager }}
   resource_manager.yaml: | {{ tpl (toYaml .) $ | nindent 4 }}
--- a/charts/flyte-core/values-eks.yaml
+++ b/charts/flyte-core/values-eks.yaml
@@ -55,6 +55,7 @@ flytescheduler: {}
 #
 
 datacatalog:
+  enabled: true
   replicaCount: 2
   serviceAccount:
     # -- If the service account is created by you, make this false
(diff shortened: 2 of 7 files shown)


charts/flyte-core/templates/propeller/configmap.yaml
```diff
@@ -33,9 +36,6 @@ data:
 {{- with .Values.configmap.otel }}
   otel.yaml: | {{ tpl (toYaml .) $ | nindent 4 }}
 {{- end }}
-{{- with .Values.configmap.qubole }}
```
Sovietaced, line 36: unused

charts/flyte-core/templates/propeller/configmap.yaml
```diff
@@ -9,11 +9,14 @@ data:
...
 {{- end }}
+{{- if .Values.datacatalog.enabled }}
 {{- with .Values.configmap.catalog }}
   catalog.yaml: | {{ tpl (toYaml .) $ | nindent 4 }}
+{{- else }}
+  catalog.yaml: |
+    catalog-cache:
+      type: noop
 {{- end }}
-{{- with .Values.configmap.catalog_cache }}
```
Sovietaced, line 15: unused

pingsutw (approved):

Tom-Newton:
Is there a new recomendation for how to configure multi-cluster Flyte deployments, where we have the datacatalog enabled in the control plane and dissabled in all the data planes?

Sovietaced:
> Is there a new recomendation for how to configure multi-cluster Flyte deployments, where we have the datacatalog enabled in the control plane and dissabled in all the data planes?

I'm sure you're aware but this change has been reverted

PR #7438, opened 2026-05-28
ci: add flyteidl-release workflow to main

pingsutw:
## Tracking issue

N/A

The `Release flyteidl` workflow (`.github/workflows/flyteidl-release.yml`) only exists on the `master` branch. Since `main` is now the default branch, `workflow_dispatch` does not surface this workflow in the Actions UI, and `gh workflow run` cannot find it either. As a result, the flyteidl release pipeline cannot be triggered.

Copy `.github/workflows/flyteidl-release.yml` from `master` to `main` verbatim. Once merged, the workflow becomes dispatchable from the default branch, and can be run against any ref (e.g. `master`):

```
gh workflow run flyteidl-release.yml --ref master -f version=v1.2.3
```

## How was this patch tested?

No functional change — workflow file copied unchanged from `master`. Will be exercised the next time we cut a flyteidl release.

### Labels

- **added**

- [x] I updated the documentation accordingly. (N/A — CI-only)
- [x] All new and existing tests passed. (N/A — workflow file copy)

### DIFF
--- a/.github/workflows/flyteidl-release.yml
+++ b/.github/workflows/flyteidl-release.yml
@@ -1,0 +1,78 @@
+name: Release flyteidl
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "version. Do *not* use the `flyteidl/` prefix, e.g. `flyteidl/v1.2.3`, instead use only `v1.2.3` (including the `v`)"
+        required: true
+
+jobs:
+  push-flyteidl-tag:
+    name: Push git tag containing the `flyteidl/` prefix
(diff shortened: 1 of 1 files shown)


.github/workflows/flyteidl-release.yml
```diff
@@ -0,0 +1,78 @@
+name: Release flyteidl
```
pingsutw, line 2: Copy from v1 so that I can trigger it from github action UI

Sovietaced (dismissed):

AdilFayyaz (approved):

PR #7461, opened 2026-05-30
fix: deflake ExampleNewAutoRefreshCache

pingsutw:

`ExampleNewAutoRefreshCache` in `flytestdlib/autorefreshcache` is flaky in CI ([failing run](https://github.com/flyteorg/flyte/actions/runs/26675665878/job/78626794222)):

```
--- FAIL: ExampleNewAutoRefreshCache (0.01s)
got:
Current status for item1 is Not-started
want:
Current status for item1 is Completed
```

The example slept a fixed `resyncPeriod * 10` (10ms) and then asserted that `item1` had reached the terminal `Completed` status. That transition is performed by the cache's **async background refresh worker**, so 10ms is not guaranteed to be enough on a slow/contended CI runner — the item is still `Not-started` when the assertion runs. This is a classic fixed-duration-wait-for-async-work race.

Replace the fixed `time.Sleep` with **condition-based polling**: poll `cache.Get` every `resyncPeriod` until `item1` reaches its terminal state (`Completed`), guarded by a 10s safety deadline. This removes the timing dependency while keeping the example's `// Output:` deterministic.

## How was this patch tested?

- `go build` / `go vet` of the package pass; `go test -race ./flytestdlib/autorefreshcache/` passes.
- Ran the example 200x via separate `go test` invocations under heavy CPU load (both before and after the change). **Note:** I was not able to reproduce the original CI flake locally even under load — the failure appears specific to the CI runner's scheduling. The change is therefore justified primarily by the root cause (a fixed 10ms wait for asynchronous work), not by a local repro.
  - Caveat worth knowing for reviewers: Go's `testing` package does **not** honor `-count` for `Example` functions (verified separately), so stressing an example requires repeated `go test` invocations rather than `-count=N`.

### Labels

- **fixed**: deflakes a CI example test.

### DIFF
--- a/flytestdlib/autorefreshcache/auto_refresh_example_test.go
+++ b/flytestdlib/autorefreshcache/auto_refresh_example_test.go
@@ -128,8 +128,8 @@ func ExampleNewAutoRefreshCache() {
 			break
 		}
 		if time.Now().After(deadline) {
-			fmt.Printf("Current status for item1 is %v", item.(*ExampleCacheItem).status)
-			break
+			cancel()
+			panic(fmt.Sprintf("timed out waiting for item1 to reach %v; last status=%v", ExampleStatusSucceeded, item.(*ExampleCacheItem).status))
 		}
 		time.Sleep(resyncPeriod)
 	}
--- a/flytestdlib/autorefreshcache/auto_refresh_example_test.go
+++ b/flytestdlib/autorefreshcache/auto_refresh_example_test.go
@@ -112,13 +112,26 @@ func ExampleNewAutoRefreshCache() {
 		fmt.Printf("unexpected error in create; err1: %v, err2: %v", err1, err2)
 	}
 
-	// wait for the cache to go through a few refresh cycles and then check status
-	time.Sleep(resyncPeriod * 10)
-	item, err := cache.Get(item1.ID())
-	if err != nil && errors.IsCausedBy(err, ErrNotFound) {
-		fmt.Printf("Item1 is no longer in the cache")
-	} else {
-		fmt.Printf("Current status for item1 is %v", item.(*ExampleCacheItem).status)
+	// Poll until the cache's background worker refreshes item1 to its terminal state. Polling against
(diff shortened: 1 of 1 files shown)


flytestdlib/autorefreshcache/auto_refresh_example_test.go
```diff
@@ -112,13 +112,38 @@ func ExampleNewAutoRefreshCache() {
...
-	item, err := cache.Get(item1.ID())
-	if err != nil && errors.IsCausedBy(err, ErrNotFound) {
-		fmt.Printf("Item1 is no longer in the cache")
-	} else {
-		fmt.Printf("Current status for item1 is %v", item.(*ExampleCacheItem).status)
+	// Poll until the cache's background worker refreshes item1 to its terminal state. Polling against
+	// the expected condition (instead of sleeping a fixed duration) keeps this example deterministic
+	// even when the async worker is slow under load.
+	var item Item
+	var lastStatus ExampleItemStatus = ExampleStatusNotStarted
```
Sovietaced: Is `require.Eventually` an option?

Now that I look at this code it probably shouldn't be using panic at all huh

pingsutw: We cannot use require.Eventually since there is no *testing.T. It's actually an example for the NewAutoRefreshCache.

Screenshot 2026-05-30 at 10 46 46 PM

pingsutw: i can replace the panic with print error if you want

Sovietaced: > We cannot use require.Eventually since there is no *testing.T. It's actually an example for the NewAutoRefreshCache.
> 
> 
> 
> Screenshot 2026-05-30 at 10 46 46 PM
> 
> 

I'll take a look later, this seems strange

Sovietaced:
TIL about example based tests

Sovietaced (approved):

Issue #7462, opened 2026-05-30
clarify on "main" vs "master" branch


yarikoptic:
"inspired" by 
- https://github.com/flyteorg/flyte/pull/7121#issuecomment-4573929221

and then observing that there are both active but different `main` and `master` and seems no clear description in README.md on the difference or workflow

<img width="3636" height="794" alt="Image" src="https://github.com/user-attachments/assets/bb632648-3cae-407b-9ce0-d048c9022c9c" />

yarikoptic:
roger that

```
❯ git describe --tags --match 'v*' origin/master
v1.16.7
❯ git describe --tags --match 'v*' origin/main
v2.0.18-14-g2fc9a4907
```

so -- `main` it should  be

PR #7463, opened 2026-05-30
.codespellrc: document each skip pattern and whitelisted word

yarikoptic:

`.codespellrc`'s `ignore-words-list` had grown to 9 short, opaque entries
(`astroid,bootup,decorder,fo,lightyear,nd,notin,ser,te`). Readers and
reviewers cannot tell at a glance why each is whitelisted, which makes it
hard to audit additions or remove dead entries over time.

Switch `ignore-words-list` to the multi-line config-file form so each
entry can be preceded by a one-line rationale (Go variable abbreviation,
library name, test-fixture string, etc.). Same pattern as
[restic/restic#21807](https://github.com/restic/restic/pull/21807).
Also annotate the `skip` patterns inline.

No behavior change: same words skipped/whitelisted, just documented.

## How was this patch tested?

`uvx codespell` passes with zero errors after the reformat. Each entry
was also verified to be still needed (removing any one of them causes
codespell to flag a real occurrence in the tree).

### Labels

- **changed**

feel welcome to just close since I am not sure if worth it and not sure what is difference between main and master

- https://github.com/flyteorg/flyte/issues/7462

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,6 +1,33 @@
 [codespell]
+# Skip:
+# - .git, *.pdf, *.svg: VCS metadata and binaries
+# - *.pb: compiled protobuf descriptors
+# - go.sum, go.mod, *.lock, *.lock.yaml, *requirements.txt: dependency lock files
+# - gen: generated code (protobuf bindings etc.)
+# - monodocs-environment.lock.yaml: conda lock file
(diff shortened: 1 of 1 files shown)


pingsutw (approved):

pingsutw:
@yarikoptic thank you

PR #7464, opened 2026-06-01
fix: populate LogContext from live JobSet pods for clustered tasks

AdilFayyaz:

The clustered (JobSet) plugin only emitted templated log URLs synthesized from predicted pod names, which requires a pod-log template to be configured and breaks when real pod names carry the Job-assigned random suffix. The console needs a structured LogContext to fetch logs natively regardless of log-template config.

  - Add getLogContext, which lists a JobSet's live child pods (via the jobset.sigs.k8s.io/jobset-name label) and builds a core.LogContext from the real pods using flytek8s.BuildPodLogContext.
  - Identify the primary (rank-0) pod by name prefix and exclude Pending pods that have no logs/container IDs yet; best-effort — returns nil on list error or when no pods are ready, falling back to the existing templated Logs path.
  - Wire LogContext into TaskInfo in GetTaskPhase.
  - Refactor test helper dummyPluginCtx to accept a client.Reader and add emptyK8sReader for tests that don't inspect pods.

## How was this patch tested?

  - New TestGetTaskPhase_LogContext verifies a running JobSet with live worker pods produces a LogContext where the rank-0 pod is marked primary (matched despite its random suffix) and Pending pods are excluded.
  - Existing phase tests updated to pass an explicit K8sReader and continue to pass.

## Stack

### DIFF
--- a/flyteplugins/go/tasks/plugins/k8s/clustered/logs.go
+++ b/flyteplugins/go/tasks/plugins/k8s/clustered/logs.go
@@ -3,15 +3,24 @@ package clustered
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	jobsetv1alpha2 "sigs.k8s.io/jobset/api/jobset/v1alpha2"
 
 	"github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/logs"
+	"github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/flytek8s"
 	"github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/k8s"
 	"github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/tasklog"
+	"github.com/flyteorg/flyte/v2/flytestdlib/logger"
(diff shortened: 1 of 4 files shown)


flyteplugins/go/tasks/plugins/k8s/clustered/logs.go
```diff
@@ -76,3 +85,55 @@ func getTaskLogs(ctx context.Context, pluginContext k8s.PluginContext, jobSet *j
...
+// Unlike getTaskLogs (which synthesizes templated URIs from *predicted* pod names and
+// requires a pod-log template to be configured in cluster config), this uses the *real*
+// pods — actual names (including the Job-assigned random suffix), namespace, primary
+// container, and per-container IDs — so the console can fetch logs natively regardless
+// of log-template config. Best-effort: returns nil on list error or when no pods are
+// ready yet, leaving the templated Logs path as the fallback.
+func getLogContext(ctx context.Context, pluginContext k8s.PluginContext, jobSet *jobsetv1alpha2.JobSet) *core.LogContext {
+	podList := &v1.PodList{}
+	if err := pluginContext.K8sReader().List(ctx, podList,
+		client.InNamespace(jobSet.Namespace),
```
pingsutw: We don't need to pass namespace and labels here, the plugin k8s reader already did that. https://github.com/pingsutw/flyte/blob/1dd046406db056f08d27adf2299d2527be9bdffb/flyteplugins/go/tasks/pluginmachinery/flytek8s/plugin_exec_context.go#L159-L168

pingsutw (approved):

PR #7459, opened 2026-05-29
Add support for OTEL metrics

Sovietaced:
Fixes #7447

## Why are the changes needed?
This pull request generally adds support for emitting OTEL metrics. Additionally, it uses `otelconnect` to instrument all of the connect RPC clients/servers with OTEL metrics. Without configuration all trace/metric related calls will be a noop so this won't change the default behavior.

## What changes were proposed in this pull request?
* Generate meter providers alongside existing trace providers
* Instrument RPC clients/servers with trace and meter providers
 
## How was this patch tested?
We have a variation of this running in our fork in production. It still uses plain gRPC instead of connect RPC but we have used `otelconnect` in other capacities. 

### Labels

Please add one or more of the following labels to categorize your PR:
- **added**: For new features.
- **changed**: For changes in existing functionality.
- **deprecated**: For soon-to-be-removed features.
- **removed**: For features being removed.
- **fixed**: For any bug fixed.
- **security**: In case of vulnerabilities

This is important to improve the readability of release notes.

## Check all the applicable boxes 

- [x] I updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] All commits are signed-off.

## Stack

If you do use [`git town`](https://www.git-town.com/introduction) to manage PR Stacks, the stack relevant to this PR
will show below. Otherwise, you can ignore this section.

PR #7483, opened 2026-06-08
feat(executor): add OTel metrics for the TaskAction controller

pingsutw:
## Tracking issue

Related to #7459

Screenshot 2026-06-10 at 5 18 58 PM
Screenshot 2026-06-10 at 5 19 10 PM
Screenshot 2026-06-10 at 5 19 24 PM

#7459 added the OpenTelemetry **meter-provider plumbing** (`otelutils.GetMeterProvider`)
and instrumented the connect RPC clients/servers, but added no **business metrics** for
the TaskAction controller — operators have no visibility into how many TaskAction CRDs
exist, what phase they're in, how large they are, or how long their Kubernetes API
operations take.

It also surfaced a cardinality problem: the otelconnect server interceptors tag every
`rpc_server_*` metric with the caller's ephemeral source port, which is unbounded.

### 1. TaskAction controller metrics

Three custom OTel meters, registered on the **executor meter provider that is injected**
into the reconciler (`NewTaskActionReconciler(... , meterProvider, cache)` →
`registerTaskActionMetrics(provider, cachedPhaseCounter(cache))`), so they flow through
the same OTLP pipeline as the RPC metrics and the helper is testable without globals:

- **`taskaction.active`** (`Int64ObservableGauge`) — number of TaskAction CRDs by
  `phase`. Observed asynchronously via a callback that counts straight from the
  controller cache's informer **indexer** (`indexer.List()` returns the cached object
  pointers, so there is no per-collection deep-copy of every CRD — unlike `client.List`,
  this stays O(N) pointer reads even with many TaskActions). Off the reconcile hot path;
  runs on the SDK's collection goroutine ~once per export.
- **`taskaction.crd.size_bytes`** (`Int64Histogram`) — serialized (JSON) size of a
  TaskAction CRD, recorded once per reconcile.
- **`taskaction.k8s.duration`** (`Float64Histogram`, `op`/`error` attributes) —
  per-operation latency of TaskAction CRD calls to the Kubernetes API
  (`get`/`update`/`status_update`), timed **inline** at the call sites and recorded via
  `recordK8sOp`.

Registration short-circuits to a no-op when metrics are disabled (the meter provider is
a noop provider), and every record helper is nil-safe, so the controller degrades to no
custom metrics rather than failing setup.

### 2. Reduce server RPC metric cardinality

Add `otelconnect.WithoutServerPeerAttributes()` to the **server** interceptors across all
services (runs, actions, app, internal-app, dataproxy, cache, secret, events, executor).
otelconnect otherwise tags `rpc_server_*` metrics/spans with `net.peer.name` and
`net.peer.port` — on the server side these are the **caller's IP and ephemeral source
port**, so the port is a new value per connection. That makes each series a one-shot
(`rate()` ≈ 0), explodes cardinality, and grows the cumulative OTLP payload past the
collector's default 4 MiB gRPC limit. This drops those two attributes from **server-side**
telemetry only; client-side telemetry (where `net.peer.*` is the server's fixed address)
is untouched, and no metric is removed — only two high-cardinality labels.

### Deliberately *not* re-implemented (already provided)

| Already available | Source |
|---|---|
| reconcile rate/errors, workqueue work/queue duration | controller-runtime metrics server (`:10254`) |
| event → event-proxy send latency | otelconnect-wrapped events client — `rpc_client_duration{rpc_method="Record"}` |

## How was this patch tested?

- **Unit tests** (`metrics_test.go`): `TestRegisterTaskActionMetrics` (gauge observes
  the injected phase counter), `TestCountByPhase` (pure tally: empty phase → `Unknown`,
  nil entries skipped), `TestObserveCRDSize` (records and is nil-safe), `TestRecordK8sOp`
  (records latency and is nil-safe).
- `go build ./...`, `go vet`, and `gofmt` clean.
- Verified live on a dev cluster: the `taskaction_*` metrics flow through an OTel
  collector → Prometheus and render in Grafana next to the controller-runtime and RPC
  panels.

### Labels

added


executor/pkg/controller/metrics.go
```diff
@@ -0,0 +1,167 @@
...
+func (m *taskActionMetrics) observeCRDSize(ctx context.Context, ta *flyteorgv1.TaskAction) {
+	if m == nil || m.crdSizeBytes == nil {
+		return
+	}
+	if b, err := json.Marshal(ta); err == nil {
+		m.crdSizeBytes.Record(ctx, int64(len(b)))
+	}
+}
+
+// recordK8sOp records the latency of a Kubernetes API operation against the
```
pingsutw, line 156: i'm going to run some load tests, will remove this if it causes too much overhead

executor/pkg/controller/taskaction_controller.go
```diff
@@ -197,22 +203,34 @@ func (r *TaskActionReconciler) finalizePermanentFailure(
...
 	dataStore *storage.DataStore,
 	eventsClient workflowconnect.EventsProxyServiceClient,
 	cluster string,
+	meterProvider metric.MeterProvider,
+	cache ctrlcache.Cache,
 ) *TaskActionReconciler {
+	metrics, err := registerTaskActionMetrics(meterProvider, cachedPhaseCounter(cache))
+	if err != nil {
+		// Non-fatal: degrade to no custom metrics rather than failing controller setup.
+		log.Log.Error(err, "failed to register TaskAction OTel metrics")
```
Sovietaced, line 224: You could argue it should fail. I think something really bad would have to happen for this to error

Sovietaced (approved):

executor/pkg/controller/metrics.go
Sovietaced: This will definitely cause a good amount of load. We find that de(serialization) is most of the CPU load in propeller v1

PR #7485, opened 2026-06-09
feat(flyte-binary): manage the task-pod default ServiceAccount

pingsutw:

Task pods run as the namespace **`default`** ServiceAccount — the executor returns an
empty service account name (`executor/pkg/plugin/task_exec_metadata.go`,
`GetK8sServiceAccount() string { return "" }`), so Kubernetes assigns `default`.

The chart had no way to put annotations on that ServiceAccount. On EKS this means task
pods can't be given an IRSA role (`eks.amazonaws.com/role-arn`), so they run without
cloud credentials and fail S3 operations with `403 Forbidden` (e.g. downloading the
fast-register code bundle).

- **`templates/task-serviceaccount.yaml`** (new) — manages the task ServiceAccount
  (`default`, configurable) in the release namespace so it can carry annotations/labels.
- **`values.yaml`** — adds a `taskServiceAccount` block, **disabled by default**
  (`create: false`); set `create: true` and `annotations` to attach an IRSA role:
  ```yaml
  taskServiceAccount:
    create: true
    name: default
    annotations:
      eks.amazonaws.com/role-arn: "arn:aws:iam::<acct>:role/<worker-role>"
  ```

Note: because Kubernetes auto-creates the `default` ServiceAccount, an existing one
must carry Helm ownership metadata (`app.kubernetes.io/managed-by: Helm`,
`meta.helm.sh/release-name`/`-namespace`) for Helm to adopt it on upgrade.

## How was this patch tested?

- `helm template --show-only templates/task-serviceaccount.yaml` renders the `default`
  ServiceAccount with the configured IRSA annotation when enabled, and nothing when
  `create: false`.
- `helm lint` passes.
- Verified live on a dev cluster: a task pod using the annotated `default` SA assumes
  the worker IRSA role and the previously-403'ing `HEAD` on the code bundle returns 200.

### Labels

added

### DIFF
--- a/charts/flyte-binary/templates/task-serviceaccount.yaml
+++ b/charts/flyte-binary/templates/task-serviceaccount.yaml
@@ -1,0 +1,29 @@
+{{- if .Values.taskServiceAccount.create }}
+{{- /*
+The executor runs task pods as the namespace `default` ServiceAccount
+(executor returns an empty service account name, so Kubernetes uses `default`).
+This manages that ServiceAccount so it can carry annotations such as an IRSA
+role-arn, giving task pods cloud credentials (e.g. S3 access).
+*/}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.taskServiceAccount.name | default "default" }}
+  namespace: {{ .Values.taskServiceAccount.namespace | default .Release.Namespace | quote }}
+  labels: {{- include "flyte-binary.labels" . | nindent 4 }}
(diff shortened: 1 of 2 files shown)


pingsutw:
Superseded by #7486 — instead of managing the auto-created `default` ServiceAccount via Helm (which hits the resource-adoption problem), the executor now lets task pods run as a configurable, already-annotated ServiceAccount (`executor.defaultK8sServiceAccount`).

PR #7486, opened 2026-06-09
feat(executor): configurable default service account for task pods

pingsutw:

Task pods always ran as the namespace **`default`** ServiceAccount — the executor's
`GetK8sServiceAccount()` returned `""`, so Kubernetes assigned `default`. On EKS that
SA typically has no IRSA `eks.amazonaws.com/role-arn`, so task pods get no cloud
credentials and fail S3 operations (e.g. downloading the fast-register code bundle)
with `403 Forbidden`.

The auto-created `default` SA is awkward to annotate via Helm (it pre-exists and Helm
won't adopt an unowned resource), so being able to point task pods at an existing,
already-annotated ServiceAccount is a cleaner fix.

- **`executor/pkg/config/config.go`** — add `defaultK8sServiceAccount` to the executor
  config. Also change `MaxSystemFailures` from `uint32` to `int32`: the pflags generator
  doesn't support `uint32`, so `config_flags.go` couldn't be regenerated and its flag
  entry had been hand-added. It's cast back to `uint32` at the reconciler boundary
  (`executor/setup.go`), where it's compared against the CRD's `Status.SystemFailures uint32`.
- **`executor/pkg/config/config_flags.go` / `config_flags_test.go`** — regenerated via
  `go generate ./executor/pkg/config/...` (genuine generator output now, including the new
  `defaultK8sServiceAccount` flag).
- **`executor/pkg/plugin/task_exec_metadata.go`** — `GetK8sServiceAccount()` now resolves:
  1. the ServiceAccount on the task's `SecurityContext.RunAs.K8SServiceAccount` if set,
  2. else `executor.defaultK8sServiceAccount`,
  3. else `""` → Kubernetes uses the `default` ServiceAccount of the pod's own namespace
     (unchanged behavior).

A deployment can now set e.g. `executor.defaultK8sServiceAccount: <flyte-binary SA>` so
task pods run as the chart's IRSA-annotated ServiceAccount.

> Note: pointing task pods at the server ServiceAccount grants task code that SA's IAM
> role and RBAC. Prefer a dedicated, least-privilege worker ServiceAccount in production;
> this option simply makes the account configurable.

## How was this patch tested?

- Unit test (`TestResolveServiceAccount`): a run-specified SA always wins; with no run SA
  it falls back to the executor config default (asserted with a non-empty configured
  value), and resolves to `""` when neither is set. Getters are nil-safe.
- `go generate ./executor/pkg/config/...`, `go build ./executor/...`, `go vet`, `gofmt`,
  and the `config` pflag tests all pass.
- Verified live on a dev cluster: with `defaultK8sServiceAccount` set to an
  IRSA-annotated SA, task pods assume that role and S3 access succeeds.

### Labels

added

### DIFF
--- a/executor/pkg/config/config.go
+++ b/executor/pkg/config/config.go
@@ -79,6 +79,11 @@ type Config struct {
 	// Cluster is the cluster identifier attached to action events.
 	Cluster string `json:"cluster" pflag:",Cluster identifier for action events"`
 
+	// DefaultK8sServiceAccount is assigned to task pods when the task's security
+	// context does not specify one. Empty leaves the pod on the namespace's
+	// `default` ServiceAccount.
+	DefaultK8sServiceAccount string `json:"defaultK8sServiceAccount" pflag:",Default Kubernetes service account for task pods when the task does not set one"`
+
 	// MaxSystemFailures bounds consecutive system-level failures (Plugin.Handle Go
 	// errors and plugin-reported system-retryable failures) before a TaskAction is
 	// converted to a permanent failure.
--- a/executor/pkg/config/config.go
+++ b/executor/pkg/config/config.go
@@ -80,8 +80,8 @@ type Config struct {
 	Cluster string `json:"cluster" pflag:",Cluster identifier for action events"`
 
 	// DefaultK8sServiceAccount is assigned to task pods when the task's security
-	// context does not specify one. Empty leaves the pod on the namespace's
-	// `default` ServiceAccount.
+	// context does not specify one. Empty means no account is set, so Kubernetes
+	// assigns the `default` ServiceAccount of the pod's own namespace.
 	DefaultK8sServiceAccount string `json:"defaultK8sServiceAccount" pflag:",Default Kubernetes service account for task pods when the task does not set one"`
 
 	// MaxSystemFailures bounds consecutive system-level failures (Plugin.Handle Go
--- a/executor/pkg/config/config.go
+++ b/executor/pkg/config/config.go
@@ -87,7 +87,7 @@ type Config struct {
 	// MaxSystemFailures bounds consecutive system-level failures (Plugin.Handle Go
 	// errors and plugin-reported system-retryable failures) before a TaskAction is
 	// converted to a permanent failure.
-	MaxSystemFailures uint32 `json:"maxSystemFailures" pflag:",Max consecutive system-level failures before forcing permanent failure"`
+	MaxSystemFailures int32 `json:"maxSystemFailures" pflag:",Max consecutive system-level failures before forcing permanent failure"`
 
 	// GC configures the garbage collector for terminal TaskActions.
 	GC GCConfig `json:"gc" pflag:",Garbage collector configuration for terminal TaskActions"`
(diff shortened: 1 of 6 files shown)


executor/pkg/config/config_flags.go
```diff
@@ -63,6 +63,7 @@ func (cfg Config) GetPFlagSet(prefix string) *pflag.FlagSet {
 	cmdFlags.Bool(fmt.Sprintf("%v%v", prefix, "enableHTTP2"), defaultConfig.EnableHTTP2, "Enable HTTP/2 for metrics and webhook servers")
 	cmdFlags.String(fmt.Sprintf("%v%v", prefix, "EventsServiceURL"), defaultConfig.EventsServiceURL, "URL of the Event Service for action event updates")
 	cmdFlags.String(fmt.Sprintf("%v%v", prefix, "cluster"), defaultConfig.Cluster, "Cluster identifier for action events")
+	cmdFlags.String(fmt.Sprintf("%v%v", prefix, "defaultK8sServiceAccount"), defaultConfig.DefaultK8sServiceAccount, "Default Kubernetes service account for task pods when the task does not set one")
```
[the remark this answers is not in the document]
machichima: I think we should run `go generate ./executor/pkg/config/...` rather than adding it manually?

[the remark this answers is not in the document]
pingsutw: Good call — and you're right that it shouldn't be hand-edited. The reason it was: pflags doesn't support `uint32`, and the pre-existing `MaxSystemFailures uint32` field made `go generate ./executor/pkg/config/...` fail with `only these basic kinds are allowed. given [uint32]` (true even on the latest flytestdlib v1.16.7). So the prior `maxSystemFailures` flag entry was itself hand-added.

Fixed properly now: changed `MaxSystemFailures` to `int32` (cast back to `uint32` at the reconciler boundary in `executor/setup.go`, where it's compared against the CRD's `Status.SystemFailures uint32`) and reran `go generate`. `config_flags.go` / `config_flags_test.go` are genuine generator output now, including the new `defaultK8sServiceAccount` flag.

executor/pkg/plugin/task_exec_metadata.go
```diff
@@ -58,6 +59,14 @@ type taskExecutionMetadata struct {
 	envVars         map[string]string
 	interruptible   bool
 	securityContext *core.SecurityContext
+	serviceAccount  string
+}
+
+func resolveServiceAccount(securityContext *core.SecurityContext) string {
```
machichima: Can we pass in the default SA here rather than read default from global config `executorconfig.GetConfig().DefaultK8sServiceAccount`

Something like this

```go
func resolveServiceAccount(securityContext *core.SecurityContext, defaultSA string) string {
      if sa := securityContext.GetRunAs().GetK8SServiceAccount(); sa != "" {
          return sa
      }
      return defaultSA
  }
```

In this case we do not need to do this in test

```go
	cfg := executorconfig.GetConfig()
	prev := cfg.DefaultK8sServiceAccount
	defer func() { cfg.DefaultK8sServiceAccount = prev }()
```

pingsutw: Done — `resolveServiceAccount` now takes the default as a parameter, and the caller (`NewTaskExecutionMetadata`) passes `executorconfig.GetConfig().DefaultK8sServiceAccount`. The helper is pure now, so the test no longer mutates global config. Thanks!

executor/pkg/config/config.go
```diff
@@ -79,10 +79,15 @@ type Config struct {
...
+	// DefaultK8sServiceAccount is assigned to task pods when the task's security
+	// context does not specify one. Empty means no account is set, so Kubernetes
+	// assigns the `default` ServiceAccount of the pod's own namespace.
+	DefaultK8sServiceAccount string `json:"defaultK8sServiceAccount" pflag:",Default Kubernetes service account for task pods when the task does not set one"`
+
 	// MaxSystemFailures bounds consecutive system-level failures (Plugin.Handle Go
 	// errors and plugin-reported system-retryable failures) before a TaskAction is
 	// converted to a permanent failure.
-	MaxSystemFailures uint32 `json:"maxSystemFailures" pflag:",Max consecutive system-level failures before forcing permanent failure"`
+	MaxSystemFailures int32 `json:"maxSystemFailures" pflag:",Max consecutive system-level failures before forcing permanent failure"`
```
pingsutw, line 90: changed to int32 so that we can run go generate

AdilFayyaz (approved):

PR #7490, opened 2026-06-10
feat(copilot): preserve nested subdirectories in multipart blob uploads

k1sauce:

Shell Glob matching in the SDK extras should work (pathlib.Path.glob supports **), [but Glob serializes to a Dir on the wire](https://github.com/flyteorg/flyte-sdk/blob/3c69976cd58820fad411c0e606b280e98ac7dc50/src/flyte/extras/shell/_runtime.py#L72), and in real remote execution flytecopilot uploads /var/outputs/<name>/ as a MULTIPART blob. This is because copilot's multipart uploader drops nested subdirectories. Glob and a bare Dir output fail identically here.

Walk into directories and key each file by its relative path from the blob root (filepath.Rel + ConstructReference with the path segments), mirroring the download side, which already reconstructs nested keys. Top-level files are unaffected — their relative path is just the basename, so keys are unchanged.

## How was this patch tested?

Add an upload-multipart-nested test covering a root file, nested files, and a cross-subdir basename collision.

### Labels

## Stack

### DIFF
--- a/flytecopilot/data/upload.go
+++ b/flytecopilot/data/upload.go
@@ -8,6 +8,7 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
+	"strings"
 
 	"github.com/golang/protobuf/proto" //nolint: staticcheck
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


k1sauce:
Test by Samhita:
image

pingsutw (approved):

PR #7495, opened 2026-06-10
fix(flytestdlib): mark batch head item as processing in auto-refresh cache

1fanwang:

The auto-refresh cache tracks which items a worker is currently syncing so the periodic resync does not enqueue an item again while it is still in flight. The first item of every enqueued batch was never recorded as in-flight, so the resync loop kept re-enqueueing items that were already being synced — the same key handled by multiple workers at once, duplicate `syncCb` calls, and wasted workqueue throughput. With the default single-item batching this affects every item the cache tracks.

Record every item in an enqueued batch as in-flight, including the first one.

## How was this patch tested?

Added `TestEnqueueBatches_MarksSingleItemBatchHead` to both `flytestdlib/cache` and `flytestdlib/autorefreshcache`. It drives the real enqueue path with the default single-item batching and asserts the enqueued item is marked in-flight.

Reproduce:

```
go test ./flytestdlib/cache/ ./flytestdlib/autorefreshcache/ -run TestEnqueueBatches_MarksSingleItemBatchHead -v
```

Before:

```
--- FAIL: TestEnqueueBatches_MarksSingleItemBatchHead
    single-item batch head must be marked processing after enqueue
    Should be true
FAIL    github.com/flyteorg/flyte/v2/flytestdlib/cache
--- FAIL: TestEnqueueBatches_MarksSingleItemBatchHead
    single-item batch head must be marked processing after enqueue
    Should be true
FAIL    github.com/flyteorg/flyte/v2/flytestdlib/autorefreshcache
```

After:

```
--- PASS: TestEnqueueBatches_MarksSingleItemBatchHead
ok      github.com/flyteorg/flyte/v2/flytestdlib/cache
--- PASS: TestEnqueueBatches_MarksSingleItemBatchHead
ok      github.com/flyteorg/flyte/v2/flytestdlib/autorefreshcache
```

Both packages also pass with `-race`.

### Labels
fixed

### DIFF
--- a/flytestdlib/autorefreshcache/auto_refresh.go
+++ b/flytestdlib/autorefreshcache/auto_refresh.go
@@ -283,7 +283,7 @@ func (w *autoRefresh) enqueueBatches(ctx context.Context) error {
 	for _, batch := range batches {
 		b := batch
 		w.workqueue.AddRateLimited(&b)
-		for i := 1; i < len(b); i++ {
+		for i := 0; i < len(b); i++ {
 			w.processing.Store(b[i].GetID(), w.clock.Now())
 		}
 	}
--- a/flytestdlib/autorefreshcache/auto_refresh_test.go
+++ b/flytestdlib/autorefreshcache/auto_refresh_test.go
@@ -281,3 +281,24 @@ func TestInProcessing(t *testing.T) {
 	_, found := cache.processing.Load("test1")
 	assert.False(t, found)
 }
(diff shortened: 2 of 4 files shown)


Sovietaced:
This looks like an old bug that was fixed on the master branch. https://github.com/flyteorg/flyte/pull/6725

Sovietaced (approved):

PR #7498, opened 2026-06-10
dataproxy: fall back to RunService.GetActionData for inline literals

SVilgelm:
## Overview

`DataProxyService.GetActionData` materialises action inputs/outputs by fetching `inputs.pb`/`outputs.pb` via the URIs returned by `RunService.GetActionDataURIs`. For **condition** actions (`ACTION_TYPE_CONDITION`) the literals don't live in object storage at all — the signal payload lives inline on the action row (`RunInfo.Output`). The run service signals this case by returning **both URIs empty**.

Today the dataproxy hits that empty-URIs case, skips both S3 fetches, and returns its zero-value skeleton `{Inputs: {}, Outputs: {}}` — losing the literal even though the run service has it ready to hand back inline.

This PR adds a small fallback: when both URIs come back empty, call `RunService.GetActionData` once and pass its inline `Inputs`/`Outputs` straight through. Tasks/traces are unaffected — they always have a non-empty inputs URI from RecordAction time.

## Why not unconditionally?

We don't want to lose the dataplane-fetch optimisation for tasks/traces (they hold large literals in S3 and the dataproxy is co-located with the bucket). The fallback only triggers when `urisResp.inputs_uri == "" && urisResp.outputs_uri == ""` — which today only happens for condition actions.

## Changes

- `flyteidl2/dataproxy/service/dataproxy_service.go` — new condition fallback branch in `Service.GetActionData`, gated on both URIs empty. Logs at debug level so the call site is easy to spot in traces.
- The `RunService.GetActionData` call is marked deprecated (in favour of `DataProxyService.GetActionData`), but calling ourselves would loop, so a `//nolint:staticcheck` directive explains the intentional reach-into.

## Test plan

- `make test` (flyte2 unit tests).
- Manually verified on dogfood-cloud-staging: a signaled BOOLEAN condition that previously returned `{"inputs":{}, "outputs":{}}` now returns `{"outputs":{"literals":[{"name":"<condition_name>","value":{"scalar":{"primitive":{"boolean":true}}}}]}}`.

## Upstream context

This is part of a larger end-to-end fix in `unionai/cloud` ([PR #16408](https://github.com/unionai/cloud/pull/16408)) that plumbs condition signal payloads inline through:

1. actions-service `Update.output` proto field (closed-source)
2. RedisStream replicator + workflow consumer carrying the literal
3. workflow service merging the literal into `RunInfo.Output`
4. **dataproxy fallback (this PR)** to surface the inline literal to the UI

### DIFF
--- a/dataproxy/service/dataproxy_service.go
+++ b/dataproxy/service/dataproxy_service.go
@@ -483,6 +483,28 @@ func (s *Service) GetActionData(
 		return nil, err
 	}
 
+	// Condition actions carry their inputs/outputs inline (the ConditionAction
+	// spec on RunInfo and the signal Literal on RunInfo.Output) rather than
+	// materialising inputs.pb / outputs.pb to object storage. RunService
+	// signals this by returning both URIs empty; fall back to RunService's
+	// GetActionData, which returns the inline literals directly.
+	if urisResp.Msg.GetInputsUri() == "" && urisResp.Msg.GetOutputsUri() == "" {
+		logger.Debugf(ctx, "GetActionData: both URIs empty for action %s, falling back to RunService.GetActionData for inline literals", actionId.String())
+		//nolint:staticcheck // RunService.GetActionData is deprecated in favour of
+		// DataProxyService.GetActionData (this method) — but calling ourselves would
+		// loop, so we must reach into the run service for the inline literal.
+		dataResp, err := s.runClient.GetActionData(ctx, connect.NewRequest(&workflow.GetActionDataRequest{
+			ActionId: actionId,
+		}))
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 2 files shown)


SVilgelm:
Closing — not needed. Our deployment serves the V2 `flyteidl2.dataproxy.DataProxyService` from the cloud2 `dataproxy/service/dataproxy_flyte.go` binary (compiled into both control-plane `dataproxy` and the `unionoperator` image's `dataproxy` sub-binary). The fix lives there: https://github.com/unionai/cloud/pull/16408. The flyte2 path in `flyteidl2/dataproxy/service/dataproxy_service.go` isn't reached in our setup, so the fallback here is dead code on top of an unmerged change. Sorry for the churn.

PR #6998, opened 2026-03-09
[V2] Add AuthMetadataService + wire external OAuth2 auth into runs service

pingsutw:
## Architecture

The v2 runs service acts as an OAuth2 **resource server**. It validates JWTs issued by a configured external OAuth2 authorization server and runs the standard OIDC code flow against an OpenID Connect IdP (e.g. Okta) for browser users. v2 does **not** mint tokens itself — `authCfg.AppAuth.AuthServerType` is `External`, and `externalAuthServer.baseUrl` points at the issuer whose `.well-known/oauth-authorization-server` + `jwks_uri` v2 reads at boot.

```mermaid
graph TB
    Client[UI / flyte-sdk / task pods]
    ALB["Ingress / load balancer"]
    IDP["OIDC IdP (Okta)<br/>signin.example.com<br/>client id 0oak...5d6"]
    AUTHSRV["External OAuth2 authorization server<br/>.well-known/oauth-authorization-server<br/>jwks_uri"]
    MW["runs GetAuthenticationHTTPInterceptor<br/>bearer + cookie + loopback gate"]
    OIDC["OIDC browser handlers<br/>/login /callback /logout<br/>.well-known/openid-configuration"]
    RPC["v2 connect-rpc handlers<br/>RunService ProjectService TaskService<br/>TriggerService TranslatorService<br/>AuthMetadataService IdentityService ..."]
    INT["intra-cluster connect-rpc<br/>ActionsService InternalRunService<br/>(ClusterIP only, allowlisted)"]
    COOKIE["CookieManager<br/>AES-256 + HMAC-SHA256<br/>cookie_hash_key cookie_block_key"]
    S1["auth secret<br/>cookie_hash_key / cookie_block_key"]
    S2["OIDC client secret<br/>oidc_client_secret"]

    Client --> ALB
    ALB -->|"/login /callback /logout /healthcheck<br/>/flyteidl2.*/*"| MW
    ALB -.->|"/v2/* static SPA"| MW
    Client -.->|"task-pod ClusterIP calls"| INT
    MW --> OIDC
    MW --> RPC
    OIDC <-->|"authorization code flow"| IDP
    OIDC --> COOKIE
    MW -.->|"boots JWKS"| AUTHSRV
    MW -.->|"validate bearer token signature"| AUTHSRV
    COOKIE -.->|"reads"| S1
    OIDC -.->|"reads"| S2
```

## Component data flow

[...]

PR #7507, opened 2026-06-11
feat(runs): serve external OAuth2 authorization-server metadata

pingsutw:
## Tracking issue

Related to #6998

The v2 runs service exposes `flyteidl2.auth.AuthMetadataService`, but
`GetOAuth2Metadata` is an unimplemented stub (returns `Unimplemented`/501), and
there is no `/.well-known/oauth-authorization-server` HTTP route. As a result,
OAuth2/OIDC discovery clients (flyte-sdk) and upstream proxies cannot
learn which authorization server to use from a v2 deployment.

This is needed when a deployment fronts its API with an **external** identity
provider (e.g. Okta) — for example when an ALB validates Okta-issued JWTs in
front of the API. Clients that discover auth at the deployment host must be
pointed at the external IdP so they obtain tokens the upstream actually accepts.

Implement OAuth2 authorization-server metadata in the runs service by **proxying
an external authorization server's metadata document**:

- **`runs/service/auth_metadata_service.go`** — implement `GetOAuth2Metadata`:
  when an external auth server is configured, fetch its
  `.well-known/oauth-authorization-server` (with retry) and return it. Unknown
  fields are discarded (`protojson` with `DiscardUnknown`) so real servers like
  Okta — which return many extra metadata fields — parse cleanly. When not
  configured, it returns `Unimplemented`. Also adds `OAuth2MetadataHTTPHandler`,
  which serves the document at the RFC 8414 well-known path as proto3 JSON.
- **`runs/setup.go`** — pass the new config into `NewAuthMetadataService` and
  mount `GET /.well-known/oauth-authorization-server` on the mux.
- **`runs/config/config.go`** — add `AuthMetadataConfig`
  (`externalAuthServerBaseUrl`, `externalMetadataUrl`, `retryAttempts`,
  `retryDelay`) under the `runs` config section.

The external-fetch/unmarshal/retry logic is adapted from #6998; this PR ports
just the metadata-serving slice (no token issuance / cookie / middleware stack).

## How was this patch tested?

Unit tests in `runs/service/auth_metadata_service_test.go`:
- not-configured → `Unimplemented` (RPC) / `501` (HTTP)
- external proxy returns the upstream document, including snake_case + extra
  fields (Okta-shaped) via `DiscardUnknown`
- custom metadata path resolution
- upstream `5xx` → `Unavailable`
- HTTP handler returns `200` with camelCase proto3 JSON

`go build ./runs/...`, `go vet`, and `gofmt` clean.

### Labels

added


pingsutw:
Triage of the Copilot review comments as of `84ec9d46c`:

**Fixed in 84ec9d46c**
- Shared `http.Client` (10s timeout) instead of per-request construction
- Upstream metadata read bounded to 1 MiB
- Successful metadata fetches cached for 5 min (public endpoint; avoids an outbound fetch + retries per discovery call)
- Non-retryable upstream 4xx now classified `Internal` (was blanket `Unavailable`); pre-classified connect codes propagate

runs/config/config.go
```diff
@@ -62,6 +62,61 @@ type Config struct {
...
+	RetryDelay time.Duration `json:"retryDelay" pflag:",Delay between external metadata fetch attempts"`
+
+	// AuthorizationMetadataKey is the header/metadata key clients should place
+	// tokens in, returned by GetPublicClientConfig (default "authorization").
+	AuthorizationMetadataKey string `json:"authorizationMetadataKey" pflag:",Header key clients should use for tokens"`
+
+	// FlyteClient is the public (CLI/SDK) OAuth2 client configuration returned
+	// by GetPublicClientConfig.
+	FlyteClient FlyteClientConfig `json:"flyteClient"`
+}
```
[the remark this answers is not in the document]
pingsutw: Done in 62fce0c09 — `go generate ./runs/config` was actually failing on pre-existing unsupported types (`[]DomainConfig`, raw `time.Duration`, `float64`), which is why the file was stale. Converted `authMetadata.retryDelay` to `config.Duration` (keeps the flag) and excluded the unsupported pre-existing fields with `pflag:"-"`; regenerated `config_flags.go` + `config_flags_test.go` (adds `authMetadata.*` and the previously missing `triggerScheduler.*` flags).

runs/service/auth_metadata_service.go
```diff
@@ -2,33 +2,302 @@ package service
...
 var _ authconnect.AuthMetadataServiceHandler = (*AuthMetadataService)(nil)
 
+// GetPublicClientConfig returns the public (CLI/SDK) OAuth2 client settings.
 func (s *AuthMetadataService) GetPublicClientConfig(
 	ctx context.Context,
 	req *connect.Request[auth.GetPublicClientConfigRequest],
 ) (*connect.Response[auth.GetPublicClientConfigResponse], error) {
+	authMetadataKey := s.cfg.AuthorizationMetadataKey
+	if authMetadataKey == "" {
+		authMetadataKey = "authorization"
```
EngHabu, line 73: Move to const.
Shouldn't it be `Authorization` (capitalized)?

runs/service/auth_metadata_service.go
```diff
@@ -2,33 +2,302 @@ package service
...
+	}
+
+	// Serve from cache while fresh; only successful fetches are cached.
+	s.mu.Lock()
+	if s.cachedMetadata != nil && time.Now().Before(s.cacheExpiry) {
+		cached := proto.Clone(s.cachedMetadata).(*auth.GetOAuth2MetadataResponse)
+		s.mu.Unlock()
+		return connect.NewResponse(cached), nil
+	}
+	s.mu.Unlock()
```
EngHabu: Make it an RWLock instead so multiple readers don't block

runs/setup.go
```diff
@@ -112,11 +112,18 @@ func Setup(ctx context.Context, sc *app.SetupContext) error {
...
+	authMetadataSvc := service.NewAuthMetadataService(sc.BaseURL, cfg.AuthMetadata)
 	authMetadataPath, authMetadataHandler := authconnect.NewAuthMetadataServiceHandler(authMetadataSvc, connect.WithInterceptors(otelInterceptor))
 	sc.Mux.Handle(authMetadataPath, authMetadataHandler)
 	logger.Infof(ctx, "Mounted AuthMetadataService at %s", authMetadataPath)
 
+	// Serve OAuth2 authorization-server metadata at the RFC 8414 well-known path
+	// so OAuth2/OIDC discovery clients (flyte-sdk) can find it. When
+	// runs.authMetadata.externalAuthServerBaseUrl is set, this proxies the
+	// external IdP's (e.g. Okta) metadata document.
+	sc.Mux.Handle("/.well-known/oauth-authorization-server", service.OAuth2MetadataHTTPHandler(authMetadataSvc))
```
EngHabu, line 125: Do we also need to serve `/.well-known/openid-configuration`?  maybe not

runs/service/auth_metadata_service.go
```diff
@@ -2,33 +2,302 @@ package service
...
+		httpClient:      &http.Client{Timeout: externalFetchTimeout},
 	}
 }
 
 var _ authconnect.AuthMetadataServiceHandler = (*AuthMetadataService)(nil)
 
+// GetPublicClientConfig returns the public (CLI/SDK) OAuth2 client settings.
 func (s *AuthMetadataService) GetPublicClientConfig(
 	ctx context.Context,
 	req *connect.Request[auth.GetPublicClientConfigRequest],
```
popojk: nit: Can we fix the "ctx" and "req" variable name to "_" as they are unused?

runs/service/auth_metadata_service.go
```diff
@@ -2,33 +2,302 @@ package service
...
+
+func connectCodeToHTTPStatus(err error) int {
+	var connectErr *connect.Error
+	if !errors.As(err, &connectErr) {
+		return http.StatusInternalServerError
+	}
+	switch connectErr.Code() {
+	case connect.CodeUnimplemented:
+		return http.StatusNotImplemented
+	case connect.CodeInvalidArgument:
```
popojk, line 225: nit: Should we remove this case as we don't return connect.CodeInvalidArgument error in GetOAuth2Metadata function?

runs/service/auth_metadata_service.go
pingsutw: it's the same as sdk. https://github.com/flyteorg/flyte-sdk/blob/d2623981ef88a389bea521fc689f90dd58105d68/src/flyte/remote/_client/auth/_client_config.py#L23

runs/service/auth_metadata_service.go
pingsutw: fixed

runs/setup.go
pingsutw: I don't use that so far, I'll add it later if needed

runs/service/auth_metadata_service.go
pingsutw: I'll keep it. we may return that in the future

runs/service/auth_metadata_service.go
pingsutw: fixed

machichima (approved):

PR #7517, opened 2026-06-12
fix(clustered): correctly reconcile JobSet phase during restarts

AdilFayyaz:
  
The clustered (JobSet) plugin was prematurely surfacing task failures during JobSet restarts: a single worker failure
within the restart budget could fast-fail the task, and rank-0 pod lookups used an unsuffixed name that never matched the real, randomly-suffixed pod. This caused spurious failures and missed diagnostics while a JobSet was legitimately restarting.

- Gate maybeFastFailWorker0 on restart budget: worker failures only surface as RetryableFailure once Status.Restarts >= MaxRestarts; otherwise the task stays Running. Pending-pod diagnostics (e.g. ImagePullBackOff) still fast-fail regardless of budget.
- Handle the RestartingJobSet condition and the no-condition-yet case via a new hasJobSetStarted check (restarts, worker readiness/activity, or prior plugin state), reporting Running with a restart-attempt reason instead of falling back to Initializing/failure.
- Replace unsuffixed rank0PodName matching with findRank0Pod, a suffix-tolerant lookup that deterministically prefer

## How was this patch tested?
 - New unit tests cover: suffixed/deterministic rank-0 selection, failure-with-budget-remaining → Running, budget-exhausted → RetryableFailure, pending ImagePullBackOff → fast-fail regardless of budget, RestartingJobSet condition → Running (attempt N), restarts with no true condition → Running, and plugin-state read error → fallback to status.
 - Tested a failing example from the sdk on dogfood-1

## Stack

### DIFF
--- a/flyteplugins/go/tasks/plugins/k8s/clustered/util.go
+++ b/flyteplugins/go/tasks/plugins/k8s/clustered/util.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	jobsetv1alpha2 "sigs.k8s.io/jobset/api/jobset/v1alpha2"
 
 	"github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/k8s"
@@ -61,7 +62,7 @@ func shouldPreferRank0Pod(candidate, current *v1.Pod) bool {
 // Returns nil when not found or when listing pods fails.
 func findRank0Pod(ctx context.Context, pluginContext k8s.PluginContext, jobSet *jobsetv1alpha2.JobSet) *v1.Pod {
 	podList := &v1.PodList{}
-	if err := pluginContext.K8sReader().List(ctx, podList); err != nil {
+	if err := pluginContext.K8sReader().List(ctx, podList, client.InNamespace(jobSet.Namespace)); err != nil {
 		logger.Warnf(ctx, "failed to list pods for JobSet %s/%s rank-0 lookup: %v", jobSet.Namespace, jobSet.Name, err)
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


flyteplugins/go/tasks/plugins/k8s/clustered/util.go
```diff
@@ -19,14 +30,59 @@ var (
...
+	if candidateActive != currentActive {
+		return candidateActive
+	}
+	if candidate.CreationTimestamp.Time.Equal(current.CreationTimestamp.Time) {
+		return candidate.Name > current.Name
+	}
+	return candidate.CreationTimestamp.After(current.CreationTimestamp.Time)
+}
+
+// findRank0Pod locates the rank-0 pod by prefix-matching the real pod name, which includes a random suffix.
```
pingsutw: Could we update the comment here since we get the pods by labels under the hood

flyteplugins/go/tasks/plugins/k8s/clustered/util.go
```diff
@@ -19,14 +30,59 @@ var (
...
+		return nil
+	}
+
+	var selected *v1.Pod
+	for i := range podList.Items {
+		pod := &podList.Items[i]
+		if !isRank0PodName(jobSet.Name, pod.Name) {
+			continue
+		}
+		if selected == nil || shouldPreferRank0Pod(pod, selected) {
```
pingsutw, line 81: qq: does it mean we may have two rank-0 pods, when will it happen?

AdilFayyaz: yes but only for a brief period during a restart. `ShouldPreferRank0Pod` would then just mean that we pick the new one instead of the failed one

AdilFayyaz: Because the old one could still be aborting and not deleted yet so made sense to check the active one and keep only that

pingsutw (approved):

PR #7525, opened 2026-06-16
refactor(flyte-binary): consolidate DB config to runs.database

pingsutw:
## Tracking issue

N/A — chart cleanup.

The flyte-binary chart configured the database in **two** places:
- `flyte-core-components.runs.database` — what the unified binary actually
  connects through (`runsconfig.GetConfig().Database` in `manager/cmd/main.go`,
  registered under the `runs` config section).
- `configuration.database` — a separate top-level block that the unified binary
  does **not** read. It only fed the `waitForDB` init container, the
  `002-database.yaml` config file (a top-level `database:` section read only by
  the standalone `runs`/`cache_service` cmd binaries, never the unified binary),
  and an optional password Secret.

Keeping both invites drift (the two could point at different DBs) and is
confusing. This consolidates to a single, flat source of truth:
`configuration.database`. Users set the DB in one conventional place, and the
chart maps it to `runs.database` — the path the unified binary actually reads —
at render time.

- **`values.yaml`** — removed the `flyte-core-components.runs.database` block.
  `configuration.database` is now the single DB config (with `maxIdleConnections`,
  `maxOpenConnections`, `connMaxLifeTime`, and the `postgres` sub-block).
- **`configmap.yaml`** — `002-database.yaml` now renders `configuration.database`
  **under `runs.database`** (instead of the old top-level `database:`), so the
  unified binary reads it at `runsconfig.GetConfig().Database`. The config dir is
  loaded via `--config /etc/flyte/config.d/*.yaml`, so this file deep-merges with
  the `runs:` section in `000-core.yaml` and the password in
  `012-database-secrets.yaml`. The password field is stripped before `toYaml`
  (`deepCopy` + `unset`) so it is never written to the ConfigMap in plaintext —
  it is supplied via the Secret instead.
- **`config-secret.yaml`** — the optional DB-password Secret
  (`012-database-secrets.yaml`) now reads `configuration.database.postgres.password`
  and renders it as `runs.database.postgres.password`.
- **`deployment.yaml`** — the `waitForDB` init container derives its `pg_isready`
  target from `configuration.database.postgres`, so it waits on exactly the DB
  the binary connects to.

`configuration.database` remains the single, required DB config.

## How was this patch tested?

- `helm template` renders cleanly both with a values override and with bare
  chart defaults.
- Verified `002-database.yaml` renders `runs.database` and that the password is
  **never** present in the ConfigMap (0 plaintext occurrences) while still being
  carried by `012-database-secrets.yaml`.
- Verified the `wait-for-db` init container resolves to the configured
  `configuration.database` host/port/user.
- `helm lint` passes.

### Labels

changed


machichima:
Should we also update 

https://github.com/flyteorg/flyte/blob/9c41e6edf941d7fe709464b97588e3724b531b2a/charts/flyte-binary/README.md?plain=1#L46-L52

charts/flyte-binary/values.yaml
```diff
@@ -55,23 +42,22 @@ flyte-core-components:
...
+  # DB config the binary reads (runsconfig.GetConfig().Database). The waitForDB
+  # init container and the optional DB-password Secret both derive from here.
   database:
-    # username Name for user to connect to database as
-    username: postgres
-    # password Password to connect to database with
-    # If set, a Secret will be created with this value and mounted to Flyte pod
-    password: ""
-    # passwordPath Mountpath of file containing password to be added to Flyte deployment
-    passwordPath: ""
```
machichima, line 66: I think this is still supported? Should we add it back?

pingsutw: Good catch — yes, it's still supported. `passwordPath` is read by the binary (`flytestdlib/database/config.go`: `PostgresConfig.PasswordPath`, with "Either Password or PasswordPath must be set"), and the legacy flat `database.passwordPath` is still mapped into `Postgres.PasswordPath` in `GetConfig()`.

Added it back under `configuration.database.postgres.passwordPath` in 689f63ab7. It's the file-based password option (point at a file you mount yourself via `deployment.extraVolumes`/`extraVolumeMounts`, e.g. from an external secret manager) — distinct from `password`, which the chart turns into a managed Secret; the two are mutually exclusive. It flows straight through to `runs.database.postgres.passwordPath` via the existing configmap rendering (only `password` is stripped), so no template change was needed.

machichima (approved):

PR #7536, opened 2026-06-17
Add disable identity enrichment flag for ListApps

pmahindrakar-oss:
  Some AppService List callers only need app metadata/spec/status and do not need display-oriented identity enrichment in the response.
  This adds an explicit request-level opt-out so those callers can avoid unnecessary identity enrichment work while preserving the existing default behavior for current callers.

  * Add `disable_identity_enrichment` to `flyteidl2.app.ListRequest`.
  * Regenerate the Go protobuf and validation artifacts.

  The new field defaults to `false`, so existing clients continue to receive the current behavior unless they explicitly opt out.

## How was this patch tested?

  ```bash
  go test ./gen/go/flyteidl2/app

### Labels

  changed

### Setup process

  N/A

### Screenshots

  N/A

## Related PRs

  Union cloud PR will consume this request flag.

## Stack
  <!-- branch-stack -->

## Docs link

  N/A


mhotan (dismissed):

katrogan (dismissed):
LGTM but I think you need to regen protos in all languages, right?

katrogan (dismissed):

katrogan (approved):

PR #7546, opened 2026-06-17
feat(flyte-binary): add optional JWT and auth-discovery ingresses

pingsutw:
> [!NOTE]
> **Stacked on #7543** — it uses the `ingress.apiPaths` helper introduced there. Base is `flyte-binary-remove-grpc-ingress`; will retarget to `main` once #7543 merges. Review the [stacked diff](https://github.com/flyteorg/flyte/compare/flyte-binary-remove-grpc-ingress...flyte-binary-auth-discovery-ingresses).

On ingress controllers like AWS ALB, a single ingress resource can't combine different auth behaviors — cookie-OIDC (browser) vs JWT (token) — and the unauthenticated OAuth-discovery endpoints must be exposed *without* auth, separately from the auth-gated API. The working pattern is three ingresses in one ALB group, ordered by precedence:

| Ingress | precedence | auth | paths |
|---|---|---|---|
| wellknown | highest | **none** | `/.well-known/oauth-authorization-server`, `AuthMetadataService` |
| api-jwt | middle | **JWT** (only `Authorization: Bearer*`) | the flyteidl2 API services (incl. `IdentityService`, `SettingsService`) |
| http | lowest | OIDC cookie (`authenticate`) | `/v2` + the same API services |

So: only OAuth discovery is public, Bearer→JWT, everything else→cookie. Until now the first two were hand-rolled as out-of-band `kubectl` ingresses, so every operator had to recreate them. This makes them first-class, **opt-in** chart resources.

Only the OAuth-discovery paths are public — clients need them to bootstrap before they hold a token. Everything that returns user/tenant state (`IdentityService`, `SettingsService`) stays behind auth.

- **`templates/ingress/api-jwt.yaml`** (new) — gated on `ingress.apiJwtIngress.enabled`. Serves the shared `ingress.apiPaths` backed by the http service. The controller-specific bits (precedence, the `Authorization: Bearer*` match condition, JWT validation config) are supplied via `ingress.apiJwtIngress.annotations`. Reusing `apiPaths` means **all** API services — including `RunLogsService` — are JWT-validated for Bearer clients (the old hand-made ingress omitted `RunLogsService`, so Bearer log-streaming fell through to the cookie ingress and got 302'd).
- **`templates/ingress/wellknown.yaml`** (new) — gated on `ingress.wellknownIngress.enabled`. Serves the unauthenticated OAuth-discovery paths (`/.well-known/oauth-authorization-server`, `AuthMetadataService`) via the new `ingress.wellknownPaths` helper.
- **`_helpers.tpl`** — `IdentityService` and `SettingsService` live in `apiPaths` (auth-gated via both the JWT and http ingresses), **not** in `wellknownPaths`. They return the authenticated user's identity / settings, so they must not be public.
- **`values.yaml`** — `ingress.apiJwtIngress.*` and `ingress.wellknownIngress.*`, both **disabled by default**; `annotations` / `ingressClassName` are pass-through so the chart stays controller-agnostic.
- **`README.md`** — regenerated parameter rows.

Both ingresses render only when `ingress.create` is also true.

## How was this patch tested?

No Helm unit-test harness exists for this chart, so verified by render + live apply on the `flyte-development` ALB cluster:

- `helm lint` + `helm template` clean; all three ingresses render with the correct paths and `group.order`.
- Applied to the live release and smoke-tested the full dual-auth routing:

  | request | expected | result |
  |---|---|---|
  | `AuthMetadataService`, no auth | 200 (public) | ✅ 200 |
  | `/.well-known/oauth-authorization-server`, no auth | 200 (public) | ✅ 200 |
  | API call, no auth | 302 (cookie ingress) | ✅ 302 |
  | API call, `Authorization: Bearer` | 401 (JWT ingress validates/rejects, not 302) | ✅ 401 |
  | `RunLogsService` + Bearer | 401 (now JWT-validated) | ✅ 401 |
  | `IdentityService`, no auth | 302 (now auth-gated) | ✅ 302 |
  | `IdentityService` + Bearer | 401 (JWT) | ✅ 401 |
  | `SettingsService`, no auth | 302 (now auth-gated) | ✅ 302 |

### Labels

- **added**


Sovietaced (commented):
One comment for now

Sovietaced (commented):
I have a feeling we will need TLS support on all of these?

It looks like you've dropped support for terminating TLS at the ingress entirely

Sovietaced (approved):

Issue #7455, opened 2026-05-29
[flyte2] Executor: add custom domain metrics (TaskAction reconcile, cache, GC, plugins)


pingsutw:
> Part of #7445. Best paired with #7453 (which makes promutils-scoped metrics actually scrapeable on the executor endpoint).

## Summary

The executor emits **no custom domain metrics**. It gets controller-runtime built-ins for free (reconcile counts/errors/duration, workqueue depth/latency), but there's no visibility into the executor's own behavior: TaskAction reconcile outcomes, cache effectiveness, garbage-collection activity, or plugin execution latency. Add these using the metrics `Scope` that's already plumbed through.

## Background

The plumbing exists but is unused: `executor/pkg/plugin/setup_context.go` already exposes `MetricsScope() promutils.Scope`, and `executor/setup.go` constructs `promutils.NewScope("executor")`. But a grep for metric instruments (`MustNewCounter`/`MustNewGauge`/`MustNewStopWatch`, `.Inc()`/`.Observe()`/`.Set()`/`.Start()`) across the executor's own code returns nothing — the controllers don't emit any.

## What to do

Add metrics (under dedicated sub-scopes, e.g. `scope.NewSubScope("taskaction")`, `"cache"`, `"gc"`) to the core executor logic:

- **TaskAction controller** (`executor/pkg/controller/taskaction_controller.go`): reconcile outcomes labeled by result/phase (success/error/requeue), and per-reconcile latency. (Note: controller-runtime already gives generic reconcile totals — add only what the generic metrics don't cover, e.g. terminal phase counts.)
- **TaskAction cache** (`executor/pkg/controller/taskaction_cache.go`): cache hit / miss / eviction counters, and current size gauge.
- **Garbage collector** (`executor/pkg/controller/garbage_collector.go`): objects deleted (counter), deletion errors (counter), and GC sweep duration.
- **Plugin execution** (via the registry / `setupContext.MetricsScope()`): per-plugin execution latency and error counts, if not already covered.

## Acceptance criteria

- [ ] The above metrics appear on the executor metrics endpoint (depends on #7453 being resolved so default-registry metrics are exposed; until then they can be verified via the default registry in a unit test).
- [ ] Metrics use dedicated sub-scopes and are created once (no duplicate-registration panics).
- [ ] Unit tests assert the relevant counters/gauges update (e.g. cache hit increments on a hit; GC deletion counter increments on delete).

## Pointers

[...]

Issue #7445, opened 2026-05-29
[flyte2] Emit Prometheus metrics from v2 services (runs service + app framework)


pingsutw:
## Summary

The v2 services built on the `flytestdlib/app` framework (starting with the **runs service**) currently expose **no Prometheus metrics**. This is an umbrella issue tracking the work to add observability, broken into small, independent tasks that are great for first-time contributors.

## Background

When you run the runs service (`runs/cmd/main.go` → `app.App.Run()`), the only HTTP endpoints served are `/healthz` and `/readyz` (`flytestdlib/app/app.go:94-109`). There is **no `/metrics` endpoint**, and nothing emits Prometheus metrics:

- `app.SetupContext` has a `Scope promutils.Scope` field (`flytestdlib/app/context.go:62-63`), but it is **never assigned**.
- `runs/cmd/main.go:43` passes `promutils.NewTestScope()` into the `DataStore` — a throwaway scope that registers nothing.
- `runs/setup.go` registers no counters/gauges/timers and no RPC interceptors.
- The old v1 pattern that served `/metrics` (`flytestdlib/profutils/server.go:118`, `promhttp.Handler()`) was **not ported** into the v2 `app` framework.

So today a Prometheus scrape of the runs service returns nothing useful.

## Goal

Give every v2 service that uses the `app` framework a Prometheus `/metrics` endpoint, a real metrics `Scope`, and meaningful instrumentation on the runs service hot paths.

## Tasks

This work is split so multiple contributors can pick up pieces in parallel. **Task 1 is the foundation and must land first**; tasks 2–4 depend on it and can then proceed independently.

- [ ] #7446 (**foundation**): Add a `/metrics` endpoint and initialize the metrics `Scope` in the `app` framework — _blocker for the rest_
- [ ] #7447: Add gRPC/Connect RPC metrics interceptor to the runs service _(depends on #7446)_
- [ ] #7448: Instrument the runs DB repository layer with Prometheus metrics _(depends on #7446)_
- [ ] #7449: Instrument the runs reconcilers (abort-reconciler) with Prometheus metrics _(depends on #7446)_
- [ ] #7450: Instrument the actions service (watcher metrics + dropped-updates counter) _(depends on #7446)_

**Executor** (controller-runtime based — already has a `/metrics` endpoint, but gaps remain):
- [ ] #7453: Expose promutils (default-registry) metrics on the controller-runtime metrics endpoint — _scoped metrics are collected but not scraped_
[...]

PR #7550, opened 2026-06-18
feat(executor): add OTel metrics for the TaskAction garbage collector

davidlin20dev:

Adds OpenTelemetry metrics for the executor's TaskAction garbage collector, registered on the executor's existing meter provider.

| metric | type | labels |
|---|---|---|
| `taskaction.gc.deletions` | `Int64Counter` | `outcome` = `deleted` \| `already_gone` \| `failed` |
| `taskaction.gc.sweep.duration` | `Float64Histogram` (ms) | `error` = `true` \| `false` |

This replaces the promutils implementation this PR originally carried, per review feedback that OTel is the direction for executor metrics. The pattern follows `executor/pkg/controller/metrics.go` from #7483.

A few design notes:

- **Registration failure is non-fatal.** If instrument registration fails, the GC runs unmetered rather than blocking startup, matching `registerTaskActionMetrics`.
- **Deletions are labeled by `outcome`, not a boolean.** A delete attempt has three results: it succeeded, the object was already gone (Kubernetes cascade-deleted it when the GC removed its parent earlier in the same sweep), or it failed. A boolean can't express the middle case, and skipping it undercounts what the sweep actually removed.
- **The sweep histogram keeps a boolean `error` label**, since a sweep has only two outcomes.
- **The histogram overrides the SDK's default bucket boundaries**, which top out at 10s. A sweep over a large backlog runs for minutes, so the defaults would collapse every slow sweep into the overflow bucket and make it impossible to tell a 12-second sweep from a 4-minute one.

The executor gets reconcile counts and workqueue metrics for free from controller-runtime, but nothing reports garbage collection activity: how much is being deleted, whether deletes are failing, or how long a sweep takes. See #7455.

## How was this patch tested?

- Unit tests for the instruments: a noop provider returns no metrics, deletions record under each of the three outcomes, sweep duration records with its error label, and recording through a nil metrics struct is a safe no-op.
- An envtest integration spec that runs the real `collect()` against a live API server with a real meter provider and asserts the recorded values, so the call sites in `collect()` are covered rather than just the helpers.
- Verified the spec catches regressions: it fails if a recording call is removed from `collect()`.

## Labels

Related to #7455

### DIFF
--- a/executor/pkg/controller/garbage_collector.go
+++ b/executor/pkg/controller/garbage_collector.go
@@ -4,11 +4,31 @@ import (
 	"context"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	flyteorgv1 "github.com/flyteorg/flyte/v2/executor/api/v1"
+	"github.com/flyteorg/flyte/v2/flytestdlib/promutils"
 )
+
+// gcMetrics holds the Prometheus instruments for the garbage collector. They are
+// created once (in newGCMetrics) and only updated thereafter, never re-registered.
+type gcMetrics struct {
+	deleted   prometheus.Counter
+	errors    prometheus.Counter
+	sweepTime promutils.StopWatch
+}
+
+// newGCMetrics builds the garbage collector instruments under the given scope.
+// Call exactly once per scope to avoid duplicate registration panics.
+func newGCMetrics(scope promutils.Scope) gcMetrics {
+	return gcMetrics{
+		deleted:   scope.MustNewCounter("objects_deleted", "Total TaskActions deleted by the garbage collector"),
+		errors:    scope.MustNewCounter("deletion_errors", "Total errors encountered while deleting expired TaskActions"),
+		sweepTime: scope.MustNewStopWatch("sweep_duration", "Duration of a full garbage collection sweep", time.Millisecond),
+	}
+}
 
 // GarbageCollector periodically deletes terminal TaskActions that have exceeded their TTL.
 // It implements the controller-runtime manager.Runnable interface.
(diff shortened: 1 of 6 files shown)


Sovietaced:
Is it possible to rework these using OTEL instead of prometheus?

davidlin20dev:
> Is it possible to rework these using OTEL instead of prometheus?

Hi @Sovietaced, happy to switch these to OTel.

I went with Prometheus because #7445 and #7455 (umbrella issue) are both centered around promutils, so that's
the direction I followed. Before I rework it, though, I just want to make sure I'm building the right thing.

@pingsutw, since you opened these issues: could you confirm whether OTel is the preferred
direction now, and whether the issues should be updated to reflect that? Happy to rework this
on the meter provider (mirroring #7483) once I know which way to go.

Thanks both, appreciate the help!

pingsutw:
@davidlin20dev thanks for working it. Sorry for the late reply, I was OOO. yes, we decide to use OTEL

davidlin20dev:
> @davidlin20dev thanks for working it. Sorry for the late reply, I was OOO. yes, we decide to use OTEL

@pingsutw thanks for confirming! quick question: should I wait for the issue to be updated, or is it fine to start the OTel rework now?

pingsutw:
It's fine to start the work now

davidlin20dev:
Hi @pingsutw, I'm working on the OTel rework now, and there's one design decision I'm debating on the deletions counter.

Right now I'm labeling it `error=true|false`, and skipping `NotFound` entirely. If I'm reading the comment in `collect()` right, those are usually objects k8s already cascade-deleted when the parent was removed in the same sweep, so the count might end up lower than what actually got cleaned up.

I'm thinking of labeling by `outcome` with three values (`deleted`, `already_gone`, `failed`) instead, so the total reflects everything that was actually deleted. Let me know if this makes sense, and I'd appreciate your input on this!

davidlin20dev:
Pushed the OTel rework. On the question above, I went ahead and implemented the outcome version (deleted / already_gone / failed) so there's something concrete to look at. I can switch it back to the boolean if needed. Please take a look.

pingsutw (approved):
LGTM, thanks. Could you also test it in the devbox and help update the [grafana dashboard](https://github.com/flyteorg/flyte/blob/main/monitoring/dashboards/flyte-execution.json)?
I added a [make target](https://github.com/flyteorg/flyte/blob/a803bddc6e8bda4c562d69108a7f2a900d11e158/Makefile#L54-L55) for the devbox to install the grafana stack, so you can easily to test it in the devbox.

davidlin20dev:
Thanks! Yeah sure, I'll take a look at the devbox and the dashboard.

davidlin20dev:
> LGTM, thanks. Could you also test it in the devbox and help update the [grafana dashboard](https://github.com/flyteorg/flyte/blob/main/monitoring/dashboards/flyte-execution.json)? I added a [make target](https://github.com/flyteorg/flyte/blob/a803bddc6e8bda4c562d69108a7f2a900d11e158/Makefile#L54-L55) for the devbox to install the grafana stack, so you can easily to test it in the devbox.

Hi @pingsutw , I just submitted a PR addressing this: #7836. Please take a look. Thank you!

PR #7554, opened 2026-06-18
Fix: Log links with showWhilePending: true are not shown while pendin…

fg91:

Log links can be shown before a task reaches the `Running` phase by setting `showWhilePending: true`. For the ray plugin, this doesn't actually work. This PR fixes this and adds a unit test that would have caught this bug.

## How was this patch tested?

* Added a unit test
* Tested flytepropeller with this fix in our staging cluster

### DIFF
--- a/flyteplugins/go/tasks/plugins/k8s/ray/ray.go
+++ b/flyteplugins/go/tasks/plugins/k8s/ray/ray.go
@@ -671,14 +671,12 @@ func (plugin rayJobResourceHandler) GetTaskPhase(ctx context.Context, pluginCont
 		return pluginsCore.PhaseInfoUndefined, err
 	}
 
-	if len(rayJob.Status.JobDeploymentStatus) == 0 {
-		return pluginsCore.PhaseInfoQueuedWithTaskInfo(pluginsCore.DefaultPhaseVersion, "Scheduling", info), nil
-	}
-
 	var phaseInfo pluginsCore.PhaseInfo
 
 	// KubeRay creates a Ray cluster first, and then submits a Ray job to the cluster
 	switch rayJob.Status.JobDeploymentStatus {
+	case rayv1.JobDeploymentStatusNew:
+		phaseInfo, err = pluginsCore.PhaseInfoQueuedWithTaskInfo(pluginsCore.DefaultPhaseVersion, "Scheduling", info), nil
 	case rayv1.JobDeploymentStatusInitializing:
 		phaseInfo, err = pluginsCore.PhaseInfoInitializing(pluginsCore.DefaultPhaseVersion, "cluster is creating", info), nil
 	case rayv1.JobDeploymentStatusRunning:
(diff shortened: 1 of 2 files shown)


flyteplugins/go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -671,14 +671,12 @@ func (plugin rayJobResourceHandler) GetTaskPhase(ctx context.Context, pluginCont
...
-		return pluginsCore.PhaseInfoQueuedWithTaskInfo(pluginsCore.DefaultPhaseVersion, "Scheduling", info), nil
-	}
-
 	var phaseInfo pluginsCore.PhaseInfo
 
 	// KubeRay creates a Ray cluster first, and then submits a Ray job to the cluster
 	switch rayJob.Status.JobDeploymentStatus {
+	case rayv1.JobDeploymentStatusNew:
+		phaseInfo, err = pluginsCore.PhaseInfoQueuedWithTaskInfo(pluginsCore.DefaultPhaseVersion, "Scheduling", info), nil
 	case rayv1.JobDeploymentStatusInitializing:
```
fg91, line 680: Task info including log links is only sent to flyteadmin when either the task phase changes (e.g. queued -> initializing) or when the so-called phase-version within the same phase changes (due a different "reason").

An initial event with phase `Queued` and `pluginsCore.DefaultPhaseVersion` is sent to flyteadmin before `GetTaskPhase` is called/the log links are constructed.

Subsequent calls to `GetTaskPhase` never reach `phaseVersionUpdateErr := k8s.MaybeUpdatePhaseVersionFromPluginContext(&phaseInfo, &pluginContext)` where the phase version would be bumped when necessary due to the early exit that this PR removes.

Because of this, log links with `showWhilePending: true` are constructed by `GetTaskPhase` but are never sent to flyteadmin.

By removing the early exit and ensuring that `MaybeUpdatePhaseVersionFromPluginContext` is reached, this problem is avoided.

flyteplugins/go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -671,14 +671,12 @@ func (plugin rayJobResourceHandler) GetTaskPhase(ctx context.Context, pluginCont
...
 
-	if len(rayJob.Status.JobDeploymentStatus) == 0 {
-		return pluginsCore.PhaseInfoQueuedWithTaskInfo(pluginsCore.DefaultPhaseVersion, "Scheduling", info), nil
-	}
-
 	var phaseInfo pluginsCore.PhaseInfo
 
 	// KubeRay creates a Ray cluster first, and then submits a Ray job to the cluster
 	switch rayJob.Status.JobDeploymentStatus {
+	case rayv1.JobDeploymentStatusNew:
```
fg91, line 678: `const rayv1.JobDeploymentStatusNew rayv1.JobDeploymentStatus = ""` so this is equivalent to the prior `len(rayJob.Status.JobDeploymentStatus) == 0`.

Sovietaced (approved):

Sovietaced:
Makes sense to me. We use the phase version to trigger updates in the UI ourselves..

PR #7565, opened 2026-06-19
Register the connector-service (webapi) plugin in the Flyte 2 executor

### LINKED ISSUE #7564: Register the connector-service (webapi) plugin in the Flyte 2 executor
The Flyte 2 executor doesn't register the connector-service (webapi) plugin, so it can't run connector tasks. The Python SDK already ships the connector interfaces (`AsyncConnector` and friends), but nothing on the executor side runs them: the plugin that delegates to a connector deployment is never registered.

I ran into this building a connector for Armada. The task type resolves to `connector-service`, but the executor has no plugin by that name, so the task never routes.

Registering the plugin isn't quite enough on its own. The webapi machinery also pulls a ResourceManager and a SecretManager from the plugin setup context, and the executor provides neither, so the allocation-token path nil-derefs.

I opened #7565 with the change. It registers connector-service in the executor and supplies a no-op ResourceManager and registrar so the allocation path doesn't panic. With it, a connector works end to end: https://github.com/dejanzele/armada-flyte runs plain `@env.task` functions as Armada jobs through it (fan-out, gang scheduling, a small ML pipeline).

A couple of questions for reviewers:

1. Is registering connector-service in the executor the intended path for v2, or is connector execution meant to be wired up somewhere else?
2. Is a no-op ResourceManager and SecretManager acceptable in the executor, or should it provide real implementations?

This isn't Armada-specific. Any connector backend (Databricks, Ray, etc.) needs the same plugin registered to run on v2.

dejanzele:
Registers the connector-service (webapi) plugin in the Flyte 2 executor so it can run connector tasks. Today a task whose type resolves to `connector-service` has no plugin to handle it, so it never runs.

The webapi machinery also expects a ResourceManager and a SecretManager from the plugin setup context, which the executor doesn't supply. This adds a no-op ResourceManager and registrar so the allocation-token path doesn't nil-dereference.

Closes [#7564.](https://github.com/flyteorg/flyte/issues/7564)

Tested with unit tests for the no-op ResourceManager and registrar, and manually end to end: a real connector ([armada-flyte](https://github.com/dejanzele/armada-flyte)) runs `@env.task` functions as Armada jobs through this change.

### DIFF
--- a/executor/pkg/plugin/resource_manager.go
+++ b/executor/pkg/plugin/resource_manager.go
@@ -1,0 +1,45 @@
+package plugin
+
+import (
+	"context"
+
+	pluginsCore "github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/core"
+)
+
+// The webapi plugin machinery requires the host to supply a ResourceRegistrar at plugin setup and a
+// ResourceManager at task execution, for its allocation-token quota feature. FlytePropeller (v1)
+// always supplied them, defaulting to a no-op manager when no quota backend is configured. The v2
+// executor never reimplemented this, so a webapi plugin that declares ResourceQuotas dereferences a
+// nil registrar or manager. These no-op types restore the contract and grant every allocation. Swap
+// in a real ResourceManager to enforce quotas without touching any plugin.
+
+// noopResourceRegistrar accepts quota registrations without recording them.
+type noopResourceRegistrar struct{}
+
+var _ pluginsCore.ResourceRegistrar = noopResourceRegistrar{}
+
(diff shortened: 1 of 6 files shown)


dejanzele:
Thanks, good catches. Added a no-op `SecretManager` and wired it into both the setup context and the reconciler. It returns an error on lookups rather than an empty string, so a connector task that references a secret fails with a clear message instead of silently getting a blank, which felt like the safer default until there's a real backend. Also dropped the stray `Printf`.

kumare3:
@pingsutw there was some work already on this right

EngHabu (approved):

PR #7570, opened 2026-06-22
Add paused_actions_only filter to ListRunsRequest

SVilgelm:
## Tracking issue

Related to https://linear.app/unionai/issue/ENG26-652

The console wants to surface runs that need a human in the loop (e.g. a banner / filter for runs with a paused HITL gate node). `ListRunsRequest` had no way to express "only runs that currently have a PAUSED action", so the backend filter needs a request field to key on.

Adds a `bool paused_actions_only = 9` field to `workflow.ListRunsRequest`. When set, the Run Service (closed-source, in `unionai/cloud`) restricts results to runs containing at least one action in `ACTION_PHASE_PAUSED`. This PR is just the IDL + regenerated bindings (Go/Python/Rust/TS); the server-side filter lands in `unionai/cloud`.

## How was this patch tested?

IDL-only change; bindings regenerated with `make gen`. The consuming server logic and its unit tests live in the `unionai/cloud` PR.

### Labels

added


flyteidl2/workflow/run_service.proto
```diff
@@ -282,6 +282,10 @@ message ListRunsRequest {
     // Task identifier for filtering runs
     task.TaskIdentifier task_id = 8;
   }
+
+  // When true, only return runs that contain at least one action in the
+  // PAUSED phase (e.g. a human-in-the-loop gate node awaiting input).
+  bool has_paused_action = 9;
```
pingsutw: nit
```suggestion

  // When true, only return runs that contain at least one action in the
  // PAUSED phase (e.g. a human-in-the-loop gate node awaiting input).
  bool paused_actions_only = 9;
```

SVilgelm: done

pingsutw (approved):

kumare3 (approved):

PR #7572, opened 2026-06-22
fix: make devbox build work with Helm v4

pingsutw:

`make devbox-build` runs `kustomize build --enable-helm`, and kustomize < v5.8.0 probes Helm with `helm version -c --short`. Helm v4 removed the `-c`/`--client` shorthand, so the build fails for any developer who has Helm v4 installed:

```
Error: unknown shorthand flag: 'c' in -c
: unable to run: 'helm version -c --short' ... (is 'helm' installed?): exit status 1
make[1]: *** [manifests] Error 1
make: *** [devbox-build] Error 2
```

The devbox `manifests` target also called bare `kustomize` from `PATH`, so the build depended on whatever kustomize/helm version a developer happened to have installed.

- `executor/Makefile`: bump pinned kustomize `v5.7.1` → `v5.8.1`. v5.8.0 added Helm v4 support ([kustomize#6016](https://github.com/kubernetes-sigs/kustomize/pull/6016)); v5.8.1 fixes the v5.8.0 namespace-propagation regression ([kustomize#6044](https://github.com/kubernetes-sigs/kustomize/issues/6013)).
- `docker/devbox-bundled/Makefile`: the `manifests` target now uses the executor's pinned kustomize binary (`../../executor/bin/kustomize`) instead of system `kustomize`, so the devbox build is reproducible regardless of local helm/kustomize versions.

No chart changes are needed: both charts are already `apiVersion: v2` (Helm 3 + 4 compatible), and CI installs its own Helm 3.

## How was this patch tested?

Reproduced with a stub "Helm v4" that rejects the `-c` flag like the real release:

- kustomize **v5.7.1** → reproduces the exact `unknown shorthand flag: 'c' in -c` / `helm version -c --short` error.
- kustomize **v5.8.1** → clears the version probe (no `-c`).

### Labels

- **fixed**

### DIFF
--- a/docker/devbox-bundled/Makefile
+++ b/docker/devbox-bundled/Makefile
@@ -97,16 +97,26 @@ dep_update: SANDBOX_CHART_DIR := ../../charts/flyte-devbox
 dep_update: dep_build
 	cd $(SANDBOX_CHART_DIR)/charts && rm -rf docker-registry flyte-binary knative-serving && for f in *.tgz; do tar xzf "$$f"; done
 
+# Use the executor's pinned, Helm-v4-compatible kustomize instead of whatever
+# kustomize the developer happens to have on PATH. Older kustomize probes with
+# `helm version -c --short`, and Helm v4 dropped the `-c` flag, so a bare
+# `kustomize build --enable-helm` fails on machines with Helm v4 installed.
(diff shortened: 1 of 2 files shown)


katrogan (approved):
nice, thank you!

PR #7575, opened 2026-06-22
ci: bundle flyte-copilot binary into the flyte-binary-v2 image

pingsutw:
## Tracking issue

Closes [ENG26-776](https://linear.app/unionai/issue/ENG26-776/configure-ci-to-build-and-push-flytecopilot-v2-images)

The flytecopilot binary was not built by any CI in this repo. The root `Dockerfile` only built the `flyte` single binary, so the copilot init/sidecar container — which flyteplugins invokes as `/bin/flyte-copilot` (`flytek8s.CopilotCommandArgs`) — had to be built and pushed manually whenever copilot code changed. That's how it recently drifted out of sync (e.g. copilot fixes in #7534 / #7490 not reaching deployed clusters until someone hand-built an image).

Rather than maintain a second image + workflow, **bundle the copilot binary into the existing `flyte-binary-v2` image**:

- **`Dockerfile`** — build `flyte-copilot` alongside the `flyte` single binary in the same builder stage and ship both in the final image: `/usr/local/bin/flyte` (entrypoint) and `/bin/flyte-copilot` (the exact path flyteplugins calls).

The existing `flyte-binary-v2.yml` workflow already builds and pushes this image, so copilot now ships with every flyte-binary-v2 build automatically — no second Dockerfile, target, or workflow needed. The copilot init/sidecar container can reference the same image (which is already present on the nodes running flyte2, so no extra image pull).

### Follow-up (not in this PR)

To actually run copilot from this image, point the copilot image config at `flyte-binary-v2`:
- `flyteplugins/.../flytek8s/config/config.go` default (`cr.flyte.org/flyteorg/flytecopilot:v0.0.15`)
- `charts/flyte-binary/values.yaml` (`copilot.image.*`)

These are deploy-time config and can be overridden via Helm values, so they're left out of this build-focused change.

## How was this patch tested?

- `docker build .` succeeds (linux/amd64 + arm64).
- `docker run <image> --help` → flyte unified service (entrypoint unchanged).
- `docker run --entrypoint /bin/flyte-copilot <image> --help` → copilot `download` / `sidecar` subcommands, confirming the binary is at the path flyteplugins calls.

### Labels

- **added**


charts/flyte-binary/values.yaml
```diff
@@ -126,12 +126,15 @@ configuration:
...
+    # The flyte-copilot binary ships inside the flyte single-binary image
+    # (at /bin/flyte-copilot), so this defaults to the same image as the main
+    # deployment and is bumped in lockstep with it (# FLYTE_TAG).
     image:
       # repository CoPilot sidecar image repository
-      repository: cr.flyte.org/flyteorg/flytecopilot  # FLYTECOPILOT_IMAGE
+      repository: cr.flyte.org/flyteorg/flyte-binary-v2  # FLYTE_IMAGE
       # tag CoPilot sidecar image tag
-      tag: v1.16.4  # FLYTECOPILOT_TAG
+      tag: latest  # FLYTE_TAG
```
pingsutw, line 137: will update it once we have a release pipeline

AdilFayyaz (approved):

PR #7592, opened 2026-06-25
feat(related_to) add related_to field in runspec to point to parent run

afbrock:
This change adds a field to the runspec that is immutable, set at creation time for a run, linking a run to a previous run.  

## Tracking issue

https://linear.app/unionai/issue/ENG26-628/be-link-related-runs-with-related-to

This change enables customers to group together runs that are related in some way.

This PR is limited to simply adding the field to our Proto definition, and the generated files that are created from that definition.

## How was this patch tested?

Run in devbox, then staging.

## Stack


flyteidl2/task/run.proto
```diff
@@ -125,6 +126,11 @@ message RunSpec {
   // Distinct from raw_data_storage.raw_data_prefix, which controls offloaded
   // blob (user data) layout.
   string run_base_dir = 14;
+
+  // Optional pointer to the single parent run this run was derived from (re-run, recover,
+  // clone, or explicit SDK/CLI provenance). Set at creation, immutable thereafter, and scoped
+  // to the same org/project/domain as this run.
+  flyteidl2.common.RunIdentifier related_to = 15;
```
EngHabu, line 133: My gut feeling is we will want to attach some "relationship" (e.g. a Rerun, a parent relationship... etc.) so I want to say make this a message with just the RunIdentifier in it at the moment...

EngHabu (approved):

PR #7596, opened 2026-06-26
feat: emit and parse durations as ISO-8601 in launch form converter

ursucarina:

Launch form duration fields are declared in JSON Schema with `format: "duration"`, which per the spec means an ISO-8601 duration string (e.g. `P2DT3H`). The converter instead emitted Go-style strings (`time.Duration.String()`, e.g. `2h30m`) and only parsed bare-seconds or Go-style strings on the way back in. This made the converter's `DURATION` handling inconsistent with the `format: "duration"` schema it produces, and it rejected ISO-8601 input.

This makes the launch form converter (`runs/service/converter/literal_json_converter.go`) speak ISO-8601 for `DURATION`:

- `literalToJsonValues` now formats durations as ISO-8601 (`formatISO8601Duration`) instead of `Duration.String()`.
- `jsonValueToLiteral` (the `SimpleType_DURATION` case) now parses ISO-8601 first, then falls back to bare seconds and Go-style duration strings for backward compatibility.
- Added `isISO8601Duration` / `parseISO8601Duration` / `formatISO8601Duration` helpers. Year and month lengths are fixed (365-day years, 30-day months) so durations round-trip deterministically.

## How was this patch tested?

Unit tests in `runs/service/converter/literal_json_converter_test.go`:

- Updated the existing duration literal assertion to expect ISO-8601 (`PT1H30M`).
- Added `TestFormatISO8601Duration`, `TestParseISO8601Duration` (valid, invalid, and format↔parse round-trip cases), and `TestJSONValuesToLiteralsDuration` covering ISO-8601 plus the backward-compatible bare-seconds and Go-style inputs.

`go test ./runs/service/converter/` passes.

### Labels

- **changed**

* \#6583 
  - **feat: emit and parse durations as ISO-8601 in launch form converter** :point\_left:

### DIFF
--- a/runs/service/converter/literal_json_converter.go
+++ b/runs/service/converter/literal_json_converter.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"connectrpc.com/connect"
+	"github.com/senseyeio/duration"
 	"github.com/vmihailenco/msgpack/v5"
 	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/structpb"
@@ -24,9 +25,9 @@ import (
 
 var digitsOnlyRegex = regexp.MustCompile(`^\d+$`)
 
-// iso8601DurationRegex matches an ISO-8601 duration (e.g. "P2DT3H4M5S", "PT1.5S", "-PT30M").
-// The two "M" components are disambiguated by position: months precede "T", minutes follow it.
-var iso8601DurationRegex = regexp.MustCompile(`^(-)?P(?:(\d+)Y)?(?:(\d+)M)?(?:(\d+)W)?(?:(\d+)D)?(?:T(?:(\d+)H)?(?:(\d+)M)?(?:(\d+(?:\.\d+)?)S)?)?$`)
+// fractionalSecondRegex matches a fractional seconds component (e.g. "1.5" in "PT1.5S").
+// senseyeio/duration only supports integer seconds; we normalize these before parsing.
+var fractionalSecondRegex = regexp.MustCompile(`(\d+\.\d+)S`)
 
 // Fixed component lengths used to convert ISO-8601 durations to/from seconds. Years and months
 // are inherently ambiguous; these match the launch form frontend (dayjs: 365-day years,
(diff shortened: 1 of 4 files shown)


runs/service/converter/literal_json_converter.go
```diff
@@ -23,6 +24,154 @@ import (
...
+		duration.Seconds = -duration.Seconds
+		duration.Nanos = -duration.Nanos
+	}
+	return duration, nil
+}
+
+// formatISO8601Duration converts a time.Duration into an ISO-8601 duration string
+// (e.g. "P2DT3H4M5S"). Only days/hours/minutes/seconds are emitted (never years/months) so the
+// representation is unambiguous; the launch form frontend renders it back to a readable string.
+func formatISO8601Duration(d time.Duration) string {
```
EngHabu, line 97: We should use a library for this

https://github.com/dustin/go-humanize

runs/service/converter/literal_json_converter.go
```diff
@@ -23,6 +24,154 @@ import (
...
+
+// isISO8601Duration reports whether s looks like an ISO-8601 duration string.
+func isISO8601Duration(s string) bool {
+	s = strings.TrimSpace(s)
+	return strings.HasPrefix(s, "P") || strings.HasPrefix(s, "-P")
+}
+
+// parseISO8601Duration parses an ISO-8601 duration string (e.g. "P2DT3H4M5S") into a
+// durationpb.Duration. Returns an error if the string is not a valid ISO-8601 duration.
+func parseISO8601Duration(s string) (*durationpb.Duration, error) {
```
EngHabu, line 52: Can't we usea library instead?

 channelmeter/iso8601duration

Maybe...

EngHabu (commented):
The error I get even when I write 2h. Which is a valid time.ParseDuration value I thought. So I'm a bit confused

runs/service/converter/literal_json_converter.go
ursucarina: From me looking:

> Short answer: we could, but channelmeter/iso8601duration is a poor fit and would actually be a regression. It's not currently a dependency either, so we'd be adding a new (unmaintained) module.
> 
> The two dealbreakers:
> 
> - It doesn't cover the formatting direction. String() only formats the library's own Duration struct (Years/Weeks/Days/Hours/Minutes/Seconds), not a Go time.Duration. Our literalToJsonValues path needs time.Duration → ISO string, so we'd still hand-roll the decomposition — the library buys us nothing there. By its own docstring it's a "partial implementation of ISO8601 durations (no months)."
> - 
> - It drops cases we explicitly support. It rejects months, can't represent fractional seconds, and has no negatives. Our converter handles all three. Swapping it in would silently narrow behavior and break frontend round-tripping (dayjs can emit P1M).

PR #7602, opened 2026-06-29
Add cluster_pool field to RunSpec

iaroslav-ciupin:
### Labels

## Stack


flyteidl2/task/run.proto
```diff
@@ -125,6 +125,9 @@ message RunSpec {
   // Distinct from raw_data_storage.raw_data_prefix, which controls offloaded
   // blob (user data) layout.
   string run_base_dir = 14;
+
+  // Pool of the assigned queue for given run.
```
wild-endeavor, line 135: can we clarify in comment if this is set by the user (SDK) or the server?  can both cluster and cluster_pool be set?  if both are set is there validation checking?
maybe?
> Set by the server to the name of cluster pool attached to the resolved queue of this run. The resolved queue is the requested queue (this message's cluster field), or a settings default, or the system "default".

iaroslav-ciupin: aah, this is actually not something that user should specify. It will always be calculated from queue. Should it still leave in RunSpec or should we move it somewhere else?

wild-endeavor: i think leaving in runspec is okay, just a comment that this is to be set by the server. some other fields like run base dir are already set by the server.

wild-endeavor (approved):

PR #7606, opened 2026-06-30
Instrument runs service DB repository layer with Prometheus metrics

### LINKED ISSUE #7448: [flyte2] Instrument the runs service DB repository layer with OTEL metrics
> Part of #7445. **Depends on #7446** (the `/metrics` endpoint + `Scope` must exist first).

## Summary

Add Prometheus metrics to the runs service database repository layer so we can observe DB call volume, error rates, and latency per operation.

## Background

The repository implementations live in `runs/repository/impl/` (e.g. `task.go`, `trigger.go`, and the run repo). These wrap GORM/DB calls and currently emit **no metrics** — there is no visibility into how often each query runs, how long it takes, or how often it fails.

## What to do

1. Thread the metrics `Scope` (from #7446, available via `app.SetupContext.Scope`) into the repository constructor(s). `runs/setup.go` calls `repository.NewRepository(sc.DB, cfg.Database)` and `impl.NewProjectRepo(sc.DB)` — extend these (or wrap the repo) to accept a `promutils.Scope`.
2. For each DB operation (create/get/list/update/delete), record:
   - call count (labeled by operation name)
   - error count (labeled by operation name)
   - latency (a Prometheus timer / stopwatch)

   A small helper that wraps a DB call with `start := time.Now(); defer timer.Stop()` + counter increments keeps this DRY.

## Acceptance criteria

- [ ] `/metrics` exposes per-operation DB call count, error count, and latency for the runs repository.
- [ ] Metrics are created once (no duplicate-registration panics) using a dedicated sub-scope, e.g. `scope.NewSubScope("db")`.
- [ ] Unit tests assert that a repository operation increments the expected counter / records latency.

## Pointers

- `runs/repository/impl/` — repository implementations to instrument.
- `runs/repository/repository.go` (the `NewRepository` constructor) and `runs/setup.go:40` where it's called.
- `flytestdlib/promutils/scope.go` — `Scope` helpers (`MustNewCounter`, `MustNewStopWatch`, `NewSubScope`).

## Notes for contributors

- Keep label cardinality low: label by **operation name**, never by row IDs / project / user values.
- This can be done independently of #7447 (RPC interceptors); both consume the same `Scope`.

youdie006:
## Summary

Implements #7448 — instruments the flyte2 runs service DB repository layer with Prometheus metrics.

Adds a `dbMetrics` helper (`runs/repository/impl/metrics.go`) recording, per DB operation, a calls counter, an errors counter, and a latency stopwatch, built from a `promutils.Scope` sub-scope — mirroring flyte's existing metrics idiom (a `Metrics` struct from a `promutils.Scope` via a `newMetrics(scope)` constructor). A `promutils.Scope` is threaded into the action / task / project / trigger repo constructors (each gets its own named sub-scope), and every DB op is wrapped through a DRY `observe(ctx, operation, fn)` helper. Metrics are labeled by operation name only (low cardinality, never row/project/user values). A nil scope yields a no-op `dbMetrics`, so constructors stay usable in unit tests.

## Test plan

- New `runs/repository/impl/metrics_test.go` (TDD red→green): nil-scope no-op, success/error observation, operation-only labeling, plus an end-to-end `CreateProject` test against the embedded postgres DB asserting the call counter, latency, and error counter.
- `go build ./...` exit 0, `gofmt`/`go vet` clean on touched files, `go test ./runs/repository/impl/` ok.

Fixes #7448

### DIFF
--- a/runs/repository/impl/action.go
+++ b/runs/repository/impl/action.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/flyteorg/flyte/v2/flytestdlib/database"
 	"github.com/flyteorg/flyte/v2/flytestdlib/logger"
+	"github.com/flyteorg/flyte/v2/flytestdlib/promutils"
 	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/common"
 	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/core"
 	"github.com/flyteorg/flyte/v2/runs/repository/interfaces"
@@ -30,6 +31,7 @@ type actionRepo struct {
 	db       *sqlx.DB
 	dsn      string // stored for pq.Listener
 	listener *pq.Listener
+	metrics  dbMetrics
 
 	// Subscriber management for LISTEN/NOTIFY
 	runSubscribers    map[chan string]bool
@@ -41,12 +43,15 @@ type actionRepo struct {
 	runNotifyCh    chan string
 }
 
-// NewActionRepo creates a new PostgreSQL repository
-func NewActionRepo(db *sqlx.DB, dbConfig database.DbConfig) (interfaces.ActionRepo, error) {
+// NewActionRepo creates a new PostgreSQL repository. The provided scope is used
+// to register per-operation DB metrics under a "db" sub-scope; pass nil to
+// disable metrics (e.g. in unit tests).
+func NewActionRepo(db *sqlx.DB, dbConfig database.DbConfig, scope promutils.Scope) (interfaces.ActionRepo, error) {
 	dsn := database.GetPostgresDsn(context.Background(), dbConfig.Postgres)
 	repo := &actionRepo{
 		db:                db,
 		dsn:               dsn,
+		metrics:           newDBMetrics(scope),
 		runSubscribers:    make(map[chan string]bool),
 		actionSubscribers: make(map[chan string]bool),
 	}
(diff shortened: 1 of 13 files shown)


Sovietaced:
We shouldn't be manually instrumenting every single DB call. This should be provided through some PG middleware registered as a hook.

youdie006:
@Sovietaced you're right, and sorry for the slow reply.

One wrinkle worth surfacing before I rework it: `runs/repository` is built on `sqlx` (`repository.go` takes a `*sqlx.DB`), not gorm, so there are no ORM callbacks to register a hook on. The equivalent at that layer is wrapping the `database/sql` driver — register an instrumenting `driver.Driver` and build the `sqlx.DB` over it — which intercepts every query in one place and would replace essentially all of the 636 lines here.

That does mean introducing a pattern the repo does not currently have: I searched for `otelsql`, `sqlhooks`, `sql.Register` and a hand-rolled wrapper and found none, so this would be the first driver-level instrumentation in the tree, and it would sit in `flytestdlib/database` rather than in `runs/`.

So it is your call which you would rather have:

1. I close this and open a much smaller PR adding a wrapped driver in `flytestdlib/database`, with the per-call instrumentation here deleted.
2. You would rather not take a driver wrapper (extra dependency or a preference for OTel's `otelsql` specifically), in which case I will just close this.

Happy either way — I would rather not build option 1 speculatively if the answer is 2.

Sovietaced:
> @Sovietaced you're right, and sorry for the slow reply.
> 
> One wrinkle worth surfacing before I rework it: `runs/repository` is built on `sqlx` (`repository.go` takes a `*sqlx.DB`), not gorm, so there are no ORM callbacks to register a hook on. The equivalent at that layer is wrapping the `database/sql` driver — register an instrumenting `driver.Driver` and build the `sqlx.DB` over it — which intercepts every query in one place and would replace essentially all of the 636 lines here.
> 
> That does mean introducing a pattern the repo does not currently have: I searched for `otelsql`, `sqlhooks`, `sql.Register` and a hand-rolled wrapper and found none, so this would be the first driver-level instrumentation in the tree, and it would sit in `flytestdlib/database` rather than in `runs/`.
> 
> So it is your call which you would rather have:
> 
> 1. I close this and open a much smaller PR adding a wrapped driver in `flytestdlib/database`, with the per-call instrumentation here deleted.
> 2. You would rather not take a driver wrapper (extra dependency or a preference for OTel's `otelsql` specifically), in which case I will just close this.
> 
> Happy either way — I would rather not build option 1 speculatively if the answer is 2.
> 

Looking at the issue I think the implementation is ok for now. I will review it later when I look to implement metrics internally.

PR #7611, opened 2026-07-01
[flytepropeller] Preserve deck URI when a node is aborted

### LINKED ISSUE #7610: [BUG] Aborting an execution wipes the deck URI, so decks vanish from the UI on abort
### Flyte & Flytekit version

Flyte v1.16.x through current master; flytekit 1.16.24

### Describe the bug

Aborting a running node wipes its Flyte Deck from the UI, even though the deck was showing a moment earlier and the `deck.html` file still exists in blob storage.

Abort takes a different path than the other terminal states. Normal transitions attach the deck URI in the task handler, which is why decks show for running, succeeded, and failed nodes. But `nodeExecutor.Abort` skips the handler and builds its own node event with no deck URI, and flyteadmin then overwrites the stored URI with that empty value. The blob is untouched; only the pointer is lost.

### Expected behavior

Aborting a node keeps its deck viewable, the same as a failed node does. The recovery path already does this: it checks the deck file exists and re-attaches the URI. The abort path should do the same.

### Additional context / impact

This is a hard blocker for eager workflows. Each eager child is its own execution, and the eager task's deck is the only place the UI surfaces the children and their links. Aborting the parent, the normal way to stop a run, wipes the deck and with it the trail to every child execution. There is no way back to them from the UI, which makes eager unusable for any run you might need to cancel.

Steps to reproduce:

1. Run a task (or eager workflow) that publishes a deck and stays running.
2. Confirm the deck shows in the UI while the node is RUNNING.
3. Abort the execution.
4. The node goes ABORTED, the deck is gone, but `deck.html` is still in blob storage.

moldhouse:

Aborting a running node wipes its Flyte Deck from the UI, even though the deck was showing a moment earlier and `deck.html` still exists in blob storage. Abort takes a different path than the other terminal states: `nodeExecutor.Abort` builds its own `NodeExecutionEvent` with no deck URI, and flyteadmin then overwrites the stored URI with that empty value.

This is a hard blocker for eager workflows. Each eager child is its own execution, and the eager task's deck is the only place the UI surfaces the children and their links. Aborting the parent, the normal way to stop a run, wipes the deck and with it the trail to every child execution.

On abort, re-attach the deck URI to the ABORTED event when the deck file exists, using the same `GetDeckPath()` the running task writes to. This mirrors the recovery path, which already checks the deck file and re-attaches it. It only sets the URI when the file is present, so it never resurrects a deck for a disabled or never-published task, and a `Head` error is tolerated so abort always proceeds.

## How was this patch tested?

Added a unit test that drives `Abort` and asserts the emitted event carries the deck URI; confirmed it fails without the change and passes with it. The full `flytepropeller/pkg/controller/nodes` suite passes.

Fixes #7610

### DIFF
--- a/flytepropeller/pkg/controller/nodes/executor.go
+++ b/flytepropeller/pkg/controller/nodes/executor.go
@@ -933,6 +933,14 @@ func (c *nodeExecutor) Abort(ctx context.Context, h interfaces.NodeHandler, nCtx
 		}
 
 		targetEntity := common.GetTargetEntity(ctx, nCtx)
+
+		// This event replaces the running one, so re-attach the deck URI or it is lost.
+		deckFile := ioutils.NewRemoteFileOutputPaths(ctx, nCtx.DataStore(), nCtx.NodeStatus().GetOutputDir(), nil).GetDeckPath()
+		var deckURI string
+		if md, err := nCtx.DataStore().Head(ctx, deckFile); err == nil && md.Exists() {
+			deckURI = deckFile.String()
+		}
+
 		err := nCtx.EventsRecorder().RecordNodeEvent(ctx, &event.NodeExecutionEvent{
 			Id:         nodeExecutionID,
 			Phase:      core.NodeExecution_ABORTED,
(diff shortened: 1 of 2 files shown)


Sovietaced:
Unfortunately Flyte V1 CI seems broken. I'm not too familiar with what is going on but I'll take a look

Sovietaced:
Instead of making a datastore call on every abort node we could alternatively update the flyte admin RPC layer handle the abort phase in a special way so it doesn't blow out the deck uri.

moldhouse:
In that case, I'd suggest closing this PR. We saw it as a quick side-win for eager workflows on our way to Flyte2, but have since also found other issues around eager that prevent us from using it.

PR #7616, opened 2026-07-02
fix(webhook): honor disableCreateMutatingWebhookConfig, add namespaceSelector

### LINKED ISSUE #7614: [BUG] Executor pod webhook self-registers cluster-wide; disableCreateMutatingWebhookConfig is never read
**Problem**

The v2 executor always registers its pod webhook cluster-wide at startup (observed on v2.0.26 and master):

- `executor/pkg/webhook/entrypoint.go`: `Setup` calls `createMutationConfig` unconditionally.
- `webhook.disableCreateMutatingWebhookConfig` exists in the config struct (`flyteplugins/go/tasks/pluginmachinery/secret/config/config.go`) but is read nowhere.
- The generated `MutatingWebhookConfiguration` (`executor/pkg/webhook/pod.go`) has no `namespaceSelector`, matches `inject-flyte-secrets: "true"` in all namespaces, and uses `failurePolicy: Fail`.

**Impact**

Running Flyte 2 alongside a Flyte 1 deployment in the same cluster (a recommended migration setup) breaks Flyte 1: v1 task pods carry the same `inject-flyte-secrets` label, so the v2 webhook intercepts them, cannot resolve v1 secrets, and rejects the pods (`SecretsPodMutator.Mutate` returns an admission error when no injector succeeds). `failurePolicy: Fail` additionally couples v1 pod creation to v2 webhook uptime.

No workaround exists: removing the RBAC permission makes `Setup` fail fatally, and manual edits to the webhook config are lost because the binary replaces `.webhooks` on every startup.

**Proposed fix**

Honor `disableCreateMutatingWebhookConfig` in `Setup`, so operators can bring their own namespace-scoped `MutatingWebhookConfiguration`. Happy to send a PR.

Alternatively or additionally, a config option for a `namespaceSelector` on the generated webhook would avoid the hand-managed config object entirely.

Note: v1 has the same unscoped self-registration and no flag at all, so the reverse direction (v1 intercepting v2 secret-using pods) would need the same treatment in a v1 maintenance release.

moldhouse:
## Tracking issue
Closes #7614

Running Flyte v2 alongside a Flyte v1 deployment in the same cluster breaks v1: the executor always registers a MutatingWebhookConfiguration that matches the `inject-flyte-secrets` label in all namespaces with `failurePolicy: Fail`, so it intercepts and can reject v1 task pods. `webhook.disableCreateMutatingWebhookConfig` exists in the config but was never read.

Two independent commits:
1. Honor `disableCreateMutatingWebhookConfig` in webhook `Setup`, so operators can manage a scoped MutatingWebhookConfiguration themselves.
2. Add optional `webhook.namespaceSelector` applied to the generated MutatingWebhookConfiguration. Nil keeps the current behavior of matching all namespaces.

## How was this patch tested?
Added `executor/pkg/webhook/pod_test.go` covering selector propagation and the nil default; `go test ./executor/pkg/webhook/ ./flyteplugins/go/tasks/pluginmachinery/secret/config/`.

### Labels
- **fixed**: `webhook.disableCreateMutatingWebhookConfig` is now honored
- **added**: `webhook.namespaceSelector` to scope the generated MutatingWebhookConfiguration

### DIFF
--- a/executor/pkg/webhook/entrypoint.go
+++ b/executor/pkg/webhook/entrypoint.go
@@ -35,7 +35,9 @@ func Setup(ctx context.Context, kubeClient kubernetes.Interface, cfg *webhookCon
 		return fmt.Errorf("webhook: failed to create pod mutator: %w", err)
 	}
 
-	if err := createMutationConfig(ctx, kubeClient, podMutator, defaultNamespace); err != nil {
+	if cfg.DisableCreateMutatingWebhookConfig {
+		logger.Infof(ctx, "Skipping MutatingWebhookConfiguration creation, disabled by config")
+	} else if err := createMutationConfig(ctx, kubeClient, podMutator, defaultNamespace); err != nil {
 		return fmt.Errorf("webhook: failed to create MutatingWebhookConfiguration: %w", err)
 	}
 
--- a/executor/pkg/webhook/pod.go
+++ b/executor/pkg/webhook/pod.go
@@ -127,6 +127,7 @@ func (pm PodMutator) CreateMutationWebhookConfiguration(namespace string) (*admi
 				FailurePolicy:           &fail,
 				SideEffects:             &sideEffects,
 				AdmissionReviewVersions: []string{"v1", "v1beta1"},
+				NamespaceSelector:       pm.cfg.NamespaceSelector,
 				ObjectSelector: &metav1.LabelSelector{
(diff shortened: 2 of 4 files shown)


Sovietaced (approved):

moldhouse:
@Sovietaced thanks for the review! Anything else needed to get this in?

PR #7619, opened 2026-07-02
fix(webhook): allow disabling or scoping the self-registered MutatingWebhookConfiguration

### LINKED ISSUE #7618: [BUG] propeller pod webhook self-registers cluster-wide; cannot be scoped or disabled
V1 counterpart of #7614 (the Flyte 2 side was fixed in #7616).

**Problem**

The propeller webhook always creates a MutatingWebhookConfiguration with no `namespaceSelector`, matching `inject-flyte-secrets: "true"` in all namespaces with `failurePolicy: Fail` (`flytepropeller/pkg/webhook/entrypoint.go`, `pod.go`). No config option disables or scopes it, and manual edits are lost because the binary replaces `.webhooks` on every startup.

**Impact**

This blocks the recommended migration setup of running Flyte 1 and Flyte 2 in the same cluster. #7616 stops the v2 webhook from intercepting v1 pods; this issue is the reverse direction: the v1 webhook intercepts v2's secret-requesting task pods (same label), cannot resolve their secrets, and rejects them. Side-by-side operation is only safe once both sides can be scoped.

**Proposed fix**

Mirror #7616 on `master`: add a `webhook.disableCreateMutatingWebhookConfig` option (same key as v2) and support `webhook.namespaceSelector` on the generated configuration. PR follows.

moldhouse:
## Tracking issue
Closes #7618

Related to #7614, v1 counterpart of #7616.

Running Flyte 1 and Flyte 2 in the same cluster is the recommended migration setup, but the propeller webhook always registers a MutatingWebhookConfiguration matching the `inject-flyte-secrets` label in all namespaces with `failurePolicy: Fail`, so it intercepts and rejects the other control plane's secret-requesting task pods. #7616 fixed the v2 side; this is the v1 side.

Two independent commits, mirroring #7616:
1. Add `webhook.disableCreateMutatingWebhookConfig` (same key as v2), skipping self-registration so operators can manage a scoped MutatingWebhookConfiguration out of band. Includes the regenerated pflags files.
2. Add optional `webhook.namespaceSelector` applied to the generated MutatingWebhookConfiguration. Nil keeps the current behavior of matching all namespaces.

## How was this patch tested?
Extended `Test_CreateMutationWebhookConfiguration` with selector propagation and nil-default cases; `go test ./pkg/webhook/...` in `flytepropeller`.

### Labels
- **added**: `webhook.disableCreateMutatingWebhookConfig` and `webhook.namespaceSelector` options to scope or disable the self-registered MutatingWebhookConfiguration

### DIFF
--- a/flytepropeller/pkg/webhook/config/config.go
+++ b/flytepropeller/pkg/webhook/config/config.go
@@ -5,6 +5,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/flyteorg/flyte/flytestdlib/config"
 )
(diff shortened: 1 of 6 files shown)


Sovietaced (approved):

moldhouse:
@Sovietaced thanks for getting this in. any chance of a v1.16.8 soon? we'd love to pick this fix up from a tagged release.

Sovietaced:
> @Sovietaced thanks for getting this in. any chance of a v1.16.8 soon? we'd love to pick this fix up from a tagged release.

Yeah. I just need to review some other pull requests.

PR #7621, opened 2026-07-03
fix multipart blob dimensionality for directory uploads

ApfelPresse:

Directory blob uploads currently return a blob literal with SINGLE dimensionality, even though uploading a directory represents a multipart blob.

This PR updates the directory upload path in handleBlobType to return a blob literal with MULTIPART dimensionality.

It also adds a regression test that verifies a directory-based blob upload produces output metadata with:

dimensionality: MULTIPART

## How was this patch tested?

Added a unit test for multipart blob uploads.

Tested with:

go test ./flytecopilot/data

Result:

PASS
ok      github.com/flyteorg/flyte/flytecopilot/data

In addition, we built a Flyte container image with this patch applied to our current Flyte version and tested it in our environment. The patched version correctly returns multipart blob outputs with MULTIPART dimensionality for directory uploads.

### Labels

fixed

### DIFF
--- a/flytecopilot/data/upload.go
+++ b/flytecopilot/data/upload.go
@@ -61,61 +61,65 @@ func (u Uploader) handleBlobType(ctx context.Context, localPath string, toPath s
 	if err != nil {
 		return nil, err
 	}
-	if info.IsDir() {
-		var files []dirFile
-		err := filepath.Walk(localPath, func(path string, info os.FileInfo, err error) error {
+	if !info.IsDir() {
+		size := info.Size()
+		// Should we make this a go routine as well, so that we can introduce timeouts
+		return coreutils.MakeLiteralForBlob(toPath, false, ""), UploadFileToStorage(ctx, fpath, toPath, size, u.store)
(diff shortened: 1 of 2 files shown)


flytecopilot/data/upload.go
```diff
@@ -107,7 +109,9 @@ func (u Uploader) handleBlobType(ctx context.Context, localPath string, toPath s
 			}
 		}
 
-		return coreutils.MakeLiteralForBlob(toPath, false, ""), nil
+		logger.Infof(ctx, "Successfully uploaded multipart blob from [%s] to [%s]", localPath, toPath)
```
Sovietaced: nit: it was actually hard to review this and see what was going on. Can you just add a condition up top with `!info.IsDir()` which short circuits to the single blob type call and then let the rest of the code be the more complex multi part?

Sovietaced:
Also please run `make lint-fix` inside the `flytecopilot` directory.

Sovietaced (approved):

PR #7625, opened 2026-07-05
fix: respect plugin GeneratedNameMaxLength in k8s PluginManager

### LINKED ISSUE #7607: [BUG] Flyte v2 bug: RayJob name exceeds KubeRay's 47-char limit for nested (sub-action) Ray tasks
I ran into an issue with ray task nested inside a regular task. Claude says it found the bug in the source code and "fixed" it, so wanted to surface since it could be real.

# Claude description

# Flyte v2 bug: RayJob name exceeds KubeRay's 47-char limit for nested (sub-action) Ray tasks

## TL;DR

When a Ray task runs as a **sub-action** (i.e. it is launched from a parent
workflow/driver task rather than being the root action), Flyte's backend names
the KubeRay `RayJob` object `{run}-{action}-{attempt}`. With the default
auto-generated run name (20 chars) and a sub-action name (a fixed 25-char hash),
the assembled name is **48 chars**, which KubeRay rejects:

```
The RayJob metadata is invalid flyte/rqfn7vfrdp9j84q86stm-a0d7xyznuli5is847rvhzlkly-0: RayJob name should be no more than 47 characters
```

This causes the task to hang forever.

## Environment

| | |
|---|---|
| Flyte SDK | `flyte` 2.5.4 |
| Ray plugin | `flyteplugins-ray` 2.5.4 |
| Cluster | sandbox-uc1 (org `sandbox-uc1-flyte`, project `flytesnacks`, domain `development`) |
| Ray operator | KubeRay (RayJob CRD) |
| Repro workflow | `src/orthlyml/mlops/flyte/examples/ray_example.py` |

# Code
```python
``python
@ray_env.task
def ray_square_task(n: int = 10) -> list[int]:
    """Pure Ray compute: square ``0..n-1`` on the Ray cluster and return the results."""
    futures = [_square.remote(i) for i in range(n)]
    return ray.get(futures)


@env.task
def sum_list(values: list[int]) -> int:
    """Sum ``values`` (plus 1, just to make the reduce non-trivial)."""
    return sum(values) + 1


@env.task
def ray_workflow(n: int = 10) -> int:
    """Driver/entrypoint: square ``0..n-1`` on Ray, then reduce once."""
    squares = ray_square_task(n=n)
    return sum_list(values=squares)
```
```

### PR DESCRIPTION
**Pull request by falloficaruss**
Fixes #7607

## Tracking issue
Fixes #7607

When a Ray task runs as a sub-action (i.e. it is launched from a parent workflow or driver task rather than being the root action), the generated resource name follows the format  {run}-{action}-{attempt}  (e.g.,  rqfn7vfrdp9j84q86stm-a0d7xyznuli5is847rvhzlkly-0 ), which is typically 48 characters.

However, `KubeRay` has a strict metadata validation constraint restricting  `RayJob` names to a maximum of 47 characters. When this limit is exceeded, ` KubeRay` rejects the metadata creation request, causing the workflow task to hang indefinitely.

To prevent this, the Ray task plugin declares a ` GeneratedNameMaxLength`  of  47  in its properties. But the unified Kubernetes plugin manager ( `PluginManager` ) in the `Flyte v2` executor was assigning the object name using  `taskCtx.GetTaskExecutionID().GetGeneratedName()`  directly without checking or applying  `GeneratedNameMaxLength`.

This PR modifies the plugin manager to check  `GeneratedNameMaxLength`  and call `GetGeneratedNameWith`  instead, ensuring that generated resource names are properly bounded and truncated within the limits of the target plugin.

This PR updates the Kubernetes plugin metadata builder to respect the name length limits specified by plugins:

- Respect  `GeneratedNameMaxLength` : In `plugin_manager.go`, updated the  `addObjectMetadata`  function to inspect  `pm.plugin.GetProperties().GeneratedNameMaxLength`.
- Bounded Name Generation: If a maximum length limit is specified by the plugin properties, the manager now invokes  `GetGeneratedNameWith(0, maxLength)`
  instead of  GetGeneratedName()  to securely bound and truncate the resource name.
- Compatibility Safety: Bounding the name prior to running `validation.IsDNS1123Subdomain`  ensures that trailing hyphens or dot separators resulting from truncation are automatically cleaned up.

## How was this patch tested?
1. Unit Tests:
Ran the test suites for the executor plugin package and the Ray task plugin package to ensure that there are no regressions:
    **Run executor plugin tests**
    `go test -v ./pkg/plugin/...`

    **Run Ray plugin tests**
    `go test -v ./go/tasks/plugins/k8s/ray/...`

2. Compilation:
Verified that both the manager and executor entry points compile successfully:
    `go build ./executor/cmd/main.go`
    `go build ./manager/cmd/main.go`

### Labels
- **fixed**

## Stack

### DIFF
--- a/executor/pkg/plugin/k8s/plugin_manager.go
+++ b/executor/pkg/plugin/k8s/plugin_manager.go
@@ -80,7 +80,13 @@ func (pm *PluginManager) addObjectMetadata(taskCtx pluginsCore.TaskExecutionMeta
 	o.SetNamespace(taskCtx.GetNamespace())
 	o.SetAnnotations(pluginsUtils.UnionMaps(cfg.DefaultAnnotations, o.GetAnnotations(), pluginsUtils.CopyMap(taskCtx.GetAnnotations())))
 	o.SetLabels(pluginsUtils.UnionMaps(cfg.DefaultLabels, o.GetLabels(), pluginsUtils.CopyMap(taskCtx.GetLabels())))
-	o.SetName(taskCtx.GetTaskExecutionID().GetGeneratedName())
+	name := taskCtx.GetTaskExecutionID().GetGeneratedName()
+	if pm.plugin.GetProperties().GeneratedNameMaxLength != nil {
+		if truncatedName, err := taskCtx.GetTaskExecutionID().GetGeneratedNameWith(0, *pm.plugin.GetProperties().GeneratedNameMaxLength); err == nil {
+			name = truncatedName
+		}
+	}
+	o.SetName(name)
 
 	if !pm.plugin.GetProperties().DisableInjectOwnerReferences && !cfg.DisableInjectOwnerReferences {
 		o.SetOwnerReferences([]metav1.OwnerReference{taskCtx.GetOwnerReference()})
--- a/executor/pkg/plugin/task_exec_metadata.go
+++ b/executor/pkg/plugin/task_exec_metadata.go
@@ -12,6 +12,7 @@ import (
 	flyteorgv1 "github.com/flyteorg/flyte/v2/executor/api/v1"
 	executorconfig "github.com/flyteorg/flyte/v2/executor/pkg/config"
 	pluginsCore "github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/core"
+	"github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/encoding"
 	"github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/flytek8s"
 	flytesecret "github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/secret"
 	pluginsUtils "github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/utils"
@@ -32,11 +33,7 @@ func (t *taskExecutionID) GetGeneratedName() string {
 }
 
 func (t *taskExecutionID) GetGeneratedNameWith(minLength, maxLength int) (string, error) {
-	name := t.generatedName
-	if len(name) > maxLength {
-		name = name[:maxLength]
-	}
-	return name, nil
+	return encoding.FixedLengthUniqueID(t.generatedName, maxLength)
 }
 
 func (t *taskExecutionID) GetID() *core.TaskExecutionIdentifier {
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: executor/pkg/plugin/k8s/plugin_manager.go [hunk 1]
```diff
@@ -80,7 +80,13 @@ func (pm *PluginManager) addObjectMetadata(taskCtx pluginsCore.TaskExecutionMeta
 	o.SetNamespace(taskCtx.GetNamespace())
 	o.SetAnnotations(pluginsUtils.UnionMaps(cfg.DefaultAnnotations, o.GetAnnotations(), pluginsUtils.CopyMap(taskCtx.GetAnnotations())))
 	o.SetLabels(pluginsUtils.UnionMaps(cfg.DefaultLabels, o.GetLabels(), pluginsUtils.CopyMap(taskCtx.GetLabels())))
-	o.SetName(taskCtx.GetTaskExecutionID().GetGeneratedName())
+	name := taskCtx.GetTaskExecutionID().GetGeneratedName()
+	if pm.plugin.GetProperties().GeneratedNameMaxLength != nil {
+		if truncatedName, err := taskCtx.GetTaskExecutionID().GetGeneratedNameWith(0, *pm.plugin.GetProperties().GeneratedNameMaxLength); err == nil {
```
machichima, line 85: Great choice! However I'd prefer refactoring `GetGeneratedNameWith` to use `encoding.FixedLengthUniqueID`, so that the uniqueness is preserved

https://github.com/flyteorg/flyte/blob/d7fafbeee67d2779913407e6f86fc680addea85c/flyteplugins/go/tasks/pluginmachinery/encoding/encoder.go?plain=1#L48

cc @pingsutw to confirm

pingsutw: Yes, using FixedLengthUniqueID is better

falloficaruss: I replaced `GetGeneratedNameWith` with `encoding.FixedLengthUniqueID`

executor/pkg/plugin/task_exec_metadata.go
```diff
@@ -32,11 +33,7 @@ func (t *taskExecutionID) GetGeneratedName() string {
 }
 
 func (t *taskExecutionID) GetGeneratedNameWith(minLength, maxLength int) (string, error) {
```
popojk, line 35: Great job. Seems like this function is not covered by unit test yet, would you mind adding a test to make sure it functions correctly?

popojk (approved):

PR #7626, opened 2026-07-06
oss hitl

machichima:

Support native condition (HITL; human in the loop) in Flyte

A condition is a TaskAction CR with no plugin and no pod. It will transit from `Paused` to `Signaled` (Succeeded) on signal, or `TimedOut` (Failed) past the deadline.

- Proto changes: add `output`/`principal`/`abort_reason` to `UpdateActionStatusRequest` and `condition`/`output`/`principal` to `RunInfo`

- New condition reasone: `Paused`, `Signaled`, and `TimedOut`.

- Add `actionType`/`conditionSpec` (spec) and `signalValue/signalledBy/signalledAt` (status) to the TaskAction CRD; new `reconcileCondition` drives the reconcile process.
- Implement `ActionsService.Signal`: validates and writes the value onto the CR status; `Enqueue` now creates condition CRs; `WatchForUpdates` carries the resolved value inline so the awaiting parent SDK resumes.
- Implement `RunService.SignalEvent` to record conditions and persist the resolved value/actor into the runs DB

## How was this patch tested?

Test with devbox and run flyte-sdk example:

1. Start devbox (`make devbox-build` -> `make devbox-run`)
2. Run example with `python examples/apps/hitl_condition.py`

```py
import flyte

env = flyte.TaskEnvironment(
    name="hitl-condition",
    resources=flyte.Resources(cpu=1, memory="512Mi"),
)


@env.task
async def task1() -> int:
    return 42


@env.task
async def main() -> int:
    x = await task1()
    print(f"task1 completed: x = {x}")

    condition = await flyte.new_condition.aio(
        "integer_input",
        prompt="What should I add to x?",
        data_type=int,
        timeout=600,
    )
    y = await condition.wait.aio()
    print(f"condition signaled: y = {y}")

    return x + y


if __name__ == "__main__":
    flyte.init_from_config()
    run = flyte.with_runcontext(disable_run_cache=True).run(main)
    print(f"Run name: {run.name}")
    print("Signal with: flyte signal condition", run.name, "<action-name> 8")
    run.wait()
    print(f"Result: {run.outputs()}")
```

3. `flyte get action <run-name>` to get the paused action name, set input for the condition action through `flyte signal condition <run-name> <action-name> 8`

image

NOTE: OSS UI does not support HITL now (cannot render phase "Paused" and cannot type input in UI directly)

### Labels

## Stack


actions/k8s/client.go
```diff
@@ -43,6 +45,9 @@ type ActionUpdate struct {
 	TaskType         string
 	ShortName        string
 	ErrorState       *executorv1.ErrorState
+	// Value is the resolved signal payload of a condition action, carried
+	// inline (never an outputs.pb). Nil for tasks and unsignalled conditions.
+	Value *core.Literal
```
pingsutw: Could we rename it to Literal or SignalValue or something else?

EngHabu (dismissed):
Reviewed IDL

runs/service/internal_run_service.go
```diff
@@ -252,9 +263,41 @@ func (s *RunService) updateSingleActionStatus(ctx context.Context, req *workflow
...
+		return err
+	}
+	info := &workflow.RunInfo{}
+	if len(model.DetailedInfo) > 0 {
+		if err := proto.Unmarshal(model.DetailedInfo, info); err != nil {
+			return err
+		}
+	}
+	info.Output = req.GetOutput()
+	info.Principal = req.GetPrincipal()
```
pingsutw: should we do the nil check here?

pingsutw: so it won't override the Principal value if req is nil

pingsutw (commented):
Some nits, overall LGTM, thank you!!

actions/k8s/client.go
machichima: Sure! Changed to SignalValue

runs/service/internal_run_service.go
machichima: Sure! Added the nil check

pingsutw (approved):

PR #7639, opened 2026-07-08
update proto

popojk:
## Tracking issue

Add a field in run details API to let the UI know if the run task is deployed

protobuf change

### Labels

## Stack


flyteidl2/workflow/run_definition.proto
```diff
@@ -39,6 +39,10 @@ message RunDetails {
 
   // Detailed information about the root action.
   ActionDetails action = 2;
+
+  // Computed at read time: whether the run's task version currently exists in the
+  // task store, so the rerun UI can send a TaskId instead of the full inline spec.
+  bool task_is_registered = 3;
```
EngHabu: "deployed" is the verb I would use not "registered"
```suggestion

  // Computed at read time: whether the run's task version currently exists in the
  // task store, so the rerun UI can send a TaskId instead of the full inline spec.
  bool task_is_deployed = 3;
```

I would recommend we actually make it an enum `TaskSpecSource task_spec_source = 3;` with values UNSPECIFIED, DEPLOYED, EPHEMERAL. or keep it as boolean you have...

EngHabu:
Can we store that enum as part of the RunSpec.. And let the BE populate it based on whether the run was created from a TaskIdentifier or TaskSpec..

EngHabu (approved):

PR #7640, opened 2026-07-08
fix(executor): batch TaskAction event writes to unblock reconciles at scale

pingsutw:

Under high held-action concurrency (tens of thousands of live `TaskActions`), the executor's reconcile throughput collapsed and runs wedged under overload. Root cause: every reconcile that changed status blocked **synchronously** on `eventsClient.Record`, which became a **single-row `INSERT`** per event. At scale this pegged all reconcile workers (~2s each during the completion burst), so newly-created actions were starved of a worker and their pods were created 30+ minutes late.

Notably it is **not** commit-amplification, CPU throttling, or DB write saturation (all ruled out by measurement): the DB insert itself is <100ms. The cost was **flush/worker starvation on the synchronous per-event path**.

Two changes that make a reconcile cheap at scale (both keep semantics unchanged):

1. **Batched event writes** (`event_batcher.go`, wired in `taskaction_controller.go`): a small batcher coalesces concurrent reconciles' `ActionEvent`s into a few multi-event `Record` RPCs. Each caller **still blocks until its batch commits**, so delivery is unchanged (**at-least-once**): on error the reconcile requeues and re-emits, and re-emission is idempotent via the existing `ON CONFLICT DO NOTHING`. Flush concurrency is bounded independently of reconcile fan-out.
2. **Multi-row `InsertEvents`** (`runs/repository/impl/action.go`): a batch becomes one multi-row `INSERT` (one transaction/commit) instead of one commit per event.

Plus a small related optimization:
3. **Skip the cache-hit `Catalog.Get` for already-RUNNING actions** (`taskaction_cache.go`): a cache hit is moot mid-execution, so this avoids a redundant blocking cache-service RPC on every reconcile of a held action. The serializable-reservation heartbeat is preserved.

## How was this patch tested?

- Unit tests added: `event_batcher_test.go` (coalescing delivers every event exactly once, batch cap respected, errors propagate, context-cancel unblocks — run with `-race`); `taskaction_cache_test.go` (running actions skip `Get`). Existing `InsertEvents` tests pass against real Postgres (multi-row path).
- End-to-end on a 40k-held benchmark (dev cluster): per-reconcile event-write latency dropped from **~1980ms (p50) to <300ms** with **zero** slow reconciles through the completion burst; server-side inserts stayed **<100ms**. Runs stopped wedging on the control plane.

### Labels
fixed

### DIFF
--- a/executor/pkg/controller/event_batcher.go
+++ b/executor/pkg/controller/event_batcher.go
@@ -1,0 +1,111 @@
+package controller
+
+import (
+	"context"
+	"time"
+
+	"connectrpc.com/connect"
+
+	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/workflow"
+	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/workflow/workflowconnect"
+)
+
+// Event batching bounds. Coalescing concurrent reconciles' action events into a
+// few multi-event Record RPCs (each becoming one multi-row INSERT server-side)
+// amortizes the per-event DB commit cost that otherwise dominates reconcile
+// latency at high held-action counts. Callers still block until their batch
+// commits, so delivery semantics are unchanged versus a direct Record.
+const (
+	eventBatchMaxSize  = 400
+	eventBatchMaxDelay = 25 * time.Millisecond
+	eventFlushWorkers  = 32
+	eventQueueDepth    = 8192
+	eventFlushTimeout  = 30 * time.Second
+)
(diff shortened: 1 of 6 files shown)


pingsutw:
Superseded by a 3-way split for easier review:
- #7657 — bump watch workers & buffer size
- #7658 — batch TaskAction event writes
- #7659 — coalesce watch events + record true action start

Closing in favor of these.

PR #7653, opened 2026-07-10
Add common.Relation (related_to + relation_type) to RunSpec and Actio…

afbrock:
…nMetadata

- New common.Relation message and RelationType enum (RERUN, RECOVER) in flyteidl2/common/run.proto.
- RunSpec: deprecate related_to (15), add relation (18); server normalizes legacy related_to into relation with RELATION_TYPE_RERUN.
- ActionMetadata: add relation (17) so provenance is returned by ListRuns and ListActions, which build responses from columns rather than the run_spec blob.

Signed-off-by: Adam Brock <adam@union.ai>

## Tracking issue

https://linear.app/unionai/issue/ENG26-870/be-make-relation-available-in-list-runs

These changes are needed to be able to re-use the relation field for recover, and also make the data available for list runs.

A relation that is specified in the runspec in create run will now be specified with a type as well.

## How was this patch tested?
unit tests, devbox, staging

### Labels

## Stack


flyteidl2/workflow/run_definition.proto
```diff
@@ -195,6 +196,10 @@ message ActionMetadata {
 
   // Who initiated this run
   RunSource source = 16;
+
+  // Provenance link to the parent run this run was derived from (re-run or
```
katrogan: since this is in ActionMetadata, do we need to update the comment?

katrogan: Also in your PR comment you mention you add this to the ActionMetadata so we can build provenance responses from columns. but the db columnset doesn't have to map to the protobuf definition necessarily, so could we omit this addition altogether and just store the relation purely in a column or do we want this to be part of the action data itself too?

flyteidl2/common/run.proto
```diff
@@ -12,3 +13,24 @@ message OffloadedInputData {
...
+}
+
+// Relation is a provenance link to the single parent run this run was derived from.
+// Set at creation, immutable thereafter.
+message Relation {
+  // The parent run. Scoped to the same org/project/domain as the referencing run.
+  RunIdentifier related_to = 1 [(buf.validate.field).required = true];
+
+  // How this run was derived from the parent.
+  RelationType relation_type = 2 [(buf.validate.field).required = true];
```
katrogan: should we add validation that this isn't ever of type UNSPECIFIED too?

flyteidl2/workflow/run_definition.proto
afbrock: How do you mean updating the comment?  is it that it is on an action and not a run?  
In terms of the addition in runspec, we want the relation to be present in runspec for the create run call, but we want to return the data in action metadata in a list call so that we dont make postgres deserialize the runspec for all the records.

katrogan: oh I just mean "Provenance link to the parent run this run was derived from" -> "Provenance link to the parent run this _action_ was derived from"

katrogan (dismissed):

katrogan (dismissed):

katrogan (approved):

PR #7666, opened 2026-07-14
fix(runs): honor Offset in ListActions so WatchActions returns all actions

pingsutw:
## Problem

`WatchActions` only ever returns ~100 actions for a run, even when the run has many more (e.g. 1,000). Observed on a run with 1,000 actions — the client/console only sees the first ~100.

## Root cause

`ListResourceInput` supports two mutually-exclusive pagination modes: keyset (`CursorToken`) and `Offset`. `actionRepo.ListActions` implements **only** the cursor path — it builds `... LIMIT ?` and never emits an `OFFSET`, so the `Offset` field is silently ignored.

`WatchActions`' initial full listing (`listAndSendAllActions`) paginates by **`Offset`** (`offset += pageSize`). Because `ListActions` ignored it, every page re-read the same first `~Limit` rows, so only the first page was ever delivered.

## Fix

Emit `OFFSET ?` when `input.Offset > 0`. Cursor-based callers (the `ListActions` RPC handlers) pass `Offset == 0` and are unchanged; only the offset-based `WatchActions` initial listing is affected.

```go
queryBuilder.WriteString(" LIMIT ?")
args = append(args, input.Limit+1)

if input.Offset > 0 {
    queryBuilder.WriteString(" OFFSET ?")
    args = append(args, input.Offset)
}
```

## Test

`TestListActions_OffsetPagination` — creates 5 root actions and asserts `offset=N` returns `total-N` rows (order-independent counts). These assertions fail on the old behavior (offset ignored → every page returned all 5). Existing `TestListRuns` (cursor pagination) still passes, confirming the change is additive.

```
ok  github.com/flyteorg/flyte/v2/runs/repository/impl  12.176s
```

https://claude.ai/code/session_01AjhEAjx4yGQY9MhpY5DEXd

### DIFF
--- a/runs/repository/impl/action.go
+++ b/runs/repository/impl/action.go
@@ -346,6 +346,15 @@ func (r *actionRepo) ListActions(ctx context.Context, input interfaces.ListResou
 	queryBuilder.WriteString(" LIMIT ?")
 	args = append(args, input.Limit+1)
 
+	// Offset-based pagination (mutually exclusive with CursorToken, per
+	// ListResourceInput). Without emitting OFFSET, callers that page by Offset
+	// (e.g. WatchActions' initial full listing) silently re-read the first page
+	// on every iteration, capping the result at a single page (~Limit rows).
+	if input.Offset > 0 {
+		queryBuilder.WriteString(" OFFSET ?")
+		args = append(args, input.Offset)
+	}
+
 	query := sqlx.Rebind(sqlx.DOLLAR, queryBuilder.String())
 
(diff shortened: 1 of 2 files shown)


pingsutw:
Closing as a duplicate of #7585, which is the same core fix (add OFFSET to ListActions) and also correctly adds the `name` tiebreaker to the WatchActions snapshot sort — needed so OFFSET paging over map-task children with identical created_at doesn't skip/duplicate rows. Landing #7585 instead.

PR #7671, opened 2026-07-15
Treat K8s Invalid/BadRequest on resource create as permanent user failure

SVilgelm:
## Tracking issue

Related to [RUN-53](https://linear.app/unionai/issue/RUN-53/run-with-invalid-pod-spec-shows-queued-but-error-and-burns-30-system)

A run submitted with a deterministically invalid pod spec (e.g. an empty container image, as in `playground/flytesnacks/development/run-labels-smoke-6a40e9`) is rejected by K8s with 422 on every create attempt:

```
Pod "..." is invalid: spec.containers[0].image: Required value
```

`launchResource` only special-cases `IsForbidden` and `IsRequestEntityTooLargeError`; `IsInvalid` falls through to the generic system-error path. Leasor then system-retries up to `max_system_retries` (30) at ~1/min, and each retry flips the action back to Queued while the previous attempt's error is still on the record — the run shows "Queued but error" in the UI for ~35 minutes before permanently failing, and inflates active-run counts in queue metrics.

Since the same spec is rejected on every attempt, retrying can never succeed.

In `executor/pkg/plugin/k8s/plugin_manager.go` `launchResource`, handle `k8serrors.IsInvalid` and `k8serrors.IsBadRequest` on resource create the same way as `IsRequestEntityTooLargeError`: a permanent user failure (`PhaseInfoFailure`, code `InvalidResource`) that fails the action immediately instead of burning system retries.

This mirrors upstream flyte1's `transitionFromError` `IsBadRequest || IsInvalid` branch (where the permanent-failure transition is commented out), and acts as the catch-all for all K8s plugins and dynamically spawned child actions. Front-door validation of the task spec at CreateRun is tracked separately in RUN-53.

## How was this patch tested?

Added `plugin_manager_test.go` covering `launchResource` create-error handling: Invalid and BadRequest → permanent USER failure with code `InvalidResource`; Forbidden → retryable failure (unchanged); other errors (e.g. internal) → system error (unchanged); successful create → Queued.

### DIFF
--- a/executor/pkg/plugin/k8s/plugin_manager.go
+++ b/executor/pkg/plugin/k8s/plugin_manager.go
@@ -113,6 +113,12 @@ func (pm *PluginManager) launchResource(ctx context.Context, tCtx pluginsCore.Ta
 		if k8serrors.IsRequestEntityTooLargeError(err) {
 			return pluginsCore.DoTransition(pluginsCore.PhaseInfoFailure("EntityTooLarge", err.Error(), nil)), nil
 		}
+		// K8s rejected the resource as invalid (e.g. HTTP 422 for a pod spec with an empty
+		// container image). The same spec will be rejected on every attempt, so retrying as a
+		// system failure only burns retries while the task appears Queued with an error attached.
+		if k8serrors.IsInvalid(err) || k8serrors.IsBadRequest(err) {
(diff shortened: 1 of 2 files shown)


wild-endeavor (commented):
is this being used at all?

SVilgelm:
I'm not sure, it is black box to me, but claude suggested this change, so let me ask it :))))

SVilgelm:
okay, it is not in use in union, but it is OSS version and needs to be fixed as well I think

PR #7672, opened 2026-07-16
[Feat] add sentry for create run

machichima:

- Track errors when sentry for `CreateRun` RPC

#### Error metrics

1. Build and run devbox `make devbox-build devbox-run`
2. To trigger an error, alter the action table to another name

```sh
docker run --rm postgres:16-alpine psql "postgresql://postgres:postgres@host.docker.internal:30001/runs" \
    -c "ALTER TABLE actions RENAME TO actions_sentry_test_bak;"
```

3. Run an example: `python examples/basics/hello_v2.py`
4. We can see the error is recorded

image

5. After testing, alter the action table back

```sh
docker run --rm postgres:16-alpine psql "postgresql://postgres:postgres@host.docker.internal:30001/runs" \
    -c "ALTER TABLE actions_sentry_test_bak RENAME TO actions;"
```

#### Succeed counter

1. Build and run devbox `make devbox-build devbox-run`
2. Run an example: `python examples/basics/hello_v2.py`
3. We can see one success counter

image

### Labels

## Stack

### DIFF
--- a/runs/setup.go
+++ b/runs/setup.go
@@ -12,6 +12,7 @@ import (
 	"connectrpc.com/connect"
 	"connectrpc.com/otelconnect"
 	"github.com/getsentry/sentry-go"
+	"github.com/getsentry/sentry-go/attribute"
 	"github.com/flyteorg/flyte/v2/flytestdlib/app"
 	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/actions/actionsconnect"
 	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/auth/authconnect"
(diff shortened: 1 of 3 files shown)


runs/setup.go
```diff
@@ -32,6 +35,34 @@ import (
...
+
+// sentryDisabled honors FLYTE_DISABLE_SENTRY, unset or unparsable means enabled.
+func sentryDisabled() bool {
+	disabled, _ := strconv.ParseBool(os.Getenv("FLYTE_DISABLE_SENTRY"))
+	return disabled
+}
+
+// sentryInterceptor reports server-side CreateRun failures to Sentry. Client-caused
+// errors (invalid argument, not found, ...) are intentionally not reported.
+func sentryInterceptor() connect.UnaryInterceptorFunc {
```
pingsutw, line 52: Will we also get the metrics if there is no error?

machichima: Sure, let me add it

machichima: Added in https://github.com/flyteorg/flyte/pull/7672/commits/61e51827a1947d99cd47f9152bc7743aa1e87866

runs/setup.go
```diff
@@ -53,6 +97,26 @@ func Setup(ctx context.Context, sc *app.SetupContext) error {
...
 	}
 
+	// Sentry is on by default with the hardcoded DSN; FLYTE_DISABLE_SENTRY=true opts out.
+	runsInterceptors := []connect.Interceptor{otelInterceptor}
+	if !sentryDisabled() {
+		if err := sentry.Init(sentry.ClientOptions{
+			Dsn:         sentryDSN,
+			Environment: otelServiceName,
+		}); err != nil {
+			return fmt.Errorf("initializing sentry: %w", err)
```
pingsutw: nit: could we just log the error here? We don’t want the cluster startup to fail if Sentry can’t be initialized for some reason.

runs/setup.go
```diff
@@ -32,6 +36,46 @@ import (
...
+					switch connect.CodeOf(err) {
+					case connect.CodeInternal, connect.CodeUnknown, connect.CodeDataLoss:
+						hub := sentry.CurrentHub().Clone()
+						hub.Scope().SetTag("procedure", req.Spec().Procedure)
+						hub.CaptureException(err)
+					}
+				} else {
+					attrs = append(attrs, attribute.String("status", "success"))
+				}
+				sentry.NewMeter(ctx).Count("flyte.operation", 1, sentry.WithAttributes(attrs...))
```
pingsutw, line 72: If the run is created by trigger. will we get the metrics as well?

machichima: It should. Let's me test it!

machichima: Yes it does

Ensure that it's from my local flyte-binary pod

```sh
❯ k get pods -n flyte | grep flyte-binary
flyte-binary-68d484c4cb-rcd7d          1/1     Running     0          31m
flyte-binary-console-6d6cdb9c5-ggdk9   1/1     Running     0          31m
```

image

image

runs/setup.go
machichima: Yes! Updated in https://github.com/flyteorg/flyte/pull/7672/commits/f8e195b68c9a765ab526271b23bd646555478639

runs/setup.go
pingsutw: nice, thank you

pingsutw (approved):

PR #7683, opened 2026-07-20
fix: prevent composed clustered tasks from getting stuck in RUNNING due to long JobSet pod names

AdilFayyaz:
Composed/nested clustered tasks generate long action names. The JobSet plugin derived its resource name directly from that generated name, so JobSet's admission webhook rejected the child pods (`<jobSetName>-workers-0-<podIdx>-<suffix>` exceeded the 63-char DNS-1035 label limit). The plugin retried the identical, always-invalid object forever, leaving the execution stuck in RUNNING.

  - clustered/util.go: Add buildJobSetName, which sanitizes and truncates the generated name (appending an FNV hash for uniqueness) so the longest pod name JobSet derives stays within 63 chars. The bound reserves room for the worst-case pod index so the name is derivable from the generated name alone.
  - clustered/build.go + plugin.go: Use buildJobSetName on both the create (BuildResource) and lookup/abort (BuildIdentityResource) paths so they always agree on the name.
  - plugin_manager.go – addObjectMetadata: Only default the object name to the generated name when the plugin didn't set one, instead of unconditionally clobbering it (which would undo the bounded JobSet name).
  - plugin_manager.go – launchResource: Fast-fail on IsInvalid admission errors (InvalidResource) instead of looping via UnknownTransition — a generic safety net so any deterministic validation rejection surfaces as a failure rather than hanging in RUNNING.

## How was this patch tested?

  - TestBuildJobSetName_LongNameFitsPodLimit: a ~140-char name keeps the longest pod name ≤63 and DNS-1035-valid across replica counts 1…100000 (validated with jobset's own placement.GenPodName +
  validation.IsDNS1035Label).
   - TestBuildResourceAndIdentityNameMatch: create and lookup paths produce the identical (truncated) name for a long generated name.
  - Plus sanitization, uniqueness, and trailing-separator cases.
  - go test ./executor/pkg/plugin/k8s/...
  - Covers the addObjectMetadata name-preservation and launchResource IsInvalid fast-fail behavior.
  - To verify end-to-end: run a composed clustered task whose nesting produces a long action name; confirm the JobSet is admitted, pods schedule, and the execution progresses past RUNNING (previously it hung).

## Stack


executor/pkg/plugin/task_exec_metadata.go
```diff
@@ -33,7 +33,13 @@ func (t *taskExecutionID) GetGeneratedName() string {
 }
 
 func (t *taskExecutionID) GetGeneratedNameWith(minLength, maxLength int) (string, error) {
-	return encoding.FixedLengthUniqueID(t.generatedName, maxLength)
+	// Plugins that set GeneratedNameMaxLength (ray, clustered) do so because k8s derives
+	// Services and child pod names from the object name, which must be DNS-1035 labels
+	// (leading letter, no dots). FixedLengthUniqueID passes names that already fit through
+	// verbatim — without sanitizing first, a generated name with a dot or leading digit
```
pingsutw: Do we really need ConvertToDNS1035LabelCompatibleString? dot and leading digit  aren't not allow in the run name now in cloud. I think we need to add the same check in oss run service as well

pingsutw: we can add that in a separate PR

AdilFayyaz: OSS doesnt have any validation for run name rn 
Makes sense to add it but that change should land before this PR gets merged. I'll create a new one for this.

pingsutw (approved):

PR #7686, opened 2026-07-21
settings: remove dead fields; add RELATION_TYPE_SPAWN

afbrock:
Settings proto cleanup, now that the fields have shipped unused:

- StorageSettings.run_base_dir (field 2) is removed and reserved. Nothing reads it; the live field is RunSettings.run_base_dir.
- StringListSetting is removed. No settings field uses it and no merge implements it.
- RunSettings.run_base_dir's description is corrected — it described code bundling, which was wrong; it is the run-metadata base dir. That text was the one previously (correctly) on the now-removed storage field.

Also adds RelationType.RELATION_TYPE_SPAWN (3): a run spawned as a child of the parent run it relates to.

Only the Go generated code is regenerated here (unionai/cloud consumes it). Run `make gen` with the pinned Docker toolchain to regenerate ts/python/rust before merging, to keep all languages consistent.

## How was this patch tested?

run in devbox

## Stack


katrogan (approved):
thank you for cleaning up!

Issue #7703, opened 2026-07-23
[V2][Bug] Trigger task executed right away after deployed


popojk:
When deployed a trigger, the task will be executed right away. However, it should be first executed until the time interval reached.

PR #7704, opened 2026-07-23
[V2][Bug] don't fire triggers immediately on registration

popojk:
## Tracking issue

#7703 

Problem:
Deploying a schedule trigger caused it to fire once immediately, instead of waiting for its next scheduled time.

In UpdateSchedules, advance the computed start time along the schedule to the first tick strictly after now before registering the job. This leaves the (lastExec, now] window to CatchupAll and starts the steady-state job at the next future tick.

- Cron and fixed-rate schedules: advanced via sched.Next, preserving alignment.
- Future start_time (fixed-rate): unchanged, since it's already after now.

## How was this patch tested?
After the fix, the task will not be fired right after trigger deployed
Screenshot 2026-07-23 at 8 53 57 PM

### Labels

## Stack

### DIFF
--- a/docker/devbox-bundled/manifests/complete.yaml
+++ b/docker/devbox-bundled/manifests/complete.yaml
@@ -7199,6 +7199,14 @@ rules:
   - get
 - apiGroups:
   - ""
+  resources:
+  - podtemplates
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
   - events.k8s.io
   resources:
   - events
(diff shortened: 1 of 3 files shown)


docker/devbox-bundled/manifests/complete.yaml
```diff
@@ -7197,6 +7197,14 @@ rules:
   - pods/log
```
popojk, line 7197: This file change is not relevant to our PR fix. Just automatically update it according to the latest helm chart.

pingsutw (approved):

PR #7723, opened 2026-07-29
feat(runs): implement IdentityService.UserInfo from forwarded auth headers

pingsutw:

`IdentityService.UserInfo` was a stub carrying a `TODO: Wire with real auth`, so it
returned an empty `UserInfoResponse` to every caller. `flyte whoami` calls exactly this
RPC and prints the response verbatim, so it printed an empty record no matter who ran it.

The runs service already resolves the caller's identity for run attribution
(`executed_by`), from the auth headers the proxy/LB forwards. `UserInfo` simply wasn't
using it.

Implement `UserInfo` on top of the identity path that already exists, rather than adding
a second one:

1. `identityFromHeaders` over the configured proxy-forwarded headers — the ALB claims JWT
   (`X-Amzn-Oidc-Data`), a plain subject header (`X-Amzn-Oidc-Identity` /
   `X-Auth-Request-User`), or an `Authorization: Bearer` token.
2. `identityEnricher.enrich` on the Bearer path, where the access token carries only a
   subject and name/email come from the OIDC userinfo endpoint.
3. Map the resulting `common.User` onto the OIDC-shaped `UserInfoResponse`.

Because both `UserInfo` and `CreateRun` now read the same resolver, whoami and a run's
`executed_by` cannot disagree about who someone is.

`NewIdentityService` now takes the same three identity settings `NewRunService` does
(auth-server base URL, `trustForwardedIdentityHeaders`, `identityHeaders`); `runs/setup.go`
passes them from config.

Behaviour change worth calling out: `UserInfo` returns `Unauthenticated` when the request
carries no identity, or when `trustForwardedIdentityHeaders` is off (the caller cannot be
established at all). Previously it returned an empty response. This makes `flyte whoami`
report not-logged-in rather than printing `{}`.

`name` is derived from first/last since `UserSpec` has no full-name field.

## How was this patch tested?
Screenshot 2026-07-29 at 9 52 00 AM

Unit tests (`runs/service/identity_service_test.go`) — 14 cases covering each header path
(ALB claims JWT, ALB subject-only, oauth2-proxy plain values, Bearer), both unauthenticated
cases (no headers, malformed bearer), untrusted-headers rejection, userinfo enrichment
(including that userinfo's subject wins over the token's), graceful degradation when
userinfo is down, and the name/handle/photo field mapping.

```
$ go test ./runs/service/ -run 'TestIdentityService|TestUserInfoFromUser'
ok  	github.com/flyteorg/flyte/v2/runs/service	7.532s     # all 14 pass

$ go test ./runs/...        # full package, all green
$ go vet ./runs/...         # clean (one pre-existing warning in run_service_test.go, untouched)
```

End-to-end against the real CLI. I mounted the handler locally behind a shim that injects
the ALB header the way `authenticate-oidc` does, and pointed `flyte whoami` at it:

```
$ flyte --endpoint localhost:8099 --insecure whoami
{
  "subject": "00u3ipl8rr92YyW525d7",
  "name": "Kevin Su",
  "givenName": "Kevin",
  "familyName": "Su",
  "email": "kevin@union.ai"
}
```

And with no identity forwarded:

```
$ flyte --endpoint localhost:8098 --insecure whoami
ConnectError: no authenticated identity on request
```

Not tested against a deployed cluster — that needs an image build/rollout.

### Labels

**added**

## Related PRs

Builds on the identity resolution added in #7547.

### DIFF
--- a/runs/service/identity_service.go
+++ b/runs/service/identity_service.go
@@ -2,28 +2,88 @@ package service
 
 import (
 	"context"
+	"errors"
+	"strings"
 
 	"connectrpc.com/connect"
 
 	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/auth"
 	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/auth/authconnect"
+	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/common"
+	"github.com/flyteorg/flyte/v2/runs/config"
 )
 
-// IdentityService implements the IdentityServiceHandler interface.
-type IdentityService struct{}
+// errNoIdentity is returned when a request carries no resolvable caller identity.
+var errNoIdentity = errors.New("no authenticated identity on request")
 
-// NewIdentityService creates a new IdentityService instance.
-func NewIdentityService() *IdentityService {
-	return &IdentityService{}
+// IdentityService implements the IdentityServiceHandler interface.
+type IdentityService struct {
+	// enricher fills name/email from the OIDC userinfo endpoint on the Bearer path,
+	// where the access token carries only a subject. nil when no auth server is set.
+	enricher *identityEnricher
+
+	// trustHeaders gates resolving the caller from proxy-forwarded auth headers.
+	// See Config.TrustForwardedIdentityHeaders.
+	trustHeaders bool
+
+	// identityHeaders names the proxy-forwarded headers identity is read from.
+	identityHeaders config.IdentityHeadersConfig
+}
+
+// NewIdentityService creates a new IdentityService instance. It takes the same
+// identity settings as RunService so whoami reports exactly the identity that would
+// be recorded as a run's executed_by.
+func NewIdentityService(
+	authServerBaseURL string,
+	trustForwardedIdentityHeaders bool,
+	identityHeaders config.IdentityHeadersConfig,
+) *IdentityService {
+	return &IdentityService{
+		enricher:        newIdentityEnricher(authServerBaseURL),
+		trustHeaders:    trustForwardedIdentityHeaders,
+		identityHeaders: identityHeaders,
+	}
 }
 
 var _ authconnect.IdentityServiceHandler = (*IdentityService)(nil)
 
-// UserInfo returns information about the currently logged in user.
-// TODO: Wire with real auth to populate user info from the authenticated context.
+// UserInfo returns information about the currently logged in user, resolved from the
+// auth headers the proxy forwards — the same path run attribution uses, so whoami and
+// a run's executed_by can never disagree.
+//
+// Returns Unauthenticated when the request carries no identity, or when forwarded
+// headers aren't trusted (in which case the caller cannot be established at all), so
+// `flyte whoami` reports not-logged-in rather than printing an empty record.
 func (s *IdentityService) UserInfo(
 	ctx context.Context,
 	req *connect.Request[auth.UserInfoRequest],
 ) (*connect.Response[auth.UserInfoResponse], error) {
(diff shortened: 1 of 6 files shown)


runs/service/identity_service.go
```diff
@@ -2,28 +2,98 @@ package service
...
+		return nil, connect.NewError(connect.CodeUnauthenticated, errNoIdentity)
+	}
+
+	// Only enrich when the bearer token was the actual identity source.
+	if token := bearerToken(h); token != "" {
+		claimsHdr := s.identityHeaders.ClaimsJWTHeader
+		subjHdr := s.identityHeaders.SubjectHeader
+		if (claimsHdr == "" || identityFromJWT(h.Get(claimsHdr)) == nil) && (subjHdr == "" || strings.TrimSpace(h.Get(subjHdr)) == "") {
+			// On the Bearer path the token holds only a subject; userinfo supplies name/email.
+			id = s.enricher.enrich(ctx, token, id)
```
EngHabu: This might be expensive.. I think we call this endpoint a lot in the UI (/me).. what is that enricher? when do we use it elsewhere? does it cache?

runs/service/identity_service.go
```diff
@@ -2,28 +2,98 @@ package service
...
+	id := identityFromHeaders(h, s.identityHeaders)
+	if id == nil {
+		return nil, connect.NewError(connect.CodeUnauthenticated, errNoIdentity)
+	}
+
+	// Only enrich when the bearer token was the actual identity source.
+	if token := bearerToken(h); token != "" {
+		claimsHdr := s.identityHeaders.ClaimsJWTHeader
+		subjHdr := s.identityHeaders.SubjectHeader
+		if (claimsHdr == "" || identityFromJWT(h.Get(claimsHdr)) == nil) && (subjHdr == "" || strings.TrimSpace(h.Get(subjHdr)) == "") {
```
popojk: I just noticed that in [runs service](https://github.com/flyteorg/flyte/blob/d572f9f9d6b43422a6b38c6d81d8ce6f9dd82b14/runs/service/run_service.go#L354) we don't do this check before calling enrich. Should we unify the behavior?

popojk: nit: Could we wrap this if statement into a helper function to make it more readable?

runs/service/identity_service.go
pingsutw: We use it in the run service as well to get the email and user name. when people use sdk to run/deploy the task, there is only subject id in the headers
https://github.com/flyteorg/flyte/blob/d572f9f9d6b43422a6b38c6d81d8ce6f9dd82b14/runs/service/run_service.go#L354

pingsutw: yes, we do cache it for 24 hours, same as cloud.
https://github.com/flyteorg/flyte/blob/95a145a8e2c3f95c485fb1c3129bb5e6ff091cb4/runs/service/identity_enricher.go#L87

runs/service/identity_service.go
pingsutw: Unified in 3595d38b5 — extracted `resolveIdentity` (identityFromHeaders + the enrich-only-on-Bearer-path guard) and both `UserInfo` and `CreateRun` use it now. This also fixes a real gap in `CreateRun`: if a proxy header supplied the identity but the request also carried a Bearer token, enrich would adopt userinfo's subject and override the proxy-verified one.

pingsutw: fixed

pingsutw: fixed, thanks, good call

popojk (approved):
LGTM thanks! Let's see if @EngHabu  has further questions.

PR #7725, opened 2026-07-29
feat(secret): invalidate webhook secret cache on write

pingsutw:

The pod webhook's embedded secret manager caches resolved secret values so it doesn't hit the
backing store on every pod admission, with a **24h default TTL**. Nothing ever invalidated that
cache, so after a user wrote a secret through `SecretService`, task pods kept getting the **old**
value for up to a day. Users hit this as "I rotated my secret and my tasks still fail auth."

Two sharper cases fall out of the same gap:

**Cascade shadowing.** `lookUpSecret` resolves through a project+domain → domain → org cascade and
caches under whichever scope the fetcher hit. One logical secret can therefore be cached under any
of `(org, domain, project)`, `(org, domain, "")`, or `(org, "", "")`. Clearing only the exact tuple
misses a value cached at a broader scope, and the cascade keeps serving it. Concretely: an
org-scoped secret is cached, a user then creates a project-scoped override, the project key is
invalidated, the org key survives, and the cascade still returns the old value. This is why
**create** invalidates too, not just update.

**Deleted secrets staying live.** If the Secret is already gone from Kubernetes (deleted
out-of-band), `DeleteSecret` returns `NotFound`. Returning early there left the webhook serving the
cached value — a secret the user believes is deleted keeps being injected into pods until TTL.

**Webhook side — clearing the cache**

- `InvalidateCache` is part of the `SecretsInjector` interface, so adding a caching injector
  without wiring up invalidation is a compile error rather than a silently reintroduced staleness
  bug. Read-through injectors (global, k8s, aws, gcp, azure, vault) implement it as a no-op.
- `EmbeddedSecretManagerInjector.InvalidateCache` clears all three cascade keys — the caller can't
  know which scope the cached copy lived at, and the deletes are cheap.
- `SecretsPodMutator.InvalidateCache` fans out across enabled injectors.
- A plain-HTTP server on its own port (`webhook.cacheInvalidationPort`, default 9444) exposes
  `POST /invalidate-secret`, so invalidation keeps working once the webhook no longer shares a
  process with the secret service. `0` disables it.

**Secret service side — triggering it**

- Posts to the webhook after every write: create, update, and delete. Delete invalidates on its
  `NotFound` path too, via `defer`.
- Resolves the configured host and posts to **every** address rather than issuing one request to
  the name. Against a headless Service the name yields an A record per pod and Go's client picks
  one, so a single request would invalidate one arbitrary replica and leave the rest stale — a
  failure that looks exactly like success. A ClusterIP Service or bare host resolves to one
  address, so both topologies share the path.
- Best-effort by design: the secret is already durably written when invalidation runs, so failures
  are logged with an `n/m` pod count and the write RPC still succeeds. The degraded state is
  "old value until TTL", which is exactly the pre-PR behavior.

**Chart**

- A headless Service (`clusterIP: None`, port 9444) resolving to every webhook pod. Added
  **alongside** the existing webhook Service rather than converting it — that one is what the
  `MutatingWebhookConfiguration` points the API server at for admission and must keep its
  ClusterIP.
- `flyte-core-components.secret.webhookURL` points at it; the Go default stays `localhost` so
  local and devbox runs need no DNS.

**Incidental**

- The org is part of the secret cache key, so the value stamped on pod labels, used to encode the
  secret name, and passed to `InvalidateCache` must agree. They were three independent `"flyte"`
  literals in three packages; hoisted to `flytesecret.DefaultOrganization`.

### Security note

`/invalidate-secret` is plain HTTP and unauthenticated. It is bound to its own port and included
only in the cluster-internal Services, never routed through an ingress. It carries no secret
material in either direction — the request names a secret, the response is a status code — but it
is a cache-eviction primitive and should stay cluster-internal.

### A naming trap worth knowing

The webhook now serves two ports: **9443 for admission, 9444 for invalidation**. `webhookURL` does
not say which, and a URL with no port defaults to 9444. The field comment, config comment, pflag
help, and values.yaml comment all state this explicitly.

## How was this patch tested?

**Unit tests**

- `InvalidateCache` clears exactly the three cascade keys.
- Create, update, and delete each invalidate with the right `(org, domain, project, name)`.
- Delete invalidates even when the Secret is missing, while still reporting `NotFound`.
- Invalidation failure (disabled, unreachable, 5xx) never fails the write.
- Fan-out reaches every resolved address; target construction covers port defaulting, IPv6
  bracketing, trailing slash, and no-host rejection.
- Server: valid request 200, missing name 400, malformed body 400, `GET` 405.

`go build ./...` and `go test ./flyteplugins/go/tasks/pluginmachinery/secret/... ./executor/... ./secret/...`
pass. `helm lint` and `helm template` clean; chart README regenerated with helm-docs.

**End-to-end on a live single-binary cluster**

Rotation, against the 24h TTL that would otherwise make a stale read permanent:

| Step | Action | Task saw |
|---|---|---|
| 1 | create secret `v1`, run task | `v1` (now cached) |
| 2 | rotate to `v2` via the API, run task | `v2` — invalidation worked |
| 3 | patch the k8s Secret directly to `v3`, bypassing the service | `v2` — **stale, proving the cache is real** |
| 4 | `POST /invalidate-secret` by hand, run task | `v3` |

Step 3 is the control; without it step 2 proves nothing.

Deleted-secret fix, same shape:

| Step | Action | Result |
|---|---|---|
| 1 | create secret, run task | value injected, now cached |
| 2 | `kubectl delete` the Secret out-of-band | k8s empty, cache warm |
| 3 | run task | **still injected** — the bug |
| 4 | `DeleteSecret` on the now-missing secret | `NotFound` returned **and** invalidation fired |
| 5 | run task | pod never created; admission denied |

Fan-out was verified against a headless Service with two endpoints — one real pod and one dead
address — since the binary is not yet multi-replica-ready (see below). A single write produced an
invalidation on the real pod *and* a connection attempt to the dead one, ending in
`failed on 1/2 webhook pod(s)`: both addresses attempted from one write, which a single-request
implementation cannot produce.

## Known limitations

- **Fan-out is not exercised by real replicas yet.** `flyte-binary` is hardcoded to `replicas: 1`,
  and running more today would be unsafe for unrelated reasons: leader election defaults to off,
  and with `localCert: true` each replica mints its own webhook cert and races on the
  `MutatingWebhookConfiguration` CA bundle. The headless Service and fan-out are in place so
  invalidation is already correct when those are fixed; they are not a claim that multi-replica
  works today.
- **Invalidation is best-effort.** A wedged or unreachable webhook degrades to TTL behavior with a
  warning log rather than failing the write.

### Labels

**added**, **fixed**

https://claude.ai/code/session_01AHoR3R93whTNKzGhP2Mi4w

### DIFF
--- a/charts/flyte-binary/templates/service/webhook.yaml
+++ b/charts/flyte-binary/templates/service/webhook.yaml
@@ -11,8 +11,3 @@ spec:
       port: 443
       targetPort: 9443
       protocol: TCP
-    # Plain HTTP, unauthenticated: cluster-internal only, never routed through an ingress.
-    - name: cache-internal
-      port: 9444
-      targetPort: 9444
-      protocol: TCP
--- a/charts/flyte-binary/templates/service/webhook.yaml
+++ b/charts/flyte-binary/templates/service/webhook.yaml
@@ -11,3 +11,8 @@ spec:
       port: 443
       targetPort: 9443
       protocol: TCP
+    # Plain HTTP, unauthenticated: cluster-internal only, never routed through an ingress.
+    - name: cache-internal
+      port: 9444
+      targetPort: 9444
+      protocol: TCP
--- a/executor/pkg/webhook/cache_invalidation_server.go
+++ b/executor/pkg/webhook/cache_invalidation_server.go
@@ -1,0 +1,72 @@
+package webhook
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/secret"
+	"github.com/flyteorg/flyte/v2/flytestdlib/logger"
+)
+
+// InvalidateSecretPath is the HTTP path for the cache invalidation endpoint.
+const InvalidateSecretPath = "/invalidate-secret"
+
+// InvalidateRequest is the body of a POST to InvalidateSecretPath.
+type InvalidateRequest struct {
+	Org     string `json:"org"`
+	Domain  string `json:"domain"`
+	Project string `json:"project"`
+	Name    string `json:"name"`
+}
+
+// StartCacheInvalidationServer starts a plain HTTP server that listens for cache invalidation
+// requests. It exposes POST /invalidate-secret so other services (the secret service) can drop
+// secret values cached in the webhook process instead of waiting out the cache TTL.
+//
+// The server is plain HTTP and unauthenticated, so it must stay cluster-internal: it is bound to
+// its own port and never routed through an ingress. Blocks until ctx is cancelled.
+func StartCacheInvalidationServer(ctx context.Context, port int, mutator *secret.SecretsPodMutator) error {
+	mux := http.NewServeMux()
+	mux.HandleFunc(InvalidateSecretPath, func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodPost {
+			http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+			return
+		}
+
+		var req InvalidateRequest
+		if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+			http.Error(w, fmt.Sprintf("invalid request body: %v", err), http.StatusBadRequest)
+			return
+		}
+
+		if req.Name == "" {
+			http.Error(w, "name is required", http.StatusBadRequest)
+			return
+		}
+
+		mutator.InvalidateCache(r.Context(), req.Org, req.Domain, req.Project, req.Name)
+		w.WriteHeader(http.StatusOK)
+	})
+
+	addr := fmt.Sprintf(":%d", port)
+	server := &http.Server{ // #nosec G112 -- internal-only server, not exposed outside the cluster
+		Addr:    addr,
+		Handler: mux,
+	}
+
+	go func() {
+		<-ctx.Done()
+		if err := server.Shutdown(context.Background()); err != nil {
+			logger.Errorf(ctx, "Failed to shutdown cache invalidation server: %v", err)
+		}
+	}()
+
+	logger.Infof(ctx, "Starting cache invalidation server on %s", addr)
+	if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		return fmt.Errorf("cache invalidation server failed: %w", err)
+	}
+
+	return nil
+}
--- a/flyteplugins/go/tasks/pluginmachinery/secret/config/config_flags_test.go
+++ b/flyteplugins/go/tasks/pluginmachinery/secret/config/config_flags_test.go
@@ -10,7 +10,7 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/mitchellh/mapstructure"
+	"github.com/go-viper/mapstructure/v2"
 	"github.com/stretchr/testify/assert"
 )
 
--- a/flyteplugins/go/tasks/pluginmachinery/secret/config/config_flags_test.go
+++ b/flyteplugins/go/tasks/pluginmachinery/secret/config/config_flags_test.go
@@ -10,7 +10,7 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/go-viper/mapstructure/v2"
+	"github.com/mitchellh/mapstructure"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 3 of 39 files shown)


flyteplugins/go/tasks/pluginmachinery/secret/config/config_flags_test.go
```diff
@@ -10,7 +10,7 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/go-viper/mapstructure/v2"
+	"github.com/mitchellh/mapstructure"
```
katrogan: why did this change?

secret/service/secret_service.go
```diff
@@ -11,6 +18,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/flyteorg/flyte/v2/executor/pkg/webhook"
```
katrogan, line 21: hm now we're depending this package which is going to make "point it at the headless webhook Service once they are split" from the PR description messy

Can we import flyteplugins/go/tasks/pluginmachinery/secret, which is where InvalidateCache is defined?

secret/service/secret_service.go
```diff
@@ -97,6 +222,8 @@ func (s *SecretService) CreateSecret(ctx context.Context, req *connect.Request[s
 		return nil, connect.NewError(connect.CodeInternal, err)
 	}
 
+	s.invalidateWebhookSecretCache(ctx, req.Msg.GetId())
```
katrogan, line 237: if a caller cancels the request but we've already persisted the right then the canceled context means we never get here. Maybe we can use `context.WithoutCancel(ctx)` to ensure this gets hit regardless?

charts/flyte-binary/templates/service/webhook.yaml
```diff
@@ -11,3 +11,8 @@ spec:
       port: 443
       targetPort: 9443
       protocol: TCP
+    # Plain HTTP, unauthenticated: cluster-internal only, never routed through an ingress.
+    - name: cache-internal
```
katrogan: who is using this today? don't we want the headless service so we can modify the cache across replicas?

executor/pkg/webhook/cache_invalidation_server.go
```diff
@@ -0,0 +1,72 @@
...
+			http.Error(w, "name is required", http.StatusBadRequest)
+			return
+		}
+
+		mutator.InvalidateCache(r.Context(), req.Org, req.Domain, req.Project, req.Name)
+		w.WriteHeader(http.StatusOK)
+	})
+
+	addr := fmt.Sprintf(":%d", port)
+	server := &http.Server{ // #nosec G112 -- internal-only server, not exposed outside the cluster
```
katrogan, line 54: why is this necessary? what config is missing here?

secret/service/secret_service.go
```diff
@@ -37,11 +55,118 @@ var _ secretconnect.SecretServiceHandler = (*SecretService)(nil)
...
+// Resolution rather than a single request is what makes this correct against a headless Service.
+func (s *SecretService) invalidationTargets(ctx context.Context) ([]string, error) {
+	base, err := url.Parse(strings.TrimSuffix(s.webhookURL, "/"))
+	if err != nil {
+		return nil, fmt.Errorf("parsing %q: %w", s.webhookURL, err)
+	}
+
+	host := base.Hostname()
+	if host == "" {
+		return nil, fmt.Errorf("no host in %q", s.webhookURL)
```
katrogan, line 138: won't this reject endpoints w/o a scheme like localhost?

katrogan (dismissed):

flyteplugins/go/tasks/pluginmachinery/secret/config/config_flags_test.go
pingsutw: it's autogenerated. it changes since we are still using plag from v1 stdlib to generate the code. I used v2 plag to generate locally
https://github.com/flyteorg/flyte/blob/2187f5470c3dadd32d1a748ee8a5e14c6e4f02ba/boilerplate/flyte/golang_support_tools/go.mod#L7

There are some drifts in other files, I'll create a follow-up PR/issue for it

secret/service/secret_service.go
pingsutw: We always call the webhook endpoint to invalidate the cache, so the code will be the same in both single-binary and multi-pod environments.

charts/flyte-binary/templates/service/webhook.yaml
pingsutw: Removed it. I was using it to test locally before adding headless service. sorry for the confusion.

secret/service/secret_service.go
pingsutw: good call, thanks, added context.WithoutCancel() in the invalidateWebhookSecretCache

executor/pkg/webhook/cache_invalidation_server.go
pingsutw: We need to start an HTTP server here, no? The code is the same as what we use in the cloud.

secret/service/secret_service.go
pingsutw: if it doesn't have schema (http://). I added a default schema if it's missing in the URL

katrogan (approved):

PR #7731, opened 2026-07-30
feat: add ReusePolicy to TaskTemplate

pingsutw:
## Tracking issue

Related to first-class environment reuse for tasks (reusable containers and cluster-backed plugins).

Today a task's environment-reuse configuration (replica bounds, concurrency, idle/scaledown TTLs) is encoded inside the untyped `TaskTemplate.custom` `Struct`. That works but is stringly-typed: every producer and consumer has to agree on ad-hoc key names and value encodings, and it can't be validated or discovered from the schema.

This adds a structured `ReusePolicy` message and a `reuse_policy` field on `TaskTemplate`, so the policy can be expressed as a first-class field. Reusable and cluster-backed plugins can then read it directly instead of parsing `custom`.

### `ReusePolicy` message (`flyteidl2/core/tasks.proto`)

```proto
message ReusePolicy {
  option (buf.validate.message).cel = {
    id: "reuse_policy.replicas"
    message: "max_replicas must be >= min_replicas"
    expression: "this.max_replicas >= this.min_replicas"
  };

  enum Scope {
    GLOBAL = 0;  // reuse one environment across all runs (default)
    RUN = 1;     // restrict reuse to a single run; each run gets its own environment
    // PROJECT and DOMAIN scoping are reserved for future use.
  }

  int32 min_replicas = 1 [(buf.validate.field).int32.gte = 0];   // minimum warm replicas (fixed pool: min == max)
  int32 max_replicas = 2 [(buf.validate.field).int32.gte = 1];   // auto-scaling upper bound
  int32 concurrency = 3 [(buf.validate.field).int32.gte = 1];    // max concurrent tasks per replica (async only for > 1)
  google.protobuf.Duration idle_ttl = 4;                         // tear the whole environment down after it's idle this long
  google.protobuf.Duration scaledown_ttl = 5;                    // min delay before an individual idle replica scales down
  Scope scope = 6 [(buf.validate.field).enum.defined_only = true];
}
```

Total concurrent capacity is `max_replicas * concurrency`. `idle_ttl` governs the whole environment; `scaledown_ttl` governs an individual replica during auto-scaling.

- **`Scope`** bounds how widely a reusable environment may be shared: `GLOBAL` (default, reuse across all runs) or `RUN` (each run gets its own environment). `PROJECT`/`DOMAIN` are reserved for later.
- **`buf.validate` constraints** document the field invariants at the schema level — `min_replicas >= 0`, `max_replicas >= 1`, `concurrency >= 1`, a defined-only `scope`, and a message-level CEL enforcing `max_replicas >= min_replicas`. These are ready for any consumer that runs `protovalidate`.

### `TaskTemplate.reuse_policy` (field 19)

```proto
ReusePolicy reuse_policy = 19;
```

Purely additive and wire-compatible — existing tasks leave it unset, and consumers that read the policy from `custom` are unaffected. Generated bindings (Go, Python, TypeScript, Rust) are regenerated.

## How was this patch tested?

- `make buf-lint` passes; `make buf-go buf-python buf-ts buf-rust` regenerate cleanly; `go build ./...` passes.
- The `buf.validate` constraints were verified enforced at runtime via `protovalidate.Validate`: a valid policy passes, while `min_replicas > max_replicas`, `max_replicas < 1`, `concurrency < 1`, and an undefined `scope` are each rejected.
- Additive proto change (new message + new field number), so no behavior changes and nothing to migrate.


flyteidl2/core/tasks.proto
```diff
@@ -211,6 +212,55 @@ message TaskMetadata {
...
+  int32 min_replicas = 1 [(buf.validate.field).int32.gte = 0];
+
+  // Maximum number of replicas; the upper bound for demand-based auto-scaling.
+  int32 max_replicas = 2 [(buf.validate.field).int32.gte = 1];
+
+  // Maximum number of concurrent tasks a single replica may handle. Values > 1 are only
+  // meaningful for async tasks.
+  int32 concurrency = 3 [(buf.validate.field).int32.gte = 1];
+
+  // How long the whole reusable environment may sit idle (no running tasks) before it is torn
```
pingsutw, line 248: same here https://github.com/flyteorg/flyte-sdk/blob/78968759f1a9a44af678c2c96d17c3c200666b18/src/flyte/_internal/runtime/reuse.py#L143-L150

kumare3 (approved):

PR #7732, opened 2026-07-30
 feat: add total app count to Apps List API

AdilFayyaz:
## Tracking issue
In reference to the following issue on Linear: [ENG26-969](https://linear.app/unionai/issue/ENG26-969/add-total-app-count-to-apps-list-api-for-ui-display)

The console needs to display the total number of apps, but AppService/List returns only a page of results plus a pagination token, so an infinite query with `limit: 10` can never know the total. This adds an opt-in count to the list response.

  - Add bool `include_total_count = 6` to `flyteidl2.app.ListRequest` — opt-in so the extra `COUNT(*)` stays off pagination pages, poll ticks, and internal callers (same pattern as the existing `disable_identity_enrichment flag`).
  - Add `int64 total_count = 3` to `flyteidl2.app.ListResponse`, populated only when the flag is set. The count honors the request's filters, so it respects the UI's name search for free.
  - Regenerate all committed bindings: Go (app_payload.pb.go, .pb.validate.go), TypeScript, Python (_pb2.py/.pyi), and Rust.

  IDL-only — server-side wiring and the frontend hook land in the cloud repo once this merges and the submodule pointer is bumped.

## How was this patch tested?

  - `make buf-go buf-rust buf-python` and a local buf generate --template `buf.gen.ts.yaml` reproduce the committed generated files with no further diff.
  - `go build ./gen/go/flyteidl2/app/...` passes.
  - `buf format -d flyteidl2/app/app_payload.proto` is clean.
  - `buf lint --exclude-path flytestdlib/` reports only the two pre-existing `flyteidl2/plugins/clustered.proto` errors (verified present on a clean tree).

## Stack


flyteidl2/app/app_payload.proto
```diff
@@ -82,6 +82,11 @@ message ListRequest {
 
   // Disable caller identity enrichment in the response.
   bool disable_identity_enrichment = 5;
+
+  // Include the total number of apps matching the request's filters in the response.
+  // This performs an additional count query, so it should typically only be set on
+  // the first page of a paginated listing.
```
pingsutw: What if people deploy another app after the first page of a paginated listing

AdilFayyaz: the doc comment can be softened. on the BE any client can re-request it any point in time.

pingsutw (approved):

PR #7737, opened 2026-07-31
feat: add LocalRunService for reporting locally-orchestrated runs

EngHabu:
## Overview

Adds `flyteidl2.workflow.LocalRunService`: clients that orchestrate actions **outside** the platform (e.g. on a user's machine) register a run and report action state, while the platform only records and serves it — no queueing, no dataplane.

Design highlights:
- The read/watch surface (`GetRunDetails`/`WatchRunDetails`, `GetActionDetails`/`WatchActionDetails`, `ListRuns`/`WatchRuns`, `ListActions`/`WatchActions`) reuses RunService's request/response messages **verbatim**, so existing clients and UIs work against local runs with only a service swap. The two new shapes are `CreateLocalRunRequest` and `ReportLocalActionsRequest` (batch, idempotent on the event's `(attempt, version, phase)`).
- `DataProxyService` gains the control-plane-direct data surface for local runs: `UploadMetadata` (signed PUT URLs for inputs.pb/outputs.pb/report.html), `GetLocalActionData`, `CreateLocalDownloadLink`, plus `ArtifactType` `INPUTS`/`OUTPUTS` values.
- `RunSource.RUN_SOURCE_LOCAL` marks these runs.

buf lint's message-reuse rules are ignored for the affected files (same precedent as cacheservice/project).

## Test Plan

Codegen (`make buf-*` + mockery) committed for go/ts/python/rust; consumed and exercised end-to-end by the union cloud implementation (service + integration harness driving create → report → watch streams → reads).

## Rollout Plan

IDL + generated code only; no server behavior in this repo changes. Server implementation lands in the consuming repo.

## Rollback Plan

Revert; no persisted state depends on this.


EngHabu:
Reworked per review discussion: the three dedicated local-run dataproxy RPCs are gone. Local-run data now flows through the existing APIs — `CreateUploadLocation` for uploads and `GetActionData`/download links for reads — routed client-side via `SelectCluster` with the new `OPERATION_LOCAL_RUN_DATA` (server answers a dataplane endpoint + cluster name for zero-trust orgs, its own endpoint otherwise). Two small extensions to existing messages carry the routing: `SelectClusterResponse.cluster` (reporters stamp it on attempt events; reads route by the recorded cluster) and `GetActionDataRequest.run_source` (run names aren't unique across platform/local, so resolution needs the hint). Net −2k lines. Verified live end-to-end against the cloud/SDK branches.

flyteidl2/workflow/local_run_service.proto
```diff
@@ -0,0 +1,149 @@
...
+  // honored on the first report of an action.
+  string parent_name = 2;
+
+  // Group this action belongs to, if applicable. Only honored on the first report of an action.
+  string group = 3;
+
+  // The action's spec. Should be provided on the first report of an action; ignored afterwards.
+  oneof spec {
+    TaskAction task = 4;
+    TraceAction trace = 5;
```
kumare3, line 128: how do we restrict condition action, we should test that it fails gracefully

kumare3 (approved):

ursucarina (approved):

PR #7738, opened 2026-07-31
Do not leak implementation details in internal error messages

### LINKED ISSUE #7736: Don't leak implementation details in internal errors sent to clients
Flyte has this pattern where often times an error is logged and then also returned somewhere. Ideally errors should just be returned until the point where they are finally handled and they should be logged at that point. We should have an error interceptor that does this logging.

Additionally, we should return unhandled errors as INTERNAL instead of allowing the default behavior of returning unhandled errors as UNKNOWN back to users.

Sovietaced:
## Tracking issue
Closes #7736 

The Flyte backend currently has some behavior that I don't think are ideal.
1. Internal errors are largely passed through directly to the client which is a [security issue](https://owasp.org/www-community/Improper_Error_Handling). 
2. In many places errors are logged to stdout and then duplicated in the returned error which is an anti pattern
3. Non connect RPC errors are currently returned as UNKNOWN and directly pass through error details to the user. Ultimately these are internal server issues.  

Add an error interceptor that..
1. Logs non-INTERNAL connect RPC errors at WARN level so every error is visible in logs for debugging
2. Converts non connect RPC errors into INTERNAL and omits error messages to avoid leaking implementation details
3. Logs INTERNAL errors at ERROR level since this represents a server side issue

Also updates existing RPCs that return internal to return a non connect RPC error with additional context that will get caught by the error interceptor. Note: I could potentially just keep the existing INTERNAL calls and just drop the error details from being passed through in the interceptor logic instead to be more explicit for RPC handler authors. 

Also removes `toConnectError` helper that was used in one area that basically does what the error interceptor does but at each RPC call site.

## How was this patch tested?
We use this error interceptor internally.

### Labels

## Stack

### DIFF
--- a/actions/service/actions_service.go
+++ b/actions/service/actions_service.go
@@ -2,7 +2,6 @@ package service
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"connectrpc.com/connect"
@@ -31,8 +30,7 @@ func (s *ActionsService) Signal(ctx context.Context, req *connect.Request[action
 	}
 
 	if err := s.client.Signal(ctx, req.Msg.ActionId, req.Msg.Value, principalSubject(req.Msg.GetSignalledBy())); err != nil {
-		logger.Errorf(ctx, "Failed to signal action: %v", err)
-		return nil, toConnectError(err)
+		return nil, fmt.Errorf("failed to signal action: %w", err)
 	}
 
 	return connect.NewResponse(&actions.SignalResponse{}), nil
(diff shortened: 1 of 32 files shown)


pingsutw (commented):
We should also update [this](https://github.com/flyteorg/flyte/blob/1dd046406db056f08d27adf2299d2527be9bdffb/runs/service/abort_reconciler.go#L286-L293) since it no longer get connect error

Sovietaced:
Discussed over slack. There are some things that depend on error messages. Also we should pass the request ID of trace ID in the error messages if possible.

PR #7739, opened 2026-07-31
feat: Adding cluster_id to SelectClusterRequest

notSoWiseOldMan:

`SelectClusterRequest`'s `resource` oneof can name a run/action, an app, a project, a
domain, or a cluster pool — but not a single cluster. Every existing variant is resolved
*to* a cluster by the service, which is the right model when any healthy cluster serving
that resource will do.

Not every operation works that way. Some are inherently cluster-scoped: the caller already
knows which cluster it needs, and no other cluster can answer the question. Today such a
caller has to pick a resource variant that resolves to *some* cluster in the pool, which
returns a different cluster's data and looks like a success. There is no way to express
"this cluster, or fail".

Adds a cluster-scoped member to the `resource` oneof in
`flyteidl2/cluster/payload.proto`:

```proto
// Names one specific cluster. Used by operations that are inherently
// cluster-scoped, where any other cluster in the pool would serve the
// wrong data.
flyteidl2.common.ClusterIdentifier cluster_id = 10;
```

Plus the regenerated Go bindings, validators, and swagger. `ClusterIdentifier` already
exists in `flyteidl2/common` and is org-scoped (`organization` + `name`), matching how
clusters are keyed elsewhere.

This is additive and backward compatible: field 10 is unused, no existing field is renamed
or renumbered, and no existing oneof member changes meaning. Servers that predate this
change see the new variant as an unset oneof and reject it at validation, which callers can
detect and fall back on.

No behavior change in this repo — nothing here reads `cluster_id` yet. It is up to an
implementing service to define how the variant resolves; the intent is that it routes to
the named cluster and fails rather than substituting another cluster when the named one is
unavailable.

### Labels

**added**

## How was this patch tested?

Codegen only — `buf generate`, with the generated Go, validators, and swagger checked in.
There is no new logic in this repo to unit test.

## Stack


flyteidl2/cluster/payload.proto
```diff
@@ -20,6 +20,10 @@ message SelectClusterRequest {
     flyteidl2.app.Identifier app_id = 6;
     flyteidl2.common.ClusterPoolIdentifier cluster_pool_id = 7;
     flyteidl2.common.DomainIdentifier domain_id = 9;
+    // Names one specific cluster. Used by operations that are inherently
```
katrogan: Just a point of clarification, but some of the resources above (action attempt id and app id) when used in conjunction with a tail logs operation below, are implicitly pinned to a cluster too, we just don't require the user to specify the cluster since we do the lookup on the backend.

katrogan (dismissed):

flyteidl2/cluster/payload.proto
notSoWiseOldMan: Updated to be more accurate

katrogan (dismissed):

pingsutw (approved):

PR #7744, opened 2026-07-31
[Spark] Upgrade client library

wild-endeavor:
First step in spark upgrade: Swap the spark plugin's client library from the 2020 era `GoogleCloudPlatform/spark-on-k8s-operator` to `kubeflow/spark-operator/v2 v2.4.0`, without requiring any cluster's operator or CRDs to be upgraded first.

## How it works
On startup, first test to see if the `SparkApplication` CRD has a driver pod field called `template`. The [old one](https://github.com/kubeflow/spark-operator/blob/spark-operator-chart-1.1.27/charts/spark-operator-chart/crds/sparkoperator.k8s.io_sparkapplications.yaml#L101) does not, and the [new one does](https://github.com/kubeflow/spark-operator/blob/v2.4.0/charts/spark-operator-chart/crds/sparkoperator.k8s.io_sparkapplications.yaml#L5027).

Then keep writing every `v1beta2` field it writes today.  If the CRD detected is old (i.e. doesn't have the `template` field), the serialized objects are _nearly_ byte-identical to before (see the breaking section below).  If the CRD does have that field, then the plugin additionally passes the full flyte-built pod spec through as the driver/executor pod template.

On clusters with the old CRDs, the k8s api server just prunes the template field away.  A manual kill switch is also provided: `plugins.spark.enable-pod-template: false`

### What's breaking

One field: the pod-level security context. The old struct serialized it as `securityContext`; the new one is `podSecurityContext`, which old CRDs prune. So on a cluster with old CRDs, a security context coming from platform config or pod templates silently stops reaching spark pods.  It's used again (via the new `template` field) once that cluster's CRDs & operator are upgraded.

### Non-breaking changes

The new API moved/removed two fields we set:
- `DriverSpec.ServiceAccount` moved into the embedded `SparkPodSpec` set inside `sparkOp.DriverSpec`.
- The top-level `spec.serviceAccount` is gone. We checked the operator version actually deployed with the 1.1.x charts (v1beta2-1.3.8): it never reads that key — the driver SA comes from `spec.driver.serviceAccount`, which we still send. Dead field, safe to drop.

Side-note from that dig: the old client never set an executor service account (the field didn't exist in its API), so executors have always run as the namespace default SA. On template-capable clusters we now set the task SA on executors too.

## Testing

Unit tests cover both gate positions, including an assertion that the with-template object minus the template is deeply equal to the without-template object. Runtime verification (old operator + new client on a live cluster, pod-template disk/security-context checks) is planned via `flyte-sdk/examples/plugins/spark_migrate/`.


flyteplugins/go/tasks/plugins/k8s/spark/podtemplate.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+	}
+
+	clientset, err := apiextensionsclientset.NewForConfig(restConfig)
+	if err != nil {
+		logger.Warnf(ctx, "Cannot build apiextensions client to inspect the SparkApplication CRD, assuming no pod template support: %v", err)
+		return false
+	}
+
+	getCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
EngHabu: move to config?

EngHabu (dismissed):

flyteplugins/go/tasks/plugins/k8s/spark/podtemplate.go
wild-endeavor: done

pmahindrakar-oss (approved):

PR #7749, opened 2026-08-03
Add artifact trigger automation: TYPE_ARTIFACT + ArtifactTrigger

kumare3:
## Summary

Adds the third `TriggerAutomationSpec` automation type for **artifact triggers**: fire a run when a new version of a named artifact is created.

- `TriggerAutomationSpecType.TYPE_ARTIFACT = 3`
- `ArtifactTrigger{artifact_name, version, input_arg}` in the `TriggerAutomationSpec.automation` oneof — `version` optionally pins an exact version (empty = any new version); `input_arg` names the task input that receives the triggering artifact's value at fire time, mirroring `Schedule.kickoff_time_input_arg`
- `RunSource.RUN_SOURCE_ARTIFACT_TRIGGER = 4` so artifact-triggered runs carry their provenance

Backend consumption lands separately (union cloud); the SDK side adds `flyte.OnArtifact(...)` + a `flyte.TriggeredArtifact` input sentinel on top of these protos.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- `main` 
  - **Add artifact trigger automation: TYPE\_ARTIFACT + ArtifactTrigger** :point\_left:
    - \#7760

PR #7760, opened 2026-08-04
First-class artifact info: Literal.artifact_id + Outputs.produced_artifacts

kumare3:
## Summary

Stacked on #7749. Replaces the implicit metadata-key contract (magic `_ua`/`_uap` keys inside `Literal.metadata` carrying JSON payloads that the Python SDK and Go backends must keep byte-in-sync) with typed fields, split by lifecycle:

**Identity → the value** (permanent, value-intrinsic):

- `core.ArtifactVersionId { ArtifactKey key; string version; }` — minimal identity, deliberately not the v1-heritage `ArtifactID` (no partition machinery)
- `core.Literal.artifact_id` — stamped once at artifact registration (and on artifact-bound inputs); travels with the value through every copy; participates in cache-key hashing by design (two versions of an artifact are distinct inputs even when bytes coincide)

**Registration declaration → the output envelope** (transient, slot-scoped):

- `task.ProducedArtifact { output, name, version?, core.ArtifactInfo info, core.LiteralType type }` + `task.Outputs.produced_artifacts = 3` — an action-scoped instruction that never enters the value stream or downstream hashes. Field 3 (not 2) because `outputs.pb` blobs are cross-decoded as `task.Inputs`, where field 2 is `context`. The declared `type` makes the declaration self-contained (backends no longer need a task-interface lookup with a silent nil-type drop).

**Shared descriptive payload**:

- `core.ArtifactInfo { description, user_metadata, card }` + `core.ArtifactCard` (hoisted from `artifact.Card`, identical shape) — used verbatim by both `task.ProducedArtifact.info` and the restructured `artifact.ArtifactSpec { value, type, info, source }`, and excluding identity so declaration and spec can never drift.

Consumers (union cloud leaseworker/artifact service, flyte-sdk) migrate in their respective stacks; the metadata keys are deleted outright there since the artifact stack is unreleased.

- `main` 
  - \#7749
    - **First-class artifact info: Literal.artifact\_id + Outputs.produced\_artifacts** :point\_left:


flyteidl2/core/artifact_id.proto
```diff
@@ -14,6 +15,42 @@ message ArtifactKey {
...
+
+  // Free-form, user-supplied key/value metadata.
+  map<string, string> user_metadata = 2;
+
+  // Optional card describing this artifact.
+  ArtifactCard card = 3;
+}
+
 // Only valid for triggers
 message ArtifactBindingData {
```
wild-endeavor, line 55: is this used?

flyteidl2/core/artifact_id.proto
```diff
@@ -14,6 +15,42 @@ message ArtifactKey {
   string org = 4;
```
wild-endeavor, line 15: should we add validation to these fields?

wild-endeavor (approved):

PR #7764, opened 2026-08-04
feat(devbox): optional Grafana/Prometheus/OTel stack

pingsutw:
Screenshot 2026-08-04 at 12 01 06 AM

Reading the RPC metrics from a devbox is all manual today: stand up an OTel collector, Prometheus and Grafana by hand, then port the dashboard across, every time. The dashboard itself only lived on one dev cluster, so there was no version-controlled copy to point an AWS or GCP deployment at either.

An **opt-in** monitoring stack for the devbox, plus the dashboard as a repo asset:

```bash
make devbox-run           # unchanged, still 1.7 GB
make devbox-monitoring    # adds collector -> Prometheus -> Grafana + the dashboard
make devbox-monitoring-stop
```

**Why opt-in and not bundled.** Preloading the three images into the devbox image measures at **+1.43 GB unpacked** (Grafana 704 MB, Prometheus 404 MB, collector 334 MB), taking it from **1.7 GB → 3.13 GB**; compressed, 514 MB → 835 MB. That is 84% larger for something most devbox users never open, so they are pulled on demand instead — the default image is untouched and the pods are up in ~20 s.

**The dashboard lives at `monitoring/dashboards/flyte-execution.json`**, not under `docker/devbox-bundled/`, because it charts the `rpc_*` metrics any flyte2 emits — the same JSON is what you want on an AWS or GCP install. `monitoring/README.md` covers the three ways to use it, including the ConfigMap labelled `grafana_dashboard: "1"` that a kube-prometheus-stack Grafana sidecar provisions.

It is kept as a real `.json` file rather than embedded in the manifest, so it diffs as JSON; the ConfigMap is built from the directory at apply time, and dropping another `.json` there is enough to add a dashboard.

**Deployment shape:** no operator and no CRDs — a static scrape config is enough for a single node. Two collector settings are load-bearing:

- `max_recv_msg_size_mib: 64` — otelconnect's cumulative payload exceeds the 4 MiB default and exports fail with `ResourceExhausted`.
- `resource_to_telemetry_conversion` — without it `service.name` never becomes a label, so the dashboard's `service` variable is empty and every per-service panel has nothing to group by. It surfaces only as an empty dropdown, which reads as "the dashboard is broken" rather than "the config is wrong".

Grafana runs with anonymous **Viewer** access, not Admin: k3s publishes NodePorts on `0.0.0.0`, so this is LAN-reachable, and a Grafana Admin can add datasources — which makes the pod issue arbitrary requests to whatever it can reach. Viewing is all the stack is for; basic auth stays enabled so an admin login is still available for editing panels.

`start` also publishes `30300` for the Grafana NodePort. A devbox created before this change needs recreating for the URL to work — `make devbox-monitoring` detects that and says so instead of printing a dead link.

## How was this patch tested?

No unit tests: this is Makefile plumbing plus a manifest, so it was tested by running it.

**On a fresh devbox from the unmodified image** (the path a user actually takes):

```bash
make devbox-run
make devbox-monitoring
```

- Images pulled on demand, all three pods Running in 22 s
- Grafana healthy, `Flyte Execution` dashboard (uid `oss`) provisioned, datasource health OK
- Prometheus target `otel-collector:8889` → `up`
- End to end: 30 RPCs in → `rpc_server_duration_milliseconds_count{service_name="runs-service", rpc_method="ListProjects"} = 30`
- `/d/oss/flyte-execution` → HTTP 200, 24 panels, both template variables

**The security posture** was verified after the change: `/d/oss/flyte-execution` and `/api/search` still 200 anonymously; `POST /api/datasources` returns 403 where it previously succeeded.

**The ConfigMap path** was verified by deleting the generated ConfigMap and re-running the target — recreated from the JSON file, key follows the filename, Grafana re-provisions.

**Also verified the bundled-image variant** to get the size numbers: images preloaded as tars, pods Running in 21 s with zero pull events, i.e. fully offline. Those changes are not in this PR.

### Labels

- **added**: the optional monitoring stack and the versioned dashboard

### Setup process

```bash
make devbox-run
make devbox-monitoring
open http://localhost:30300/d/oss/flyte-execution   # anonymous viewer
```

Metrics appear ~60 s after the first RPC traffic.


pingsutw:
A background security review flagged the Grafana auth config in `monitoring.yaml`, and it was a fair hit — fixed in f798d1080.

The stack shipped **anonymous Admin with basic auth disabled**. Since k3s publishes NodePorts on `0.0.0.0` rather than localhost, that is LAN-reachable, not just local to the machine running the devbox. A Grafana Admin can create datasources, which makes the Grafana pod issue arbitrary requests to whatever it can reach (the flyte API, Postgres, the registry) — a pivot into the devbox rather than just dashboard reading. And with basic auth off there was no real admin login either, so the role could not be tightened without editing the manifest.

Anonymous access is now **Viewer**, and basic auth is re-enabled so an admin login remains available for editing panels. Viewing without a login — the whole point of the stack — is unaffected.

Verified against a running devbox:

| request | before | after |
|---|---|---|
| `GET /d/oss/flyte-v2` | 200 | 200 |
| `GET /api/search` | 200 | 200 |
| `POST /api/datasources` | 200 | **403** |

Sovietaced (approved):

PR #7773, opened 2026-08-04
Move settings desc field option off extension number 50000

afbrock:
The desc field option on google.protobuf.FieldOptions sat at 50000, in the range protobuf reserves for internal use within a single organization. That range is only safe when you control every .proto sharing a descriptor pool. flyteidl2 is published to buf.build, PyPI and npm, so it routinely shares a pool with third-party protos, and desc collided with another published extension claiming the same tag. The collision fails extension registration at import time for anyone importing both schemas.

Move desc to 1364, from a block of 1364-1373 requested for Flyte in the protobuf global extension registry
(protocolbuffers/protobuf#29013). The block deliberately starts after 1314-1363, which is claimed by the currently open
protocolbuffers/protobuf#28919.

This is backwards compatible. desc is a descriptor-level annotation living in FieldOptions, so it is never serialized into any RPC payload, and every consumer references the generated symbol rather than the literal number. The only fallout is transient and cosmetic: during version skew, a mixed-version reader of a serialized FileDescriptorProto drops the annotation as an unknown field.

Also correct the usage docblock, which referenced the org package and org.E_Desc. The extension lives in flyteidl2.settings and generates as settings.E_Desc.

Customer moving to flyte2 encountererd a collision on field 50000, so we need to properly reserve a the field number since they are globally defined.  

Updating the description field number to a number inside our reserved range per the protobuf PR.  

## How was this patch tested?

run in devbox, ran edit settings from multiple versions of sdk.

## Stack


afbrock:
re-opening

afbrock:
trying to re-open this pr

PR #7127, opened 2026-03-31
refactor settings api to be simpler and more type safe

afbrock:
## Why are the changes needed?

The previous API was clunky when we went to start using it, so before it gets any adoption we are changing the interface to be simpler, easier to understand and use, and type safe.

## What changes were proposed in this pull request?

This PR removes the 'raw' unstructured APIs and instead only allows get/getForEdit/Update using the structured and typed settings message.

## How was this patch tested?

Unit tests and devbox testing with the included buf curl scripts.  (along with the cloud pr: <https://github.com/unionai/cloud/pull/15164> )

### Labels

Please add one or more of the following labels to categorize your PR:

- **added**: For new features.
- **changed**: For changes in existing functionality.
- **deprecated**: For soon-to-be-removed features.
- **removed**: For features being removed.
- **fixed**: For any bug fixed.
- **security**: In case of vulnerabilities

This is important to improve the readability of release notes.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

## Docs link

- `main` 
  - \#6583
    - **refactor settings api to be simpler and more type safe** :point\_left:

PR #7775, opened 2026-08-04
docs(rfc): Settings Service — hierarchical runtime configuration

pingsutw:
## Why

The settings API (`flyteidl2/settings/settings_service.proto`, `settings_definition.proto`) was designed, reviewed, and merged in #7127 — with generated clients for Go, TypeScript, Python, and Rust — but **no server implements it yet**. This RFC proposes implementing the `SettingsService`: instance/domain/project-scoped, strongly-typed runtime configuration (default queue, task resource guardrails, env vars, labels/annotations, service account, storage paths) with well-defined inheritance semantics.

## What this RFC covers

- **Why we need it** — today every runtime knob is either static server config (global, redeploy to change) or repeated in user code; there is no way to set per-domain defaults or enforce platform guardrails centrally.
- **When we need it** — concrete examples: per-domain env var defaults, org-wide resource ceilings, default queue routing, and `UNSET` blocking inheritance.
- **Phased implementation plan** — ~15 small, mostly independent tasks across 4 phases (persistence → resolution engine + service → applying settings on the run path → CLI/UI/docs), each with an existing in-repo pattern to copy from, so many contributors can pick up pieces in parallel. Phase 1 tasks are good first issues.

The existing walkthrough doc `flyteidl2/settings/settings_customer_flow.md` doubles as the acceptance-test spec: Phase 2 is done when every example in it reproduces against a running server.

The only proto change proposed is a single new `pod_template_name` StringSetting (restores per-project/domain PodTemplate selection, which single-namespace Flyte 2 lost); everything else is server/consumer implementation work.


Sovietaced:
This makes sense to me.

davidlin20dev:
I'm currently implementing 2.3 Validator, and wanted to check the reasoning behind the bounds.
I understand the formula: we reject 1 and only allow 0 or any number greater than 1
My question is: why would a cap of 1 deadlock? and what does 0 mean in here?

> {0} ∪ [2, MaxUint32] (a cap of 1 would deadlock any run with more than one action)

Sovietaced (approved):

Issue #7758, opened 2026-08-03
feat: add metrics for the runs-service NOTIFY pump (pending depth, notify latency, coalescing)


pingsutw:
## Why

The runs-service NOTIFY pump — a single goroutine that turns action updates into PostgreSQL
`NOTIFY` calls — has **no metrics at all**. It recently turned out to be the throughput limit for
action updates, and finding that took manual work that no dashboard could have replaced:

- Reading `pg_stat_activity` on the production database to spot the pump's dedicated connection
  (`pid 7330 | idle | ClientRead | SELECT pg_notify($1, $2)`).
- Hand-timing `pg_notify` from a pod in the same VPC to establish the ceiling: **4.1 ms** per
  round-trip statement versus **32 µs** for the call itself in a server-side loop — so the pump
  tops out near **240 notifications/second**.
- Applying Little's law to RPC metrics to show that ~78 requests were in flight while the database
  (14 of 100 connections, all `ClientRead`) and the CPU (1.34 of 4 cores) were both idle.

What the existing dashboards *did* show was only the symptom — `UpdateActionStatus` at **101.8
req/s with p95 3085 ms and mean 780 ms** — with nothing to explain it:

| RPC | rate | p50 | p95 | mean |
|---|---|---|---|---|
| `InternalRunService.UpdateActionStatus` | 101.8/s | 394 ms | 3085 ms | 780 ms |
| `InternalRunService.RecordActionEvents` | 25.6/s | 652 ms | 2418 ms | 800 ms |

Three small instruments would have pointed straight at it, and would catch a regression
automatically next time. This is also the natural way to verify the two companion fixes: batching
should collapse notify latency per item, and the non-blocking change should show drops instead of
stalls.

## What to change

Add instrumentation around the pump in `runs/repository/impl/action.go`:

- **Pending depth** (gauge) — how many notifications are waiting to be sent (`len(r.actionNotifyCh)`
  / `len(r.runNotifyCh)` today; the size of the pending set if the companion issue lands first).
  Non-zero and climbing means the pump is behind; pinned at capacity means callers are blocking.
- **Notify round-trip latency** (histogram) — around the `ExecContext` inside `execNotify`
  (line ~1070). This is the number that turned out to matter; today it is invisible.
- **Coalesced notifications** (counter) — how many redundant notifications were merged because the
  same action was already pending. This is the metric that shows the companion issue working, and
  it is the one number that says how much of the notify load is genuinely redundant.

[...]

Issue #7756, opened 2026-08-03
perf: batch pg_notify round trips in the runs-service NOTIFY pump


pingsutw:
## Why

The runs-service sends one PostgreSQL `NOTIFY` per action update, and each one costs a full
network round trip on a single dedicated connection. That serial round trip — not Postgres, not
CPU — is what limits action-update throughput today.

Measured on a dev cluster during a 100,000-action benchmark:

| RPC | rate | p50 | p95 | mean |
|---|---|---|---|---|
| `InternalRunService.UpdateActionStatus` | **101.8/s** | 394 ms | **3085 ms** | 780 ms |
| `EventsProxyService.Record` | 25.6/s | 653 ms | 2418 ms | 801 ms |
| `InternalRunService.RecordActionEvents` | 25.6/s | 652 ms | 2418 ms | 800 ms |
| `ActionsService.Enqueue` | 57.5/s | — | 1351 ms | 375 ms |
| `InternalRunService.RecordAction` | 52.9/s | 326 ms | 588 ms | 288 ms |

The mean is 780 ms, so this is the whole distribution being slow, not a tail. And it is **not**
the usual suspects:

- **Not Postgres** — RDS `WriteLatency` 0.9 ms, CPU 58–72%, and only **14 database connections
  open** against a pool max of 100. Every connection sat in `ClientRead`, i.e. the database was
  waiting on the application.
- **Not CPU** — the pod used 1.34 of its 4 cores with CFS throttling at 0.09% of periods.

By Little's law, 101.8 rps × 0.78 s ≈ **78 requests in flight** — with an idle CPU and an idle
connection pool. Those requests are blocked on something inside the process.

They are blocked on the NOTIFY pump. `pg_stat_activity` shows the single dedicated connection:

```
 pid  | state | wait_event | query
 7330 | idle  | ClientRead | SELECT pg_notify($1, $2)     <- the pump, one connection
```

Timing `pg_notify` from a pod in the same VPC shows where the cost actually is:

| | total | per notify |
|---|---|---|
| 1,000 × `pg_notify` inside one server-side `DO` loop | 32.3 ms | **32 µs** |
| 1 × `pg_notify` as its own statement | 4.1 ms | **4.1 ms** |

`pg_notify` itself costs 32 µs; the other ~4.07 ms is pure round trip. Since the pump pays that
per payload, serially, its ceiling is **~240 notifications/second on an idle database** — and it
[...]

PR #7787, opened 2026-08-05
perf(runs): don't block action writes on the NOTIFY pump

### LINKED ISSUE #7757: perf: don't block action writes on the NOTIFY pump
## Why

A PostgreSQL `NOTIFY` in the runs-service is a **wakeup hint** — the payload carries only an action
identifier, and every listener re-reads state from the database when it wakes. Losing one is
recoverable. But today, sending one can block a write RPC indefinitely.

[`notifyActionUpdate`](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L1111-L1120) does a **blocking** channel send:

```go
select {
case r.actionNotifyCh <- payload:
case <-ctx.Done():
    logger.Errorf(ctx, "Action NOTIFY send cancelled for %s: %v", payload, ctx.Err())
}
```

The channel is [buffered at **256**](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L54-L55) and drained by a single goroutine that round-trips to
Postgres once per payload — a ceiling of roughly **240 notifications/second** (measured: 4.1 ms per
`pg_notify` round trip from the same VPC, versus 32 µs for the `pg_notify` call itself). When demand
exceeds that, the buffer fills and the send blocks — **inside the RPC handler**, because
[`UpdateActionPhase` calls it](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L482-L485) right after the row is written.

### What is *not* at risk: the database write

Worth stating up front, because it is the first question everyone asks. The status update is
**already committed before the notification is attempted** — every one of the seven notify call
sites runs after its DB statement returns, and [`UpdateActionPhase`](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L429) additionally
gates on `rowsAffected > 0`:

```go
result, err := r.db.ExecContext(ctx, queryBuilder.String(), args...)  // committed here
if err != nil { return err }                                         // no notify on failure
rowsAffected, err := result.RowsAffected()
if rowsAffected > 0 {
    r.notifyActionUpdate(ctx, actionID)   // only after the row is written
}
```

These are autocommit single statements — there is no transaction spanning the write and the
notification. So blocking on the channel does **not** make the write more likely to succeed; it
only delays the RPC's return after the write already succeeded.

What *is* at risk is watcher liveness: `WatchActionUpdates` / `WatchAllActionUpdates` take their
wakeup from this notification and re-read state from the database when woken. Lose a wakeup and a
stream — and therefore the phase a user sees in the UI — can lag until something else touches that
action. So the fix must not block the write path **and** must not lose a wakeup. The design below
does both; see "What to change".

The result is that watch fan-out latency becomes write latency. Measured on a dev cluster under a
100,000-action benchmark:

| RPC | rate | p50 | p95 | mean |
|---|---|---|---|---|
| `InternalRunService.UpdateActionStatus` | **101.8/s** | 394 ms | **3085 ms** | 780 ms |
| `InternalRunService.RecordActionEvents` | 25.6/s | 652 ms | 2418 ms | 800 ms |
| `ActionsService.Enqueue` | 57.5/s | — | 1351 ms | 375 ms |

Meanwhile the database was idle (`WriteLatency` 0.9 ms, **14 connections** open out of 100, all in
`ClientRead`) and the pod used 1.34 of 4 cores with 0.09% CPU throttling. About **78 requests were
in flight** (Little's law: 101.8 rps × 0.78 s) waiting on neither CPU nor Postgres.

Batching the pump's round trips (see the companion issue) raises the ceiling, but the structural
problem remains: **a durable write should not wait on a best-effort notification.** As long as the
send can block, any future slowdown in the notify path silently becomes user-visible write latency.

## What to change

The fix must satisfy two constraints at once: **never block the write path**, and **never lose a
wakeup** — a lost wakeup means a run sits at a stale phase in the UI until something else happens
to that action.

Both are achievable, because the notification payload is an *identity*
(`project/domain/run/action`), not a state. Listeners re-read from the database when woken. So
collapsing five notifications for the same action into one is **not** a partial delivery — the
single wakeup still causes a read of the latest state. What must never be lost is the *last*
wakeup for a given action.

That points at replacing the FIFO channel with a **coalescing pending set**:

```go
// today:
actionNotifyCh chan string          // 256 FIFO slots, blocking send

// proposed:
pendingActions map[string]struct{}  // mutex-guarded; key = the payload
pendingCh      chan struct{}        // capacity 1, non-blocking "something is pending" signal
```

(Name the signal channel to match the surrounding style — `pendingCh` alongside `pendingActions`
reads as "what is pending" plus "a nudge that something is pending". The existing fields use the
`...Ch` suffix, so `wakeCh`/`signalCh` would also be fine; just avoid a bare `wake`.)

- **Writer** ([`notifyActionUpdate`](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L1111-L1120)): take the lock, insert the key, unlock, then a
  non-blocking send on `pendingCh`. It cannot block and cannot discard a key. A signal that is
  already buffered needs no second one — the map is the queue, the channel is only a nudge.
- **Pump** ([`runNotifyLoop`](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L1039)): on each signal, swap the map out under the lock, then emit
  the whole set — ideally as one batched statement, which is exactly what the companion batching
  issue does. The two changes compose.

### Delivery: keep the key until the NOTIFY succeeds

"If the writer doesn't block, how do we know the NOTIFY was actually sent?" — worth answering
directly, because **blocking does not answer it today either**. [`execNotify`](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L1062-L1076)
discards the payload on any failure:

```go
if conn == nil {
    logger.Errorf(ctx, "No NOTIFY connection available, dropping %s notification", channel)
    return                                    // payload lost
}
if _, err := conn.ExecContext(ctx, "SELECT pg_notify($1, $2)", channel, payload); err != nil {
    logger.Errorf(ctx, "Failed to NOTIFY %s: %v", channel, err)
    if isConnError(err) { reconnect() }
    // payload lost — no retry
}
```

The blocking send only guarantees that the payload reached a 256-slot buffer. Once the pump takes
it, a connection blip drops it silently. So delivery is best-effort today, and a run can already
sit at a stale phase in the UI after a transient database hiccup.

The pending set fixes this, and that is the main reason to prefer it over a bigger buffer:

1. The pump swaps the pending set out into a batch under the lock.
2. It attempts the batched `pg_notify`.
3. On success the keys are gone.
4. **On failure it merges the keys back into the pending set**, reconnects, and retries with backoff.

Step 4 is safe *because it is a set*: re-adding a key is idempotent, so a retry cannot duplicate or
double-count, and any new updates for the same action that arrived meanwhile merge into the same
entry. A FIFO would force a choice between dropping and unbounded re-queueing. The result is
**at-least-once delivery for the life of the process** — the right target, since a notification is
an idempotent wakeup and a duplicate costs one redundant re-read.

**What about a runs-service restart?** In-memory pending entries are lost, but that does not strand
the UI, because every new watch stream starts from the database rather than from notifications.
[`WatchActions`](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/service/run_service.go#L1336) subscribes first, *then* sends a full snapshot paged out of the
`actions` table, and only then streams deltas:

```go
// Start watching for updates from DB first to prevent event miss
go s.repo.ActionRepo().WatchAllActionUpdates(ctx, runID, updatesCh, errsCh)
...
if err := s.listAndSendAllActions(ctx, runID, rsm, stream); err != nil {   // full snapshot
```

So a restart breaks the streams, clients reconnect, and each reconnect re-reads authoritative state
that was committed long before any notification was attempted. Losing in-flight wakeups costs a
reconnect window, not a permanently stale run. This is also *why* the notification path can be
best-effort at all: it is an optimization over polling, and Postgres is the source of truth.

Why this is better than a bigger buffer or a drop policy:

| | today | drop when full | pending set |
|---|---|---|---|
| Blocks the write path | **yes** | no | no |
| Can lose a wakeup | yes, on `ctx.Done()` | yes, under pressure | **no** |
| Queue bounded by | 256 updates | 256 updates | distinct *actions* |

The memory bound is the key property: the set can never hold more entries than there are distinct
actions with an unsent notification, no matter how fast updates arrive. At 65,000 live actions
that is a few MB.

It also removes a loss path that exists **today**: the current `select` discards the notification
on `ctx.Done()`, so a client disconnecting mid-request can lose a wakeup that other watchers
needed. A non-blocking send has no reason to consult the request context at all.

Files:

- `runs/repository/impl/action.go` — [`notifyActionUpdate`](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L1111-L1120),
  [`notifyRunUpdate`](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L952-L961), the [channel declarations](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L40-L41) and their
  [construction](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L54-L55), [`runNotifyLoop`](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L1039) and [`drainAndExec`](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L1078-L1090).
- `runs/repository/impl/action_test.go` — cover the properties below.

Ordering note: `notifyRunUpdate` needs the same treatment, and the two sets can share one pump.

## Outcome

- [ ] `UpdateActionPhase` never blocks on notification delivery, however slow or stalled the pump is
- [ ] No wakeup is lost: for every action that was updated, at least one notification is delivered
      after its last update
- [ ] A failed `pg_notify` retries rather than dropping the payload — keys stay pending until the
      statement succeeds (strictly better than today's behaviour)
- [ ] Repeated notifications for the same action between drains collapse into one
- [ ] Request-context cancellation no longer discards a notification
- [ ] Memory is bounded by distinct pending actions rather than by update volume
- [ ] Tests cover: pump stalled → writer returns immediately and the row is still written;
      N updates to one action between drains → one notification; every updated action is
      represented in the delivered set

## Getting started

- **Reproduce:** generate ~100 action-updates/second against a dev cluster and watch
  `rpc_server_duration_milliseconds` for `UpdateActionStatus`. The
  [flyteorg/benchmark](https://github.com/flyteorg/benchmark) suite does this with
  `uv run scripts/v2/swarm.py --k 25 --n 2000`.
- **Relevant code:** [`notifyActionUpdate`](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L1111-L1120), then the pump in
  [`runNotifyLoop`](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L1039), then the listener side in [`processNotifications`](https://github.com/flyteorg/flyte/blob/a2d7e82be425372e6d94252932b31800b34992f3/runs/repository/impl/action.go#L901).
- **How to test:** `go test ./runs/repository/impl/...` — needs a local Postgres on port **15432**
  (`user/password: postgres`, db `flyte_runs_test`; see `testDbConfig` in `action_test.go`).

Related — three issues on the same bottleneck:
- #7756 batch the pump's round trips
- #7757 don't block action writes on the pump (this one)
- #7758 instrument the pump

stantheman0128:
## Tracking issue

Closes #7757

`notifyActionUpdate` and `notifyRunUpdate` did a blocking send on a 256-slot channel from inside write RPCs, right after the row was already committed. When the pump falls behind, the buffer fills and the send blocks in the RPC handler, so a best-effort wakeup becomes user-visible write latency. The issue measures that as a p95 of 3085 ms on `UpdateActionStatus` while the database itself sat idle.

There is a second, quieter problem in the same six lines. The old `select` had a `<-ctx.Done()` arm, so a client that disconnected mid-request discarded a notification that other watchers still needed.

The FIFO channels are replaced with a coalescing pending set per notification kind, plus one capacity-1 signal channel shared by both, as the issue proposes.

|  | before | after |
| --- | --- | --- |
| Blocks the write path | yes | no |
| Can lose a wakeup | yes, on `ctx.Done()` and on any `pg_notify` failure | no |
| Queue bounded by | 256 updates | distinct pending actions |

The writer takes the lock, inserts the payload as a key, unlocks, and does a non-blocking send on `pendingCh`. It cannot block and it cannot discard a key. A nudge that is already buffered needs no second one: the sets are the queue, the channel only wakes the pump.

Dropping a nudge is safe because of the ordering. The insert always happens before the signal attempt, and the pump always swaps the sets out after it has consumed a token. If the send finds the buffer full, a token was outstanding at that instant, so the pump has not yet woken for it; when it does, its swap happens after our insert and picks the key up. If the send succeeds, the same argument applies to the token we just put there. Either way a key cannot sit in the set with no pending wakeup.

The pump swaps both sets out under the lock and emits them. Payloads it could not deliver are merged back into the pending set and retried with backoff, so a transient database problem costs a delay rather than a lost wakeup. Re-adding a key is idempotent, so a retry cannot duplicate work, and updates that arrived meanwhile merge into the same entry.

### Retrying forever is its own failure mode

An earlier version of this patch retried unconditionally, which is what the issue asks for. That turns out to have a sharp edge, and I only found it because I went looking for a payload that fails for a reason other than a connection blip.

Postgres rejects a `pg_notify` payload of 8000 bytes or more, permanently. Measured against the package's own test database:

```
len=7999 err=<nil>
len=8000 err=ERROR: payload string too long (SQLSTATE 22023)
```

`isConnError` correctly says that is not a connection problem, so the connection stays up and the payload fails the same way on every attempt. With an unconditional retry the key never leaves the pending set, the success branch never runs, and the backoff pins at its 5 s ceiling. One such payload adds seconds of wakeup latency to every other watcher for the life of the process. That is worse than the old behaviour, which logged once and dropped it.

`RunIdentifier.Name` has no length validation (`identifier.pb.validate.go` says as much), and the root-action `notifyRunUpdate` in `UpdateActionPhase` is not gated on `rowsAffected`, so a caller can queue such a payload without a row existing. Reachable, not theoretical.

Two limits keep the retry honest:

A payload only spends its retry budget when the connection was healthy and it failed anyway, because that is the case where the payload itself is what Postgres rejected. Connection-level failures cost nothing, so a genuine outage still retries for as long as it lasts. After `notifyRetryLimit` payload-specific failures the key is dropped with an error log.

A drain that delivered anything resets the backoff. Otherwise one failing payload would slow every healthy notification behind it. The effect is measurable: the test covering this drops from 16.4 s to 1.0 s once the reset is in, because the poison key burns its budget at the minimum interval while ordinary traffic keeps flowing.

### Two smaller notes

A lost connection aborts the current drain. `execNotify` reconnects when `conn` is nil, so without an early exit a dead connection would make every remaining payload in a large batch trigger its own reconnect attempt, which is a storm precisely when the database is already unhappy. The undelivered payloads stay pending and the backoff handles recovery instead.

`runNotifyLoop` now takes a context rather than exiting when a channel is closed. Closing a channel that writers send on would panic the write path, and the loop still needs a stop condition the tests can drive. Production passes `context.Background()`, so the running behaviour is unchanged.

### Level of fix

Fixed at the notify layer inside `action.go`, which owns both writers and the pump, rather than at either call site. I enumerated the class of the same defect shape across `runs/`:

- `updates <- run` / `updates <- action` in the four `Watch*` methods (`action.go:707`, `:760`, `:826`, `:884`) block, but each runs in its own stream-serving goroutine. Blocking there is backpressure against a slow client and is correct. Not the same defect.
- `ch <- notif.Extra` in `processNotifications` is already non-blocking with a `default` that drops. That is the subscriber side, and #7643 is already open to instrument those drops. Left alone.
- `dedupeQueue.push` in `runs/service/abort_reconciler.go:70-88` **is the same defect and this PR does not fix it.** It has the same shape, a blocking send into a 1000-slot buffer guarded only by `<-ctx.Done()`, and `AbortRun` calls it at `run_service.go:530` after the row is already marked ABORTED at line 521. So a backed-up reconciler blocks that RPC too, and a disconnecting client drops the abort task *and* deletes its dedup key, so nothing re-pushes it. I left it alone because fixing it means touching a different subsystem and its tests in a PR scoped to the files this issue names. Happy to open a follow-up issue, or to fold it in here if you would rather see them land together.

So the write path has two members of this class and this PR fixes the notify one. I considered lifting a shared coalescing-pending-set type that both could use. The cost is a new internal package plus a rewrite of the reconciler's tests, which puts two subsystems in one review, so I would rather do it as a follow-up once the shape here has been reviewed.

Batching the pump's round trips (#7756) composes with this cleanly. The pump already hands `emit` a whole set, so batching becomes a change inside `emit` rather than a change to the structure. I have not touched instrumentation, since #7758 covers it.

## How was this patch tested?

The package brings up an embedded PostgreSQL on port 15432 by itself, so the delivery tests below issue real `pg_notify` statements and read them back through the real listener. The blocking, coalescing, and retry tests are white box and touch no database.

**Red then green.** Both new blocking tests fail on unmodified `main` at `2ad38dca6`:

```
=== RUN   TestNotifyActionUpdate_DoesNotBlockOnStalledPump
    action_test.go:921: notifyActionUpdate blocked while the notify pump was stalled
--- FAIL: TestNotifyActionUpdate_DoesNotBlockOnStalledPump (5.00s)
=== RUN   TestNotifyRunUpdate_DoesNotBlockOnStalledPump
    action_test.go:948: notifyRunUpdate blocked while the notify pump was stalled
--- FAIL: TestNotifyRunUpdate_DoesNotBlockOnStalledPump (5.00s)
```

and pass after the change, along with the rest of the new coverage:

```
--- PASS: TestNotifyActionUpdate_DoesNotBlockOnStalledPump (0.00s)
--- PASS: TestNotifyRunUpdate_DoesNotBlockOnStalledPump (0.00s)
--- PASS: TestUpdateActionPhase_CompletesWithStalledPump (0.00s)
--- PASS: TestNotifyActionUpdate_CoalescesRepeats (0.00s)
--- PASS: TestNotifyPump_CoalescesRepeatsOnTheWire (2.03s)
--- PASS: TestNotifyActionUpdate_KeepsWakeupAfterContextCancel (0.00s)
--- PASS: TestRunNotifyLoop_RetriesUndeliveredPayloads (0.00s)
--- PASS: TestRunNotifyLoop_DropsPayloadPostgresWillNeverAccept (1.00s)
--- PASS: TestNotifyPump_ConcurrentWritersDeliverEveryAction (0.20s)
--- PASS: TestWatchActionUpdates_DeliversPhaseChange (0.05s)
```

Mapping tests to the outcomes listed in the issue:

| Outcome | Test |
| --- | --- |
| Writer never blocks on a stalled pump | `TestNotifyActionUpdate_DoesNotBlockOnStalledPump`, `TestNotifyRunUpdate_DoesNotBlockOnStalledPump` |
| Pump stalled, writer returns and the row is still written | `TestUpdateActionPhase_CompletesWithStalledPump` |
| No wakeup lost, every updated action represented | `TestNotifyPump_ConcurrentWritersDeliverEveryAction` |
| Failed `pg_notify` retries rather than dropping | `TestRunNotifyLoop_RetriesUndeliveredPayloads` |
| N updates to one action produce one notification | `TestNotifyPump_CoalescesRepeatsOnTheWire` (on the wire), `TestNotifyActionUpdate_CoalescesRepeats` (in the set) |
| Context cancellation no longer discards | `TestNotifyActionUpdate_KeepsWakeupAfterContextCancel` |
| Product path still delivers | `TestWatchActionUpdates_DeliversPhaseChange` |
| Retry cannot become a permanent stall | `TestRunNotifyLoop_DropsPayloadPostgresWillNeverAccept` |

Two of these are worth a note.

`TestNotifyPump_CoalescesRepeatsOnTheWire` queues all 500 updates for one action before the pump starts, so the whole burst is guaranteed to be one drain. It then asserts one delivery arrives and that a second never does. That makes the collapse observable on the wire rather than by asserting that a Go map deduplicates.

`TestRunNotifyLoop_DropsPayloadPostgresWillNeverAccept` was the one that nearly fooled me. Its first version polled the pending set for the poison key's absence, and it passed even with the retry limit raised to 100000, because the pump briefly holds a taken batch outside the pending set and a 20 ms poll lands in that window eventually. It now requires 25 consecutive absent readings. Mutation checked both ways: with `notifyRetryLimit` raised to 100000 it fails after 15 s, and at 10 it passes in 1 s.

**Race detector**, since this is a concurrency change:

```
$ go test -race -count=2 ./runs/...
ok      github.com/flyteorg/flyte/v2/runs/config                  1.025s
ok      github.com/flyteorg/flyte/v2/runs/repository/impl        29.421s
ok      github.com/flyteorg/flyte/v2/runs/repository/transformers 1.025s
ok      github.com/flyteorg/flyte/v2/runs/scheduler/core          1.620s
ok      github.com/flyteorg/flyte/v2/runs/service                13.954s
ok      github.com/flyteorg/flyte/v2/runs/test/api               13.714s
```

**Full tree** without the race detector:

```
$ go test -count=1 ./runs/...
ok      github.com/flyteorg/flyte/v2/runs/config                  0.718s
ok      github.com/flyteorg/flyte/v2/runs/repository/impl        21.284s
ok      github.com/flyteorg/flyte/v2/runs/repository/transformers 1.365s
ok      github.com/flyteorg/flyte/v2/runs/scheduler/core          1.948s
ok      github.com/flyteorg/flyte/v2/runs/service                12.511s
ok      github.com/flyteorg/flyte/v2/runs/test/api               12.527s
```

For a baseline I ran `./runs/repository/impl/...` on a clean checkout of `2ad38dca6` before touching anything: `ok ... 25.838s`. There were no pre-existing failures in this package to work around, so every red result quoted above came from the change itself.

### Labels

- [ ] I updated the documentation accordingly. No user-facing documentation covers this internal path; the reasoning lives in the code comments.

### What was not tested

The 100,000-action benchmark from `flyteorg/benchmark` that produced the p95 numbers in the issue. I have no dev cluster to run `swarm.py` against, so the latency improvement is argued from the structure and from the concurrency test rather than re-measured end to end. If someone with a cluster reruns it, the number to watch is `rpc_server_duration_milliseconds` for `UpdateActionStatus`.

A real connection loss or failover. `TestRunNotifyLoop_RetriesUndeliveredPayloads` uses a nil connection with no database to reconnect to, which exercises the same branch, but it is not the same as pulling the plug on a live Postgres mid-drain.

The pending set still has no upper bound, which is what the issue proposes and I did not change. During a long outage it is bounded by the number of distinct actions being updated rather than by anything the process controls, whereas the old 256-slot buffer at least applied backpressure. At the issue's own figure of 65,000 live actions that is a few MB, so I left it alone, but if you would rather have a cap plus a dropped-notification counter I am happy to add it, and #7758 looks like the natural home for the counter.

I did not check whether `InternalRunService.UpdateActionStatus` is reachable by anything other than the in-cluster operator, so I cannot say whether the oversized-payload path above is only an operational footgun or something a caller could aim. The retry limit bounds the damage either way.

## Related PRs

- #7756 batch the pump's round trips, which composes with this
- #7758 instrument the pump
- #7643 counter for dropped subscriber updates, the other side of the notification path

---

_This change was developed with AI assistance from Claude. I reviewed and verified it before submitting._

### DIFF
--- a/runs/repository/impl/action.go
+++ b/runs/repository/impl/action.go
@@ -36,10 +36,36 @@ type actionRepo struct {
 	actionSubscribers map[chan string]bool
 	mu                sync.RWMutex
 
-	// Dedicated channels for async NOTIFY to avoid pool contention
-	actionNotifyCh chan string
-	runNotifyCh    chan string
+	// Pending NOTIFY payloads, drained by a dedicated pump so that write RPCs
+	// never wait on a notification. A payload is an identity rather than a
+	// state and every listener re-reads from the database when it wakes, so
+	// repeated updates for the same key collapse into a single delivery. That
+	// makes a set the right structure: it bounds memory by the number of
+	// distinct pending actions instead of by update volume, and it lets a
+	// failed pg_notify be retried by merging the keys back in.
+	// The value is the number of times this payload has failed for a reason
+	// that was not the connection's fault, which bounds how long a payload
+	// Postgres will never accept can hold up everything else.
+	notifyMu       sync.Mutex
+	pendingActions map[string]int
+	pendingRuns    map[string]int
+	// pendingCh carries a single "something is pending" nudge. The sets above
+	// are the queue; this channel only wakes the pump.
+	pendingCh chan struct{}
 }
+
+const (
+	// Backoff bounds for retrying a pg_notify that failed, so a broken
+	// connection cannot turn the pump into a hot loop.
+	notifyRetryMinBackoff = 50 * time.Millisecond
+	notifyRetryMaxBackoff = 5 * time.Second
+	// notifyRetryLimit bounds retries of a payload that keeps failing while
+	// the connection is healthy, which means the payload itself is the
+	// problem. pg_notify rejects payloads of 8000 bytes or more permanently,
+	// for example. Retrying such a payload forever would hold every other
+	// watcher's wakeup behind it, which is worse than losing one hint.
+	notifyRetryLimit = 10
+)
 
 // NewActionRepo creates a new PostgreSQL repository
 func NewActionRepo(db *sqlx.DB, dbConfig database.DbConfig) (interfaces.ActionRepo, error) {
@@ -51,8 +77,9 @@ func NewActionRepo(db *sqlx.DB, dbConfig database.DbConfig) (interfaces.ActionRe
 		actionSubscribers: make(map[chan string]bool),
 	}
 
-	repo.actionNotifyCh = make(chan string, 256)
-	repo.runNotifyCh = make(chan string, 256)
+	repo.pendingActions = make(map[string]int)
+	repo.pendingRuns = make(map[string]int)
+	repo.pendingCh = make(chan struct{}, 1)
 
 	if err := repo.startPostgresListener(); err != nil {
 		return nil, fmt.Errorf("failed to start postgres listener: %w", err)
@@ -947,16 +974,84 @@ func (r *actionRepo) processNotifications() {
 	}
 }
 
-// notifyRunUpdate sends a notification about a run update via the
-// dedicated notify channel, avoiding connection pool contention.
-func (r *actionRepo) notifyRunUpdate(ctx context.Context, runID *common.RunIdentifier) {
+// notifyRunUpdate queues a notification about a run update for the dedicated
+// notify pump, avoiding connection pool contention.
+//
+// ctx is deliberately not consulted: the row is already committed by the time
+// this runs, and dropping the wakeup because the caller's request was
+// cancelled would leave other watchers looking at a stale phase.
+func (r *actionRepo) notifyRunUpdate(_ context.Context, runID *common.RunIdentifier) {
 	payload := fmt.Sprintf("%s/%s/%s", runID.Project, runID.Domain, runID.Name)
 
-	select {
-	case r.runNotifyCh <- payload:
-	case <-ctx.Done():
-		logger.Warnf(ctx, "Run NOTIFY send cancelled for %s: %v", payload, ctx.Err())
+	r.notifyMu.Lock()
+	markPending(r.pendingRuns, payload)
+	r.notifyMu.Unlock()
+
+	r.signalPending()
+}
+
+// markPending queues payload, keeping any attempt count it has already
+// accumulated. Resetting the count here would let a payload that can never be
+// delivered stay pending forever, since a busy action keeps re-queuing it.
+func markPending(pending map[string]int, payload string) {
+	if _, queued := pending[payload]; !queued {
+		pending[payload] = 0
 	}
+}
+
+// signalPending nudges the notify pump without ever blocking. A nudge that is
+// already buffered needs no second one: the pump re-reads the whole pending
+// set once it wakes, so one outstanding signal covers any number of keys.
+func (r *actionRepo) signalPending() {
+	select {
+	case r.pendingCh <- struct{}{}:
+	default:
+	}
+}
+
+// takePendingNotifications swaps the pending sets out for empty ones and
+// returns what was queued.
+func (r *actionRepo) takePendingNotifications() (actions, runs map[string]int) {
+	r.notifyMu.Lock()
+	defer r.notifyMu.Unlock()
+
+	actions, runs = r.pendingActions, r.pendingRuns
+	r.pendingActions = make(map[string]int)
+	r.pendingRuns = make(map[string]int)
+	return actions, runs
+}
+
+// requeuePendingNotifications merges undelivered payloads back so the pump can
+// retry them. Re-adding a key is idempotent, so a retry cannot duplicate work,
+// and updates that arrived meanwhile merge into the same entry.
+func (r *actionRepo) requeuePendingNotifications(actions, runs map[string]int) {
+	if len(actions) == 0 && len(runs) == 0 {
+		return
+	}
+
+	r.notifyMu.Lock()
+	r.pendingActions = mergePending(r.pendingActions, actions)
+	r.pendingRuns = mergePending(r.pendingRuns, runs)
+	r.notifyMu.Unlock()
+
+	r.signalPending()
+}
+
+// mergePending folds src into dst and returns the map to keep. It copies the
+// smaller side so that requeuing a large failed batch into a nearly empty
+// pending set holds the lock for the size of the smaller map, not the batch.
+// Where both sides hold a payload the higher attempt count wins, so a new
+// update arriving for a failing action cannot refill its retry budget.
+func mergePending(dst, src map[string]int) map[string]int {
+	if len(src) > len(dst) {
+		dst, src = src, dst
+	}
+	for payload, attempts := range src {
+		if existing, queued := dst[payload]; !queued || attempts > existing {
+			dst[payload] = attempts
+		}
+	}
+	return dst
 }
 
 // ListRootActions lists root actions (runs) matching scope and date filters.
@@ -1012,7 +1107,7 @@ func (r *actionRepo) startNotifyLoop() error {
 		return fmt.Errorf("acquire dedicated NOTIFY connection: %w", err)
 	}
 
-	go r.runNotifyLoop(sqlDB, conn)
+	go r.runNotifyLoop(context.Background(), sqlDB, conn)
 	return nil
 }
 
@@ -1033,10 +1128,13 @@ func isConnError(err error) bool {
 	return false
 }
 
-// runNotifyLoop processes notify channels using the given connection.
-// On connection errors it attempts to reconnect; if reconnection fails
-// the error is logged and the notification is skipped.
-func (r *actionRepo) runNotifyLoop(sqlDB *sql.DB, conn *sql.Conn) {
+// runNotifyLoop drains the pending notification sets using the given
+// connection. On connection errors it attempts to reconnect; payloads it could
+// not deliver stay pending and are retried with backoff, so a transient
+// database problem costs a delay rather than a lost wakeup.
+//
+// It returns when ctx is done.
+func (r *actionRepo) runNotifyLoop(ctx context.Context, sqlDB *sql.DB, conn *sql.Conn) {
 	defer func() {
 		if conn != nil {
 			conn.Close() //nolint:errcheck
@@ -1052,69 +1150,126 @@ func (r *actionRepo) runNotifyLoop(sqlDB *sql.DB, conn *sql.Conn) {
 			return
 		}
 		var err error
-		conn, err = sqlDB.Conn(context.Background())
+		conn, err = sqlDB.Conn(ctx)
 		if err != nil {
-			logger.Errorf(context.Background(), "Failed to re-acquire NOTIFY connection: %v", err)
+			logger.Errorf(ctx, "Failed to re-acquire NOTIFY connection: %v", err)
 			conn = nil
 		}
 	}
 
-	execNotify := func(channel, payload string) {
+	// execNotify reports whether the payload was delivered. A false result
+	// keeps the payload pending for the next attempt.
+	execNotify := func(channel, payload string) bool {
 		if conn == nil {
 			reconnect()
 		}
 		if conn == nil {
-			logger.Errorf(context.Background(), "No NOTIFY connection available, dropping %s notification", channel)
-			return
+			logger.Errorf(ctx, "No NOTIFY connection available, keeping %s notification pending", channel)
+			return false
 		}
-		if _, err := conn.ExecContext(context.Background(), "SELECT pg_notify($1, $2)", channel, payload); err != nil {
-			logger.Errorf(context.Background(), "Failed to NOTIFY %s: %v", channel, err)
+		if _, err := conn.ExecContext(ctx, "SELECT pg_notify($1, $2)", channel, payload); err != nil {
+			logger.Errorf(ctx, "Failed to NOTIFY %s: %v", channel, err)
 			if isConnError(err) {
 				reconnect()
 			}
+			return false
 		}
+		return true
 	}
 
-	drainAndExec := func(channel, firstPayload string, ch <-chan string) {
-		execNotify(channel, firstPayload)
-		for {
-			select {
-			case payload, ok := <-ch:
-				if !ok {
-					return
-				}
-				execNotify(channel, payload)
-			default:
-				return
+	// emit delivers every payload in the set. It returns the payloads worth
+	// another attempt, with their attempt counts advanced, and how many were
+	// delivered. A payload only spends its retry budget when the connection
+	// was healthy and it failed anyway, because that means the payload itself
+	// is what Postgres rejected.
+	emit := func(channel string, payloads map[string]int) (retry map[string]int, delivered int) {
+		keep := func(payload string, attempts int) {
+			if retry == nil {
+				retry = make(map[string]int, len(payloads))
 			}
+			retry[payload] = attempts
 		}
+
+		connectionLost := false
+		for payload, attempts := range payloads {
+			// Once the connection is gone, trying the rest would just make
+			// every remaining payload attempt its own reconnect. Keep them
+			// and let the backoff below handle recovery.
+			if connectionLost {
+				keep(payload, attempts)
+				continue
+			}
+			if execNotify(channel, payload) {
+				delivered++
+				continue
+			}
+			if conn == nil {
+				connectionLost = true
+				keep(payload, attempts)
+				continue
+			}
+
+			attempts++
+			if attempts >= notifyRetryLimit {
+				logger.Errorf(ctx, "Dropping %s notification after %d failed attempts: %s",
+					channel, attempts, payload)
+				continue
+			}
+			keep(payload, attempts)
+		}
+		return retry, delivered
 	}
 
+	backoff := notifyRetryMinBackoff
 	for {
 		select {
-		case payload, ok := <-r.actionNotifyCh:
-			if !ok {
-				return
-			}
-			drainAndExec("action_updates", payload, r.actionNotifyCh)
-		case payload, ok := <-r.runNotifyCh:
-			if !ok {
-				return
-			}
-			drainAndExec("run_updates", payload, r.runNotifyCh)
+		case <-ctx.Done():
+			return
+		case <-r.pendingCh:
+		}
+
+		actions, runs := r.takePendingNotifications()
+		retryActions, deliveredActions := emit("action_updates", actions)
+		retryRuns, deliveredRuns := emit("run_updates", runs)
+
+		if len(retryActions) == 0 && len(retryRuns) == 0 {
+			backoff = notifyRetryMinBackoff
+			continue
+		}
+
+		r.requeuePendingNotifications(retryActions, retryRuns)
+
+		// Delivering anything at all means the pump is healthy and something
+		// payload-specific failed, so do not slow every other watcher down.
+		if deliveredActions+deliveredRuns > 0 {
+			backoff = notifyRetryMinBackoff
+		}
+
+		// Wait before the retry so a broken connection cannot spin the pump.
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(backoff):
+		}
+		backoff *= 2
+		if backoff > notifyRetryMaxBackoff {
+			backoff = notifyRetryMaxBackoff
 		}
 	}
 }
 
-// notifyActionUpdate sends a notification about an action update via the
-// dedicated notify channel, avoiding connection pool contention.
-func (r *actionRepo) notifyActionUpdate(ctx context.Context, actionID *common.ActionIdentifier) {
+// notifyActionUpdate queues a notification about an action update for the
+// dedicated notify pump, avoiding connection pool contention.
+//
+// This is called from inside write RPCs, right after the row is committed, so
+// it must never block. ctx is deliberately not consulted; see notifyRunUpdate.
+func (r *actionRepo) notifyActionUpdate(_ context.Context, actionID *common.ActionIdentifier) {
 	payload := fmt.Sprintf("%s/%s/%s/%s",
 		actionID.Run.Project, actionID.Run.Domain, actionID.Run.Name, actionID.Name)
 
-	select {
-	case r.actionNotifyCh <- payload:
-	case <-ctx.Done():
-		logger.Errorf(ctx, "Action NOTIFY send cancelled for %s: %v", payload, ctx.Err())
-	}
+	r.notifyMu.Lock()
+	markPending(r.pendingActions, payload)
+	r.notifyMu.Unlock()
+
+	r.signalPending()
 }
--- a/runs/repository/impl/action_test.go
+++ b/runs/repository/impl/action_test.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"net"
 	"sort"
+	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -821,11 +823,7 @@ func TestInsertEvents_Empty(t *testing.T) {
 }
 
 func TestNotifyActionUpdate_PayloadWithSpecialChars(t *testing.T) {
-	r := &actionRepo{
-
-		actionNotifyCh: make(chan string, 256),
-		runNotifyCh:    make(chan string, 256),
-	}
+	r := newNotifyTestRepo()
 
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
@@ -843,20 +841,12 @@ func TestNotifyActionUpdate_PayloadWithSpecialChars(t *testing.T) {
 
 	r.notifyActionUpdate(ctx, actionID)
 
-	select {
-	case payload := <-r.actionNotifyCh:
-		assert.Equal(t, "proj/domain/run'; DROP TABLE actions; --/action", payload)
-	case <-ctx.Done():
-		t.Fatal("timed out waiting for payload on actionNotifyCh")
-	}
+	actions, _ := r.takePendingNotifications()
+	assert.Contains(t, actions, "proj/domain/run'; DROP TABLE actions; --/action")
 }
 
 func TestNotifyRunUpdate_PayloadWithSpecialChars(t *testing.T) {
-	r := &actionRepo{
-
-		actionNotifyCh: make(chan string, 256),
-		runNotifyCh:    make(chan string, 256),
-	}
+	r := newNotifyTestRepo()
 
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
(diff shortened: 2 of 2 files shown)


stantheman0128:
Genuine contributor — happy to trim the PR description if maintainers prefer a shorter body. The automated flag looks driven by description length, fork rate, and the username heuristic on stantheman0128 rather than the code change itself. Fixes #7757; tests and CI are green.

pingsutw:
> The 100,000-action benchmark from flyteorg/benchmark that produced the p95 numbers in the issue. I have no dev cluster to run swarm.py against, so the latency improvement is argued from the structure and from the concurrency test rather than re-measured end to end. If someone with a cluster reruns it, the number to watch is rpc_server_duration_milliseconds for UpdateActionStatus.


Could you try to test it in the debbox? I added a make target for installing grafana dashboard in the devbox
https://github.com/flyteorg/flyte/blob/fd2a82aa2a959f373f5f3441e1328940c641e42f/Makefile#L53-L54

stantheman0128:
Thanks, that target is exactly what I was missing. I will bring the devbox up with the Grafana stack and report the before and after for `rpc_server_duration_milliseconds` on UpdateActionStatus, running the swarm against master and this branch.

Will follow up here with the numbers.

stantheman0128:
Following up on the devbox + Grafana request.

## Setup

- Devbox from this branch (`make devbox-run`, `make devbox-monitoring`)
- Grafana: `http://localhost:30300/d/oss/flyte-execution`
- Workload: `flyteorg/flyte-benchmarks` `scripts/v2/swarm.py --k 50 --n 2000` (100k actions, same shape as the issue)
- Metric: Prometheus `rpc_server_duration_milliseconds` for `UpdateActionStatus`
- Query (example): `histogram_quantile(0.95, sum(rpc_server_duration_milliseconds_bucket{rpc_method=UpdateActionStatus}) by (le))`

## This branch (commit `90fafa19f`, fix/7757-notify-pending-set)

After a full swarm on the devbox with this PR's binary:

| Stat | UpdateActionStatus |
|------|-------------------|
| samples | 57,630 |
| p50 | 9.7 ms |
| p95 | 149.3 ms |
| p99 | 1205.8 ms |

## Main baseline

I imported `origin/main` (`41e562c57`) flyte-binary into the same devbox image slot and tried to rerun the identical swarm for a paired measurement. On a freshly recreated devbox volume I hit repeated `rustfs-svc.flyte:9000` connection refused errors during run submit (inputs upload), so I do not have a clean main histogram from this session to put beside the branch numbers.

For main comparison I am still using the table from #7757 (same workload class, dev cluster): p95 **3085 ms** at ~101.8 rps on `UpdateActionStatus`.

## Takeaway

On the branch side, the devbox run moved p95 from the issue's ~3s main figure down to ~150 ms under swarm load, which matches the non-blocking notify intent. Happy to rerun the main side on request once devbox storage is stable on my box, or if you have a CI devbox run from `devbox-functional-tests` with metrics exported.

Commands and raw query output from the branch run are in the PR branch worktree notes if useful.

runs/repository/impl/action.go
```diff
@@ -947,18 +974,86 @@ func (r *actionRepo) processNotifications() {
...
+	markPending(r.pendingRuns, payload)
+	r.notifyMu.Unlock()
+
+	r.signalPending()
+}
+
+// markPending queues payload, keeping any attempt count it has already
+// accumulated. Resetting the count here would let a payload that can never be
+// delivered stay pending forever, since a busy action keeps re-queuing it.
+func markPending(pending map[string]int, payload string) {
```
0yukali0, line 996:
```golang
  func (r *actionRepo) markRunPending(payload string) {
  	r.notifyMu.Lock()
  	defer r.notifyMu.Unlock()

  	markPending(r.pendingRuns, payload)
  }

  func (r *actionRepo) markActionPending(payload string) {
  	r.notifyMu.Lock()
  	defer r.notifyMu.Unlock()

  	markPending(r.pendingActions, payload)
  }

  func markPending(pending map[string]int, payload string) {
  	if _, queued := pending[payload]; !queued {
  		pending[payload] = 0
  	}
  }
```

runs/repository/impl/action.go
```diff
@@ -947,18 +974,86 @@ func (r *actionRepo) processNotifications() {
...
+// notifyRunUpdate queues a notification about a run update for the dedicated
+// notify pump, avoiding connection pool contention.
+//
+// ctx is deliberately not consulted: the row is already committed by the time
+// this runs, and dropping the wakeup because the caller's request was
+// cancelled would leave other watchers looking at a stale phase.
+func (r *actionRepo) notifyRunUpdate(_ context.Context, runID *common.RunIdentifier) {
 	payload := fmt.Sprintf("%s/%s/%s", runID.Project, runID.Domain, runID.Name)
 
+	r.notifyMu.Lock()
```
0yukali0, line 986: r.markRunPending(payload)

runs/repository/impl/action.go
```diff
@@ -1052,69 +1150,126 @@ func (r *actionRepo) runNotifyLoop(sqlDB *sql.DB, conn *sql.Conn) {
...
+func (r *actionRepo) notifyActionUpdate(_ context.Context, actionID *common.ActionIdentifier) {
 	payload := fmt.Sprintf("%s/%s/%s/%s",
 		actionID.Run.Project, actionID.Run.Domain, actionID.Run.Name, actionID.Name)
 
-	select {
-	case r.actionNotifyCh <- payload:
-	case <-ctx.Done():
-		logger.Errorf(ctx, "Action NOTIFY send cancelled for %s: %v", payload, ctx.Err())
-	}
+	r.notifyMu.Lock()
```
0yukali0, line 1270: r.markActionPending(payload)

runs/repository/impl/action.go
```diff
@@ -947,18 +974,86 @@ func (r *actionRepo) processNotifications() {
...
+
+// requeuePendingNotifications merges undelivered payloads back so the pump can
+// retry them. Re-adding a key is idempotent, so a retry cannot duplicate work,
+// and updates that arrived meanwhile merge into the same entry.
+func (r *actionRepo) requeuePendingNotifications(actions, runs map[string]int) {
+	if len(actions) == 0 && len(runs) == 0 {
+		return
+	}
+
+	r.notifyMu.Lock()
```
0yukali0, line 1032: r.mergePendingActions(actions)
r.mergePendingRuns(runs)

runs/repository/impl/action.go
```diff
@@ -947,18 +974,86 @@ func (r *actionRepo) processNotifications() {
...
+
+	r.signalPending()
+}
+
+// mergePending folds src into dst and returns the map to keep. It copies the
+// smaller side so that requeuing a large failed batch into a nearly empty
+// pending set holds the lock for the size of the smaller map, not the batch.
+// Where both sides hold a payload the higher attempt count wins, so a new
+// update arriving for a failing action cannot refill its retry budget.
+func mergePending(dst, src map[string]int) map[string]int {
```
0yukali0, line 1045:
```golang
func (r *actionRepo) mergePendingActions(src map[string]int) {
  	if len(src) == 0 {
  		return
  	}

  	r.notifyMu.Lock()
  	defer r.notifyMu.Unlock()

  	r.pendingActions = mergePending(r.pendingActions, src)
  }

  func (r *actionRepo) mergePendingRuns(src map[string]int) {
  	if len(src) == 0 {
  		return
  	}

  	r.notifyMu.Lock()
  	defer r.notifyMu.Unlock()

  	r.pendingRuns = mergePending(r.pendingRuns, src)
  }

  func mergePending(dst, src map[string]int) map[string]int {
  	if len(src) > len(dst) {
  		dst, src = src, dst
  	}

  	for payload, attempts := range src {
  		if existing, queued := dst[payload]; !queued || attempts > existing {
  			dst[payload] = attempts
  		}
  	}

  	return dst
  }
```

runs/repository/impl/action.go
```diff
@@ -1052,69 +1150,126 @@ func (r *actionRepo) runNotifyLoop(sqlDB *sql.DB, conn *sql.Conn) {
...
+				connectionLost = true
+				keep(payload, attempts)
+				continue
+			}
+
+			attempts++
+			if attempts >= notifyRetryLimit {
+				logger.Errorf(ctx, "Dropping %s notification after %d failed attempts: %s",
+					channel, attempts, payload)
+				continue
```
0yukali0, line 1216: I’d prefer to keep failed notifications pending until delivery succeeds so that connected clients can eventually fetch the latest status.

0yukali0 (commented):
From an eventual-consistency perspective, I don’t think notifications should be dropped.

0yukali0:
I ran the tests and `tc qdisc replace dev eth0 root netem delay 15ms` the Postgres network interface.
I got the following results.

le of InternalRunService.UpdateActionStatus and InternalRunService.RecordActionEvents decreases but p95 of ActionsService.Enqueue increases.

Screenshot 2026-08-29 at 2 00 11 AM
Screenshot 2026-08-29 at 2 00 55 AM
Screenshot 2026-08-29 at 2 01 44 AM

stantheman0128:
@0yukali0 sorry for the late reply.

On dropping notifications: a wakeup that can still be delivered already stays pending. The attempt counter only moves when `pg_notify` fails and the connection is still healthy, and we drop only after 10 of those. The case that forced the limit is a payload Postgres rejects forever (`pg_notify` is capped at 8000 bytes). Keeping that key pending forever puts every other watcher behind a 5s backoff.

So the question is whether you want that poison case dropped, or kept with some other isolation so it cannot stall the rest of the set. I would rather not remove the limit without that. The original bug was write RPCs stuck on notify, and an unbounded retry of a dead payload recreates a milder version of the same stall. Connection-loss already keeps the payload and does not spend the budget.

On Enqueue p95 rising under `tc netem delay 15ms`: that matches the design. `UpdateActionStatus` / `RecordActionEvents` no longer wait for the pump, so their p95 should fall. Enqueue is on the other side of the wakeup, so it sees whatever delay the coalesced pump plus backoff introduces. I do not have a second run that isolates how much of your Enqueue rise is the cap-1 nudge versus the 15ms delay on Postgres itself.

If the Enqueue rise is a blocker, I would look at skipping backoff when the failed payload is new, or at isolating poison keys so a healthy batch is not delayed. I do not want to promise another paired swarm from this box. The last main-side run died on rustfs, and the numbers were not repeatable.

The `markPending` / `mergePending` helpers you sketched are fine as methods if you would rather have them that way. They do not change the drop or Enqueue questions.

PR #7797, opened 2026-08-06
fix: allow paused actions to reach terminal phases

anxkhn:
## Tracking Issue

NA

`actionRepo.UpdateActionPhase` guards writes with `phase <= $requested OR phase = ANY(retryablePhases)`, where the retryable set is `FAILED` and `TIMED_OUT`. `ACTION_PHASE_PAUSED` is enum value 9, while `ACTION_PHASE_SUCCEEDED` is 5 and `ACTION_PHASE_TIMED_OUT` is 8, so a paused condition action can never leave `PAUSED`: both the signal path and the timeout path fail the predicate. The statement updates zero rows, the repository still returns `nil`, and `UpdateActionStatus` reports success while the actions informer goes on to persist the signal output against a row that is still `PAUSED`.

The practical effect is that human-in-the-loop conditions stay listed and filtered as awaiting input after they have been answered or timed out, no end time is ever recorded, and historical action details plus paused-run listings are permanently wrong.

Reproduction: persist a condition action, move it `QUEUED -> PAUSED`, then call `UpdateActionStatus` with `SUCCEEDED` or `TIMED_OUT`. The RPC returns OK, but `GetAction` still returns `PAUSED`.

- Replace the ordinal phase comparison in `UpdateActionPhase` with an explicit map of allowed source phases per target phase. Forward progress through the running phases and the existing retry-from-`FAILED`/`TIMED_OUT` behavior are preserved, `PAUSED` becomes a valid source for `SUCCEEDED`, `TIMED_OUT`, `ABORTED`, `FAILED` and the retry phases, and completed actions can no longer be moved backwards.
- Stop reporting a rejected update as success. When the statement matches no row, the repository checks whether the action exists and returns `ErrPhaseTransitionRejected` (guard rejected the transition) or `sql.ErrNoRows` (no such action) instead of `nil`.
- `updateSingleActionStatus` treats `ErrPhaseTransitionRejected` as a no-op and returns before writing the output and run info, so a stale or out-of-order event can no longer overwrite the output of an action it was not allowed to transition.
- Tests: repository coverage for `PAUSED -> SUCCEEDED/TIMED_OUT/ABORTED` including the persisted end time, tightened assertions on the existing backwards-transition tests, and a condition service test asserting that a rejected transition does not persist the signal output.

### Note for reviewers

The explicit table also rejects a few transitions that the ordinal check happened to permit, notably `SUCCEEDED -> FAILED/ABORTED/TIMED_OUT` and moving into `RECOVERED` from a terminal phase. That follows from treating completed actions as final, but if any of those are relied on in practice I am happy to add the specific entries back.

## How was this patch tested?

`go test ./runs/repository/impl ./runs/service`

The new repository test drives the real `QUEUED -> PAUSED -> terminal` sequence against the test database and asserts both the stored phase and that `ended_at` is set. The service test uses the mocked action repo to assert that a rejected transition short-circuits before the output write and still returns an OK status to the caller.

## Setup process

No setup or migration changes; the fix is confined to the update predicate and its error handling.

## Screenshots

NA

- [x] To the best of my knowledge, the proposed patch does not introduce new security risks.
- [ ] Documentation updates are needed (no user-facing docs cover this guard).


pingsutw:
Do you have an example that I can repro?

anxkhn:
@pingsutw thanks for the review. i also addressed the remaining findings: paused actions can enter retry phases and failed, missing actions return not found, and the transition matrix has direct coverage. the minimal repro is paused followed by succeeded, failed, or timed_out; previously the rpc returned ok while the stored phase remained paused. rebased onto the latest main and the focused repository and service tests pass.

PR #7798, opened 2026-08-06
fix: honor cron trigger timezone when scheduling runs

anxkhn:

The structured `Cron` schedule carries a `timezone`, and trigger validation already prepends `CRON_TZ=` before parsing it. The scheduler does not: `ParseSchedule` passes only `Cron.expression` to `cron.ParseStandard`, and the cron runner defaults to UTC. Every accepted non-UTC trigger therefore executes in UTC.

Concretely, a trigger deployed as `0 9 * * *` with timezone `America/Los_Angeles` passes validation but fires at 09:00 UTC instead of 09:00 Pacific. Users get runs several hours early or late, the offset shifts again across daylight-saving transitions, and catch-up time calculations inherit the same error. `CRON_TZ` is the per-schedule timezone override documented by robfig/cron, so the information is available, it is just dropped on the runtime path.

- Add a shared `ParseCron` helper that handles both schedule forms: the structured `Cron` message (applying `CRON_TZ=<timezone>` when a timezone is set) and the legacy `cron_expression` string.
- Use it in `ParseSchedule` so live scheduling and catch-up calculations respect the declared timezone.
- Use it in `validateCronExpression` so deploy-time validation and runtime parsing can no longer disagree about what a schedule means.

One behavior change worth calling out: because validation now goes through the same parser, an invalid legacy `cron_expression` is rejected at registration. Previously only the structured `Cron` form was validated, and a bad legacy expression was accepted and then failed later inside the scheduler.

## How was this patch tested?

New table tests in the scheduler core cover UTC, a non-UTC zone, and a DST boundary (7-8 March 2026 in `America/Los_Angeles`), asserting both the next fire time and the catch-up times. New validation tests cover structured cron with a timezone, legacy cron, and invalid expressions in both forms.

```
go test ./runs/service ./runs/scheduler/core
```

## Related PRs

None.

## Docs link

None.

### DIFF
--- a/runs/schedule/cron.go
+++ b/runs/schedule/cron.go
@@ -1,0 +1,35 @@
+package schedule
+
+import (
+	"fmt"
+
+	"github.com/robfig/cron/v3"
+
+	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/task"
+)
+
+// ParseCron parses either form of cron schedule, including its timezone.
+func ParseCron(schedule *task.Schedule) (cron.Schedule, error) {
+	if schedule == nil {
+		return nil, nil
+	}
+
+	var expression string
+	switch value := schedule.GetExpression().(type) {
+	case *task.Schedule_CronExpression:
+		expression = value.CronExpression
+	case *task.Schedule_Cron:
+		expression = value.Cron.GetExpression()
+		if timezone := value.Cron.GetTimezone(); timezone != "" {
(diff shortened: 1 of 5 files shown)


pingsutw (approved):
Thanks for fixing it

anxkhn:
updated cron validation errors to report the raw expression and timezone, and removed the duplicate invalid cron wording while preserving trigger context. rebased onto the latest main; the schedule, scheduler, and service tests pass. a maintainer review would be appreciated.

Issue #7772, opened 2026-08-04
[Ray Plugin] Make SubmissionMode configurable


evan-sadler:
Using `HttpMode` avoids an extra pod and potential pod eviction on k8s taking down the entire cluster. We found it to be a bit more reliable. If it is easy to add - that would be great.

davidlin20dev:
#take

PR #7799, opened 2026-08-06
[V2][Ray] Make KubeRay submission mode configurable

davidlin20dev:

Adds a `submissionMode` setting to the Ray plugin configuration, letting platform operators choose how the KubeRay operator submits Ray jobs:

- **`K8sJobMode`** (default): a submitter Kubernetes Job runs `ray job submit` against the RayCluster. This is the existing behavior and remains the default.
- **`HTTPMode`**: the KubeRay operator submits the job via an HTTP request directly to the Ray head node. No submitter pod is created, so job failures caused by submitter pod eviction are eliminated.
- **`SidecarMode`**: the KubeRay operator injects the submitter as a sidecar container in the Ray head pod. The submitter shares the head pod's fate, so it cannot be evicted independently of the job. Requires the head pod's `restartPolicy` to be `Never`, which Flyte already sets by default.

```yaml
plugins:
  ray:
    submissionMode: HTTPMode
```

In `HTTPMode` and `SidecarMode`, the plugin skips building the submitter pod template, since no separate submitter pod exists. An invalid `submissionMode` value fails task construction with an explicit error rather than silently falling back.

In `K8sJobMode`, the submitter pod can be evicted independently of the Ray cluster (e.g., during node drains), causing KubeRay to mark an otherwise healthy job as failed. `HTTPMode` and `SidecarMode` both remove the standalone submitter pod from the picture. See #7772.

Per review discussion, the submission mode is a platform-level reliability concern, so it is exposed via plugin configuration (decided by the platform team) rather than per-task API. A per-task proto override can be added in the future if needed.

## How was this patch tested?

- Unit tests cover: explicit `K8sJobMode`, unset config (defaults to `K8sJobMode`), `HTTPMode` (asserts no submitter pod template is set), `SidecarMode` (asserts no submitter pod template is set), and an invalid value (asserts a hard error).
- `config_flags.go` regenerated with the `pflags` generator.
- Full `flyteplugins/.../k8s/ray` package test suite passes locally.

## Labels

Related to #7772


flyteidl2/plugins/ray.proto
```diff
@@ -6,6 +6,17 @@ import "flyteidl2/core/tasks.proto";
 
 option go_package = "github.com/flyteorg/flyte/v2/gen/go/flyteidl2/plugins";
 
+// SubmissionMode specifies how the KubeRay operator submits the Ray job to the RayCluster.
+enum SubmissionMode {
```
pingsutw: I think we just need to update the ray config here. https://github.com/flyteorg/flyte/blob/1dd046406db056f08d27adf2299d2527be9bdffb/flyteplugins/go/tasks/plugins/k8s/ray/config.go#L61-L63
The submission mode should be decided by the platform team

davidlin20dev: @pingsutw Thanks for the review! Just want to make sure I'm on the right track before I continue. 

@popojk  mentioned in Slack that users should be able to specify the mode, with both SDK and backend changes (which is what this PR implements), so I assumed we'd let users override the mode from the SDK. And what you mean is that only the platform team should pick the setting, is that correct?

So, should I do config-only, per-task only, or both (config default + per-task override, like ShutdownAfterJobFinishes)?

pingsutw: yup, ray plugin config is usually set by the platform team, so we only need to do config-only

pingsutw: We can add it to the proto in the future if someone really wants to override it per task

davidlin20dev: I see, thanks for the clarification!

popojk:
>SidecarMode: The KubeRay operator injects a container into the Ray head Pod to submit the Ray job. This mode does not support clusterSelector, submitterPodTemplate, and submitterConfig, and requires the head Pod’s restart policy to be Never.

According to Kuberay [doc](https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/rayjob-quick-start.html#rayjob-configuration), submitter also support sidecar mode, which means the ray operator will inject submitter into header node as sidecar. @pingsutw  Do you think we should also support it as the config settings seems easy?

pingsutw:
yup, let's add that too

flyteplugins/go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -304,7 +304,25 @@ func constructRayJob(ctx context.Context, taskCtx pluginsCore.TaskExecutionConte
...
+	submissionMode := rayv1.K8sJobMode
+	switch cfg.SubmissionMode {
+	case string(rayv1.HTTPMode):
+		submissionMode = rayv1.HTTPMode
+	case string(rayv1.SidecarMode):
+		submissionMode = rayv1.SidecarMode
+	case string(rayv1.K8sJobMode), "":
+		// submissionMode already defaults to K8sJobMode
+	default:
+		return nil, fmt.Errorf("invalid ray submission mode %q: must be K8sJobMode, HTTPMode or SidecarMode", cfg.SubmissionMode)
```
pingsutw: We could return `flyteerr.Errorf(flyteerr.BadTaskSpecification, )`, so the task won't be retried on this error

davidlin20dev: Done. Thanks for the pointer! One question, I'm curious: I traced the code and understand `flyteerr` classifies the error with a code, but I couldn't find where that code prevents retries. It looks like plugin errors all go through `recordSystemError` and get retried regardless of code. Am I missing something?

pingsutw: ah, seems like we don't check the error code in v2? would you mind creating a separate PR to fix it? task should not retry on BadTaskSpecification, MetadataTooLarge, CorruptedPluginState error

pingsutw (approved):

flyteplugins/go/tasks/plugins/k8s/ray/ray.go
davidlin20dev: yeah, I can take this one

PR #7803, opened 2026-08-07
Add InstanceInfo to ActionEvent and ActionAttempt protos

pvditt:

We want action events to record which cloud instance (node) an action-attempt's pod actually ran on — e.g. `p4d.24xlarge` — so it can be shown per attempt and used for GPU/cost attribution. Today nothing captures this at runtime.

Proto/plugin-core surface only (no executor or run-service behavior change in this repo):

- New `InstanceInfo` message: instance type, instance id, accelerator, capacity type (spot/on-demand), zone, node name, plus an attributes map for future use.
- `ActionEvent.instance_info = 17` — set by an executor when it knows the node.
- `ActionAttempt.instance_info = 15` — read-side surface, filled by whatever run service merges events.
- `TaskInfo.NodeName` in pluginmachinery — lets single-pod execution paths pass the scheduled node name along, so an executor can resolve node metadata off the hot path.

The flyte2 executor/run service does not populate or merge the field yet — consumers of these protos (Union leaseworker) do.

## How was this patch tested?

- `buf lint` passes; Go bindings regenerated with the pinned buf templates.
- `go build` + existing test suites for `runs/service` and `flyteplugins` pass.
- Exercised end-to-end from the Union leaseworker implementation against a local devbox: events carried the field and it round-tripped through `GetActionDetails`.

### Labels

added

https://claude.ai/code/session_01PMNn4hYoYsVZWdUB4Cs9iJ


wild-endeavor (commented):
add region and provider which is an enum

flyteidl2/workflow/run_definition.proto
```diff
@@ -370,6 +370,44 @@ message ActionAttempt {
...
+  string instance_type = 1;
+
+  // Cloud provider instance ID parsed from the node's spec.providerID,
+  // e.g. "i-0abc123def456".
+  string instance_id = 2;
+
+  // Accelerator device attached to the node, from the cloud accelerator label
+  // (e.g. "cloud.google.com/gke-accelerator" on GCP). Typically empty on AWS,
+  // where the instance type alone identifies the GPU.
+  string accelerator = 3;
```
wild-endeavor, line 395: @jeevb @EngHabu should we just copy this https://github.com/unionai/flyte/blob/3a2f9ef33da55a017ad447c96523fcd0d3f5d7ec/flyteidl/protos/flyteidl/core/tasks.proto#L51

PR #7715, opened 2026-07-28
Add lastStartedAt field in app status

popojk:
## Why are the changes needed?
We need to add a last_started_at field for app service to inform the run service for persisting such an information.

## What changes were proposed in this pull request?
Idl change

### Labels

Please add one or more of the following labels to categorize your PR:
- **added**: For new features.
- **changed**: For changes in existing functionality.
- **deprecated**: For soon-to-be-removed features.
- **removed**: For features being removed.
- **fixed**: For any bug fixed.
- **security**: In case of vulnerabilities

This is important to improve the readability of release notes.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

## Stack

If you do use [`git town`](https://www.git-town.com/introduction) to manage PR Stacks, the stack relevant to this PR
will show below. Otherwise, you can ignore this section.

PR #7808, opened 2026-08-08
Bind app inputs to a pinned artifact version, and list apps by consumed artifact

kumare3:
## Tracking issue

Related to the Union app-artifact lineage work (unionai/cloud).

An app input could reference an artifact two ways, and neither was the right shape:

- `artifact_query` — an `ArtifactQuery` the server was expected to resolve at some unspecified later time.
- `artifact_id` — an `ArtifactID`, whose `version` is optional.

Both describe a **moving reference**. An app that binds an input to "the latest artifact named X" can silently start running against different data than it was deployed with, with nothing recording which version it actually used. What an app input actually needs is one exact artifact version: the client resolves the artifact at deploy time, and the app runs against that version until it is redeployed.

Neither field was reachable from a released client, so no app in existence uses either one. That makes this the last cheap moment to fix the shape.

Separately, there was no way to ask the reverse question — *which apps consume this artifact?* — which is what makes the recorded lineage useful to anyone.

**`app_definition.proto`**

Replace both fields in the `Input` oneof with a single `artifact_id` of type `flyteidl2.core.ArtifactVersionId` (field 6). `ArtifactVersionId` carries the version as a required part of its identity, so the moving-reference case is unrepresentable.

The field is recorded for lineage only — the server does not resolve or validate it, and may hold a reference to an artifact that no longer exists.

```protobuf
// ArtifactId is the exact artifact version bound to this input, pinned by the
// client at deploy time. Recorded for lineage only; the server does not
// validate or resolve it.
flyteidl2.core.ArtifactVersionId artifact_id = 6;
```

The old fields are **removed outright rather than reserved**. Both were declared in #7715 on 2026-07-29 and never implemented: no server ever read them and no client ever set them, so there is no wire data in existence to protect against. Reserving would only have added permanent noise to the schema for a hazard that does not exist.

The new field does take **6** rather than reusing 3 or 4, so the numbers that did ship in `flyteidl2` 2.0.37–2.0.39 are not given a new meaning.

**`app_payload.proto`**

Add `ConsumedArtifactFilter` as a new `ListRequest.filter_by` variant (field 7), so callers can list the apps consuming a given artifact. An empty `version` matches any version; the artifact org is taken from the caller's organization.

## How was this patch tested?

Generated code was regenerated with `buf` for all targets (Go, Python, Rust, TS) and is included in this PR.

The proto change itself is exercised end-to-end by the two consumers built on top of it:

- **unionai/cloud** — records one lineage row per artifact-bound input, and serves `ConsumedArtifactFilter` through the app `List` RPC. Covered by repository and service unit tests, plus authz middleware table tests.
- **flyteorg/flyte-sdk** — pins the resolved artifact version at deploy time and sends it as `artifact_id`. Its assertion test is currently skipped pending a `flyteidl2` release carrying this change, and unskips automatically once it lands.

### Labels

- **changed** — `Input.artifact_query` and the old `Input.artifact_id` are replaced by a pinned `ArtifactVersionId`.
- **added** — `ConsumedArtifactFilter` on app `ListRequest`.

- [x] I updated the documentation accordingly. *(field-level proto comments; no separate docs page covers these fields)*

## Related PRs

- unionai/cloud — lineage recording, list-by-artifact filter, and console tab (pins the submodule at this branch).
- flyteorg/flyte-sdk — pins the resolved artifact version at deploy time.


flyteidl2/app/app_definition.proto
```diff
@@ -344,15 +344,13 @@ message Input {
     // StringValue is a plain string value for the input
     string string_value = 2 [(buf.validate.field).string.min_len = 1];
 
-    // ArtifactQuery is that should result in a single artifact that will be used as the input to the app at runtime.
-    // The artifact will be pinned at the time of the app creation.
-    flyteidl2.core.ArtifactQuery artifact_query = 3;
-
-    // ArtifactId is an identifier of an artifact that will be used as the input to the app at runtime.
-    flyteidl2.core.ArtifactID artifact_id = 4;
-
```
EngHabu, line 353: Are these not used at all?

EngHabu (approved):

flyteidl2/app/app_definition.proto
kumare3: nope this is flyteidl2

Issue #7806, opened 2026-08-08
Human-in-the-loop conditions are unresolvable on the devbox


francescorubbo:
**Env**: Flyte SDK 2.5.18, local devbox.

**What we're doing**: a task calls `flyte.new_condition(name, prompt=..., data_type=str)` then `await cond.wait()` to pause a run for a human label.

```python
@env.task
async def human_label(item: dict) -> str:
    cond = await flyte.new_condition.aio("label-123", prompt="positive/negative?", data_type=str)
    return await cond.wait.aio()
```

**Symptom**: the run pauses at the condition and never resumes. The Console shows no input/approval widget — the run just reads as _Running_ indefinitely, with no indication a human is needed.

**The backend is actually fine.** The condition registers and the sub-action reaches `PAUSED`; the SDK sees it:

```python
remote.Action.listall(for_run_name=RUN, in_phase=("PAUSED",))
# -> ['curu7kmp1mzzeb0de88wogmvk']
```

**But every human-facing path to resolve it fails:**

1. **Console UI** — no input affordance rendered for a paused condition action; the run surfaces as plain `RUNNING`, so there's no signal that input is required.
2.` flyte get condition <run>` — errors before returning anything: `Invalid argument… failed to convert filters: invalid filter field: action_type` (the command sends a filter field the backend's filter converter doesn't support).
3. `flyte signal condition <run> <action> <value>` — reads the condition correctly (prints its prompt and declared type from the backend) but the signal RPC fails: `Requested object NOT FOUND… condition action <run>-<action> not found`
The read path resolves the condition; the write path looks it up under a different (run-name-prefixed) identifier and misses.

**Net**: the pause half of the HITL primitive works; none of the resume/discovery surfaces do, so a paused run is a dead end unless you hand-write SDK calls.

Happy to open PR(s) to address these if not already in the works (assuming this is not an user-error issue).

pingsutw:
@francescorubbo we have addressed this issue, could you give the latest devbox a try

francescorubbo:
Thank you, @pingsutw ! I tried with 2.6.3 (and pulling the latest dev box image) and the UI works now.

However the CLI seems to be still broken
```
uv run flyte get condition rjqjkbjqb8pgh5vk4pf4 -p flytesnacks -d development
╭─ Error ──────────────────────────────────────────────────────────────────╮
│ Invalid argument provided. Please check your input. Error: failed to     │
│ convert filters: invalid filter field: action_type                       │
[...]

PR #7813, opened 2026-08-08
fix: unblock condition action signalling

Rohithmatham12:
## Summary
- allow action_type as a ListActions filter column so condition queries do not fail validation before reaching the repository
- let condition signalling fall back to an already-materialized TaskAction name when the primary run/name lookup misses
- add regression coverage for both the filter allowlist and raw TaskAction-name signal path

## Issue
Addresses #7806.

## Testing
- GOCACHE=/private/tmp/flyte-go-cache GOMODCACHE=/private/tmp/flyte-go-mod-cache go test ./actions/k8s -run TestSignal
- git diff --check

Attempted but locally blocked:
- GOCACHE=/private/tmp/flyte-go-cache GOMODCACHE=/private/tmp/flyte-go-mod-cache go test ./runs/repository/impl -run TestConvertProtoFilters_ActionTypeColumn

The repository package TestMain starts embedded Postgres on hard-coded port 15432; my local run failed before executing the pure filter test with: process already listening on port 15432.

### DIFF
--- a/actions/k8s/client.go
+++ b/actions/k8s/client.go
@@ -310,29 +310,31 @@ func (c *ActionsClient) Signal(ctx context.Context, actionID *common.ActionIdent
 
 	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		taskAction := &executorv1.TaskAction{}
-		resolvedTaskActionName, err := c.getTaskActionForSignal(ctx, actionID, taskActionName, taskAction)
-		if err != nil {
-			return err
+		if err := c.k8sClient.Get(ctx, client.ObjectKey{Name: taskActionName, Namespace: c.namespace}, taskAction); err != nil {
+			if apierrors.IsNotFound(err) {
+				return connect.NewError(connect.CodeNotFound, fmt.Errorf("condition action %s not found", taskActionName))
+			}
+			return fmt.Errorf("failed to get TaskAction %s: %w", taskActionName, err)
 		}
 		if taskAction.Spec.ActionType != executorv1.ActionTypeCondition {
 			return connect.NewError(connect.CodeFailedPrecondition,
-				fmt.Errorf("action %s is not a condition", resolvedTaskActionName))
+				fmt.Errorf("action %s is not a condition", taskActionName))
 		}
 		if existing := SignalValueFromStatus(ctx, taskAction); existing != nil {
 			if proto.Equal(existing, value) {
 				return nil // idempotent retry of the same signal
 			}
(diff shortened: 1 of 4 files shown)


actions/k8s/client.go
```diff
@@ -348,11 +346,36 @@ func (c *ActionsClient) Signal(ctx context.Context, actionID *common.ActionIdent
...
+	taskAction *executorv1.TaskAction,
+) (string, error) {
+	key := client.ObjectKey{Name: taskActionName, Namespace: c.namespace}
+	if err := c.k8sClient.Get(ctx, key, taskAction); err == nil {
+		return taskActionName, nil
+	} else if !apierrors.IsNotFound(err) {
+		return "", fmt.Errorf("failed to get TaskAction %s: %w", taskActionName, err)
+	}
+
+	if actionID.GetName() != "" && actionID.GetName() != taskActionName {
```
popojk: Do we really need this fall back? When we use `flyte signal condition`, The action name was fetched from the backend DB and filled into signal API call request. The taskActionName should always be <run_name>-<action_name>. I tested in main branch with bellow script, it works actually.
```
"""Minimal HITL condition, for exercising the pause/signal round trip on the devbox.

Run it, note the run name it prints, then resolve the condition from another shell:

    flyte get condition <run-name>
    flyte signal condition <run-name> <action-name> yes

The task blocks in `cond.wait()` until then. A 30 minute timeout keeps a forgotten
run from pausing forever.
"""
import asyncio
import logging

import flyte

env = flyte.TaskEnvironment(
    name="hitl_demo",
)


@env.task
async def ask_human(item: str) -> str:
    cond = await flyte.new_condition.aio(
        "approve-item",
        prompt=f"Approve {item}? (yes/no)",
        data_type=str,
        description="Demo condition for exercising the signal path on the devbox.",
        timeout=timedelta(minutes=30),
    )
    print(f"waiting for a signal on '{item}' — action: {flyte.ctx().action}")
    answer = await cond.wait.aio()
    print(f"got: {answer!r}")
    return answer


@env.task(entrypoint=True)
async def main(item: str = "widget-42") -> str:
    answer = await ask_human(item=item)
    return f"{item} -> {answer}"


if __name__ == "__main__":
    flyte.init_from_config()
    run = flyte.with_runcontext(log_level=logging.INFO).run(main, item="widget-42")
    print(f"run name: {run.name}")
    print(f"resolve with: flyte signal condition {run.name} <action-name> yes")
    flyte.remote.Action.listall(in_phase=("PAUSED",))

```

pingsutw (approved):

PR #7821, opened 2026-08-10
ci: cache devbox image layers

pingsutw:
## Summary

- Stop passing a timestamp `CACHEBUST` to the devbox image build in GitHub Actions so the preloaded image layer can restore from buildx cache.
- Keep local `make devbox-build` behavior unchanged by continuing to refresh floating preloaded images outside CI.
- Move the flyte binary image `go mod download` layer before source copies and remove ineffective module cache mounts.

## Root Cause

The devbox workflow already configures buildx `type=gha` cache, but the top-level `devbox-build` target always passed `CACHEBUST=$(date +%s)`. That invalidated the `RUN ./preload manifest.txt` layer on every CI run, forcing the bundled image preload work to repeat.

The binary image also ran `go mod download` after copying source directories, so source changes invalidated dependency download reuse.

## Validation

- `git diff --check`
- `GITHUB_ACTIONS=true make -n devbox-build FLYTE_ARCHES=amd64` shows GHA cache flags and no `--build-arg CACHEBUST=...`.
- `make -n devbox-build FLYTE_ARCHES=amd64` still shows `CACHEBUST=...` and `--build-arg CACHEBUST=...` locally.
- Targeted preload-stage cache test: first local run `43s`, second local run `3s`; second run showed `RUN echo "cachebust=0" && ./preload manifest.txt` as `CACHED`.
- `docker buildx build --builder flyte-devbox --platform linux/amd64 --target flytebuilder --progress=plain --output=type=cacheonly .` succeeded.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,9 @@ ENV CGO_ENABLED=0
 
 WORKDIR /flyteorg/build
 
+COPY go.mod go.sum ./
+RUN mkdir flyteidl2 flyteplugins flytestdlib && go mod download
+
 COPY dataproxy dataproxy
 COPY executor executor
 COPY flytecopilot flytecopilot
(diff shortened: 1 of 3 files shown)


pingsutw:
CI validation for the devbox cache change:

| Attempt | Job ID | Result | Total job | Build devbox image |
| --- | ---: | --- | ---: | ---: |
| 1 | 93595408654 | success | 11m35s | 7m58s |
| 2 | 93598619380 | success | 7m16s | 3m47s |
| 3 | 93600687325 | success | 7m51s | 4m01s |
| 4 | 93603114458 | success | 7m27s | 3m43s |

Baseline from the referenced run was `9m53s` total with `Build devbox image` at about `6m33s`.

The CI command no longer passes `CACHEBUST` in GitHub Actions, and all reruns passed the full SDK example flow. The rerun logs still show the final devbox preload `RUN echo "cachebust=0" && ./preload manifest.txt` executing instead of being restored as a cached Docker `RUN` layer, so this PR improves the build materially but does not fully eliminate that preload work from GitHub's buildx cache.

Sovietaced (approved):

PR #7833, opened 2026-08-11
fix(pytorch): propagate pod metadata to the PyTorchJob CR

evan-sadler:
BuildResource returned the PyTorchJob without an ObjectMeta, so pod-level metadata (pod_template / k8s_pod labels and annotations) only reached the replica pod templates via common.ToReplicaSpec and never the CR itself. Execution-level labels were unaffected -- the plugin manager stamps those on every resource in addObjectMetadata.

Set ObjectMeta from a fresh flytek8s.ToK8sPodSpec call, matching what the ray and dask plugins already do. Deriving it from a replica spec would be wrong: ToReplicaSpecWithOverrides builds a per-replica-group task context, so group-specific labels would leak onto the CR.

### Labels

## Stack

### DIFF
--- a/flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
+++ b/flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
@@ -136,12 +136,22 @@ func (pytorchOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx
 		delete(jobSpec.PyTorchReplicaSpecs, kubeflowv1.PyTorchJobReplicaTypeMaster)
 	}
 
+	// Propagate the task's pod metadata (k8s_pod / pod_template labels and annotations) onto the CR
+	// itself, as the ray and dask plugins do. The replica pod templates already receive this metadata
+	// via common.ToReplicaSpec; without this the CR only gets the execution-level labels that the
+	// plugin manager stamps on every resource.
+	_, objectMeta, _, err := flytek8s.ToK8sPodSpec(ctx, taskCtx)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


flyteplugins/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
```diff
@@ -136,12 +136,22 @@ func (pytorchOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx
 		delete(jobSpec.PyTorchReplicaSpecs, kubeflowv1.PyTorchJobReplicaTypeMaster)
 	}
 
+	// Propagate the task's pod metadata (k8s_pod / pod_template labels and annotations) onto the CR
+	// itself, as the ray and dask plugins do. The replica pod templates already receive this metadata
+	// via common.ToReplicaSpec; without this the CR only gets the execution-level labels that the
+	// plugin manager stamps on every resource.
+	_, objectMeta, _, err := flytek8s.ToK8sPodSpec(ctx, taskCtx)
+	if err != nil {
+		return nil, flyteerr.Errorf(flyteerr.BadTaskSpecification, "Unable to create pod spec: [%v]", err.Error())
```
pingsutw, line 145: I think we also need to merge metadata in task ctx, like https://github.com/flyteorg/flyte/blob/a3093e708433edfe32046aefd9512e8163065916/flyteplugins/go/tasks/plugins/k8s/kfoperators/common/common_operator.go#L286-L287

pingsutw: nvm, I think we probably don't really need it for now

pingsutw: I don't see we add it in the ray plugin

pingsutw (approved):

PR #7834, opened 2026-08-11
[FIX] reference secrets with secretKeyRef instead of putting values in the pod spec

### LINKED ISSUE #7824: [V2] Secret value is showing in pod spec ENVVAR directly
## Issue

When injecting a secret into task pod like this:
```
env = flyte.TaskEnvironment(
    name="hello_world",
    image=image,
    secrets=[
        flyte.Secret(key="my_secret", as_env_var="MY_SECRET_KEY"),
    ]
)
```

When getting pod spec with `kubectl describe` command, we will directly see the secret value in ENVVAR, which is not secure:

Image

## Suggestion
We use webhook to inject secret into task pod. We should insert EnvVarSource instead of [EnvVar value](https://github.com/flyteorg/flyte/blob/aa38af043e04220f91466132cd744666e0f571d8/flyteplugins/go/tasks/pluginmachinery/secret/embedded_secret_manager.go#L360) directly

0yukali0:
## Tracking issue
Close #7824 

`flyte.Secret(key="my_secret", as_env_var="MY_SECRET_KEY")` will trigger that the backend injects the plain value from secret to the pod.
That's meaning who use `kubectl describe <pod>` command to able to know the actual secret value.
  
```python
env = flyte.TaskEnvironment(
    name="hello_world",
    image=image,
    secrets=[
        flyte.Secret(key="my_secret", as_env_var="MY_SECRET_KEY"),
    ]
)
```

1. After the existing cascade lookup resolves which scope the secret lives at, check that it is actually stored in a Kubernetes Secret in the pod's namespace (envVarSecretRef: object EncodeK8sSecretName(secretID), data key secretID).
2. If it is, point the env var at it with an EnvVarSource/secretKeyRef instead of the plain value; if it is not, return an error instead of falling back to the value.

## How was this patch tested?
1. Unit tests commands 
```bash
make test_unit
```

2. quickstart.py is following, run the commands in the `setup section` to check the results.
```python
import os
import flyte

env = flyte.TaskEnvironment(
    name="my_env",
    resources=flyte.Resources(memory="250Mi"),
    image=(
        flyte.Image.from_debian_base(
            name="my-image",
            python_version=(3, 13),
            registry="localhost:30000",  # the devbox registry, so the cluster can pull the build
        )
        .with_apt_packages("libopenblas-dev")
        .with_pip_packages("numpy")
        .with_env_vars({"OMP_NUM_THREADS": "4"})
    ),
    secrets=[
        flyte.Secret(key="MY_SECRET_KEY", as_env_var="MY_SECRET_KEY"),
    ],
)

@env.task
def fn(x: int) -> int:
    print(f"MY_SECRET_KEY={os.environ.get('MY_SECRET_KEY')!r}", flush=True)
    slope, intercept = 2, 5
    return slope * x + intercept

@env.task
def main(x_list: list[int] = [1, 2, 3]) -> float:
    print(f"MY_SECRET_KEY={os.environ.get('MY_SECRET_KEY')!r}", flush=True)
    y_list = list(flyte.map(fn, x_list))
    return sum(y_list) / len(y_list)

if __name__ == "__main__":
    flyte.init_from_config()
    run = flyte.run(main)
    print(run.name)
    print(run.url)
```

### Labels
- **fix**:

### Setup process
```bash
make devbox-build
make devbox-start

# Use kubectl or k9s to view the pod spec, secret resources and action logs.
# Case 1：org scope，--value
flyte create secret MY_SECRET_KEY --value my_secret_value
flyte run -p flytesnacks -d development quickstart.py main

# Case 2：org scope，--from-file
flyte delete secret MY_SECRET_KEY
flyte create secret MY_SECRET_KEY --from-file ./my_secret_file
flyte run -p flytesnacks -d development quickstart.py main

# Case 3：project scope
flyte create project --name my_project  --id my_project
flyte create secret MY_SECRET_KEY --value my_secret_value --project my_project --domain development
flyte run -p flytesnacks -d development quickstart.py main

```

### Screenshots
Screenshot 2026-08-12 at 6 28 01 AM

## Stack

### DIFF
--- a/flyteplugins/go/tasks/pluginmachinery/secret/embedded_secret_manager.go
+++ b/flyteplugins/go/tasks/pluginmachinery/secret/embedded_secret_manager.go
@@ -137,13 +137,13 @@ func encodeSecretName(components *SecretNameComponents) string {
 	return EncodeSecretName(components.Org, components.Domain, components.Project, components.Name)
 }
 
-type resolvedSecret struct {
-	value               *SecretValue
-	secretID            string
-	imagePullSecretName string
+type resolvedUserSecret struct {
+	value       *SecretValue
+	secretID    string
+	isK8sSecret bool
 }
 
-func (i *EmbeddedSecretManagerInjector) lookUpSecret(ctx context.Context, components *SecretNameComponents) (*resolvedSecret, error) {
+func (i *EmbeddedSecretManagerInjector) lookUpSecret(ctx context.Context, components *SecretNameComponents) (*resolvedUserSecret, string, error) {
 	componentsCopy := *components // Create a copy to avoid modifying the original
 
 	// Compute encoded IDs for each scope in priority order.
@@ -167,6 +167,8 @@ func (i *EmbeddedSecretManagerInjector) lookUpSecret(ctx context.Context, compon
 		{orgScopedSecret, orgScopedImagePullSecretName},
 	}
 
+	isK8sSecret := i.cfg.Type == config.EmbeddedSecretManagerTypeK8s
+
 	// Walk scopes in priority order (project+domain → domain → org). For each
 	// scope, check the cache first; on miss, fall through to the fetchers for
 	// that same scope. Only move to the next broader scope when the current
@@ -174,7 +176,13 @@ func (i *EmbeddedSecretManagerInjector) lookUpSecret(ctx context.Context, compon
 	for _, scope := range scopes {
 		if cachedValue, err := i.secretCache.Get(ctx, scope.secretID); err == nil {
 			logger.Debugf(ctx, "Found secret [%s] in cache.", scope.secretID)
-			return &resolvedSecret{&cachedValue, scope.secretID, scope.imagePullSecretName}, nil
+			if isK8sSecret {
+				storedSecret := &corev1.Secret{}
+				err := i.k8sClient.Get(ctx, types.NamespacedName{Name: EncodeK8sSecretName(scope.secretID), Namespace: i.cfg.K8sConfig.Namespace}, storedSecret)
+				_, isK8sSecret = storedSecret.Data[scope.secretID]
+				isK8sSecret = isK8sSecret && err == nil
+			}
+			return &resolvedUserSecret{&cachedValue, scope.secretID, isK8sSecret}, scope.imagePullSecretName, nil
 		}
 
 		for _, secretFetcher := range i.secretFetchers {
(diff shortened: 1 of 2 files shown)


popojk (dismissed):
LGTM thanks! cc @pingsutw for second eye.

flyteplugins/go/tasks/pluginmachinery/secret/embedded_secret_manager.go
```diff
@@ -137,7 +137,13 @@ func encodeSecretName(components *SecretNameComponents) string {
 	return EncodeSecretName(components.Org, components.Domain, components.Project, components.Name)
 }
 
-func (i *EmbeddedSecretManagerInjector) lookUpSecret(ctx context.Context, components *SecretNameComponents) (*SecretValue, string, error) {
+type resolvedSecret struct {
+	value               *SecretValue
+	secretID            string
+	imagePullSecretName string
+}
```
pingsutw:
```suggestion
type resolvedUserSecret struct {
	value               *SecretValue
	secretID            string
	isK8sSecret.     bool
}
```

flyteplugins/go/tasks/pluginmachinery/secret/embedded_secret_manager.go
```diff
@@ -137,7 +137,13 @@ func encodeSecretName(components *SecretNameComponents) string {
...
+
+func (i *EmbeddedSecretManagerInjector) lookUpSecret(ctx context.Context, components *SecretNameComponents) (*resolvedSecret, error) {
```
pingsutw:
```suggestion
func (i *EmbeddedSecretManagerInjector) lookUpSecret(ctx context.Context, components *SecretNameComponents) (*resolvedUserSecret, string, error) {
```

flyteplugins/go/tasks/pluginmachinery/secret/embedded_secret_manager.go
pingsutw: We should check if the secret is from K8sFetcher or not. if it's not  a k8s secret, we cannot mount it

flyteplugins/go/tasks/pluginmachinery/secret/embedded_secret_manager.go
```diff
@@ -356,14 +362,48 @@ func (i *EmbeddedSecretManagerInjector) injectAsEnvVar(secret *core.Secret, secr
...
+	namespace string,
+) (*corev1.SecretKeySelector, error) {
+	name := EncodeK8sSecretName(secretID)
+
+	if namespace == "" {
+		return nil, fmt.Errorf("cannot mount secret [%s]: pod has no namespace, so Kubernetes secret [%s] was not looked up", components.Name, name)
+	}
+
+	storedSecret := &corev1.Secret{}
+	if err := i.k8sClient.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, storedSecret); err != nil {
```
pingsutw: We already lookup secret above, why do we need to get the secret again here?

0yukali0: Make sure that k8s secret exists and it is in same namespace of the pod where it mount.
I moved this to the lookupSecret  which handle the secret search.

0yukali0:
Screenshot of this change.
Screenshot 2026-08-20 at 11 38 33 PM

0yukali0:
Secret mount on applications also solved.
Screenshot 2026-08-21 at 1 06 48 AM

popojk (approved):

PR #7841, opened 2026-08-12
feat(runs): settings persistence layer (RFC #7775, 1/2)

davidlin20dev:
## Tracking issue

Related to #7775 (Settings Service RFC, first implementation PR)

The Settings API in `flyteidl2/settings/` is fully defined, but no server implements it: there is currently no way to store or read settings at all. #7775 lays out the implementation plan. This PR is the first piece of it, the persistence layer: the `settings` table and the repository that owns all SQL against it, so the CRUD service handlers (next PR) have something to build on.

This is built as a vertical slice rather than strictly phase-by-phase: this PR covers tasks 1.1-1.4, and the follow-up adds minimal versions of 2.4-2.6 (Connect handlers and `runs/setup.go` wiring) so the service is callable end to end.

- Migration for the `settings` table: one row per scope (instance / domain / project), with a `version` column for optimistic locking
- Go model and key encoder (`"v1:{org}:{domain}:{project}"`, empty org normalized to `flyte`, matching the existing secret/app-service convention)
- Repository interface (`Create/Get/GetByKeys/Update`) with sentinel errors, and the sqlx implementation; optimistic locking follows the trigger repository's pattern

Settings are stored as the raw protojson of the `Settings` message. Sparse storage (prune/hydrate) and value validation are deferred to tasks 2.1/2.3, which fit naturally with the merge-engine work.

Not included (follow-ups): CRUD handlers and registration (next PR), merge/inheritance resolution, applying settings to runs.

## How was this patch tested?

Embedded-Postgres tests (real PostgreSQL via the existing `RunTestMain` harness, all migrations applied):

- create then get round-trip (struct/row consistency, JSONB round-trip)
- duplicate create returns `ErrSettingsAlreadyExists`, original row unharmed
- get of a missing key returns `ErrSettingsNotFound`
- multi-key ancestor fetch: missing rows are absent from results, not errors
- optimistic-locking race: the stale writer gets `ErrSettingsVersionConflict`,
  the winner's write survives, and the struct version syncs only on success

The whole `impl` package passes (new tests plus all existing ones), and the key encoder has its own table test in `models`. `gofmt` and `go vet` are clean.

## Related PRs

Follow-up: CRUD service handlers + wiring, to be opened on top of this branch.

### DIFF
--- a/runs/repository/impl/settings.go
+++ b/runs/repository/impl/settings.go
@@ -25,6 +25,10 @@ func NewSettingsRepo(db *sqlx.DB) interfaces.SettingsRepo {
 }
 
 func (r *settingsRepo) CreateSettings(ctx context.Context, settings *models.Settings) error {
+	if len(settings.Data) == 0 {
+		return fmt.Errorf("settings data is required: %s", settings.Key)
+	}
+
 	now := time.Now().UTC()
 	settings.CreatedAt = now
 	settings.UpdatedAt = now
--- a/runs/repository/impl/settings.go
+++ b/runs/repository/impl/settings.go
@@ -1,0 +1,99 @@
+package impl
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/jmoiron/sqlx"
+	"github.com/lib/pq"
+
+	"github.com/flyteorg/flyte/v2/flytestdlib/database"
+	"github.com/flyteorg/flyte/v2/runs/repository/interfaces"
+	"github.com/flyteorg/flyte/v2/runs/repository/models"
+)
+
+type settingsRepo struct {
+	db *sqlx.DB
+}
+
+// NewSettingsRepo returns a SettingsRepo backed by Postgres.
+func NewSettingsRepo(db *sqlx.DB) interfaces.SettingsRepo {
+	return &settingsRepo{db: db}
+}
+
+func (r *settingsRepo) CreateSettings(ctx context.Context, settings *models.Settings) error {
+	now := time.Now().UTC()
+	settings.CreatedAt = now
+	settings.UpdatedAt = now
+	settings.Version = 1
+
+	result, err := r.db.ExecContext(ctx,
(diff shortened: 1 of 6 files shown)


runs/repository/impl/settings.go
```diff
@@ -0,0 +1,99 @@
...
+// NewSettingsRepo returns a SettingsRepo backed by Postgres.
+func NewSettingsRepo(db *sqlx.DB) interfaces.SettingsRepo {
+	return &settingsRepo{db: db}
+}
+
+func (r *settingsRepo) CreateSettings(ctx context.Context, settings *models.Settings) error {
+	now := time.Now().UTC()
+	settings.CreatedAt = now
+	settings.UpdatedAt = now
+	settings.Version = 1
```
pingsutw, line 35: Could we return error here if settings.data is nil?

runs/repository/impl/settings.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	rowsAffected, err := result.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("failed to get rows affected: %w", err)
+	}
+	// Zero rows: the caller's version is stale (or the row is gone);
+	// either way the caller must re-read, so report it as a conflict.
+	if rowsAffected == 0 {
+		return fmt.Errorf("%w: %s", interfaces.ErrSettingsVersionConflict, settings.Key)
```
pingsutw, line 103: We should return non-found error if there the row is gone

runs/repository/impl/settings.go
davidlin20dev: Added, I also applied it to UpdateSettings, same failure there: writes the whole document, so empty bytes means the caller forgets it

runs/repository/impl/settings.go
davidlin20dev: done!

davidlin20dev:
@pingsutw Thanks for the review! Just pushed, lmk if there is any questions.

pingsutw (approved):
LGTM, thanks

PR #7844, opened 2026-08-13
fix(flytestdlib): fix race condition in TestCacheFour Enqueue nothing

### LINKED ISSUE #7843: bug: TestCacheFour/"Enqueue nothing" is flaky on main — clock.Step races the timer re-arm
## Why

`TestCacheFour/Enqueue_nothing` in `flytestdlib/cache` fails intermittently on `main`, turning the
`Go Tests / test (flytestdlib)` job red on commits that have nothing to do with the cache. Recent
example on `main` (`be46993`):

https://github.com/flyteorg/flyte/actions/runs/31663249201/job/94332365727

```
--- FAIL: TestCacheFour (1.13s)
    --- FAIL: TestCacheFour/Enqueue_nothing (1.00s)
        Error Trace: flytestdlib/cache/in_memory_auto_refresh_test.go:206
        Error:       Condition never satisfied
        Test:        TestCacheFour/Enqueue_nothing
FAIL github.com/flyteorg/flyte/v2/flytestdlib/cache
```

Red CI on `main` costs everyone: contributors can't tell whether their own PR broke something, and
re-running the job becomes reflex — which is exactly how a real regression slips through.

### What's actually racing

`enqueueLoop` uses a **one-shot timer that is re-armed only after the batch is enqueued**
([`flytestdlib/cache/in_memory_auto_refresh.go:181-198`](https://github.com/flyteorg/flyte/blob/main/flytestdlib/cache/in_memory_auto_refresh.go#L181-L198)):

```go
timer := w.clock.NewTimer(w.syncPeriod)
for {
    select {
    case <-timer.C():
        err := w.enqueueBatches(ctx)   // increments enqueueCount (defer, line 263)
        ...
        timer.Reset(w.syncPeriod)      // <-- re-arm happens AFTER the counter moves
    }
}
```

The test steps a fake clock and waits on that counter
([`in_memory_auto_refresh_test.go:194-206`](https://github.com/flyteorg/flyte/blob/main/flytestdlib/cache/in_memory_auto_refresh_test.go#L194-L206)):

```go
fakeClock.Step(testResyncPeriod)
assert.Eventually(t, func() bool { return cache.enqueueCount.Load() > enqueueCount }, ...)  // L197
...
fakeClock.Step(testResyncPeriod)                                                            // L204
assert.Eventually(t, func() bool { return cache.enqueueCount.Load() > enqueueCount }, ...)  // L206 <-- flakes
```

`enqueueCount` is incremented *before* `timer.Reset` runs, so the first `Eventually` can return while
the loop has no armed timer. The second `Step` then advances the fake clock with no waiter registered;
`timer.Reset` afterwards arms relative to the already-advanced fake now, so it never fires — and
nothing steps the clock again. The counter never moves, and the assertion times out after 1s.

Also worth noting: `fakeClock` is shared by every subtest in `TestCacheFour`, so cancelled caches from
earlier subtests add more waiters to the same clock — extra noise on top of the same race.

## What to change

Fix the test's synchronization; the production `enqueueLoop` behavior is fine.

- `flytestdlib/cache/in_memory_auto_refresh_test.go` — in the `"Enqueue nothing"` (and the
  structurally identical `"Test update and delete cache"`) subtests, don't assume the loop has
  re-armed. Either:
  - wait for the timer to be armed before stepping: `assert.Eventually(t, fakeClock.HasWaiters, ...)`
    right before each `fakeClock.Step(...)`; **or**
  - step the clock *inside* the polling loop, the way `"normal operation"` already does
    (`assert.EventuallyWithT` with `fakeClock.Step(testResyncPeriod)` in the body, lines 101-110) —
    that variant is self-healing and hasn't flaked.
- Consider giving each subtest its own `testingclock.NewFakeClock(...)` so cancelled caches from
  earlier subtests can't register waiters on the shared clock.

## Outcome

- [ ] `go test -race -count=200 -run TestCacheFour ./flytestdlib/cache/...` passes with no failures
- [ ] The subtest still asserts what it was written to assert: terminal items are **not** enqueued for
      sync (`syncCount` unchanged across resync periods) — don't weaken it into a no-op
- [ ] No `time.Sleep`-based fix; the fake clock stays the only time source

## Getting started

- Reproduce: `go test -race -count=200 -run 'TestCacheFour/Enqueue_nothing' ./flytestdlib/cache/`
  (it's timing-dependent; a loaded machine or `-cpu 1` makes it fail sooner)
- Relevant code: `enqueueLoop` / `enqueueBatches` in `flytestdlib/cache/in_memory_auto_refresh.go`,
  and the subtest at `flytestdlib/cache/in_memory_auto_refresh_test.go:174-211`
- Test: `go test ./flytestdlib/cache/...`
- Setup: [CONTRIBUTING.md](https://github.com/flyteorg/flyte/blob/main/CONTRIBUTING.md)

Related, previously closed: #5711 (`[BUG] flytestdlib/cache unit tests are flaky`).

VedantMadane:
Fixes #7843

### Summary
Fixes a race condition in `TestCacheFour/Enqueue_nothing` (and related subtests in `flytestdlib/cache/in_memory_auto_refresh_test.go`) where `fakeClock.Step()` executed before `enqueueLoop` finished setting or re-arming its timer.

### Changes
1. **Isolated Fake Clocks:** Instantiated isolated `testingclock.NewFakeClock()` instances per subtest in `TestCacheFour` to prevent timer step cross-talk.
2. **Waiter Assertions:** Added `assert.Eventually(t, fakeClock.HasWaiters, ...)` checks before `fakeClock.Step(...)` calls to ensure the background `enqueueLoop` goroutine has armed/re-armed its timer.
3. **Verification:** Ran `go test -v ./flytestdlib/cache -run TestCacheFour` and confirmed all 6 subtests pass consistently.

### DIFF
--- a/flytestdlib/cache/in_memory_auto_refresh_test.go
+++ b/flytestdlib/cache/in_memory_auto_refresh_test.go
@@ -77,9 +77,9 @@ func (p *panickingSyncer) sync(_ context.Context, _ Batch) ([]ItemSyncResponse, 
 func TestCacheFour(t *testing.T) {
 	testResyncPeriod := 5 * time.Second
 	rateLimiter := workqueue.DefaultTypedControllerRateLimiter[*Batch]()
-	fakeClock := testingclock.NewFakeClock(time.Now())
 
 	t.Run("normal operation", func(t *testing.T) {
+		fakeClock := testingclock.NewFakeClock(time.Now())
 		// the size of the cache is at least as large as the number of items we're storing
 		cache, err := NewInMemoryAutoRefresh("fake1", syncFakeItem, rateLimiter, testResyncPeriod, 10, 10, promutils.NewTestScope(), WithClock(fakeClock))
 		assert.NoError(t, err)
@@ -112,6 +112,7 @@ func TestCacheFour(t *testing.T) {
 	})
 
 	t.Run("disable sync on create", func(t *testing.T) {
+		fakeClock := testingclock.NewFakeClock(time.Now())
 		cache, err := NewInMemoryAutoRefresh("fake1", syncFakeItem, rateLimiter, testResyncPeriod, 10, 10, promutils.NewTestScope(), WithClock(fakeClock), WithSyncOnCreate(false))
 		assert.NoError(t, err)
 
@@ -143,6 +144,7 @@ func TestCacheFour(t *testing.T) {
 	})
 
 	t.Run("Not Found", func(t *testing.T) {
+		fakeClock := testingclock.NewFakeClock(time.Now())
 		// the size of the cache is at least as large as the number of items we're storing
 		cache, err := NewInMemoryAutoRefresh("fake2", syncFakeItem, rateLimiter, testResyncPeriod, 10, 2, promutils.NewTestScope(), WithClock(fakeClock))
 		assert.NoError(t, err)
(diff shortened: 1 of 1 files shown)


pingsutw (commented):
@VedantMadane Thanks for your contribution, could you take a look at the unit test failure?

Sovietaced (approved):

Sovietaced:
@VedantMadane can you rebase to pick up my build fix

VedantMadane:
@Sovietaced Rebased onto latest main! Thanks.

PR #7852, opened 2026-08-13
fix(propeller): record events Admin already has in the dedup filter

1fanwang:

`adminEventSink.Sink` keeps a dedup filter so an event that has already been recorded doesn't go back to Admin. The id is only added to that filter after a clean send:

```go
	s.filter.Add(ctx, id)
	return nil
```

When Admin answers `AlreadyExists`, the sink returns the wrapped error and the id is never recorded. Every later re-emit of that event passes `filter.Contains`, takes a token from the sink's rate limiter, and makes a round trip that cannot succeed.

The limiter is shared by every execution the propeller is running. One execution re-emitting a recorded event can drain it, and the executions that needed those tokens get `ResourceExhausted`:

```
re-emits of an already-recorded event that reached Admin: 10/10
Error:      Received unexpected error:
            ResourceExhausted: Resource Exhausted, caused by [Admin EventSink throttling admin traffic]
Messages:   a healthy execution was starved of rate-limiter capacity by an already-recorded event
```

Add the id to the dedup filter when Admin reports it already has the event, so later re-emits are answered by the `filter.Contains` check instead of spending limiter capacity.

The error is still returned, and `filter.Contains` returns that same code for a locally known duplicate, so callers see no change either way.

Only `AlreadyExists` is filtered. `EventAlreadyInTerminalState` is deliberately left out: callers act on it — `nodeerrors.IllegalStateError` in `node_exec_context.go`, a forced transition to `WorkflowPhaseFailed` in `workflow/executor.go` — and folding it into the filter would replay it as `AlreadyExists`, which those same callers map to nil.

## How was this patch tested?

`TestAdminEventSink_AlreadyExistsDrainsRateLimiter` builds the sink the way `NewAdminEventSink` does in production — a real `OppoBloomFilter` and a real token-bucket limiter, with only the Admin client stubbed, answering `codes.AlreadyExists` as a real Admin does for an event it already holds. It re-emits that event ten times against a limiter with ten tokens, then checks an unrelated execution can still record its own event.

```
cd flytepropeller
git checkout upstream/master -- events/admin_eventsink.go   # drop the fix
go test ./events/ -run TestAdminEventSink_AlreadyExistsDrainsRateLimiter -count=1
git checkout HEAD -- events/admin_eventsink.go              # restore it
go test ./events/ -run TestAdminEventSink_AlreadyExistsDrainsRateLimiter -count=1 -v
go test ./events/... -count=1
```

<details><summary>Raw output</summary>

Without the fix:

```
--- FAIL: TestAdminEventSink_AlreadyExistsDrainsRateLimiter (0.00s)
    eventsink_drain_repro_test.go:79: re-emits of an already-recorded event that reached Admin: 10/10
    eventsink_drain_repro_test.go:80:
        	Error:      	Not equal:
        	            	expected: 1
        	            	actual  : 10
        	Messages:   	an already-recorded event should reach Admin once; afterwards the dedup filter should answer
    eventsink_drain_repro_test.go:85:
        	Error:      	Received unexpected error:
        	            	ResourceExhausted: Resource Exhausted, caused by [Admin EventSink throttling admin traffic]
        	Messages:   	a healthy execution was starved of rate-limiter capacity by an already-recorded event
FAIL
FAIL	github.com/flyteorg/flyte/flytepropeller/events	0.850s
```

With the fix:

```
=== RUN   TestAdminEventSink_AlreadyExistsDrainsRateLimiter
    eventsink_drain_repro_test.go:79: re-emits of an already-recorded event that reached Admin: 1/10
--- PASS: TestAdminEventSink_AlreadyExistsDrainsRateLimiter (0.00s)
PASS
ok  	github.com/flyteorg/flyte/flytepropeller/events	0.739s
```

The rest of the package, including `TestAdminAlreadyExistsError` which pins the returned error:

```
ok  	github.com/flyteorg/flyte/flytepropeller/events	0.983s
ok  	github.com/flyteorg/flyte/flytepropeller/events/errors	0.305s
```

</details>

### Labels

fixed

### DIFF
--- a/flytepropeller/events/admin_eventsink.go
+++ b/flytepropeller/events/admin_eventsink.go
@@ -66,36 +66,41 @@ func (s *adminEventSink) Sink(ctx context.Context, message proto.Message) error 
 
 	// Validate submission with rate limiter and send admin event
 	if s.rateLimiter.Allow() {
+		var sendErr error
 		switch eventMessage := message.(type) {
 		case *event.WorkflowExecutionEvent:
 			request := &admin.WorkflowExecutionEventRequest{
 				Event: eventMessage,
 			}
 
-			_, err := s.adminClient.CreateWorkflowEvent(ctx, request)
-			if err != nil {
-				return errors.WrapError(err)
-			}
+			_, sendErr = s.adminClient.CreateWorkflowEvent(ctx, request)
 		case *event.NodeExecutionEvent:
 			request := &admin.NodeExecutionEventRequest{
 				Event: eventMessage,
 			}
 
-			_, err := s.adminClient.CreateNodeEvent(ctx, request)
-			if err != nil {
-				return errors.WrapError(err)
-			}
+			_, sendErr = s.adminClient.CreateNodeEvent(ctx, request)
 		case *event.TaskExecutionEvent:
 			request := &admin.TaskExecutionEventRequest{
 				Event: eventMessage,
 			}
 
-			_, err := s.adminClient.CreateTaskEvent(ctx, request)
-			if err != nil {
-				return errors.WrapError(err)
-			}
+			_, sendErr = s.adminClient.CreateTaskEvent(ctx, request)
 		default:
 			return fmt.Errorf("unknown event type [%s]", eventMessage.String())
+		}
+
+		if sendErr != nil {
+			wrapped := errors.WrapError(sendErr)
+			if errors.IsAlreadyExists(wrapped) {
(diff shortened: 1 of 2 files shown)


Sovietaced:
I've been meaning to look into this issue after seeing HUGE amounts of `AlreadyExists` errors when rebooting propeller and wiping this in-memory send cache. Great find!

Sovietaced (approved):

PR #7859, opened 2026-08-14
feat(runs): settings service CRUD handlers (RFC #7775, 2/2)

davidlin20dev:
## Tracking issue

Related to #7775 (Settings Service RFC, second implementation PR).
Follows #7841 (persistence layer), now merged.

#7841 added the settings storage layer, but nothing served it over the API. The generated `SettingsServiceHandler` interface had no implementation, so clients could not store or read settings. This PR adds the service layer so the settings API is callable end to end.

- `CreateSettings` and `UpdateSettings`: validate the key shape by hand (the `buf.validate` annotations on these protos are not enforced by the generated Go), store the `Settings` message as protojson, and map repository sentinel errors to Connect codes: `AlreadyExists`, `NotFound`, `FailedPrecondition`.
- `GetSettingsForEdit`: reads every ancestor scope in one query and returns one `SettingsRecord` per level, broadest first. The number of levels comes from the request key rather than from what is stored, so a level with no row still appears with empty settings at version 0. That is the client's signal to call `CreateSettings` instead of `UpdateSettings`.
- `GetSettings` stays `Unimplemented` on purpose. Merged reads need the resolution engine (RFC task 2.2), which is out of scope here.
- Registration in `runs/setup.go`.
- One generated-code commit: `SettingsRepo` was added to an `all: true` mockery package in #7841 without regenerating `runs/repository/mocks/mocks.go`. This regenerates it, purely additive.
- Covers RFC tasks 2.4, the `GetSettingsForEdit` half of 2.5, and the wiring half of 2.6. Deferred: 2.1 (settings are stored as full protojson for now), 2.2, the quantity validation in 2.3, and the `runs/test/api` smoke test in 2.6.

## How was this patch tested?

Ten test functions and twelve subtests in `runs/service/settings_service_test.go`, running against the embedded Postgres this package already uses, through the real repository rather than mocks:

- create, read, update round trip, including that the version echoed back matches what the row actually holds
- error mapping: duplicate create returns `AlreadyExists`, updating a missing row returns `NotFound`, a stale version returns `FailedPrecondition`, each with a read-back confirming the rejected write changed nothing
- validation tables for all three handlers, nine cases, all `InvalidArgument`
- the levels contract: every level populated, absent levels empty at version 0, and level count following the depth of the request key

## Related PRs

Follows #7841 (persistence layer, 1/2), merged.


runs/service/settings_service.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// NewSettingsService returns the Connect handler implementation for the
+// settings service, backed by the given repository.
+func NewSettingsService(settingsRepo interfaces.SettingsRepo) *SettingsService {
+	return &SettingsService{settingsRepo: settingsRepo}
+}
+
+// GetSettings returns resolved, merged settings. Resolution requires the
+// merge engine (RFC #7775 task 2.2), which is not implemented yet.
```
pingsutw:
```suggestion
// GetSettings returns resolved, merged settings.
```

runs/service/settings_service.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	levels := make([]*settings.SettingsRecord, 0, len(levelKeys))
+	for i, lk := range levelKeys {
+		// Absent level: empty settings, version 0 — the client's signal to use
+		// CreateSettings there (see SettingsRecord in the proto).
+		record := &settings.SettingsRecord{Key: lk, Settings: &settings.Settings{}}
+		if row := rowsByKey[storageKeys[i]]; row != nil {
+			stored := &settings.Settings{}
+			if err := protojson.Unmarshal(row.Data, stored); err != nil {
```
pingsutw: We should set DiscardUnknown: true for backward compatibility. If we add a new field to the settings and store it in the database, we need to make sure the data can still be read if we roll back to a previous version.

runs/service/settings_service.go
```diff
@@ -0,0 +1,184 @@
...
+	if key == nil || req.Msg.GetSettings() == nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, errors.New("key and settings are required"))
+	}
+	if key.GetProject() != "" && key.GetDomain() == "" {
+		return nil, connect.NewError(connect.CodeInvalidArgument, errors.New("a project-scope key requires a domain"))
+	}
+
+	// Settings are stored as full protojson. Sparse storage (PruneInherited /
+	// Hydrate) is RFC #7775 task 2.1, a prerequisite of this handler that is
+	// not implemented yet.
```
pingsutw: we don't need to mention the detail of RFC here
```suggestion
```

pingsutw (commented):
LGTM, some nits

davidlin20dev:
thanks for the review!
all three addressed, ready for another look

pingsutw (approved):

PR #7868, opened 2026-08-18
fix: reference embedded env secrets from pod specs

### LINKED ISSUE #7824: [V2] Secret value is showing in pod spec ENVVAR directly

Rohithmatham12:
## Summary
- mirror embedded env-var secret values into generated Kubernetes Secret objects in the target pod namespace
- inject pod env vars with `ValueFrom.SecretKeyRef` instead of putting the raw secret value directly in the pod spec
- reuse the resolved project/domain/org secret scope for generated env-var secret names and labels
- keep existing image-pull secret lookup behavior using the same resolved scope

Fixes #7824.

## Testing
- `GOCACHE=/private/tmp/flyte-secret-go-cache GOMODCACHE=/private/tmp/flyte-secret-go-mod-cache go test ./flyteplugins/go/tasks/pluginmachinery/secret -run 'TestEmbeddedSecretManagerInjector_(Inject|InjectSecretScopedToOrganization|InjectSecretScopedToDomain|InjectSecretScopedToProject|LookUpSecret_StaleOrgCacheDoesNotMaskProjectScope|InjectImagePullSecret)$|Test_ToImagePullK8sName'`
- `GOCACHE=/private/tmp/flyte-secret-go-cache GOMODCACHE=/private/tmp/flyte-secret-go-mod-cache go test ./flyteplugins/go/tasks/pluginmachinery/secret`
- `git diff --check`

### DIFF
--- a/flyteplugins/go/tasks/pluginmachinery/secret/embedded_secret_manager.go
+++ b/flyteplugins/go/tasks/pluginmachinery/secret/embedded_secret_manager.go
@@ -1,6 +1,7 @@
 package secret
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"fmt"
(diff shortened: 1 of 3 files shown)


flyteplugins/go/tasks/pluginmachinery/secret/embedded_secret_manager.go
```diff
@@ -345,23 +349,84 @@ func (i *EmbeddedSecretManagerInjector) Inject(
...
+				Name:      k8sSecretName,
+				Namespace: pod.GetNamespace(),
+				Labels:    ToEnvVarK8sLabels(components),
+			},
+			Type: corev1.SecretTypeOpaque,
+			Data: map[string][]byte{
+				EmbeddedSecretsEnvVarKey: secretValue,
+			},
+		}
+		if err := i.k8sClient.Create(ctx, newSecret); err != nil {
```
AdilFayyaz: The `Get` on line 361 is from the infromer cache, but the the `Create` would go straight to the API server. So there is a potential window where the secret was just created and a concurrent admission in this process also tried to `Get` the secret. But this would return `NotFound`. Imagine a burst of mpa-task subtasks pods sharing one secret, they would all hit this on the first use of every secret.

flyteplugins/go/tasks/pluginmachinery/secret/embedded_secret_manager.go
```diff
@@ -345,23 +349,84 @@ func (i *EmbeddedSecretManagerInjector) Inject(
...
+		}
+		return k8sSecretName, nil
+	} else if err != nil {
+		return "", fmt.Errorf("failed to check for env var secret [%s] in namespace [%s]: %w", k8sSecretName, pod.GetNamespace(), err)
+	}
+
+	if k8sSecret.Data == nil {
+		k8sSecret.Data = map[string][]byte{}
+	}
+	k8sSecret.Data[EmbeddedSecretsEnvVarKey] = secretValue
```
AdilFayyaz, line 393: This can issue a write every time the pod is being admitted even when the value hasn't changed. So, `N` subtask pods = `N` secret writes.

flyteplugins/go/tasks/pluginmachinery/secret/embedded_secret_manager.go
```diff
@@ -345,23 +349,84 @@ func (i *EmbeddedSecretManagerInjector) Inject(
...
+	if err := i.k8sClient.Update(ctx, k8sSecret); err != nil {
```
AdilFayyaz: The `K8sSecret` comes from the informer cache, so its `resourceVersion` can be stale whenever the secret was recently written.

AdilFayyaz:
lgtm @pingsutw can I get a quick second set of eyes on this?

popojk:
This issue was fixed in this [PR](https://github.com/flyteorg/flyte/pull/7834). We could close this PR. cc @AdilFayyaz

PR #7875, opened 2026-08-19
[V2] Enable UI signal human in the loop

### LINKED ISSUE #7806: Human-in-the-loop conditions are unresolvable on the devbox

popojk:
## Tracking issue

Closes #7806 

The human in the loop signal message bar is not showing in the UI because the backend DB lacks action metadata info. This PR fixed the issue

Backend changes

## How was this patch tested?

Test with the script
```python
"""Minimal HITL condition, for exercising the pause/signal round trip on the devbox.

Run it, note the run name it prints, then resolve the condition from another shell:

    flyte get condition <run-name>
    flyte signal condition <run-name> <action-name> yes

The task blocks in `cond.wait()` until then. A 30 minute timeout keeps a forgotten
run from pausing forever.
"""

import asyncio
import logging

import flyte

image = (
    flyte.Image.from_debian_base()
    .with_pip_packages("cryptography==44.0.3", "pyOpenSSL==25.1.0")
)

env = flyte.TaskEnvironment(
    name="hitl_demo",
    image=image,
)


@env.task
async def ask_human(item: str) -> str:
    cond = await flyte.new_condition.aio(
        "approve-item",
        prompt=f"Approve {item}? (yes/no)",
        data_type=str,
        description="Demo condition for exercising the signal path on the devbox.",
        timeout=timedelta(minutes=30),
    )
    print(f"waiting for a signal on '{item}' — action: {flyte.ctx().action}")
    answer = await cond.wait.aio()
    print(f"got: {answer!r}")
    return answer


@env.task(entrypoint=True)
async def main(item: str = "widget-42") -> str:
    answer = await ask_human(item=item)
    return f"{item} -> {answer}"


if __name__ == "__main__":
    flyte.init_from_config()
    run = flyte.with_runcontext(log_level=logging.INFO).run(main, item="widget-42")
    print(f"run name: {run.name}")
    print(f"resolve with: flyte signal condition {run.name} <action-name> yes")

```
We can see the message bar now
Screenshot 2026-08-19 at 11 48 59 AM

### Labels

## Stack

### DIFF
--- a/runs/repository/impl/action.go
+++ b/runs/repository/impl/action.go
@@ -440,6 +440,20 @@ func (r *actionRepo) UpdateActionPhase(
 		int32(common.ActionPhase_ACTION_PHASE_FAILED),
 		int32(common.ActionPhase_ACTION_PHASE_TIMED_OUT),
 	}
+	// A paused action may settle into any terminal phase. This case needs stating
+	// separately because the `phase <= $n` guard below reads the enum's numeric
+	// order as the lifecycle order, and PAUSED breaks that assumption: it was
+	// appended after the terminal phases, so PAUSED (9) sorts above SUCCEEDED (5).
+	// Without this, signalling a condition updates zero rows — silently, since a
+	// no-op update also skips notifyActionUpdate, so the change never reaches the
+	// watch stream either and the console keeps offering the signal input.
+	terminalPhases := []int32{
+		int32(common.ActionPhase_ACTION_PHASE_SUCCEEDED),
+		int32(common.ActionPhase_ACTION_PHASE_FAILED),
+		int32(common.ActionPhase_ACTION_PHASE_ABORTED),
+		int32(common.ActionPhase_ACTION_PHASE_TIMED_OUT),
+		int32(common.ActionPhase_ACTION_PHASE_RECOVERED),
+	}
 
 	var queryBuilder strings.Builder
 	var args []interface{}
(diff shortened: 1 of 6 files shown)


runs/repository/impl/action.go
```diff
@@ -467,9 +481,10 @@ func (r *actionRepo) UpdateActionPhase(
 		argIdx++
 	}
 
-	queryBuilder.WriteString(fmt.Sprintf(" WHERE project = $%d AND domain = $%d AND run_name = $%d AND name = $%d AND (phase <= $%d OR phase = ANY($%d))", //nolint: staticcheck
-		argIdx, argIdx+1, argIdx+2, argIdx+3, argIdx+4, argIdx+5))
-	args = append(args, actionID.Run.Project, actionID.Run.Domain, actionID.Run.Name, actionID.Name, phase, pq.Array(retryablePhases))
+	queryBuilder.WriteString(fmt.Sprintf(" WHERE project = $%d AND domain = $%d AND run_name = $%d AND name = $%d AND (phase <= $%d OR phase = ANY($%d) OR (phase = $%d AND $%d = ANY($%d)))", //nolint: staticcheck
```
popojk, line 477: In the past we only write db when current phase > previous phase, but this makes new added phases like paused skipping db update, resulting in status unchanged in the UI. We fix that problem here.

pingsutw (approved):

PR #7889, opened 2026-08-20
fix: honor in-band rejection status from UpdateActionStatus

### LINKED ISSUE #7884: bug: actions client ignores the run service's in-band status, skipping the terminal-status retry
## Why

`InternalRunService.UpdateActionStatus` reports failures **in the response body**, not as a
transport error. The handler always returns a nil error and puts the failure in `Status`:

```go
// runs/service/internal_run_service.go:204
func (s *RunService) UpdateActionStatus(...) (*connect.Response[workflow.UpdateActionStatusResponse], error) {
	resp := s.updateActionStatus(ctx, req.Msg)   // errors become resp.Status via statusFromError
	return connect.NewResponse(resp), nil        // <- always nil
}
```

Its only caller ignores that field and keys off the transport error alone:

```go
// actions/k8s/client.go:840
if _, err := c.runClient.UpdateActionStatus(ctx, connect.NewRequest(statusReq)); err != nil {
	logger.Warnf(ctx, "Failed to update action status in run service for %s: %v", update.ActionID.Name, err)
} else if isTerminalPhase(update.Phase) && !update.IsDeleted {
	c.markTerminalStatusRecorded(ctx, taskAction)   // runs even when the run service rejected the update
}
```

So when the run service fails to record a terminal phase (a DB blip in `UpdateActionPhase`, or in
`mergeRunInfoOutput` for a signalled condition), the executor treats it as success and patches
`flyte.org/terminal-status-recorded=true` onto the TaskAction.

That label is what closes the door. `shouldSkipTaskAction` (`actions/k8s/client.go:713`) makes the
informer skip *every* subsequent event for that CR — including the Delete event the garbage
collector produces at TTL, which is the one retry the system actually has: `buildActionUpdate`
preserves the terminal phase and `SignalValue` on deletes, so that event would otherwise re-issue
the same `UpdateActionStatus`. The GC selects on `LabelTerminationStatus`/`LabelCompletedTime`
(`executor/pkg/controller/garbage_collector.go:109`), not on `terminal-status-recorded`, so the CR
is deleted regardless and the action is gone for good.

**User-visible result:** the action stays in its last non-terminal phase in the DB forever. The
console shows it as running/paused long after the pod finished, and nothing anywhere errors — the
executor logged success. Recovering requires editing the DB by hand.

### Concrete example: works after the change, broken today

Same input: a terminal `UpdateActionStatus` where the run service hits a transient DB error.

| | today | after the change |
|---|---|---|
| run service | returns `Status{Code: 13 (Internal)}` in the response body | same |
| executor sees | `err == nil` → "success" | non-zero `Status.Code` → logs a warning |
| CR label | `terminal-status-recorded=true` patched | not patched |
| GC delete event | skipped by `shouldSkipTaskAction` | delivered → status update retried |
| final DB phase | stuck at `RUNNING`/`PAUSED` | `SUCCEEDED` |

## What to change

- `actions/k8s/client.go:840` — treat a non-zero `resp.Msg.GetStatus().GetCode()` as a failure, so
  `markTerminalStatusRecorded` is skipped and the CR stays eligible for the GC delete-event retry:

  ```go
  resp, err := c.runClient.UpdateActionStatus(ctx, connect.NewRequest(statusReq))
  switch {
  case err != nil:
  	logger.Warnf(ctx, "Failed to update action status for %s: %v", update.ActionID.Name, err)
  case resp.Msg.GetStatus().GetCode() != 0:
  	logger.Warnf(ctx, "Run service rejected status update for %s: %s",
  		update.ActionID.Name, resp.Msg.GetStatus().GetMessage())
  case isTerminalPhase(update.Phase) && !update.IsDeleted:
  	if err := c.markTerminalStatusRecorded(ctx, taskAction); err != nil {
  		logger.Warnf(ctx, "Failed to mark terminal status recorded for %s: %v", update.ActionID.Name, err)
  	}
  }
  ```

- `actions/k8s/client.go:806` — the parent-promotion call has the same unchecked pattern. No label is
  patched there, so it only costs a missing log line; worth fixing in the same pass for consistency.

- `actions/k8s/client_test.go` — add a case where the mocked run client returns a response carrying a
  non-zero `Status.Code`, asserting the TaskAction is **not** labelled `terminal-status-recorded`.
  The existing tests around `TestNotifyRunService_*` (line 118 onward) show the mock setup.

## Outcome

- [ ] A terminal status update rejected by the run service does not label the TaskAction
      `flyte.org/terminal-status-recorded`
- [ ] The GC delete event therefore still reaches `notifyRunService`, retrying the status update
- [ ] The rejection is logged with the run service's status message, not silently swallowed
- [ ] A test covers the non-zero-`Status.Code` path

## Getting started

- **Reproduce (unit):** in `actions/k8s/client_test.go`, have the mock return
  `&connect.Response[workflow.UpdateActionStatusResponse]{Msg: &workflow.UpdateActionStatusResponse{
  Status: &status.Status{Code: int32(connect.CodeInternal), Message: "boom"}}}` with a nil error, then
  assert on the CR's labels. Today the label is patched; after the fix it isn't.
- **Relevant code:** `notifyRunService` in `actions/k8s/client.go` (the two call sites at 806 and
  840), `shouldSkipTaskAction` at 713, `markTerminalStatusRecorded` at 904. The producer side is
  `UpdateActionStatus`/`statusFromError` in `runs/service/internal_run_service.go`.
- **Note:** `connect.Response{}` with a nil `Msg` is safe to call `GetStatus()` on — protobuf getters
  are nil-receiver safe — so existing tests that return an empty response keep passing unchanged.
- **Test:** `go test ./actions/k8s/...`

Fayupable:
## Tracking issue

Closes #7884

`UpdateActionStatus` never returns a transport error for an in-band rejection (e.g. a DB write failure) — the run service reports it via `resp.Status` instead. `notifyRunService` only checked the transport error, so a rejected terminal status update was still marked `terminal-status-recorded`, permanently preventing further retries. This left actions stuck in a non-terminal phase in the DB with no visible error.

Check `resp.Msg.GetStatus().GetCode()` in addition to the transport error at both `UpdateActionStatus` call sites in `notifyRunService` (parent-promotion and terminal status update). If the run service reports a non-zero status code, log a warning and skip `markTerminalStatusRecorded`, so the update can be retried on the next event.

## How was this patch tested?

Added `TestNotifyRunService_RejectedStatusUpdateSkipsTerminalLabel` in `client_test.go`, which mocks `UpdateActionStatus` returning a non-zero `Status.Code` with a nil transport error, and asserts the TaskAction CR does not receive the `terminal-status-recorded` label.

### Labels

- fixed

### DIFF
--- a/actions/k8s/client.go
+++ b/actions/k8s/client.go
@@ -645,12 +645,6 @@ func (c *ActionsClient) notifyRunService(ctx context.Context, taskAction *execut
 		if resp, err := c.runClient.UpdateActionStatus(ctx, connect.NewRequest(statusReq)); err != nil {
 			logger.Warnf(ctx, "Failed to update action status in run service for %s: %v", update.ActionID.Name, err)
 		} else if code := resp.Msg.GetStatus().GetCode(); code != 0 {
-			// UpdateActionStatus never returns a transport error for an in-band
-			// rejection (e.g. a DB failure) — the run service reports it via
-			// resp.Status instead. Without this check, a rejected terminal
-			// update still gets marked terminal-status-recorded below, and the
-			// action is stuck in a non-terminal phase with no visible error
-			// and no further retries.
 			logger.Warnf(ctx, "Run service rejected action status update for %s: code=%d message=%s",
 				update.ActionID.Name, code, resp.Msg.GetStatus().GetMessage())
 		} else if isTerminalPhase(update.Phase) && !update.IsDeleted {
--- a/actions/k8s/client.go
+++ b/actions/k8s/client.go
@@ -624,8 +624,11 @@ func (c *ActionsClient) notifyRunService(ctx context.Context, taskAction *execut
 					Phase: common.ActionPhase_ACTION_PHASE_RUNNING,
 				},
 			}
-			if _, err := c.runClient.UpdateActionStatus(ctx, connect.NewRequest(parentStatusReq)); err != nil {
+			if resp, err := c.runClient.UpdateActionStatus(ctx, connect.NewRequest(parentStatusReq)); err != nil {
 				logger.Warnf(ctx, "Failed to promote parent action %s to RUNNING: %v", update.ParentActionName, err)
+			} else if code := resp.Msg.GetStatus().GetCode(); code != 0 {
+				logger.Warnf(ctx, "Run service rejected promoting parent action %s to RUNNING: code=%d message=%s",
+					update.ParentActionName, code, resp.Msg.GetStatus().GetMessage())
 			}
 		}
 	}
@@ -639,8 +642,17 @@ func (c *ActionsClient) notifyRunService(ctx context.Context, taskAction *execut
 				CacheStatus: taskAction.Status.CacheStatus,
 			},
 		}
-		if _, err := c.runClient.UpdateActionStatus(ctx, connect.NewRequest(statusReq)); err != nil {
+		if resp, err := c.runClient.UpdateActionStatus(ctx, connect.NewRequest(statusReq)); err != nil {
 			logger.Warnf(ctx, "Failed to update action status in run service for %s: %v", update.ActionID.Name, err)
+		} else if code := resp.Msg.GetStatus().GetCode(); code != 0 {
+			// UpdateActionStatus never returns a transport error for an in-band
+			// rejection (e.g. a DB failure) — the run service reports it via
+			// resp.Status instead. Without this check, a rejected terminal
+			// update still gets marked terminal-status-recorded below, and the
+			// action is stuck in a non-terminal phase with no visible error
+			// and no further retries.
+			logger.Warnf(ctx, "Run service rejected action status update for %s: code=%d message=%s",
+				update.ActionID.Name, code, resp.Msg.GetStatus().GetMessage())
 		} else if isTerminalPhase(update.Phase) && !update.IsDeleted {
 			// Skip label patching for deleted CRs — the patch would always fail
 			// with "not found" since the object is already gone.
--- a/actions/k8s/client_test.go
+++ b/actions/k8s/client_test.go
@@ -10,9 +10,14 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/genproto/googleapis/rpc/status"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/protobuf/types/known/wrapperspb"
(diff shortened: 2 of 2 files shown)


actions/k8s/client.go
```diff
@@ -639,8 +642,17 @@ func (c *ActionsClient) notifyRunService(ctx context.Context, taskAction *execut
...
-		if _, err := c.runClient.UpdateActionStatus(ctx, connect.NewRequest(statusReq)); err != nil {
+		if resp, err := c.runClient.UpdateActionStatus(ctx, connect.NewRequest(statusReq)); err != nil {
 			logger.Warnf(ctx, "Failed to update action status in run service for %s: %v", update.ActionID.Name, err)
+		} else if code := resp.Msg.GetStatus().GetCode(); code != 0 {
+			// UpdateActionStatus never returns a transport error for an in-band
+			// rejection (e.g. a DB failure) — the run service reports it via
+			// resp.Status instead. Without this check, a rejected terminal
+			// update still gets marked terminal-status-recorded below, and the
+			// action is stuck in a non-terminal phase with no visible error
+			// and no further retries.
```
pingsutw:
```suggestion
```

pingsutw: nit I think the code is self-explanatory

pingsutw (commented):
LGTM, one nit

pingsutw (dismissed):

pingsutw:
@Fayupable you should change the base branch to main

pingsutw:
can you update it or open another PR

Fayupable:
@pingsutw  Sorry, I'm a bit confused about what's needed here. I see the base branch is back on v2 now. Could you clarify exactly what you'd like me to do? Should I recreate the PR, or is the current state okay as is?

davidlin20dev:
@Fayupable My understanding is that the v2 branch(which you set as your base branch) is stale now, since it was merged into main, which is the base branch we should be developing against. Hope this helps!

Fayupable:
@davidlin20dev  @pingsutw Switching the base to main shows conflicts across ~20 unrelated files (Dockerfile, Makefile, README.md, most of /app, etc.) that I haven't touched. I'm not familiar enough with this codebase to know which side is correct, so I don't want to resolve those blindly and risk breaking something.

Could you let me know the right way to handle this? Happy to open a fresh PR rebased against main if that's preferred, or wait if there's a better approach on your end.

Fayupable:
@pingsutw @davidlin20dev Opened a fresh PR rebased cleanly against main to avoid the v2/main conflict mess: #7896

Closing this one in favor of that.

PR #7891, opened 2026-08-20
fix(executor): stop retrying Kubernetes HTTP 400 responses

1fanwang:
## Tracking issue

Related to https://github.com/flyteorg/flyte/pull/7366

A validating admission webhook can reject a Kubernetes resource with HTTP 400. The
executor currently treats that response as a transient system error.

Before this change, Flyte retries the same rejected resource until the task times out.
After this change, the task fails once as a user error and shows the API server's
message.

Classify Kubernetes bad-request responses as terminal failures with reason
`BadRequest`. Existing handling for forbidden, invalid, and unknown errors is
unchanged.

## How was this patch tested?

The real create-error classification path is exercised with webhook and bare HTTP 400
responses at the Kubernetes client boundary. No live admission webhook was used.

```console
$ git checkout upstream/main -- executor/pkg/plugin/k8s/plugin_manager.go
$ go test ./executor/pkg/plugin/k8s/... -run TestLaunchResourceErrors -v
--- FAIL: TestLaunchResourceErrors/webhook_400_fails
--- FAIL: TestLaunchResourceErrors/bare_400_fails

$ git checkout origin/1fanwang-k8s-badrequest-terminal-failure -- executor/pkg/plugin/k8s/plugin_manager.go
$ go test ./executor/pkg/plugin/k8s/... -run TestLaunchResourceErrors -v
--- PASS: TestLaunchResourceErrors/webhook_400_fails
--- PASS: TestLaunchResourceErrors/bare_400_fails
--- PASS: TestLaunchResourceErrors/invalid_fails
--- PASS: TestLaunchResourceErrors/forbidden_retries
--- PASS: TestLaunchResourceErrors/unknown_retries
```

### Labels

fixed

### Setup process

The controller-runtime fake client returns real Kubernetes status errors from Create.

- [ ] I updated the documentation accordingly. Not applicable: no documented behavior changed.

## Related PRs

- https://github.com/flyteorg/flyte/pull/7366

### DIFF
--- a/executor/pkg/plugin/k8s/plugin_manager.go
+++ b/executor/pkg/plugin/k8s/plugin_manager.go
@@ -126,6 +126,12 @@ func (pm *PluginManager) launchResource(ctx context.Context, tCtx pluginsCore.Ta
 		if k8serrors.IsInvalid(err) {
 			return pluginsCore.DoTransition(pluginsCore.PhaseInfoFailure("InvalidResource", err.Error(), nil)), nil
 		}
+		// Same for HTTP 400, which is what a validating admission webhook returns when it
+		// rejects the spec outright. Distinct code from InvalidResource because a webhook
+		// rejection and a field the apiserver itself found invalid are different diagnoses.
+		if k8serrors.IsBadRequest(err) {
+			return pluginsCore.DoTransition(pluginsCore.PhaseInfoFailure("BadRequest", err.Error(), nil)), nil
+		}
 		reason := k8serrors.ReasonForError(err)
 		logger.Errorf(ctx, "Failed to launch job, system error. err: %v", err)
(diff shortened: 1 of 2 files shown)


1fanwang:
The description that went up when I opened this was the wrong one — it described an unrelated change. I've replaced it, which is most likely what tripped the check.

The fix: when a validating admission webhook rejects a pod spec, the API server answers the create with HTTP 400. That is neither `Forbidden` nor `Invalid`, so it falls through to the catch-all as a retryable system error and the executor retries a request that can never succeed. This makes it terminal and surfaces the webhook's message. It ports behaviour already merged on v1 in https://github.com/flyteorg/flyte/pull/7366.

Red-to-green for the new test is in the description.

pingsutw (approved):

PR #7899, opened 2026-08-21
fix(flyteplugins): fail tasks whose container exits non-zero instead of relaunching forever

### LINKED ISSUE #7874: Raw ContainerTask exiting non-zero is deleted and recreated forever instead of failing the action
## Describe the bug

A raw `ContainerTask` whose command exits non-zero never fails its action. The executor deletes the pod and creates a new one under the **same pod name** with a **new uid**, indefinitely, while the action stays non-terminal and `Attempts` stays at 1. `retries` is no defence — the loop sits below it.

The identical task exiting 0 finalises correctly, so the exit code is the only variable.

I hit this on a CAD task that rejects malformed input in ~1.5s. It ran 16 times in two minutes before I killed it by hand; because the action reports RUNNING the whole time, it presents as a hang rather than a failure.

## Expected behavior

A non-zero exit should fail the action (or consume a retry and then fail), not be re-created forever.

## To Reproduce

Two tasks, identical except for the last token of the command:

```python

import flyte
from flyte.extras import ContainerTask

IMAGE = "alpine:3.20"
WRITE_OUTPUT = "mkdir -p /var/outputs; echo ok > /var/outputs/status;"

exit_zero = ContainerTask(
    name="exit_zero",
    image=flyte.Image.from_base(IMAGE),
    outputs={"status": str},
    command=["/bin/sh", "-c", f"{WRITE_OUTPUT} exit 0"],
    retries=0,
    timeout=timedelta(minutes=5),
)

exit_nonzero = ContainerTask(
    name="exit_nonzero",
    image=flyte.Image.from_base(IMAGE),
    outputs={"status": str},
    command=["/bin/sh", "-c", f"{WRITE_OUTPUT} exit 5"],
    retries=0,
    timeout=timedelta(minutes=5),
)

env = flyte.TaskEnvironment.from_task("exitrepro", exit_zero, exit_nonzero)
```

```bash
flyte run repro.py exit_zero      # SUCCEEDED in ~10s
flyte run repro.py exit_nonzero   # never terminates; abort by hand
```

Watch the pod rather than the console, which just says Running:

```bash
kubectl get pod -n flyte <run>-a0-0 \
  -o jsonpath='{.metadata.uid} {.status.phase} {.status.containerStatuses[0].restartCount}'
```

## Observed

| task | result |
|---|---|
| `exit_zero` (writes output) | `ACTION_PHASE_SUCCEEDED`, ~10s |
| `exit_nonzero` (writes output) | **never terminal.** 11 distinct pod uids in 15 samples over ~45s, one pod name, `restartCount` 0 throughout, cycling Pending → Running → deleted |
| `exit_nonzero`, no `outputs=` declared (no copilot sidecar) | **never terminal.** 30 samples, 30 distinct uids, `restartCount` 0 — so the sidecar is not involved |
| `exit_zero`, no `outputs=` declared | terminal `FAILED` in 4s: `plugin "pod" failed with system error 4 times: failed to read data from dataDir` |

That last row is the useful contrast: the **system-error** path is correctly bounded by `maxSystemFailures` and terminates. Only the non-zero exit is unbounded.

This is not kubelet restarting the container: `restartPolicy` is `Never` on every Flyte task pod (verified across all 51 in the namespace), `restartCount` stays 0, and the pod **uid changes** — something is deleting and re-creating it.

## Suspected area (not verified — pointers for whoever picks this up)

The delete-and-recreate is what the in-place restart block in `executor/pkg/controller/taskaction_controller.go` does on a USER-kind `PhaseRetryableFailure`: `p.Abort(...)`, `PluginState = nil`, transition overridden to `Queued`. But that block is supposed to be bounded by `currentAttempts < maxAttempts`, and `Status.Attempts` never advances past 1 in `flyte get action`.

Two things worth checking:

1. `observedAttempts()` is 1-based — it returns 1 when `Status.Attempts` is unset. The restart block consumes it as a count of *completed* attempts (`restartAttempts = currentAttempts + 1`). If the incremented value doesn't survive to the next reconcile, `currentAttempts` is 1 forever and the guard never trips.
2. With `retries=0` I'd expect `GetMaxAttempts()` to be 1, which should make `1 < maxAttempts` false and fail permanently on the first failure. It doesn't, so either the budget is larger than the task declares or the counter is being reset.

Related but **not** a fix for this: #7823 (v2.0.41) makes deterministic plugin errors fail immediately, and its own description notes it only catches errors returned directly by a plugin — "errors built from a failure phase go through `systemErrorFromPhaseInfo`, which flattens the code into the message text". A container exit is a failure phase, so it isn't covered. I diffed `taskaction_controller.go` between v2.0.35 and v2.0.42: the retryable-failure and attempts logic is byte-identical, so this reproduces on current main.

## Workaround

A raw container must always exit 0 and report its outcome through a declared output, which a downstream Python task reads and raises on. `pod_template` is not an option for this — setting one routes serialization through `_get_k8s_pod`, which drops `data_config` and so removes the copilot sidecar entirely.

The cost of the workaround is that a rejected case now shows SUCCEEDED in the console, so failures are invisible in the UI.

## Environment

- Flyte backend: `cr.flyte.org/flyteorg/flyte-binary-v2:v2.0.35` (also present in v2.0.42 by inspection)
- Flyte SDK: `flyte` 2.5.7
- Kubernetes: GKE v1.35.6-gke.1258000

davidlin20dev:
## Tracking issue

Closes #7874

A raw `ContainerTask` whose command exits non-zero never fails. The executor keeps deleting and recreating the pod under the same name forever, and `retries` has no effect.

When a pod fails, `DemystifyFailure` decides whether the failure is the user's fault (consumes a retry) or the system's fault (relaunch the pod, no retry consumed). It makes that call from the pod-level `status.Reason` field, but the kubelet leaves that field empty for an ordinary container failure. Finding nothing, the code assumes the node died before anything was recorded and classifies the failure as SYSTEM. The system path relaunches the pod without consuming retries, so the same failure repeats forever. The actual answer, the container's own exit code, was available the whole time but never consulted.

Before falling back to SYSTEM: if the pod has exactly one main container and it exited with a code between 1 and 127, classify the failure as USER, using the container's recorded reason (`ContainerFailed` when the runtime left it empty).

This matches `DeterminePrimaryContainerPhase` in the same file, which already treats a non-zero exit as a user error. Two guards keep the change narrow:

- Exit codes above 127 mean the process was killed by a signal (a node drain or spot eviction sends SIGTERM, which surfaces as 143). Those stay SYSTEM so infrastructure churn never burns user retries.
- Only the pod's main containers are considered. Flyte's co-pilot runs as init containers, so it can never be blamed on the user.

The new code only runs when `code` is still `UnknownError`, so it can only affect cases that previously fell into the fallback.

## How was this patch tested?

Four new unit tests in `pod_helper_test.go`; three fail without the fix (non-zero exit with no reason, exit 127, exit 5 alongside a co-pilot init container), and one pins exit codes above 127 as SYSTEM.

Also verified end to end the same way the issue was reported: the repro script from #7874 unchanged, submitted with the `flyte` CLI (Python SDK) via `flyte run repro.py ...` against a local devbox running the backend built from this branch, with the pods watched through `kubectl`:

| run | before (#7874) | after |
|---|---|---|
| `exit_zero` | succeeded | succeeded, output intact |
| `exit_nonzero`, `retries=0` | looped forever, 11 pod uids | failed in ~5s, USER error with the exit code |
| same, `retries=2` | retries had no effect | pods `-0`, `-1`, `-2`, then failed, 3 attempts used |

The recorded error also correctly separates the sidecar from the task:

    [flyte-copilot-uploader] terminated with ExitCode 0.
    [rcqzcwfvs5kx84p7chdb-a0-0] terminated with exit code (5). Reason [Error].

### Setup process

    make devbox-build FLYTE_ARCHES=arm64
    make devbox-run
    flyte create config --endpoint localhost:30080 --project flytesnacks --domain development --builder local --insecure
    flyte run repro.py exit_nonzero   # repro.py verbatim from #7874

### Screenshots
image-1787352527483
image-1787352533409

### DIFF
--- a/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -40,6 +40,16 @@ const SIGKILL = 137
 // unsignedSIGKILL = 256 - 9
 const unsignedSIGKILL = 247
 
+// ContainerFailed is reported when the task's container exited with a non-zero status of its own
+// accord and the container runtime gave no more specific reason.
+const ContainerFailed = "ContainerFailed"
+
+// maxUserExitCode is the highest exit status attributable to the application itself. A process
+// killed by signal N is reported as 128+N, so the band above 127 means something terminated the
+// process rather than the process choosing to fail. Graceful eviction, drain and preemption all
+// send SIGTERM first, which surfaces as 143.
+const maxUserExitCode = 127
+
 const defaultContainerTemplateName = "default"
 const defaultInitContainerTemplateName = "default-init"
 const primaryContainerTemplateName = "primary"
@@ -1600,6 +1610,21 @@ func DemystifyFailure(ctx context.Context, status v1.PodStatus, info pluginsCore
 					containerState.Terminated.ExitCode,
 					containerState.Terminated.Reason,
 					containerState.Terminated.Message)
+			}
+		}
+	}
+
+	// A container that exited non-zero on its own is the application reporting failure, not the
+	// missing kubelet record the fallback below assumes. Only main containers are considered: the
+	// co-pilot runs as init containers. When a pod has exactly one main container, that
+	// container's exit status is taken as the task's result.
+	if code == "UnknownError" && len(status.ContainerStatuses) == 1 {
+		if t := status.ContainerStatuses[0].State.Terminated; t != nil &&
+			t.ExitCode > 0 && t.ExitCode <= maxUserExitCode {
+			if t.Reason != "" {
+				code = t.Reason
+			} else {
+				code = ContainerFailed
 			}
 		}
 	}
--- a/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper_test.go
+++ b/flyteplugins/go/tasks/pluginmachinery/flytek8s/pod_helper_test.go
@@ -3156,6 +3156,107 @@ func TestDemystifyFailure(t *testing.T) {
 			})
 		}
 	})
+
+	t.Run("non-zero exit with no reason is a user error", func(t *testing.T) {
+		phaseInfo, err := DemystifyFailure(ctx, v1.PodStatus{
+			Phase: v1.PodFailed,
+			ContainerStatuses: []v1.ContainerStatus{
+				{
+					Name: "abc123-n0-0",
+					State: v1.ContainerState{
+						Terminated: &v1.ContainerStateTerminated{
+							ExitCode: 2,
+						},
+					},
+				},
+			},
+		}, pluginsCore.TaskInfo{}, "")
+		assert.Nil(t, err)
+		assert.Equal(t, pluginsCore.PhaseRetryableFailure, phaseInfo.Phase())
+		assert.Equal(t, ContainerFailed, phaseInfo.Err().Code)
+		assert.Equal(t, core.ExecutionError_USER, phaseInfo.Err().Kind)
+	})
+
+	t.Run("exit 127 is still a user error", func(t *testing.T) {
+		// 127 is the inclusive upper bound: a missing binary or bad entrypoint is
+		// the user's failure, not the platform's.
+		phaseInfo, err := DemystifyFailure(ctx, v1.PodStatus{
+			Phase: v1.PodFailed,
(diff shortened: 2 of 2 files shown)


popojk (approved):
LGTM. Nice work!

PR #7900, opened 2026-08-23
feat(runs): validate settings on create and update

davidlin20dev:
## Tracking issue

Related to #7775. This is task 2.3 (validators).

The `buf.validate` annotations on the settings protos are not enforced by the generated Go code, so the service currently accepts any value. `task_resource.max.memory` can be stored as `"apple"`. Nothing reads these values yet, but once the appliers land a bad value would fail deep inside a run instead of at the point someone saved it.

Validators wired into `CreateSettings` and `UpdateSettings`:

- quantities must parse with `resource.ParseQuantity`, across all eight `task_resource.{min,max}.{cpu,gpu,memory,storage}` leaves; the error names the dot-path
- `max_action_concurrency` must be `0` or between `2` and `MaxUint32`. The ceiling matches `RunSpec.max_action_concurrency`, which is a `uint32`
- key-shape checks, previously duplicated in three handlers, extracted into `validateSettingsKey`

Leaves whose state is not `VALUE` are skipped, so sparse documents stay valid.

## How was this patch tested?

Table tests per validator, plus rows added to the existing `TestCreateSettings_Validation` and `TestUpdateSettings_Validation` tables to cover the handler wiring.

## Related PRs

Follows #7841, #7859 and #7881.

### DIFF
--- a/runs/service/settings_service.go
+++ b/runs/service/settings_service.go
@@ -39,11 +39,8 @@ func (s *SettingsService) GetSettingsForEdit(
 	req *connect.Request[settings.GetSettingsForEditRequest],
 ) (*connect.Response[settings.GetSettingsForEditResponse], error) {
 	key := req.Msg.GetKey()
-	if key == nil {
-		return nil, connect.NewError(connect.CodeInvalidArgument, errors.New("key is required"))
-	}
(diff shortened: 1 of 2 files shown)


popojk (approved):
LGTM thanks! cc @pingsutw for second review.

PR #7907, opened 2026-08-24
Add support for service to service auth

### LINKED ISSUE #7908: Add support for service to service auth
Currently there is no way to configure authentication for intra-service communication that may happen across kubernetes clusters. In environments where external requests are authenticated (ie. cross cluster) this makes Flyte non functional.

Sovietaced:
## Tracking issue
Closes #7908 

If flyte components are deployed on different clusters there is no way to configure oauth credentials such that requests will pass ingress and oauth2proxy. 

Expand service URLs into a more generic service config concept where we can have common configuration for connect RPC clients. This can be extended in the future to configure TLS related settings and retry options.

This takes a similar approach to the one in Flyte V1 where there is an auth type enum. 

Note: This technically introduced breaking changes into the existing flyte-binary helm chart but I'm not sure why people would be customizing any service to service related configuration since it runs on a single binary anyway. 

## How was this patch tested?
Tested in our internal environment with our multi cluster flyte-core helm chart. 

### Labels

## Stack


flytestdlib/serviceclient/client.go
```diff
@@ -0,0 +1,162 @@
...
+		EndpointParams: endpointParams,
+	}
+
+	baseTransport := base.Transport
+	if baseTransport == nil {
+		baseTransport = http.DefaultTransport
+	}
+	// Use the base transport for token requests as well as service requests. This
+	// preserves custom TLS/proxy settings without recursively invoking auth.
+	tokenHTTPClient := *base
```
popojk, line 121: Could we set a timeout for this tokenHTTPClient, so we can at least return an error when the token service is down?

Sovietaced: Good point. I had this in our internal fork as well.

popojk (dismissed):
LGTM thanks!

popojk (approved):

PR #7910, opened 2026-08-25
fix(executor): enforce TaskMetadata.timeout as a per-attempt max runtime

### LINKED ISSUE #7901: [flyte2] TaskMetadata.timeout (max_runtime) is serialised but never enforced
## Describe the bug

`flyte.Timeout(max_runtime=...)` has no effect. A task declaring a 30 second bound ran for 325
seconds and finished `SUCCEEDED`, returning the value it should never have reached.

The SDK is not at fault — the value is on the wire. Decoding `taskTemplate` from the `TaskAction`
CRD shows `22 02 08 1e`: field 4, length 2, varint `0x1e` = 30. That matches the contract in
`flyteidl2/core/tasks.proto`:

```proto
// Wall-clock bound on a single attempt's ACTION_PHASE_RUNNING phase
// (a.k.a. max_runtime). Per attempt — resets on each retry. Enforced by
// the lease worker. Unset means unlimited.
google.protobuf.Duration timeout = 4;
```

A repo-wide search suggests it is unimplemented rather than broken: `max_runtime` appears only in
`flyteidl2/core/tasks.proto`, `flyteidl2/core/literals.proto` and generated TS/swagger, and there
are no Go references to `MaxRuntime` at all. If the lease worker is meant to enforce this, it does
not appear to read the field.

## Expected behavior

A task exceeding `max_runtime` is terminated and the attempt fails — then retries, or fails
terminally once retries are exhausted.

## Actual behavior

The bound is ignored. The task runs to completion and reports success.

```
pod : 03:39:52Z -> 03:45:17Z   exit=0     (325s)
run : ActionPhase.SUCCEEDED
out : "should have been killed at 30s"
```

## To Reproduce

```python

import flyte

env = flyte.TaskEnvironment(
    name="timeouts",
    image="<any image with the flyte SDK>",
    resources=flyte.Resources(cpu="1", memory="2Gi"),
)


@env.task(timeout=flyte.Timeout(max_runtime=timedelta(seconds=30)))
async def sleeps_past_timeout() -> str:
    import time

    time.sleep(300)
    return "should have been killed at 30s"


if __name__ == "__main__":
    flyte.init(endpoint="...", project="...", domain="...")
    r = flyte.run(sleeps_past_timeout)
    r.wait()
    print(r.phase, r.outputs())   # SUCCEEDED, "should have been killed at 30s"
```

## Why this matters, and a warning about the obvious workaround

Without an enforced bound, a hung task runs — and bills — until something external stops it.
`run.abort()` works reliably, but that requires a human or an external watchdog to notice.

The natural fallback is `activeDeadlineSeconds` on the pod template, since the kubelet enforces it
below Flyte entirely. **That makes things worse.** The kubelet does kill the container, and the
executor then recreates the pod indefinitely:

```
03:58:27  task submitted to K8s -> Queued -> Executing
          DeadlineExceeded: Pod was active on the node longer than the specified deadline
03:59:43  task submitted to K8s -> Queued -> Executing      <- same action, new pod
attempts: 1          <- never increments
errorState: None
```

Thirty phase-history entries on a single action with `attempts` stuck at 1. This looks like the same
mechanism as #7874 reached from a different trigger — I have added a comment there with the detail.

So on `v2.0.42` there is currently no safe way to bound task runtime: the supported mechanism is
ignored, and the Kubernetes-level one turns a bounded task into an unbounded one.

## Environment

| | |
|---|---|
| SDK / CLI | `2.6.3` |
| Backend / `flyte-binary` chart | `v2.0.42` |
| Platform | GKE Autopilot `1.35.6-gke.1641000`, `australia-southeast1` |
| Storage | GCS |
| Database | PostgreSQL 16 |

shaon-chowdhury-euc:
## Summary

- Enforce `TaskMetadata.timeout` (`max_runtime`) in the TaskAction executor: persist `AttemptStartedAt` / `TimeoutAt`, abort and finalize the plugin at the deadline, emit `TIMED_OUT`, and retry only within the existing attempt budget.
- Fixes [#7901](https://github.com/flyteorg/flyte/issues/7901). The SDK already serializes the bound; `v2.0.42` ignored it, so a 30s task could run for minutes and finish `SUCCEEDED`.
- Do not use `activeDeadlineSeconds` as a substitute — on current builds the kubelet kill leaves the Pod object behind and the executor recreates it without consuming an attempt.

The bound is enforced against the controller clock. A terminal transition observed before the deadline wins whatever timestamp it carries; from the deadline onward it wins only if `TaskInfo.OccurredAt` proves it finished in time. A terminal report with no usable timestamp is indistinguishable from an overrun, so the bound takes precedence.

`TIMED_OUT` was already terminal for condition actions but rare in practice. Two places that special-cased `ACTION_PHASE_FAILED` are updated to keep behaviour consistent now that tasks reach it: action-detail error resolution and task-group fail rate.

## Test plan

Verified:

- [x] `go test -race` across `./executor/...`, `./actions/...`, `./runs/...`
- [x] Controller envtest specs covering: attempt-clock anchoring (plugin-reported start, history bootstrap after an upgrade mid-attempt, rejection of a start time predating the attempt); requeue landing on the exact deadline; terminal-vs-deadline races in both directions; per-attempt clock reset across retries with the full event sequence asserted; retry exhaustion becoming terminal failure; resumption after Abort, Finalize, event-publish and status-write failures; unbounded and zero timeouts left unlimited; invalid durations rejected as `InvalidSpec`; ordinary failures not misclassified as timeouts; system retries not consuming a user attempt; serialized cache reservations held until terminal.
- [x] `make manifests generate` produces no drift; all four CRD copies carry the new status fields.

Not yet verified on a live cluster:

- [ ] A `flyte.Timeout(max_runtime=30s)` task sleeping 300s, end to end through the SDK.


shaon-chowdhury-euc:
The `Screen low-quality pull requests` check failed in **Merge checks**, not on the diff.

`peakoss/anti-slop` computes `min-global-merge-ratio` via the GitHub Search API (`author:<login> is:pr`). That call 422s for this account:

```
Validation Failed: The listed users cannot be searched either because the users
do not exist or you do not have permission to view the users.
```

Every earlier check passed (branch, size, description, commits, account age, public profile). This is a Search API limitation on the author account, not a quality problem with the change.

Happy to walk through the timeout enforcement if a maintainer can take a look — the workflow says to reply here on a false positive.

executor/pkg/controller/taskaction_controller.go
```diff
@@ -165,6 +172,347 @@ func (r *TaskActionReconciler) requeueDuration() time.Duration {
...
+		&pluginsCore.TaskInfo{OccurredAt: &deadline},
+	)
+
+	if willRetry {
+		// Publish TIMED_OUT for this attempt before advancing the status. If the
+		// status write then fails, the next reconcile re-publishes it; the events
+		// service dedupes on insert (see recordEvent), so this stays at-least-once
+		// end to end rather than pushing the problem onto consumers.
+		timeoutEvent := r.buildActionEvent(ctx, taskAction, timeoutInfo)
+		if err := r.recordEvent(ctx, timeoutEvent); err != nil {
```
popojk, line 456: When we record timeout event here, the run service will treat the action as terminal and close UI watch stream [here](https://github.com/flyteorg/flyte/blob/43a77c7b93ec116ce2ae2a8ed1aeb445f91e2a75/runs/service/run_service.go#L1274-L1276). I think we should set a gate in run service to terminate the stream only if all retries are completed.

executor/pkg/controller/taskaction_controller.go
```diff
@@ -419,31 +781,66 @@ func (r *TaskActionReconciler) reconcileTask(
...
+		}
+	}
+
+	// Enforce the per-attempt max runtime. A Handle failure at or past the
+	// deadline can no longer establish that the attempt finished in time, so it
+	// ends the attempt too instead of consuming a system-failure retry.
+	if deadline, hasDeadline := taskAttemptDeadline(taskAction, maxRuntime); hasDeadline {
+		var overran bool
+		if handleErr != nil {
+			overran = !r.now().Before(deadline)
```
popojk, line 813: Why do we need this if branch? I think attemptOverran is enough to tell if we are overran when we have handleErr?

shaon-chowdhury-euc: they compute the same thing today (with handleErr set the transition is the zero value, so attemptOverran degenerates to !now.Before(deadline)). the branch exists because transition isn't meaningful when Handle returns an error, and attemptOverran's `OccurredAt` escape (a terminal result proving it finished before the deadline) only makes sense for a successful Handle. didn't want timeout enforcement depending on what a zero-value Transition happens to return. can inline it behind a comment if you'd rather.

executor/pkg/controller/taskaction_controller.go
shaon-chowdhury-euc: agreed, this closes the stream mid-retry. i'd keep the per-attempt `TIMED_OUT` event though (silent overruns are the bug this fixes) and gate the close in run service instead: only close when the terminal attempt is also the action's current attempt. `status.Attempts` moves to N+1 in the same reconcile that emits `TIMED_OUT(N)` so a retrying attempt never satisfies that. will add the gate plus a timeout-then-retry watch test. also want to check the actions table doesn't treat the intermediate `TIMED_OUT` as sticky-terminal and reject the Queued that follows.

shaon-chowdhury-euc: pushed the gate in b2f437c ie. close only when the action phase is terminal and the last event attempt is terminal and equals `status.Attempts` (same equality `buildActionDetails` already uses). added a unit table for both write orderings plus an api test that replays the executor's event-then-row sequence; it fails on the old predicate. also checked the actions-table side: `TIMED_OUT` is already in retryablePhases so the Queued that follows is accepted.

popojk:
Tested in devbox with flyte-sdl/examples/retries_timeout/deadline.py. All good:
Screenshot 2026-09-08 at 1 48 35 PM

shaon-chowdhury-euc:
> Tested in devbox with flyte-sdl/examples/retries_timeout/deadline.py. All good: Screenshot 2026-09-08 at 1 48 35 PM

thanks @popojk!

popojk (approved):

Sovietaced (approved):
Stamping since @popojk already approved and can't be the sole approver since he pushed a commit

PR #7935, opened 2026-08-27
Add initial flyte-core helm chart

### LINKED ISSUE #7931: Flyte 2: "Flyte core" multi-pod flyte services support
https://flyte-org.slack.com/archives/C0BKN1E7494/p1786959107936389

Sovietaced:
## Tracking issue
Closes #7931 

There is no way to deploy Flyte 2 OSS in a production like manager with distributed and independently scalable services.

Adds an initial flyte-core helm chart. Modifies the manager to accept the component to run so that we can reuse the single-binary image/dockerfile. Each components gets its own folder under `/templates` with relevant resources.

Still missing a lot of features to get it production ready but those can come in future pull requests. 

## How was this patch tested?
Tested internally and running across multiple clusters.

### Labels

## Stack


charts/flyte-core/values.yaml
```diff
@@ -0,0 +1,609 @@
...
+        project: my-organization-gcp-project
+      azure:
+        account: storage-account-name
+        key: ""
+        configDomainSuffix: ""
+        configUploadConcurrency: 4
+
+  runs:
+    actionsService:
+      url: http://actions:8080
```
popojk:
```suggestion
      url: '{{ include "flyte-core.componentServiceURL" (dict "root" . "component" "actions") }}'
```
The service name might be override with a prefix, should we use the helper function to generate service url?

charts/flyte-core/values.yaml
```diff
@@ -0,0 +1,609 @@
...
+    trustForwardedIdentityHeaders: true
+    identityHeaders:
+      claimsJwtHeader: X-Amzn-Oidc-Data
+      subjectHeader: X-Amzn-Oidc-Identity
+      emailHeader: ""
+    authMetadata: {}
+
+  actions:
+    runService:
+      url: http://runs:8080
```
popojk: ditto

charts/flyte-core/values.yaml
```diff
@@ -0,0 +1,609 @@
...
+    server:
+      host: 0.0.0.0
+      port: 8080
+    watchBufferSize: 1000
+    watchWorkers: 100
+    recordFilterSize: 8388608
+
+  events:
+    runService:
+      url: http://runs:8080
```
popojk: ditto

charts/flyte-core/values.yaml
```diff
@@ -0,0 +1,609 @@
...
+  cache:
+    server:
+      host: 0.0.0.0
+      port: 8080
+    heartbeatGracePeriodMultiplier: 3
+    maxReservationHeartbeat: 10s
+
+  dataproxy:
+    runService:
+      url: http://runs:8080
```
popojk: ditto

charts/flyte-core/values.yaml
```diff
@@ -0,0 +1,609 @@
...
+
+  secret:
+    server:
+      host: 0.0.0.0
+      port: 8080
+    webhookURL: ""
+
+  executor:
+    eventsService:
+      url: http://events:8080
```
popojk: ditto

charts/flyte-core/values.yaml
```diff
@@ -0,0 +1,609 @@
...
+    cacheService:
+      url: http://cache:8080
```
popojk: ditto

charts/flyte-core/templates/secret/configmap.yaml
```diff
@@ -0,0 +1,30 @@
...
+  {{- end }}
+data:
+  000-core.yaml: |
+    {{- include "flyte-core.config.logger" . | nindent 4 }}
+    secret:
+      server:
+        {{- toYaml .Values.configuration.secret.server | nindent 8 }}
+      kubernetes:
+        {{- tpl (toYaml .Values.configuration.kubernetes) . | nindent 8 }}
+      webhookURL: {{ $webhookURL | quote }}
```
popojk:
```suggestion
      webhook:
        url: {{ $webhookURL | quote }}
```

charts/flyte-core/templates/executor/configmap.yaml
```diff
@@ -0,0 +1,50 @@
+{{- if and (not (include "flyte-core.externalConfiguration" .)) .Values.components.executor.enabled }}
+{{- $executorConfig := omit .Values.configuration.executor "metricsPort" "healthProbePort" }}
```
popojk, line 2: I'm thinking maybe we should omit `certDir` as some users may want to do it but our service does not allow so.

charts/flyte-core/templates/crds/flyte.org_taskactions.yaml
```diff
@@ -0,0 +1,343 @@
+---
```
popojk, line 1: Should we move CRD folder into templates folder so that we will be able to update the CRD when executing `helm upgrade`?

popojk (commented):
@Sovietaced Thanks for your awesome work! I just left a few comments. PTAL thanks!

charts/flyte-core/values.yaml
Sovietaced: I'm ok with using the helper with the default, I had that idea earlier as well so that the port is only defined in one place.

charts/flyte-core/templates/crds/flyte.org_taskactions.yaml
Sovietaced: Sure! I'm not aware of these details so I can move it.

Sovietaced:
> @Sovietaced Thanks for your awesome work! I just left a few comments. PTAL thanks!

Addressed

popojk (approved):
LGTM thanks! I've tested in local kind cluster and it works fine cc @pingsutw for second eye

PR #7949, opened 2026-09-01
Upgrade to Go 1.27

Sovietaced:
New versions of Go include new features, performance improvements, and security fixes. See: https://go.dev/doc/go1.27

Leveraged Go 1.27 to do the following
1. Converted futures, auto-refresh cache, and scheduler timing tests to testing/synctest.
2. Migrated service-client HTTP tests to Go 1.27’s in-memory httptest.NewTestServer.
3. Modernized reverse event traversal with slices.Backward.
4. Ran the atomictypes, slicesbackward, and unsafefuncs fixers. Only the reverse-loop change was safely applicable.

### Labels

## Stack


flyteplugins/go/tasks/pluginmachinery/tasklog/azure_plugin_test.go
```diff
@@ -1,67 +1,61 @@
...
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/core"
 )
 
 func TestAzureTemplateLogPlugin(t *testing.T) {
-	type args struct {
```
Sovietaced, line 13: Go 1.27 changed the compress/flate implementation, so gzip output can produce different—but equally valid—compressed bytes.
The Azure test previously asserted the entire generated URL, including a hard-coded base64-encoded gzip payload. After upgrading Go, that byte-for-byte value changed even though the decompressed Azure query was identical.
I changed this to verify the meaningful behavior:
1. Check the Azure URL prefix and log metadata.
2. URL-decode the query.
3. Base64-decode it.
4. Gzip-decompress it.
5. Compare the resulting query text with the expected query.
So it was a test-stability adjustment required by Go 1.27’s compression change, not a production behavior change. It also avoids future failures when compression output changes without changing its contents.

go.mod
```diff
@@ -24,6 +26,7 @@ require (
 	github.com/fergusstrange/embedded-postgres v1.34.0
 	github.com/flyteorg/stow v0.3.13
 	github.com/fsnotify/fsnotify v1.10.1
+	github.com/getsentry/sentry-go v0.48.0
```
popojk: Why do we downgrade this package from v0.49.0?

go.mod
```diff
@@ -10,6 +10,8 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.23.1
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.14.1
 	github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azsecrets v1.5.0
+	github.com/Masterminds/semver/v3 v3.5.0
+	github.com/alicebob/miniredis/v2 v2.38.0
```
popojk: Why do we downgrade this package from v2.39.0?

popojk (commented):
Thanks for your contribution. I just left 2 small comments.

go.mod
Sovietaced: Probably a bad merge conflict

Issue #7932, opened 2026-08-27
Flyte 2: platform settings service


cosmicBboy:

davidlin20dev:
Hi @cosmicBboy,
Just wanted to add some context here.
I'm currently working on this under @pingsutw's RFC: #7775
Linking the 6 related PRs below:

Merged:
- #7841
- #7859
- #7881
- #7900

In review:
- #7925
- #7927

PR #7956, opened 2026-09-01
refactor(runs): make settings resolution callable outside the handlers

davidlin20dev:
## Tracking issue

Related to #7775. Preparation for task 3.1. Related to #7932.

Task 3.1 needs to resolve settings inside `RunService.CreateRun`, but the only way to resolve them today is `GetSettings`, a Connect handler. Calling it from another service means building a request envelope for a call that never leaves the process. The work it does is not reachable on its own either, since `fetchLevels` is a method on `SettingsService` and the fetch, decode and merge steps are inlined in the handler.

- New `runs/service/settings_resolve.go`: `fetchLevels` (moved, now takes the repo as a parameter), `decodeStored` (new, one place for the row decode both read paths duplicated), and `resolveSettings` (fetch, decode, merge, returning a plain error).
- `GetSettings` is now validate, resolve, wrap. `GetSettingsForEdit` uses the moved `fetchLevels` and the shared decode.

No behavior change: same responses, same errors, same queries.

## How was this patch tested?

- The existing settings service tests pass unchanged.
- New `TestResolveSettings` resolves an org and project chain with no handler involved, asserting the merged value and its `scopeLevel`. The key names a domain with no record, so the gap case is covered.
- The `runs/test/api` suite passes, so `GetSettings` is unchanged over real HTTP.

### Labels

- **changed**

## Related PRs

Builds on #7841, #7859, #7900, #7925, #7927 and #7937. The follow-up wires this into `RunService` and applies the first setting.

### DIFF
--- a/runs/service/settings_resolve.go
+++ b/runs/service/settings_resolve.go
@@ -1,0 +1,78 @@
+package service
+
+import (
+	"context"
+
+	"google.golang.org/protobuf/encoding/protojson"
+
+	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/settings"
+	"github.com/flyteorg/flyte/v2/runs/repository/interfaces"
+	"github.com/flyteorg/flyte/v2/runs/repository/models"
+)
+
+// fetchLevels returns the scope levels covered by key, broadest first, together with
+// the stored row for each. The two slices align by position, and a nil row means no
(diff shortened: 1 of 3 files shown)


AdilFayyaz (approved):
LGTM!

PR #7964, opened 2026-09-02
artifact: bound ArtifactKey/ArtifactVersionId field lengths

EngHabu:
## Why

Copilot review feedback on #7963 that missed the merge: `parent_artifact`'s reference type (`core.ArtifactVersionId` / `core.ArtifactKey`) carried no `buf.validate` constraints, so lineage references could declare unbounded strings while the equivalent identity fields on `artifact.ArtifactName` are bounded. (The artifact service already enforces these lengths server-side; this aligns the declared contract.)

## What

- `core.ArtifactKey`: `max_len` on project/domain (64), name (255), org (63) — matching `artifact.ArtifactName`. Fields stay optional: empty means "inherit from context" (e.g. `ArtifactSpec.parent_artifact`).
- `core.ArtifactVersionId.version`: `max_len` 255.
- `ProducedArtifact.parent_artifact` doc now states the same-scope constraint explicitly: only the name may differ; non-empty org/project/domain must match the produced artifact's own scope.
- Regenerated Go/Python/TS/Rust bindings (annotation-only change; no field or wire changes).

https://claude.ai/code/session_01VzRDrkwWaxeTpqB4WS165m


EngHabu:
Superseded by #7972, which folds these constraints in (the parent field there became `repeated parent_artifacts` and rewrites the same `ProducedArtifact` doc block, so keeping this separate would just conflict).

Issue #7939, opened 2026-08-30
[Fix] Increasing k8s setting to handle overwhelming requests in devbox.


0yukali0:
# Why is this proposal?
[flyte-benchmark](https://github.com/flyteorg/flyte-benchmarks/tree/main)` triggers overloading problem in liteSQL and k8s API server.

# Related Issues
Related https://github.com/flyteorg/flyte/issues/7757

# What to do?
Update k3s setting with following arguments.
```bash
"--kube-apiserver-arg=max-requests-inflight=3000", \                                                                                                                                               
"--kube-apiserver-arg=max-mutating-requests-inflight=3000"                                                                                                                                                                                                                                                                     
"--kube-controller-manager-arg=kube-api-burst=3000"                                                                                                                                           
"--kube-controller-manager-arg=kube-api-qps=3000"                                                                                                                                         
"--kubelet-arg=max-pods=300"                                                                                                     
"--kubelet-arg=system-reserved=cpu=1500m,memory=3Gi"    #Optional, enable it when no resource reserved for kubelet                                                                                                       
"--kubelet-arg=kube-reserved=cpu=500m,memory=1Gi" #Optional, enable it when no resource reserved for kubelet         
```

# How to reproduce?
```bash
# git clone flyte
make devbox-build
make devbox-run
# git clone flyte-benchmark
uv run scripts/v2/swarm.py       --k 25 --n 2000
```

# Screenshots
-k 10 -n 2000 triggers K8S Api error from  k8s api server.
<img width="1452" height="752" alt="Image" src="https://github.com/user-attachments/assets/583b4331-57e1-4ad6-8f90-803fd7d7fea1" />

0yukali0:
#take

PR #7965, opened 2026-09-02
[Flyte 7757] Prevent NOTIFY from blocking action writes

### LINKED ISSUE #7757: perf: don't block action writes on the NOTIFY pump

0yukali0:
## Tracking issue
Close #7757 

Action and run updates could block writes when the notification queue was full, or be lost when notification delivery failed.
This issue is addressed by [PR-7787](https://github.com/flyteorg/flyte/issues/7939).

- Replaced fixed notification queues with deduplicated pending sets.
- Made update writes return immediately without waiting for notification delivery.
- Retried failed notifications instead of dropping them.
- Merged repeated updates for the same action or run.
- Added tests for stalled queues, retries, concurrent updates, and end-to-end delivery.

## How was this patch tested?
```bash
go test ./runs/repository/impl
```
### Labels

### Setup process
A stable environment to run the performance have 24 core and 64G is better.
Or building a k3s cluster with multiple machines which provides 18core and 38G.
Updating [k3s setting](https://github.com/flyteorg/flyte/issues/7939)

```bash
# git clone flyte
make devbox-build && make devbox-run && make devbox-monitoring

# ssh postgres container
 k3s_node="$(kubectl get nodes -o jsonpath='{.items[0].metadata.name}')"

  kubectl debug "node/${k3s_node}" -it \
    --image=nicolaka/netshoot --profile=sysadmin -- bash
    
 tc qdisc replace dev cni0 root handle 1: prio bands 3
  tc qdisc replace dev cni0 parent 1:3 handle 30: netem delay 15ms
  tc filter replace dev cni0 protocol ip parent 1: \
    prio 1 u32 match ip protocol 6 0xff \
    match ip sport 5432 0xffff flowid 1:3

  tc -s qdisc show dev cni0
  exit


# git clone flyte-benchmark
uv run scripts/v2/swarm.py --k 25 --n 2000
```

## Stack


0yukali0:
experiment of this pr with 15ms delay setting on Postgres.

# result based on main branch
<!DOCTYPE html>
RPC | rate | p50 | p95 | mean
-- | -- | -- | -- | --
InternalRunService.UpdateActionStatus | 7.65/s | 21.7 ms | 10s | 12.5 ms
InternalRunService.RecordActionEvents |3.87/s | 3.04s | 10s | 5.99 ms
ActionsService.Enqueue |  276/s  |  | 736ms | 20.2ms | 

Screenshot 2026-09-04 at 8 13 36 AM
Screenshot 2026-09-04 at 8 15 44 AM
Screenshot 2026-09-04 at 8 18 37 AM


# result from this branch.
<!DOCTYPE html>
RPC | rate | p50 | p95 | mean
-- | -- | -- | -- | --
InternalRunService.UpdateActionStatus | 31.8/s | 21.7 ms | 99.7 ms | 31.8 ms
InternalRunService.RecordActionEvents | 22/s | 36.1 ms | 123 ms | 41 ms
ActionsService.Enqueue |  1.23/s  |  | 3.92s | 35.4ms | 

Screenshot 2026-09-04 at 2 07 19 AM

Screenshot 2026-09-04 at 2 11 15 AM

Screenshot 2026-09-04 at 2 14 51 AM

runs/repository/impl/action.go
```diff
@@ -36,11 +36,17 @@ type actionRepo struct {
 	actionSubscribers map[chan string]bool
 	mu                sync.RWMutex
 
-	// Dedicated channels for async NOTIFY to avoid pool contention
-	actionNotifyCh chan string
-	runNotifyCh    chan string
+	notifyMu       sync.Mutex
+	pendingActions map[string]struct{}
+	pendingRuns    map[string]struct{}
```
popojk: Why do we change the data structure to map here? I think we will not be able to guarantee FIFO with a map

runs/repository/impl/action.go
```diff
@@ -51,8 +57,9 @@ func NewActionRepo(db *sqlx.DB, dbConfig database.DbConfig) (interfaces.ActionRe
 		actionSubscribers: make(map[chan string]bool),
 	}
 
-	repo.actionNotifyCh = make(chan string, 256)
-	repo.runNotifyCh = make(chan string, 256)
+	repo.pendingActions = make(map[string]struct{})
+	repo.pendingRuns = make(map[string]struct{})
```
popojk: I think we should still set a boundary for the notification buffer. In case where db shut down, the buffer will  keep accumulating the notifications to a point where the run service OOM. I think we can try to find out a buffer size that could pass our pressure test(e.g. 500k actions). WDYT?

kumare3:
This is pretty good

runs/repository/impl/action.go
0yukali0: Good catch. 
I think that set a boundary to prevent from OOM is a great idea.
When exceeding the boundary, LRU strategy remove the old notification.

runs/repository/impl/action.go
0yukali0: Using a map alone does not guarantee FIFO ordering. 
I'll add a slice to preserve the order when sending batch notifications.

0yukali0:
# Result from this branch. (heap and  FIFO try it best) (Deprecated)
<!DOCTYPE html>
RPC | rate | p50 | p95 | mean
-- | -- | -- | -- | --
InternalRunService.UpdateActionStatus | 33.6/s | 21.1 ms | 93.1 ms |  127ms
InternalRunService.RecordActionEvents | 21.2/s | 36.1 ms | 99.4 ms | 22-6 ms
ActionsService.Enqueue |  1.75/s  |  | 3.49s | 38-11ms | 

Screenshot 2026-09-05 at 9 12 38 AM
Screenshot 2026-09-05 at 9 12 45 AM
Screenshot 2026-09-05 at 9 12 51 AM

I will set 50k as the boundary
Screenshot 2026-09-05 at 10 41 27 AM

0yukali0:
# Result from this branch. FIFO and deduplicate
<!DOCTYPE html>
RPC | rate | p50 | p95 | mean
-- | -- | -- | -- | --
InternalRunService.UpdateActionStatus |29.1/s | 20.9 ms | 91.7ms |  289ms
InternalRunService.RecordActionEvents |38.3/s | 37.1 ms |93.4 ms | 24.1 ms
ActionsService.Enqueue |  0.6-1.36/s  |  | 1.91s | 45-7ms | 

Screenshot 2026-09-09 at 1 20 56 AM
Screenshot 2026-09-09 at 1 19 12 AM
Screenshot 2026-09-09 at 1 17 08 AM

runs/repository/impl/action.go
```diff
@@ -36,11 +36,21 @@ type actionRepo struct {
...
+	pendingRuns        map[string]struct{}
+	pendingRunQueue    []string
+	pendingCh          chan struct{}
 }
 
+const (
+	notificationBufferLimit     = 65_000
+	pendingNotificationCapacity = 256
+	notifyRetryMinBackoff       = 50 * time.Millisecond
+	notifyRetryMaxBackoff       = 5 * time.Second
```
popojk, line 51: Can we make these settings configurable in the config file while at the same time keeping the default settings? In this way it is easier for perf test.

popojk (commented):
Looks really good. I just have a small comment. Then we are good to go. Thanks!

0yukali0 (commented):
Hi @popojk , I think that this one is ready for review.
If there are any questions or suggestions, pls tag me.
Thanks.

0yukali0:
Ｈi
I will update this pr in Sep 24 after i resolve some tasks of my work.

PR #7988, opened 2026-09-05
proto gen changes for rust

debasisdwivedy:
## Tracking issue
Adding rust photos for backend communication rather relying on python  

Instead of using `python plugin` now the rust sdk can communicate directly with the backend  

`buf.gen.rust.yaml` , `flyteidl2/gen_utils/rust/Cargo.toml`  now removes the pyo3 ref.  
`flyteidl2/gen_utils/rust/build.rs` does not need `pyo3` referenced functions.  

## How was this patch tested?

Using `flyte-rs-sdk`. Below is a screenshot.  

### Labels

### Screenshots  

image

## Stack


wild-endeavor:
thanks for helping clean this up.  before this can go in though...
* use `make gen` to get the generate checks to pass.

Secondly, we do need the pyo3 stuff.  This crate is not completely immune to python (or any language we might ultimately want to use through FFI).  To effect this, we need to publish 1 crate, but add all the bindings as features.  two crates will never work for the setup we want to achieve - you'll end up seeing errors like
```
"expected `TaskSpec`, found `TaskSpec`"`
```

Try these steps? (claude generated)...
* `buf.gen.rust.yaml`: `type_attribute=.=#[cfg_attr(feature = "pyo3"\, pyo3::pyclass(dict\, get_all\, set_all))]`
* `Cargo.toml`: `pyo3`, `pyo3-async-runtimes`, and the `syn`/`quote`/`prettyplease`/`regex` build-deps become optional under a `pyo3` feature; default features should stay empty.
* `build.rs`: emit `boxed_impls.rs` / `serde_impls.rs` only when cargo sets the `CARGO_FEATURE_PYO3` env var.

Test by compiling the SDK against this PR...
in your sdk folder, change
```
flyteidl2 = { path = "/path/to/this/pr/gen/rust" }
```

then `cargo check` and also `maturin develop --release`

debasisdwivedy:
```
Test by compiling the SDK against this PR...
in your sdk folder, change

flyteidl2 = { path = "/path/to/this/pr/gen/rust" }
then cargo check and also maturin develop --release
```
Which SDK are we talking about. I presume the above statement is about rust-sdk. The current rs-sdk will completely change. I have removed the python-plugin-rs from it.

The only thing that needs to be checked is if these changes break the `python-sdk`

I will give it a try over the weekend. 
This is a 3 phase PR's:
1) Flyte protos (current PR)
2) Flyte SDK (python rs_controller) : [Flyte-SDK](https://github.com/flyteorg/flyte-sdk/pull/1542)  
3) Flyte-SDK-RS : To be created . I have done some basic steps to create a task/view/follow, without using python plugin. The screenshot attached above is from the RS-SDK without pyo3/python-plugin.

I need to test may be the python SDK to see how does the change break it. With the current PR's mentioned above i ran [flyte_intro](https://github.com/flyteorg/flyte-sdk) example mentioned in the README of python-sdk and was able to create the task.

wild-endeavor:
I was referring to the main flyte-sdk - the #2 in your list.  was not aware you were working on that, thank you, but we shouldn't merge this pr until that one is a good state also.  i'll check that later today.

debasisdwivedy:
Totally makes sense. I will give the python-sdk testing a try during the weekend as well. I did test one use case `flyte_intro` with the changed code, but may be a more through testing is required.
I presumed rs_controller is part of the build/test pipeline and all the test case passed.

Thanks.

PR #7997, opened 2026-09-07
fix(runs): ignore org in the settings storage key

davidlin20dev:
## Tracking issue

Related to #7775 and #7932. Addresses a review comment on #7961.

The settings storage key is built from the org in the request, which nothing else in OSS Flyte reads. #7157 removed org from every runs table and service and left the proto fields in place, unread.

The SDK derives an org from the endpoint hostname, so settings written through the CLI are stored under it. Runs created without a name and runs created by the scheduler carry no org and look under the default, so they find nothing. Whether a stored setting applies depends on how the run was started, and neither path reports an error.

- `EncodeSettingsKey` no longer takes an org, so keys go from `v1:{org}:{domain}:{project}` to `v1::{domain}:{project}`. Removing the parameter rather than ignoring it keeps the mismatch from coming back. `NormalizeOrg` and `DefaultOrg` go with it.
- The wire API is unchanged. Clients still send org, validation is untouched, and `GetSettingsForEdit` still echoes the requested org back in each level key.
- One behavior change: two clients sending different orgs for the same domain and project now collapse onto one key, so the second `CreateSettings` returns `AlreadyExists`.

## How was this patch tested?

- Added a regression test that writes under one org and resolves under another, expecting the same record. Nothing covered that before, so the bug had no test.
- The encoder table test now pins one key per scope level.
- `go test ./runs/...` passes, including the `runs/test/api` customer flow replay.

### Labels

- **fixed**

## Related PRs

#7961 is the first PR that reads these settings and applies them to a run. Builds on #7841, #7859, #7900, #7925, #7927, #7937 and #7956.


popojk (approved):
LGTM, thanks! cc @AdilFayyaz for second eye.

popojk:
In addition, I'm thinking to add a migration to convert old settings keys in db into new format for backward compatibility, but maybe it's not necessary as we did not officially released settings service yet.

PR #8000, opened 2026-09-08
fix(clustered): propagate pod metadata to JobSet

### LINKED ISSUE #7989: [clustered plugin] PodTemplate labels/annotations not propagated to JobSet ObjectMeta
### Describe the bug

`PodTemplate.labels` set via the Flyte 2 SDK's `ClusteredTaskEnvironment` are propagated to the **child pod template** (`objectMeta`) but never to the **parent JobSet's `ObjectMeta.Labels`**. This makes it impossible to use Kueue (or any admission controller that reads CRD-level labels) with clustered tasks.

### Root cause

In [`flyteplugins/go/tasks/plugins/k8s/clustered/build.go`](https://github.com/flyteorg/flyte/blob/6c14e255170c89f2893a096702f72bfd482bc010/flyteplugins/go/tasks/plugins/k8s/clustered/build.go), `BuildResource` already merges SDK labels + execution labels + default labels onto the pod template's `objectMeta` at [L63-64](https://github.com/flyteorg/flyte/blob/6c14e255170c89f2893a096702f72bfd482bc010/flyteplugins/go/tasks/plugins/k8s/clustered/build.go#L63-L64):

```go
objectMeta.Labels = utils.UnionMaps(cfg.DefaultLabels, objectMeta.Labels,
    utils.CopyMap(taskCtx.TaskExecutionMetadata().GetLabels()))
```

But the JobSet itself at [L144-149](https://github.com/flyteorg/flyte/blob/6c14e255170c89f2893a096702f72bfd482bc010/flyteplugins/go/tasks/plugins/k8s/clustered/build.go#L144-L149) hardcodes its labels:

```go
ObjectMeta: metav1.ObjectMeta{
    Name:      jobSetName,
    Namespace: taskCtx.TaskExecutionMetadata().GetNamespace(),
    Labels: map[string]string{
        "flyte.org/execution": sanitizeLabelValue(...),
    },
    ...
}
```

So any labels from the SDK's `PodTemplate` (e.g. `kueue.x-k8s.io/queue-name`) land on the pods but not on the JobSet, where Kueue and other controllers look for them.

### Proposed fix

Merge `objectMeta.Labels` into the JobSet's `ObjectMeta.Labels`, same as already done for the pod template. The `flyte.org/execution` label takes precedence (rightmost in `UnionMaps`):

```go
ObjectMeta: metav1.ObjectMeta{
    Name:      jobSetName,
    Namespace: taskCtx.TaskExecutionMetadata().GetNamespace(),
    Labels: utils.UnionMaps(objectMeta.Labels, map[string]string{
        "flyte.org/execution": sanitizeLabelValue(...),
    }),
    Annotations: utils.UnionMaps(objectMeta.Annotations, map[string]string{
        "flyte.org/task-type":      taskType,
        primaryContainerAnnotation: primaryContainerName,
    }),
},
```

This is a 2-line, non-breaking change. Existing deployments see no difference (default `objectMeta.Labels` only contains execution-id labels that are already present). Deployments setting custom labels on `PodTemplate` (e.g. for Kueue) would see those labels propagated to the JobSet.

### Expected behavior

`PodTemplate.labels` should appear on both the JobSet's `ObjectMeta.Labels` AND the child pod template. This would enable Kueue queue-name labels, gang scheduling via pod-groups, and any custom admission controller integration.

### Additional context

- This was verified empirically on Flyte 2.6.5 with `flyte-binary` v2.0.42 on GKE 1.35.
- The 1.x fix for `@task(labels=...)` ([PR #6421](https://github.com/flyteorg/flyte/pull/6421), [Issue #6238](https://github.com/flyteorg/flyte/issues/6238)) addressed a similar problem for flytekit 1.16+ and PyTorchJob, but the Flyte 2 clustered plugin's `build.go` was written separately and does not include this fix.
- The Ray plugin handles this correctly for RayJob CRDs, providing precedent.
- [Kueue](https://kueue.sigs.k8s.io/) requires `kueue.x-k8s.io/queue-name` on the parent CRD (JobSet), not on child pods.

- [X] Yes

- [X] Yes

Chiuuu0209:
## Tracking issue

Closes #7989

PodTemplate labels and annotations are propagated to the child pod
template but not to the parent JobSet. Controllers such as Kueue read
queue labels from the parent JobSet CRD, so clustered tasks using these
labels cannot be admitted correctly.

Merge the resolved PodTemplate labels and annotations into the JobSet's
ObjectMeta.

Flyte-managed metadata is applied after the PodTemplate metadata so that
the following values retain precedence:

- `flyte.org/execution`
- `flyte.org/task-type`
- `flyte.org/primary-container`

A regression test verifies that custom labels and annotations are
propagated to both the parent JobSet and the child pod template. It also
verifies that Flyte-managed metadata retains precedence over
user-provided values.

**The regression test was confirmed to fail before the fix** because the
custom metadata was absent from the parent JobSet.

## How was this patch tested?

```shell
go test -race ./flyteplugins/go/tasks/plugins/k8s/clustered \
  -run '^TestBuildResource_PropagatesPodTemplateMetadataToJobSet$' \
  -count=1

go test -race ./flyteplugins/go/tasks/plugins/k8s/clustered -count=1

go test -race -coverprofile=coverage.out \
  -covermode=atomic ./flyteplugins/...

go vet ./flyteplugins/go/tasks/plugins/k8s/clustered
```
All commands passed.

`make -C flyteplugins lint` did not reach the lint step because the local
tooling bootstrap expected `flytestdlib/bin/pflags`, which is not produced
by the current `flytestdlib` compile target.

## Labels
fixed
## Setup process
No special setup is required.
## Screenshots
Not applicable.
- [ ] I updated the documentation accordingly. Not applicable: no documented behavior changed.

### DIFF
--- a/flyteplugins/go/tasks/plugins/k8s/clustered/clustered_test.go
+++ b/flyteplugins/go/tasks/plugins/k8s/clustered/clustered_test.go
@@ -30,7 +30,7 @@ import (
 	pluginIOMocks "github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/io/mocks"
 	plugink8s "github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/k8s"
 	k8smocks "github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/k8s/mocks"
-	"github.com/flyteorg/flyte/v2/flyteplugins/go/tasks/pluginmachinery/utils"
+	"github.com/flyteorg/flyte/v2/flytestdlib/utils"
 	"github.com/flyteorg/flyte/v2/gen/go/flyteidl2/core"
 	clusteredpb "github.com/flyteorg/flyte/v2/gen/go/flyteidl2/plugins"
 )
@@ -43,7 +43,7 @@ const (
 
 // buildTaskTemplate builds a TaskTemplate with the given ClusteredTaskSpec packed into Custom.
 func buildTaskTemplate(spec *clusteredpb.ClusteredTaskSpec) *core.TaskTemplate {
-	custom, err := utils.MarshalObjToStruct(spec) //nolint:staticcheck
+	custom, err := utils.MarshalObjToStruct(spec)
 	if err != nil {
 		panic(err)
 	}
--- a/flyteplugins/go/tasks/plugins/k8s/clustered/clustered_test.go
+++ b/flyteplugins/go/tasks/plugins/k8s/clustered/clustered_test.go
@@ -62,13 +62,13 @@ func buildTaskTemplate(spec *clusteredpb.ClusteredTaskSpec) *core.TaskTemplate {
 }
 
 // dummyTaskCtx builds a minimal task execution context suitable for BuildResource tests.
-func dummyTaskCtx(taskTemplate *core.TaskTemplate) *coreMocks.TaskExecutionContext {
-	return dummyTaskCtxWithGeneratedName(taskTemplate, testJobName)
+func dummyTaskCtx(taskTemplate *core.TaskTemplate, podTemplate *core.K8SPod) *coreMocks.TaskExecutionContext {
+	return dummyTaskCtxWithGeneratedName(taskTemplate, testJobName, podTemplate)
 }
 
 // dummyTaskCtxWithGeneratedName is dummyTaskCtx with a caller-supplied generated name, used to
 // exercise the long composed/nested-name truncation path.
-func dummyTaskCtxWithGeneratedName(taskTemplate *core.TaskTemplate, generatedName string) *coreMocks.TaskExecutionContext {
+func dummyTaskCtxWithGeneratedName(taskTemplate *core.TaskTemplate, generatedName string, podTemplate *core.K8SPod) *coreMocks.TaskExecutionContext {
 	taskCtx := &coreMocks.TaskExecutionContext{}
 
 	inputReader := &pluginIOMocks.InputReader{}
@@ -115,7 +115,7 @@ func dummyTaskCtxWithGeneratedName(taskTemplate *core.TaskTemplate, generatedNam
 	})
 	overrides.EXPECT().GetExtendedResources().Return(nil)
 	overrides.EXPECT().GetContainerImage().Return("")
-	overrides.EXPECT().GetPodTemplate().Return(nil)
+	overrides.EXPECT().GetPodTemplate().Return(podTemplate)
 
 	meta := &coreMocks.TaskExecutionMetadata{}
 	meta.EXPECT().GetTaskExecutionID().Return(tID)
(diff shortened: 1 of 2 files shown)


flyteplugins/go/tasks/plugins/k8s/clustered/clustered_test.go
```diff
@@ -187,6 +187,66 @@ func TestBuildResource_HappyPath(t *testing.T) {
...
+		Replicas:     2,
+		NprocPerNode: 1,
+		Runtime: &clusteredpb.Runtime{
+			Kind: &clusteredpb.Runtime_Torchrun{
+				Torchrun: &clusteredpb.TorchRuntime{},
+			},
+		},
+	}
+
+	podSpec, err := utils.MarshalObjToStruct(corev1.PodSpec{
```
AdilFayyaz, line 201: nit: can you update the call sites to use this instead: https://github.com/flyteorg/flyte/blob/main/flytestdlib/utils/marshal_utils.go#L88

flyteplugins/go/tasks/plugins/k8s/clustered/clustered_test.go
AdilFayyaz: this as well : https://github.com/flyteorg/flyte/blob/main/flytestdlib/utils/marshal_utils.go#L88

flyteplugins/go/tasks/plugins/k8s/clustered/clustered_test.go
Chiuuu0209: Thanks for the catch!

I switched the test helpers to `flytestdlib/utils.MarshalObjToStruct` instead of the deprecated `pluginmachinery/utils` one — updated the import in `clustered_test.go` and dropped the` //nolint:staticcheck` on the `buildTaskTemplate` call. The `PodSpec` site picks it up via the same import. Pushed in `e653d43`.

Re-ran locally:
```
go test ./flyteplugins/go/tasks/plugins/k8s/clustered -count=1
```
All good.

One related thing I noticed: `build.go` still has `utils.UnmarshalStruct(...)` with `//nolint:staticcheck`, which is also on the deprecated path (`UnmarshalStructToPb` in `flytestdlib`). I left that alone for now so this PR stays focused on the JobSet metadata fix + your nit. Happy to clean it up in a follow-up, or fold it into this PR if you’d rather — up to you.

flyteplugins/go/tasks/plugins/k8s/clustered/clustered_test.go
Chiuuu0209: Same here — switched this call site over with the import change in `e653d43`.

flyteplugins/go/tasks/plugins/k8s/clustered/clustered_test.go
AdilFayyaz: Looks good, a follow up PR would help

Chiuuu0209: Sounds good — I’ll open a follow-up PR for the UnmarshalStruct cleanup. Thanks for the review!

AdilFayyaz (approved):

PR #8025, opened 2026-09-12
feat(runs): apply the pod template setting to every task in a run

davidlin20dev:
## Tracking issue

Related to #7775 and #7932. This closes task 3.2b.

A task gets its pod template by naming one itself, or it falls back to a single cluster-wide default. An admin has no way to set one per project or domain, so every task author has to remember it individually and any task that forgets silently misses out.

- The resolved name is now a field on `RunSpec`, alongside environment variables, labels and annotations. Those three reach every task in a run because `applyRunSpecToTaskAction` projects them onto the root action and `inheritRunContextFromParentTaskAction` copies them to each child. `pod_template_name` had no field on `RunSpec`, which is why it was the only pod-level setting that stopped at the first task.
- That needs two schema fields the RFC did not anticipate, `RunSpec.pod_template_name` and `TaskActionSpec.PodTemplateName`. Stamping at run creation instead, as the RFC describes, only ever reaches the task the run starts with, because child actions are created later from inside the running pod.
- Nothing changes until an admin sets the setting. Both new fields are optional and default to empty, so a run with no pod template setting behaves exactly as it does today.
- A task that named its own pod template keeps it.
- The registered task spec is untouched. The stamp happens after `persistRunModel`, on a clone, so the stored spec stays exactly as the user registered it.
- The executor needs no changes. Its existing lookup reads `TaskMetadata.pod_template_name` and does not care who wrote it.

I built the run-creation version first, which is what the RFC describes, and it only ever reached the task the run starts with. This is the alternative, and it costs two schema fields the RFC says are not needed, so let me know if you'd rather keep the schema as it is.

## How was this patch tested?

- Each hop of the journey has a test: settings to run spec, run spec to root action, parent to child.
- `applyPodTemplateName` has a table covering every branch, plus a separate test proving the caller's template is not mutated.
- A wiring test proves `embedTaskTemplate` actually calls it, which the table cannot show.
- `go test ./runs/... ./actions/... ./executor/...` passes.

### Labels

- **added**

## Related PRs

Builds on the settings service work tracked in #7932.


runs/service/settings_apply.go
```diff
@@ -25,4 +25,10 @@ func applyRunSettings(spec *task.RunSpec, resolved *settings.Settings) {
 		concurrency.GetState() == settings.SettingState_SETTING_STATE_VALUE {
 		spec.MaxActionConcurrency = uint32(concurrency.GetIntValue())
 	}
+
+	// Projected onto tasks that name no template of their own, at action creation.
+	if podTemplate := resolved.GetPodTemplateName(); spec.GetPodTemplateName() == "" &&
```
AdilFayyaz, line 30: Nothing checks the pod template name before it reaches Kubernetes. Two things can go wrong:
1. Name longer than 253 chars, K8s would reject the TaskAction but by then the run is already saved as QUEUED with no errors shown
2. Any other invalid name that could be a typo can get through, which can cause a failure downstream by returning an error like "PodTemplate does not exist." 

We can probably reject an incorrect name in CreateRun after `applyRunSettings` before the run is saved.

davidlin20dev: Good catch! Fixed. `CreateRun` now validates the name (DNS-1123 subdomain, same rule Kubernetes uses for PodTemplates) before the run is saved and returns `InvalidArgument`. A well-formed name that doesn't exist still fails at pod build.

AdilFayyaz (approved):

PR #8026, opened 2026-09-13
feat(runs): apply the run base dir and raw data path settings

davidlin20dev:
## Tracking issue

Related to #7775 and #7932. Covers task 3.4 in `CreateRun`: `storage.raw_data_path` in full, `run.run_base_dir` for run metadata. `UploadInputs` does not read settings yet, so offloaded inputs still use the cluster default. That half is a separate PR.

`run.run_base_dir` and `storage.raw_data_path` can be stored and resolved, but nothing reads them, so a run's storage paths still come only from the request or the cluster-wide default. An admin cannot point a project's run data at their own bucket.

- `applyRunSettings` fills `RunSpec.run_base_dir` and `RunSpec.raw_data_storage.raw_data_prefix` when the request leaves them unset. An explicit request value always wins, and a setting that is INHERIT or UNSET contributes nothing.
- `CreateRun` needs no logic change. It already reads both fields and falls back to the configured prefix, and `applyRunSettings` runs before that, so the request, then settings, then config precedence holds without any extra code.
- The paired TODOs are corrected. The one in `CreateRun` asked for exactly this and is narrowed to the remaining gap; the one in `UploadInputs` pointed at `StorageSettings.run_base_dir`, which is `reserved` and no longer exists, and claimed `CreateRun` still needed the work.

## How was this patch tested?

- Seven new rows in the `applyRunSettings` table, covering both fields: the setting is applied, an explicit request value wins, and INHERIT and UNSET contribute nothing. One row sends `RawDataStorage` present but with an empty prefix, which must still take the setting.
- Existing `runs/service` and `dataproxy/service` tests pass unchanged.

### Labels

- **added**

### DIFF
--- a/runs/service/settings_apply.go
+++ b/runs/service/settings_apply.go
@@ -25,4 +25,16 @@ func applyRunSettings(spec *task.RunSpec, resolved *settings.Settings) {
 		concurrency.GetState() == settings.SettingState_SETTING_STATE_VALUE {
 		spec.MaxActionConcurrency = uint32(concurrency.GetIntValue())
 	}
+
+	// Base for run metadata (inputs.pb, outputs.pb), not for user blobs.
+	if baseDir := resolved.GetRun().GetRunBaseDir(); spec.GetRunBaseDir() == "" &&
+		baseDir.GetState() == settings.SettingState_SETTING_STATE_VALUE {
+		spec.RunBaseDir = baseDir.GetStringValue()
+	}
+
+	// RawDataStorage carries only this prefix, so replacing the message loses nothing.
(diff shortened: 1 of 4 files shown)


runs/service/settings_apply.go
```diff
@@ -25,4 +25,16 @@ func applyRunSettings(spec *task.RunSpec, resolved *settings.Settings) {
 		concurrency.GetState() == settings.SettingState_SETTING_STATE_VALUE {
 		spec.MaxActionConcurrency = uint32(concurrency.GetIntValue())
 	}
+
```
popojk, line 28: Could we also validate the format of storage path [here](https://github.com/flyteorg/flyte/blob/604538daae8e58fe36c6caa5ba2805c06c9b0cca/runs/service/settings_service.go#L234)

PR #8028, opened 2026-09-14
feat(dataproxy): resolve run_base_dir from settings in UploadInputs

davidlin20dev:
## Tracking issue

Related to #7775 and #7932. Follow-up to #8026 for task 3.4.

#8026 applies `run.run_base_dir` in `CreateRun`, but `UploadInputs` still picks its own base, and every SDK run uploads its inputs through it. With the setting in place, a run's outputs land under the configured base while its inputs still go to the dataproxy default. `UploadInputsRequest.base_dir` already documents that the server should fall back to settings when the request leaves it empty.

- `UploadInputs` falls back to `run.run_base_dir` from settings when the request has no `base_dir`, so the order matches `CreateRun`: request, then settings, then the configured prefix.
- The dataproxy reads settings through a `SettingsServiceClient` on the same server as its other clients, rather than the repo `CreateRun` uses, because it has no database connection when it runs on its own.
- A failed lookup fails the upload with `CodeInternal`, matching `CreateRun`, instead of writing the inputs outside the configured base.
- The TODOs in both services that tracked this gap are removed.

## How was this patch tested?

- `TestUploadInputs_RunBaseDirFromSettings`: a `VALUE` base is used with its trailing slash trimmed, `UNSET` falls back to the default, and a failed lookup returns `CodeInternal`. The mock only answers for the upload's org, domain and project, so a swapped key fails every row.
- `TestUploadInputs` scripts the settings lookup only for rows without `base_dir`, so its existing `base_dir` row also proves an explicit base skips the lookup.
- Existing `dataproxy` and `runs/service` tests pass.

### Labels

- **added**

## Related PRs

Stacked on #8026.

### DIFF
--- a/dataproxy/service/dataproxy_service.go
+++ b/dataproxy/service/dataproxy_service.go
@@ -275,8 +275,9 @@ func (s *Service) UploadInputs(
 	// When base_dir is set it is a full path used verbatim, bucket and all, so it becomes
 	// the base reference directly (enabling cross-bucket writes); otherwise inputs go
 	// under the configured Upload.StoragePrefix within the operator's base container.
-	// TODO: consult org/project/domain settings (StorageSettings.run_base_dir) here as the
-	// middle tier once settings lookup lands; it must be applied in CreateRun too.
+	// TODO: consult org/project/domain settings (RunSettings.run_base_dir) here as the
+	// middle tier. CreateRun already does, so until this matches, a settings-provided
+	// base applies to run metadata but not to offloaded inputs.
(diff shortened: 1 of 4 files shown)


popojk:
We should set the default value of ` enable-multicontainer` to true, otherwise the stow check [here](https://github.com/flyteorg/flyte/blob/e15f3d7c11b2674593e07ea3ae0b948a9c2f922e/flytestdlib/storage/stow_store.go#L308) will always fail when the bucket config in settings service and config are different.

PR #8049, opened 2026-09-17
Idl/run spec resources

wild-endeavor:

A run's task templates only carry the resources the author declared. Anything that schedules or accounts for those tasks before they become pods (Union's leasor) sees zero for an undeclared dimension, while the dataplane fills a platform default in when it builds the pod. The two disagree for every task that declares nothing.

The org/domain/project `task_resource` setting (`min` as default requests, `max` as a ceiling) is the intended source for those values, but nothing carries the resolved setting with the run. This adds it to `RunSpec` so every consumer of the run, control plane and dataplane, reads the same numbers.

`DefaultSettings`, the run-scoped defaults wrapper from #8050, gains field 2, `task_resource_defaults`, of a new message. It is read as `RunSpec.default_settings.task_resource_defaults`:

```proto
message TaskResourceDefaults {
  repeated flyteidl2.core.Resources.ResourceEntry requests = 1;  // default request per dimension the task template leaves unset
  repeated flyteidl2.core.Resources.ResourceEntry max = 2;       // per-dimension ceiling on a task's requests and limits
}
```

A separate message rather than `flyteidl2.core.Resources` on purpose: `max` is a ceiling on both requests and limits, not a Kubernetes limit, and reusing `limits` for it would read wrong. The server resolves the field at run creation from the `task_resource` setting; a caller-supplied value is currently overwritten.

Generated outputs for Go, Python, TypeScript and Rust are regenerated with the CI image (`make gen`). Two files changed as a side effect of that generator and are not hand edits: `gen/rust/src/flyteidl2.core.rs` is a struct reorder caused by the new `core/tasks.proto` import, and `flyteplugins/go/tasks/pluginmachinery/k8s/mocks/mocks.go` is a parameter rename from the pinned mockery.

## How was this patch tested?

`make gen` in the CI image produces no further diff (check-generate). Generated Go builds. Consumers are exercised in the Union control plane change below.

### Labels

- **added**

## Related PRs

Consumer: unionai/cloud#18479 (run service writes the field, leasor and leaseworker read it).


flyteidl2/task/run.proto
```diff
@@ -181,10 +182,20 @@ message RunSpec {
   // actions are reused and only failed/changed ones re-execute. See Recover.
   Recover recover = 19;
 
-  // Run-scoped pod template name, projected onto tasks that do not name one
-  // themselves. Set by the server from settings; a task's own
-  // TaskMetadata.pod_template_name always wins.
-  string pod_template_name = 20;
+  // Resource defaults for every action in this run. Resolved by the server at creation from the org, project and
+  // domain task_resource settings; a caller-supplied value is currently overwritten.
+  TaskResourceDefaults task_resource_defaults = 20;
```
EngHabu: Are there other "settings" we need to materialize and pass on? should this be a wrapped message "Defaults", "FromSettings" or something?

EngHabu (dismissed):

fiedlerNr9 (dismissed):

SVilgelm (approved):

